PR #43, opened 2018-02-06
Add (but do not use) rustfmt configuration

smarter:
Not used because of https://github.com/rust-lang-nursery/rustfmt/issues/2418


mbebenita:
@smarter is there no way to disable that particular rule?

smarter:
I didn't manage to get it to do something better even by hacking the code

PR #46, opened 2018-02-06
Intra refresh

luctrudeau:


src/lib.rs
```diff
@@ -136,6 +136,22 @@ impl FrameConstants {
             frame_number: 0,
         }
     }
+    #[inline (always)]
```
tdaede: Is there a reason you added accessors for the frame number? Something like advance_frame(), or get_next_frame_constants() that returns a new FrameConstants might make more sense.

tdaede:
This isn't really "intra refresh", it's a forced keyframe interval. Also, I'd suggest forcing all frames to be keyframes for now via some mechanism - it's confusing for it to print something else. For example, set default interval to 0.

luctrudeau:
I'm going to rebase over PR#47

PR #52, opened 2018-02-07
Bench write_b

lu-zero:


benches/bench.rs
```diff
@@ -90,5 +90,46 @@ fn aom(b: &mut Bencher) {
...
+        mc: mc,
+    };
+
+    let mode = PredictionMode::DC_PRED;
+    let tx_type = TxType::DCT_DCT;
+
+    let sbx = 0;
+    let sby = 0;
+
+/*    for p in 0..1 {
```
tdaede: drop commented out loops?

lu-zero: Sure

mbebenita:
Do these benchmarks run via [cargo bench](https://doc.rust-lang.org/1.12.1/book/benchmark-tests.html)?

lu-zero:
Yes.

PR #59, opened 2018-02-08
Enable CDEF_SINGLEPASS.

negge:


barrbrain:
This passes my local test for desync detection. (minus commit becced10 "Show Existing Frames")
This change supports rolling past Dec 21, "Remove CDEF_SINGLEPASS defines"
https://aomedia.googlesource.com/aom/+/8322ff04

PR #60, opened 2018-02-08
build.sh defaults to debug

luctrudeau:
For a release build use build.sh --release

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -5,12 +5,12 @@ set -e
 
 #SEQ=!!!!! ENTER YOUR FAVORITE Y4M HERE !!!!!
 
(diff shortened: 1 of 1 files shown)


tdaede:
Do we still want this?

rzumer:
Sorry, I didn't mean to merge directly in this PR without a custom commit name/description. I fixed conflicts and swapped the argument to set the default build type to `release` and switch to `debug` with the `--debug` flag. I think that it is useful to change the build type easily, but debug mode is much slower, so for a quick test with the default setting, release mode should be preferable.

tdaede:
needs another rebase

PR #66, opened 2018-02-08
Add smooth-hv support.

tdaede:
Also updates to Dec 14, and fixes build on gcc 7.


barrbrain:
This passes my local desync test. (BasketballDrive_1920x1080_50_1s.y4m)

PR #69, opened 2018-02-10
Buffer last reconstructed frame

### LINKED ISSUE #67: Frames with show_existing_frame are blank in reconstruction
Since becced10, frames with show_existing_frame set are blank in the reconstruction stream.

barrbrain:
The "Show Existing Frames" feature uses this to copy into the reconstruction stream.
This demonstrates the plumbing required to resolve #67, and only covers the current behavior.
For future-proofing, an actual frame context buffer is needed.


tdaede:
Looks good as a quick fix, I'll take it. But we really need an "encoder state" struct to stuff this in.

PR #70, opened 2018-02-11
Fix the frame rate in ivf

lu-zero:
Some demuxers will cry foul if the framerate is 0/60.


tdaede:
Yeah I had to do this for my live stream but forgot to pr it.

PR #78, opened 2018-02-13
Update libaom to Dec 1.

tdaede:


barrbrain:
This passes my local desync test with BasketballDrive_1920x1080_50_1s.y4m

PR #81, opened 2018-02-14
 Use Q0 quantizer to derive lambda for Q0 pixel domain

yushincho:


yushincho:
I will do this from my primary github account, ycho. 
Also, I wanted to provide the change of coding performance but could not since the size of input and decoded are all different for test set I have. (We need exact multiple of 64 for both w and h, you know)

PR #82, opened 2018-02-14
Use Q0 quantizers to derive lambda for Q0 pixel domain

yushincho:
The quantizer tables that rav1e currently use are those from libaom,
which are all in Q3 precision (except, qindex=0, lossless mode, which is Q2).
Thus, if we derive lambda which applies to pixel domain in Q0,
we should use quantizers in Q0.


tdaede:
rav1e_test_5 -> rav1e_q0-1

-2.5974 |     N/A |     N/A |  -2.3976 | -1.0754 | -0.3277 |    -1.9909

PR #85, opened 2018-02-16
Slice before building the iterators in pred_dc

lu-zero:
rust-1.24
from
test native_trait  ... bench:     988,673 ns/iter (+/- 185,503)
to
test native_trait  ... bench:     806,088 ns/iter (+/- 150,647)

rust-1.25 (nightly)
from
test native_trait  ... bench:     653,069 ns/iter (+/- 104,601)
to
test native_trait  ... bench:     421,896 ns/iter (+/- 78,777)

### DIFF
--- a/src/predict.rs
+++ b/src/predict.rs
@@ -74,7 +74,7 @@ impl Dim for Block4x4 {
 }
 
 pub fn pred_dc_trait<D: Dim>(output: &mut [u16], stride: usize, above: &[u16], left: &[u16]) {
-    let edges = left.iter().chain(above.iter());
+    let edges = left[..D::H].iter().chain(above[..D::W].iter());
     let bw = D::W;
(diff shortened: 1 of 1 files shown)


src/predict.rs
```diff
@@ -74,7 +74,7 @@ impl Dim for Block4x4 {
 }
 
 pub fn pred_dc_trait<D: Dim>(output: &mut [u16], stride: usize, above: &[u16], left: &[u16]) {
-    let edges = left.iter().chain(above.iter());
+    let edges = left[..D::W].iter().chain(above[..D::H].iter());
```
barrbrain: If `above` is the row of pixels above the block, it should have size `D::W`.
If `left` is the column of pixels to the left of the block, it should have size `D::H`.

PR #86, opened 2018-02-16
Set codegen-units = 1

ycho:
So that upgraded rustc 1.25.0 version does not slow down
the release/test/bench binary by setting the default value of
codegen-units = 16.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,11 @@ bencher = "0.1.5"
 [[bench]]
 name = "bench"
 harness = false
+
+[profile.release]
+codegen-units = 1  # if > 1 enables parallel code generation which improves
+                   # compile times, but prevents some optimizations.
+                   # Passes `-C codegen-units`. Ignored when `lto = true`.
+
+[profile.bench]
+codegen-units = 1


Cargo.toml
```diff
@@ -24,3 +24,13 @@ bencher = "0.1.5"
 [[bench]]
 name = "bench"
 harness = false
+
+[profile.release]
+codegen-units = 1  # if > 1 enables parallel code generation which improves
+                   # compile times, but prevents some optimizations.
+                   # Passes `-C codegen-units`. Ignored when `lto = true`.
+[profile.test]
+codegen-units = 1
```
Ryman: Does the test profile need to be optimized as much?

Cargo.toml
```diff
@@ -24,3 +24,13 @@ bencher = "0.1.5"
 [[bench]]
 name = "bench"
 harness = false
+
+[profile.release]
+codegen-units = 1  # if > 1 enables parallel code generation which improves
+                   # compile times, but prevents some optimizations.
+                   # Passes `-C codegen-units`. Ignored when `lto = true`.
```
barrbrain: Nit: add blank line after comment.

ycho: Oh, yes it occurs from 1.24, not 1.25.

Cargo.toml
ycho: Good question. 
Thinking again, I think we don't want to apply the change for test build, because we will build for test very often and also want those build quick.

Cargo.toml
ycho: done.

ycho:
Hmm... I did forced push after I revised the patch. I guess that hurted my 1st patch (it seems gone away!)?

Cargo.toml
ycho: done!

tdaede:
Yeah github is not good at showing previous versions of a patch set.

PR #91, opened 2018-02-21
Intra Prediction and the Dim Trait

luctrudeau:


benches/bench.rs
```diff
@@ -63,7 +63,7 @@ fn native(b: &mut Bencher) {
 
     b.iter(|| {
         for _ in 0..MAX_ITER {
-            pred_dc(&mut o2, 32, &above[..4], &left[..4]);
+        pred_dc::<Block4x4>(&mut o2, 32, &above[..4], &left[..4]);
```
barrbrain: I think we just want to drop this bench since the approach it was measuring is deleted.

lu-zero: You are right.

src/predict.rs
```diff
@@ -80,10 +80,9 @@ pub fn pred_h<D: Dim>(output: &mut [u16], stride: usize, left: &[u16]) {
   }
 }
 
-pub fn pred_v(output: &mut [u16], stride: usize, above: &[u16], bh: usize) {
-    let bw = above.len();
-    for line in output.chunks_mut(stride).take(bh) {
-        line[..bw].clone_from_slice(above)
+pub fn pred_v<D: Dim>(output: &mut [u16], stride: usize, above: &[u16]) {
+    for line in output.chunks_mut(stride).take(D::H) {
+        line[..D::W].clone_from_slice(above)
```
lu-zero: `above` would enjoy the same change.

lu-zero:
The next step would be move everything so we have all those in the trait:
``` rust
pub trait Dim {
     const W : usize;
     const H : usize;

     fn pred_dc(...) {
       // here we call Self::W and Self::H
     }
}
```

lu-zero:
https://github.com/rust-av/rav1e/commit/721a24eaf0dc9b110942caeff8119d8292463989 this is what I had in mind.

benches/bench.rs
luctrudeau: Done

src/predict.rs
luctrudeau: I'm not sure I understand this comment.

lu-zero: I was thinking about passing `&above[..D::W]`

luctrudeau: Oh I get it, or else there's a copy operation. Bench seems to indicate a slight speed up from this.

lu-zero:
I'd merge this and see if github autorebase would graciously trim https://github.com/xiph/rav1e/pull/92

PR #94, opened 2018-02-22
Part of prep works for enabling partitioning context

ycho:


src/context.rs
```diff
@@ -318,10 +318,13 @@ pub struct BlockContext {
 
 impl BlockContext {
     pub fn new(cols: usize, rows: usize) -> BlockContext {
+        // Align power of two
+        let aligned_cols = (cols + ((1 << MAX_MIB_SIZE_LOG2) - 1)) &
```
tdaede: I think it could technically be smaller than this (I think libaom allocates more than necessary) but we can always optimize it later.

PR #96, opened 2018-02-24
Use static instead of const for array

ycho:
To avoid the multiple copies of array whenever it is referenced,
also more importantly, the generated asm code seems slower
if const is used w/o '&', i.e. referencing.


tdaede:
needs rebase

ycho:
Rebase done!

PR #110, opened 2018-03-02
Add assert to p functions

ycho:
I am not so sure we need assert functions with Rust as well.
Because I see that Rust program raises flag
if index value larger than defined is ever used during runtime.
With RUST_BACKTRACE=1, debugger also shows the call stack for back trace.

I hope there is other benefit with assert, which I don't know at the moment.

### DIFF
--- a/src/plane.rs
+++ b/src/plane.rs
@@ -54,6 +54,7 @@ impl Plane {
     }
 
     pub fn p(&self, x: usize, y: usize) -> u16 {
+        assert!(y*self.cfg.stride + x < self.data.len());
         self.data[y*self.cfg.stride + x]
     }
 }
@@ -91,6 +92,7 @@ impl<'a> PlaneSlice<'a> {
     pub fn p(&self, add_x: usize, add_y: usize) -> u16 {
         let new_y = self.y + add_y;
         let new_x = self.x + add_x;
+        assert!(new_y*self.plane.cfg.stride + new_x < self.plane.data.len());
         self.plane.data[new_y*self.plane.cfg.stride + new_x]
     }
 }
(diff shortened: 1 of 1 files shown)


lu-zero:
Asserts in rust are mainly useful for refactoring/documentation and (often placebo) compiler hints. I wouldn't use them here.

ycho:
In fact, in my rust dev env, even assert is not a minimum tool for debugging crash point and seeing call stack. Once assert is met, my IDE (Eclipse) does not show any call stack. So I always adds extra if { // with breakpoint; } to see the call stack. The RUST_BACKTRACE=1 does not work to show call stack in Eclipse as well, though it shows in command line.
I believe you have dev env which can see call stack when rust program crash, w/o any extra caveat.
I assume rust support for developers with IDE is not mature, so I believe having assert will help those folks including me.

tdaede:
I think the issue is https://github.com/rust-lang/rust/issues/21102

In particular, it seems like you just need to set a breakpoint on rust_panic, then you will see the traces in your debugger.

lu-zero:
@ycho alternatively `-C panic=abort` would make the panic a normal abort.

ycho:
I see, that is the info TD-Linux was seeing after he has instructed how to possibly live w/o asserts in rust program in Eclipse or VSCode live, both were successful. He managed to use enter the command "b rust_panic" in debug command window inside IDE, as he suggested in above reply.
So... yes, lu-zero, I think I can go ahead w/o asserts.
(And I'd ask what about other pre-existing asserts? Which are not from me...)

lu-zero:
A good deal of them are either in the test code or used to document the functions input ranges in case we want to change the input types that I can see. We could switch the latter to `debug_assert!`.

ycho:
@lu-zero btw, where should I put "-C panic=abort"? Tried put it at the end of %cargo build --bin rav1e, which did not work.
Also tried adding below to Cargo.toml:
[profile.denug]
panic = "abort"

that did not work as well.

PR #113, opened 2018-03-03
Add Plane::copy_from_raw_u8().

tdaede:
This uses iterators and zip to avoid bounds checks.

### DIFF
--- a/src/plane.rs
+++ b/src/plane.rs
@@ -56,6 +56,15 @@ impl Plane {
     pub fn p(&self, x: usize, y: usize) -> u16 {
         self.data[y*self.cfg.stride + x]
     }
+
+    pub fn copy_from_raw_u8(&mut self, source: &[u8], source_stride: usize) {
+        let stride = self.cfg.stride;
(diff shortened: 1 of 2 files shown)


src/plane.rs
```diff
@@ -56,6 +56,15 @@ impl Plane {
     pub fn p(&self, x: usize, y: usize) -> u16 {
         self.data[y*self.cfg.stride + x]
     }
+
+    pub fn copy_from_raw_u8(&mut self, source: &[u8], source_stride: usize) {
+        let stride = self.cfg.stride;
+        for (self_row, source_row) in self.data.chunks_mut(stride).zip(source.chunks(source_stride)) {
+            for (self_pixel, source_pixel) in self_row.iter_mut().zip(source_row.iter()) {
```
lu-zero: https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice might be better if you can pin the line width.

lu-zero: I overlooked the line below, patch ok as is.

tdaede:
I couldn't measure an improvement timing the whole encoder.

PR #117, opened 2018-03-12
Combine left and above reset context functions

ycho:


src/context.rs
```diff
@@ -461,12 +461,28 @@ impl BlockContext {
         self.left_coeff_context[plane][bo.y_in_sb()] = uvalue;
     }
 
-    pub fn reset_left_coeff_context(&mut self, plane: usize) {
+    fn reset_left_coeff_context(&mut self, plane: usize) {
```
lu-zero: If it is not going to be used per-plane maybe we could have

``` rust
fn reset_left_coeff_context(&mut self) {
    for p in self.left_coeff_context[0..3].iter_mut() {
        for c in p.iter_mut() { 
            *c = 0;  
        } 
    } 
}
```

lu-zero: but can be done in a second time.

tdaede:
I'll land it as is.

PR #123, opened 2018-03-14
Enable partition : part 1, split down to 8x8 blocks

ycho:
This is part 1 of ongoing work of enabling partition encoding.
Every SuperBlock is partitioned down to 8x8 partitions,
where pre-existing RDO mode decision decides intra modes
(among three, DC_PRED, HORZ_PRED. VERT_PRED) for the 8x8 partition.

If input image size is not multiple of 8 pixels,
the coded image is padded to be multiple of 8 pixels in both width and height.

TEST: subset1 images.

cargo build --bin rav1e --release
rm test.ivf test_rec.y4m test_dec.y4m
./target/release/rav1e /home/yushin/sequences/subset1-y4m/125_-_Québec_-_Pont_de_Québec_de_nuit_-_Septembre_2009.y4m -o test.ivf  -r test_rec.y4m --quantizer 50
./aom_test/aomdec test.ivf -o test_dec.y4m -v
mpv --keep-open test_dec.y4m &
mpv --keep-open test_rec.y4m &

KNOWN ISSUES:
1. If SBs on right and bottom frame borders are encoded by 8x8 partiions
   and other SBs are encoded by 64x64 partition sizes, SBs start to corrupt from
   the last SB row.

TODO:
1. Fix partition down to 4x4 blocks.
2. Fix has_chroma() function.
3  Fix KNOWN ISSUE 1.
4. RDO-based block size decision

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -12,13 +12,13 @@ use plane::*;
 const PLANES: usize = 3;
 
 const PARTITION_PLOFFSET: usize = 4;
-const PARTITION_CONTEXTS: usize = 16;
-const PARTITION_TYPES: usize = 4;
+const PARTITION_CONTEXTS: usize = 20;
+pub const PARTITION_TYPES: usize = 4;
 
-const MI_SIZE_LOG2: usize = 2;
+pub const MI_SIZE_LOG2: usize = 2;
 const MI_SIZE: usize = (1 << MI_SIZE_LOG2);
 const MAX_MIB_SIZE_LOG2: usize = (MAX_SB_SIZE_LOG2 - MI_SIZE_LOG2);
-const MAX_MIB_SIZE: usize = (1 << MAX_MIB_SIZE_LOG2);
+pub const MAX_MIB_SIZE: usize = (1 << MAX_MIB_SIZE_LOG2);
 const MAX_MIB_MASK: usize = (MAX_MIB_SIZE - 1);
 
 const MAX_SB_SIZE_LOG2: usize = 6;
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -299,92 +322,208 @@ pub fn write_b(cw: &mut ContextWriter, fi: &FrameInvariants, fs: &mut FrameState
...
+                        w as usize, h as usize);
+        let r = ((cw.w.tell_frac() - tell) as f64)/8.0;
+
+        let rd = (d as f64) + lambda*r;
+        if rd < best_rd {
+            best_rd = rd;
+            best_mode = mode;
+        }
+
+        cw.rollback(checkpoint.clone());
```
lu-zero: Why the clone?

src/lib.rs
```diff
@@ -299,92 +322,208 @@ pub fn write_b(cw: &mut ContextWriter, fi: &FrameInvariants, fs: &mut FrameState
...
+            write_sb(fi, fs, cw, subsize, bo);
+            write_sb(fi, fs, cw, subsize, &BlockOffset{x: bo.x + hbs as usize, y: bo.y});
+            write_sb(fi, fs, cw, subsize, &BlockOffset{x: bo.x, y: bo.y + hbs as usize});
+            write_sb(fi, fs, cw, subsize, &BlockOffset{x: bo.x + hbs as usize, y: bo.y + hbs as usize});
+        },
+        _ => { assert!(false); },
+    }
+
+    if bsize >= BlockSize::BLOCK_8X8 &&
+       (bsize == BlockSize::BLOCK_8X8 || partition != PartitionType::PARTITION_SPLIT) {
```
lu-zero: isn't equivalent to `bsize == BlockSize::BLOCK_8X8 || (bsize > BlockSize::BLOCK_8X8 && partition != PartitionType::PARTITION_SPLIT)` ?

src/lib.rs
```diff
@@ -299,92 +322,208 @@ pub fn write_b(cw: &mut ContextWriter, fi: &FrameInvariants, fs: &mut FrameState
...
+    let partition = PartitionType::PARTITION_NONE;
+    cw.write_partition(bo, partition, bsize);
+
+    let rdo_none = rdo_mode_decision(fi, fs, cw, bsize, bo);
+
+    write_b(fi, fs, cw, rdo_none.pred_mode, bsize, bo);
+
+    let subsize = get_subsize(bsize, partition);
+
+    if bsize >= BlockSize::BLOCK_8X8 &&
```
lu-zero: Same question as above :)

src/lib.rs
```diff
@@ -275,14 +283,29 @@ fn diff_4x4(dst: &mut [i16; 16], src1: &PlaneSlice, src2: &PlaneSlice) {
...
+                       subsampling_x: usize, subsampling_y: usize) -> bool {
+    let bw = mi_size_wide[bsize as usize] as u8;
+    let bh = mi_size_high[bsize as usize] as u8;
+    let ref_pos = ((bo.x & 0x01) == 1 || (bw & 0x01) == 0 || subsampling_x == 0) &&
+                  ((bo.y & 0x01) == 1 || (bh & 0x01) == 0 || subsampling_y == 0);
+
+    ref_pos
+}
+
+// For a trasnform block,
```
lu-zero: Typo

src/context.rs
```diff
@@ -386,25 +392,30 @@ impl BlockOffset {
...
             mode: PredictionMode::DC_PRED,
+            bsize: BlockSize::BLOCK_64X64,
+            partition: PartitionType::PARTITION_NONE,
             skip: false,
         }
     }
     pub fn is_inter(&self) -> bool {
-        false
+        if self.mode >= PredictionMode::NEARESTMV { true }
+        else { false }
```
tmatth: Could be simplified as:
```
pub fn is_inter(&self) -> bool { 
    self.mode >= PredictionMode::NEARESTMV
}

File: src/lib.rs (hunk 2 above)
Comment (ycho) [reply src/lib.rs]: You are right in its absolute meaning if the meaning there is not counted.
First, the partition info is not coded for BLOCK_4X4 since it is the smallest block, then update_partition() is only considered for BLOCK_8x8 or larger, hence the first condition check it.
Another reason is, I try to follow exactly the same condition as in libaom for easier migration from existing libaom codebase.

File: src/lib.rs (hunk 3 above)
Comment (ycho) [reply src/lib.rs]: Same reason as above.

File: src/lib.rs (hunk 1 above)
Comment (ycho) [reply src/lib.rs]: I think it want to create the struct then copy the contents from previously saved states by cw.checkpoint(); above.
Btw, this is existing code, probably written by smarter or TD-Linux.
Also, I just found serious problem in our save/restore states, that we are saving "blocks: Vec<Vec<Block>>" of "struct BlockContext", which is all the MI blocks of a frame! I will file and address separately.

File: src/lib.rs (hunk 4 above)
Comment (ycho) [reply src/lib.rs]: fixed.

File: src/context.rs (hunk 5 above)
Comment (ycho) [reply src/context.rs]: Hey- I saw that and have similar change in my other branch, but the function is not part of this patch, so I believe you could write a patch separately? thx-

Comment (ycho) [reply src/context.rs]: Oh- sorry, I just realized I touched the function.
I used it as a condition so that prediction is done for each tx block if intra mode and for whole partition if inter mode, but then erased the condition it since it is out of scope of partition task for now.

So, just fixed it any way. Thx!

File: src/lib.rs [hunk 6]
```diff
@@ -299,92 +322,206 @@ pub fn write_b(cw: &mut ContextWriter, fi: &FrameInvariants, fs: &mut FrameState
...
+    }
+
+    assert!(best_rd as i64 >= 0);
+
+    let rdo_output = RDOOutput { rd_cost: best_rd as u64,
+                                pred_mode: best_mode};
+    rdo_output
+}
+
+fn write_sb(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut ContextWriter,
```
tdaede: This function needs to be renamed. It's really writing a partition, not a superblock (sb). Maybe write_partition?

src/lib.rs
```diff
@@ -299,92 +322,206 @@ pub fn write_b(cw: &mut ContextWriter, fi: &FrameInvariants, fs: &mut FrameState
...
-                if rd < best_rd {
-                    best_rd = rd;
-                    best_mode = mode;
-                }
-
-                cw.rollback(checkpoint.clone());
+            // Encode SuperBlock
+            //if sby < fi.sb_height-1 && sbx < fi.sb_width-1 {
+            if false {
+                write_sb_no_split_64x64(fi, fs, &mut cw, BlockSize::BLOCK_64X64, &bo);
```
tdaede: I think we can just delete the old write_sb_no_split_64x64 code.

src/lib.rs
ycho: We have been already using write_partition() to encode partition symbols to bitstream buffer.
Meanwhile, I think encode_partition() or write_partition() both can have dual meanings, one for partition symbols only, another for all the info in a partition, i.e. tx splits, coeffs, everything coded to decode the pixels in the partition.
Maybe 'partition_block' can better mean the pixels in the partition. So, encode_partition_block() or write_partition_block()?
One thing is for sure. Calling encode_sb() recursively as partition keep split would be avoided.

Most preferred by me is to sync with AV1 syntax spec, which uses 'decode_partition()' and 'decode_block()', so following them would at least help people dependent on AV1 spec.

So, I am trying below changes regarding function names, that I believe matches AV1 spec:
write_sb() --> encode_partition()
write_b()  --> encode_block()
write_tx_b()  --> encode_tx_block()  // Or, encode_tx_residual?

Please let me know how this look like.

src/lib.rs
ycho: Sure- I kept it there since enabling it does not work (open bug), and fixing it now,
But, I'd better not distract reviewers, so removing it.

PR #125, opened 2018-03-14
Try not to save BlockContext

ycho:
Current checkpoint/rollback saves and restores three pieces of info:
1) States of arithmetic encoder
2) CDFs
3) All of the block contexts in a frame! (Wow.. We didn't mean this...)

So, let's try whether our RDO mode decision (esp. with future
extensions) can survide w/o saving 3).

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -528,8 +528,7 @@ impl BlockContext {
 #[derive(Clone)]
 pub struct ContextWriterCheckpoint {
     pub w: ec::WriterCheckpoint,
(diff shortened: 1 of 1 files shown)


ycho:
Right, during RDO search, each mode will overwrite to those left/above contexts for coefficients and partition. The encoder works but I think mode decision is wrong due to overwritten left/above contexts.

tdaede:
Also because the affected parts of BlockContext are partition and skip, and we don't search those yet, but it'll break really soon once we add those.

Issue #131, opened 2018-03-16
Support skip mode.


mbebenita:
- This task includes:
1) Need to add reset_skip_context() if a partition is skipped.

PR #133, opened 2018-03-16
Fix skip mode

ycho:
For #131.

Add set_skip() function which sets skip flags of all MI blocks in
a partition block size.


mbebenita:
@ycho does this fix #131? If so, please close the related issue.

ycho:
I think further testing is required to close this!

PR #134, opened 2018-03-16
Update libaom to Dec 20.

tdaede:


src/context.rs
```diff
@@ -259,6 +260,52 @@ fn get_ext_tx_set(tx_size: TxSize, is_inter: bool,
...
+];
+*/
+
+pub fn y_intra_mode_to_tx_type_context(pred: PredictionMode) -> TxType {
+    intra_mode_to_tx_type_context[pred as usize]
+}
+
+pub fn uv_intra_mode_to_tx_type_context(pred: PredictionMode)-> TxType {
+    intra_mode_to_tx_type_context[pred as usize]
+    //flip me when CfL is implemented
```
ycho: Mark it as TODO (anyone): such that is easily searchable later?

tdaede: I have the other location marked as TODO so if you flip that you'll discover this one too.

tdaede:
Yes, the rtcd functions already are internally once'd. We may want our own separate mechanism for this, but for now it's required because even the _c functions do calls via rtcd pointers.

PR #139, opened 2018-03-19
Update README.md

tdaede:


ycho:
Thanks for removing frame size restrictions!

ycho:
pls merge

ycho:
One thing, do you think "Realtime 480p encoding" still valid claim?

tdaede:
added speed update

Issue #140, opened 2018-03-19
Basic speed level based on minimum partition block size


ycho:
Speed level decides the minimum partition block size, where recursive quadri-sect splitting stops at.
0(slowest) : BLOCK_4X4
1 : BLOCK_8X8
2 : BLOCK_16X16
3 : BLOCK_32X32
4~10(fastest) : BLOCK_64X64

Encoder option : "--speed=[0..10]" or "-s [0..10]".

Exception: SuperBlock(SB)s on right or bottom frame borders always split down to BLOCK_4X4.

In the future, min partition block size may be controlled by separate encoder option.

ycho:
This seems quite done.

PR #141, opened 2018-03-20
 Basic speed level based on minimum partition block size

ycho:
This patch is for issue: #140.

Speed level decides the minimum partition block size, where recursive quadri-sect splitting stops at.
0(slowest) : BLOCK_4X4
1 : BLOCK_8X8
2 : BLOCK_16X16
3 : BLOCK_32X32
4~10(fastest) : BLOCK_64X64

Encoder option : "--speed=[0..10]" or "-s [0..10]".

Exception: SuperBlock(SB)s on right or bottom frame borders always split down to BLOCK_4X4.

In the future, min partition block size may be controlled by separate encoder option.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -124,7 +124,7 @@ use rav1e::partition::*;
 use rav1e::ec;
 
 fn write_b_bench(b: &mut Bencher) {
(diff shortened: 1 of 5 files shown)


tdaede:
can you rebase this

PR #142, opened 2018-03-20
Use File::create instead of OpenOptions

est31:
It has the very same behaviour as the code it replaces,
but it is much shorter.


lu-zero:
Thank you :)

PR #146, opened 2018-03-20
Revert " Basic speed level based on minimum partition block size"

ycho:
Issue #140 
This reverts commit 6a90f9ceae2add1025b307a200ce41264bb68e58.


ycho:
We've got real fix, so abandon this...

PR #149, opened 2018-03-20
Update libaom to Dec 21.

tdaede:


ycho:
Sorry- I've still got corruptions in decode image?

ycho:
Confirm that it works here locally! Tested with build.sh and other subset1 images for speed=0, i.e. all 4x4. 
Though I found that my new patch which fixes 4x4 somehow fails with it if speed is not 0, i.e. use larger partition sizes on non-frame-border.

Anyway, I think you can land this first.

ycho:
I've tested for all available speed levels, 0,1,2,3,10, and passed for subset1 image!
Great work and thank you! Please merge.

PR #152, opened 2018-03-22
Remove left over test code

ycho:
I'm sure tdaede has just forgotten to erase this line of code for testing! ^_^


tdaede:
I actually used to have a desync because the last coeff's context is special. That might be fixed now, but you should make sure it doesn't desync at low qp.

ycho:
Wow- the issue you observed still stay there. 
Please don't merge this patch.
So, we have issue in coefficient coding of last coeff of 4x4.

PR #154, opened 2018-03-23
Move all intra prediction functions into the Intra trait.

tdaede:


ycho:
Looks fine to me, so as I confirmed in #daala IRC, this is intended to generalize the current intra prediction functions for different sizes, i.e. 8x8, 16x16, 32x32 (was there this size for intra pred?).

ycho:
Btw, can we add asserts to check whether intra prediction sizes are power of 2, say 4x4, 8x8, 16x16, 32x32.

tdaede:
It's impossible to run them any other size than 4x4 now, it only generates code for the sizes you define.

PR #155, opened 2018-03-23
Make predict function take size parameter.

tdaede:


src/partition.rs
```diff
@@ -145,23 +145,38 @@ impl PredictionMode {
...
+                        match tx_size {
+                            TxSize::TX_4X4 => Block4x4::pred_dc(slice, stride, &above[..4], &left[..4]),
+                            _ => panic!("unimplemented prediction size")
+                        },
                 }
             },
             PredictionMode::H_PRED => {
-                Block4x4::pred_h(slice, stride, &left[..4]);
+                match tx_size {
+                    TxSize::TX_4X4 => Block4x4::pred_h(slice, stride, &left[..4]),
```
ycho: I am wondering if we will put BlockNxN for each block size?
So, each block size will have sth like : pub struct Block4x4;?
Would this be necessary, I mean can we use "pub struct Block;" which can have its size when constructed, and all its traits are usable for all block sizes as well.

tdaede: Yeah, this is super verbose, I think there is a better way to do it.

ycho: Okay, then lets land this patch since the point there is pass size as parameter, then we can revise to reduce the redundancy on Block with different sizes or improve the its design.

lu-zero: Once I'm done with this weekend I'll try to come up with something a little nicer.

Keep in mind that all of this is mainly a stop-gap for when we can take the block size (a constant) as generic parameter.

tdaede: OK. In the meantime I'll land this so I can get 8x8 (and you have something to start from that's more than 1 block size)

PR #156, opened 2018-03-24
Add transform functions that take size parameter.

tdaede:


ycho:
Looks great, esp. full size function name. :)

PR #162, opened 2018-03-26
Bubble up the Intra trait

lu-zero:
Include benchmarks.

### DIFF
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -130,29 +130,34 @@ use plane::*;
 use predict::*;
 use context::*;
 
-fn setup_left(left: &mut [u16], rec: &PlaneMutSlice, height: usize) {
-    let left_slice = rec.go_left(1);
-    for i in 0..height {
-        left[i] = left_slice.p(0, i);
-    }
-}
-
 impl PredictionMode {
     pub fn predict<'a>(&self, dst: &'a mut PlaneMutSlice<'a>, tx_size: TxSize) {
-        let mut above = [127u16; 64];
-        let mut left = [129u16; 64];
+        match tx_size {
+            TxSize::TX_4X4 => self.predict_inner::<Block4x4>(dst),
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -321,31 +321,31 @@ fn has_chroma(bo: &BlockOffset, bsize: BlockSize,
...
     let stride = fs.input.planes[p].cfg.stride;
     let rec = &mut fs.rec.planes[p];
-    let tx_size = TxSize::TX_4X4;
 
     mode.predict(&mut rec.mut_slice(po), tx_size);
 
     if skip { return; }
 
-    let mut residual = [0 as i16; 16];
+    let mut residual = [0 as i16; 64*64];
```
ycho: Any way to get the size from TxSize? Otherwise, we need to change this value again when we add bigger tx later.
Same to below, coeffs_storage[], roceffs[].
I believe easiest way is just to allocate max tx size supported though.

src/lib.rs
```diff
@@ -370,24 +370,30 @@ fn encode_block(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut ContextWrite
     }
 
     let tx_type = TxType::DCT_DCT;
+    let tx_size = match bsize {
+        BlockSize::BLOCK_8X8 => TxSize::TX_8X8,
+        _ => TxSize::TX_4X4
```
ycho: What happens to the chorma block (which is 4x4) of BLOCK_8X8 luma?
Cant use 8x8 tx for 4x4 chroma!

tdaede: See let uv_tx_size later. It was broken though, so I fixed it.

src/lib.rs
tdaede: It's stack allocated so it needs to be constant size. But I could use constant values rather than magic numbers here.

tdaede:
needs rebase

PR #163, opened 2018-03-27
Fetch the quantizer only once

ycho:


tdaede:
I think this should be optimized away, but why not.

PR #165, opened 2018-03-27
All tx sizes

tdaede:

Issue #128, opened 2018-03-16
Support multiple transform sizes.


mbebenita:
- Do not support TX_SLELECT at first, but use LARGEST_TX only.

ycho:
This task is currently blocking updating libaom further, which blocks many other mandatory features including LV-MAP.

PR #166, opened 2018-03-27
Improve coeff_context() for chroma

ycho:
For issue #165, #128.
TD-Linux: This is where I meant to apply <<xdec and <<ydec for chroma coeff context writing.
I.e. only on even places. Well, but once we start to use compact storage for chroma coeff context, then this part would go away as well.

Write the chroma coeff context only on necessary places.


tdaede:
Looks good. My unconditional fill also worked but was a bit excessive.

PR #170, opened 2018-04-01
Document dependencies required to build rav1e.

tszymczak:
I decided to try out rav1e this weekend. There were dependencies that I needed to install but weren't explained in the documentation. It's a good idea to note that we need these libraries and how to install them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,11 @@ git submodule update --init
 
 This is also required everytime you switch branch or pull code and the submodule changed.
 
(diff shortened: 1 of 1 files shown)


tdaede:
Thanks! Yeah, this happens because we now inherit all libaom deps.

tszymczak:
Cool, I'm a contributor now!

tdaede:
Actually did you need to install these deps for "cargo build" to work? I'm trying to drop at least the wx dep.

tszymczak:
My memory is fuzzy now but I think so.

Issue #160, opened 2018-03-26
Add a basic inter mode


ycho:

ycho:
In the beginning, maybe w/o pred_mv, so NEW_MV mode/

ycho:
I think not done, since we think zeromv is the minimal inter mode and that has just reverted.

rzumer:
Closed by #416.

PR #172, opened 2018-04-02
Add intra inter mode coding

ycho:
For #160.

- Add inter/intra mode symbol for blocks in inter frame.
- Add intra prediction mode coding for inter frame.

### DIFF
--- a/src/bin/rav1e.rs
+++ b/src/bin/rav1e.rs
@@ -20,6 +20,9 @@ fn main() {
 
     let mut last_rec: Option<Frame> = None;
     loop {
+        if fi.number % 30 != 0 {
+            fi.ftype = FrameType::INTER;
+        }
         if !process_frame(&sequence, &fi, &mut files.output_file, &mut y4m_dec, y4m_enc.as_mut(), &mut last_rec) {
             break;
         }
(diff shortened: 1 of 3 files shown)


ycho:
Since I moved to Level Map coding, I wanted to finish what I have worked for enabling basic intra mode coding so far and committed two required patches for inter mode.

tdaede:
Cool. You need to push your libaom changes.

ycho:
tdaede: Do I have to commit libaom changes separately (I did that way before)?
If so, should I commit it to your github libaom branch only (i.e. https://github.com/tdaede/aom) ? or any branch in git?
Once libaom is commited then I need to update (commit & push -f) my rav1e commits as well, correct?

tdaede:
You can just push it to the rav1e_10 branch on my aom, but any public branch will work.

You'll need to update your rav1e commits too, but if the hash didn't change it might not see any changes - let me know and I'll manually retrigger travis.

ycho:
The existing 'show-existing-frame logic ' should behave independently, so every even odd numbered should not be encoded as before, but show previously encoded frame. 
Ex, frame 0 encoded, frame 1 (INTER frame) NOT encoded, ....

ycho:
I think better way to monitor 'show-existing-frame flag' is to output it thru eprintln!("{}", fi); in process_frame(), which shows current frame # and intra/inter frame.

tdaede:
In fact now that we have "real inter" frames we can probably stop using show-existing-frames for now. We'll need them when we do out-of-order but that's for later.

Issue #106, opened 2018-03-02
Add lv-map support.


tdaede:
Steps:
* [ ] Understand LV_MAP_MULTI (multi symbol version of Level Map) algorithm and code in libaom.
* [ ] Force partition level skip, enable lv-map, make sure image decodes. (i.e. all changes are self contained in write_coeffs)
* [ ] Force lv-map level skip, don't desync.
* [ ] Code a +/-1 DC coefficient, don't desync within a superblock.
* [ ] Code arbitrary coefficients within a superblock.
* [ ] Code all superblocks without desyncing.

LV_Map slides
https://docs.google.com/presentation/d/1nbplj5yuq2OTNQFMbymfASRsfg58GLZjvzWE3MtoZWg/edit#slide=id.g245e1ced9c_1_5

More slides (improvements)
https://docs.google.com/presentation/d/1GcCv543QV10uIehWAaoM9xjJvJkXP-MbLV_Z6GVvIq8/edit#slide=id.p

Paper
https://usercontent.irccloud-cdn.com/file/KUIqFjBi/level_map.pdf

Remove experimental flag of LV_MAP_MULTI
78f7bb98558f513ab69b7d871bf45bf119b1fd5b
https://aomedia.googlesource.com/aom/+/78f7bb98558f513ab69b7d871bf45bf119b1fd5b

Remove experimental flag of LV_MAP
cad5ebce76b50cb413b980a17748941621a747e3
https://aomedia.googlesource.com/aom/+/cad5ebce76b50cb413b980a17748941621a747e3

ycho:
https://jointdevelopment.kavi.com/higherlogic/ws/groups/aomediacodec/download/815/lv_map_multi_v2.pdf

ycho:
https://jointdevelopment.kavi.com/higherlogic/ws/groups/aomediacodec/download/815/lv_map_multi_v2.pdf

ycho:
https://jointdevelopment.kavi.com/higherlogic/ws/groups/aomediacodec/download/815/lv_map_multi_v2.pdf

ycho:
Git commit where LV_MAP_MULTI experiment flag was removed since it is default.
https://aomedia.googlesource.com/aom/+/78f7bb98558f513ab69b7d871bf45bf119b1fd5b%5E%21/#F22

PR #177, opened 2018-04-12
[NOW READY TO MERGE] Change precision in advance for the coming Level Map.

ycho:
This patch will be part of prep work for LV_MAP_MULTI, #106.

If we ever want to merge this patch, replace CONFIG_LV_MAP_MULTI as 1 in
aom_dsp/entcode.h, so that the precision of entropy coder is enforced to change.


ycho:
Again, this patch will need : 
https://github.com/tdaede/aom/pull/5/commits 
as well.

PR #179, opened 2018-04-18
Fix reconstruction on fresh builds.

tdaede:
The internal cmake build used for CDF tables is "sticky" - it doesn't
track new flags getting turned on by default. This means that changes
will only be discovered when doing a fresh build. In this case,
the EXT_SKIP flag was missed.


ycho:
LGTM, thanks for fixing this quickly(?)!

PR #181, opened 2018-04-18
update_cdf() for LV_MAP_MULTI=1

ycho:
For #106.
This patch requires corresponding change in libaom,
https://github.com/tdaede/aom/pull/7.

### DIFF
--- a/src/ec.rs
+++ b/src/ec.rs
@@ -255,24 +255,23 @@ impl Writer {
         self.enc.od_ec_encode_bool_q15(val, f)
     }
     fn update_cdf(cdf: &mut [u16], val: u32, nsymbs: usize) {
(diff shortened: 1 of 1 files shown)


tdaede:
The git submodule hash needs to be changed to de00675f5021a73257ddaf49048969c55a5c74a0, can you push an update?

ycho:
I think it should be 6844d18? Which is already there? Otherwise, Travis failed it?

ycho:
But any way let me push -f again with libaom set as 6844d18 again.

tdaede:
The problem is what when I click the "Rebase and merge" button in the aom repo on Github, it changes the public git commit hash. So the final one in the aom repo is de00675f50

ycho:
Oh, you are right.
It should point to your rav1e_11, de00675f5021a73257ddaf49048969c55a5c74a0, not mine.

ycho:
I hope this can be more easier than now....

PR #183, opened 2018-04-20
Add QM and MAX_TILE support

tdaede:


ycho:
Tested for build.sh and --speed=0,3,10 for one of subset1 image, then worked fine.
LGTM.

PR #185, opened 2018-04-23
Add support for EXT_INTRA and EXT_INTRA_MOD.

tdaede:


ycho:
So, this patch requires the updated libaom 34ff768fe, so could not build it.
I believe the functions for pred modes are added but rav1e is not yet able to use it.
LGTM.

ycho:
Ok, so H_PRED and V_PRED are now disabled.
Pls merge.

PR #190, opened 2018-04-30
Update libaom to Feb 14.

tdaede:
This adds support for CONFIG_CFL.


ycho:
Tested update13 + rav1e_13b (libaom), then passed for speed=0,1,2,3, but FAILED for speed=10 (decoded image completely broken, seems like no chroma signal).

tdaede:
OK, this is because TX64x64 is missing. I can see if I can fix it.

PR #191, opened 2018-05-01
Variable Frame Size Bit Lengths

luctrudeau:
New write_frame_size function writes out the correct number of bits for
the frame width and height.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -242,6 +242,20 @@ pub fn write_ivf_frame(output_file: &mut Write, pts: u64, data: &[u8]) {
     output_file.write(data).unwrap();
 }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -242,6 +242,20 @@ pub fn write_ivf_frame(output_file: &mut Write, pts: u64, data: &[u8]) {
     output_file.write(data).unwrap();
 }
 
+fn write_frame_size(fi: &FrameInvariants, uch: &mut BitWriter<BE>) -> Result<(), std::io::Error> {
```
lu-zero: You might put in a trait so you have `uch.write_frame_size(fi)`;

tdaede: That would be interesting, it would be a bigger refactor though

lu-zero: I can give it a go once this is merged then.

luctrudeau: Sounds good

PR #193, opened 2018-05-01
test: Always wipe the test-build dir for aomdec

lu-zero:
That should make sure that stray defaults won't get in the way (assuming it is a problem at it is for the cmake crate).


tdaede:
It is. This looks OK, though we could be a bit more clever and check if the git rev changed.

lu-zero:
https://github.com/xiph/rav1e/pull/193/files#diff-0b83f9dedf40d7356e5ca147a077acb4L33 would do but it does not right now because of a leading space.

PR #194, opened 2018-05-01
Refactor uncompressed header

lu-zero:
Some small refactoring assuming that's the direction we want to take.


luctrudeau:
Great work, we could even have a seperate file for the bitwriter (similar to bitstream.c in libaom)

tdaede:
LGTM.

Is there a less verbose way to specify the return types for these?

lu-zero:
We could define a type for it.
``` rust
type Result = Result<(), io::Error>;
```

PR #195, opened 2018-05-10
Use the right size of CDF for uv_intra_mode.

tdaede:


ycho:
LGTM, so one smaller cdf size if CDF is disabled?

tdaede:
Yeah, the last mode is CfL.

PR #197, opened 2018-05-11
Add reset_skip_context() function

ycho:
For #131 , "Support Skip Mode".

Once a partition is skipped,
this function sets left and right coefficients contexts,
which includes dc_sign and coeff_levels, as zeros.


ycho:
This patch has been locally tested in my side. For ex, two cases, skip all partitions whose y corrd is <= 128, or skip if y coord >= 128, I've tried.

Issue #198, opened 2018-05-14
Enable TXK_SEL


ycho:
Signal tx_type for each tx block, instead of each partition block.
Signal for luma plane only.

PR #199, opened 2018-05-14
Enable TXK_SEL

ycho:
For #198.


ycho:
Locally tested for one of subset1 image, for speed=0,1,2,3,10.

PR #201, opened 2018-05-18
Remove functions unused by lv_map.

tdaede:


ycho:
Okay there they were hiding!
LGTM

PR #203, opened 2018-05-19
Debug cdf

lu-zero:
I provided two separate way to debug:
- print the `w.symbol()` caller by leveraging backtrace-rs
- lookup the address of the slice on a map field -> address range and print the field name if case

The second is more invasive and a little brittle IMO.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ rand = "0.4"
 rustyline = "1"
 y4m = "0.1"
 enum-iterator-derive = "0.1.1"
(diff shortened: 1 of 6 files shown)


tdaede:
This looks good. I think the backtraces aren't as important for lockstep comparisons with libaom. The travis run needs to pass though.

lu-zero:
Done, now everything stays hidden until you set the env variable.

tdaede:
Can you squash the last three commits together?

Issue #44, opened 2018-02-06
Add partition splitting to RDO


tmatth:
Currently in `encode_tile` for every superblock, we search for the intra mode with the best RD (based on a 64x64 SSE), for PARTITION_NONE. We should extend this search to one-level of split.

tmatth:
e.g., find the best combination of intra modes for the partitioned blocks and compare that to the PARTITION_NONE case, and pick the winner.

tmatth:
Reference: http://av1-spec.argondesign.com/av1-spec/av1-spec.html#decode-partition-syntax

negge:
The reference link has changed: http://av1-spec.argondesign.com/av1-spec/av1-spec.html#decode-partition-semantics

rzumer:
This is mostly done as of #255. There is still no RDO decision made on frame edges, and not all partition types are supported, but these issues are covered by #226, #227, #258, and #175.

PR #207, opened 2018-05-22
Add RDO decision for partition splitting up to 8x8 blocks

rzumer:
Start of work on issue #44.
Partitions with block sizes above 8x8 (achieved by raising the value at lib.rs:708) will cause desynchronization.


tdaede:
This needs to rebased.

rzumer:
@tdaede Done, sorry for the messy history.

rzumer:
Done and added comparisons between default and best-performing speed parameters.

src/ec.rs
```diff
@@ -86,7 +86,6 @@ impl od_ec_enc {
         let u: u32;
         let v: u32;
         assert!(32768 <= r);
-
```
ycho: Pls remove unnecessary changes.

src/ec.rs
```diff
@@ -299,6 +298,7 @@ impl Writer {
             });
         }
         self.cdf(s, &cdf[..nsymbs]);
+
```
ycho: same here.

src/lib.rs
```diff
@@ -605,14 +605,15 @@ fn encode_block(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut ContextWrite
 fn rdo_mode_decision(fi: &FrameInvariants, fs: &mut FrameState,
                   cw: &mut ContextWriter,
                   bsize: BlockSize, bo: &BlockOffset) -> RDOOutput {
+
```
ycho: and here.

src/lib.rs
```diff
@@ -621,6 +622,7 @@ fn rdo_mode_decision(fi: &FrameInvariants, fs: &mut FrameState,
         if fi.frame_type == FrameType::KEY && mode >= PredictionMode::NEARESTMV {
           break;
         }
+
```
ycho: and here?

src/lib.rs
```diff
@@ -640,33 +642,198 @@ fn rdo_mode_decision(fi: &FrameInvariants, fs: &mut FrameState,
...
+    
     let rdo_output = RDOOutput { rd_cost: best_rd as u64,
-                                pred_mode: best_mode};
+                                part_type: best_partition,
+                                part_modes: best_pred_modes };
+
     rdo_output
 }
 
 fn encode_partition(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut ContextWriter,
```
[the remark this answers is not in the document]
rzumer: The original logic remains in `encode_partition`. I believe it may be better to handle the case of fixed partition sizes through context parameters or arguments. The block at lines 827-834 controls when a decision is allowed to be made.

rzumer:
I have re-squashed with fixes to @ycho's comments and some more improvements to reduce duplicated code and remove RD cost computations from `rdo_partition_decision` for a performance and negligible precision improvement.

src/lib.rs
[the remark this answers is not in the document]
ycho: Agree, for the option of fixed size partition, we can do it later in the separate commit.
Probably by putting, if fi.used_fixed_size_partition block in line 770, where if true then skip rdo_partition_decision() and set partition = PARTITION_SPLIT.

src/lib.rs
```diff
@@ -680,26 +787,44 @@ fn encode_partition(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut ContextW
...
     match partition {
         PartitionType::PARTITION_NONE => {
-            // TODO(anyone): Until we have RDO-based block size decision,
-            // call rdo_mode_decision() for a partition.
-            let rdo_none = rdo_mode_decision(fi, fs, cw, bsize, bo);
-            cw.bc.set_mode(bo, bsize, rdo_none.pred_mode);
+            let pred_mode = if rdo_output.part_modes.len() > 0 { 
+                rdo_output.part_modes[0].pred_mode
+            } else { // edges
+                rdo_mode_decision(fi, fs, cw, bsize, bo).part_modes[0].pred_mode };
```
ycho: Can you move the } to new line?

tdaede:
LGTM with ycho's comments addressed. Make sure to "squash and merge".

ycho:
LGTM with my recommended manual rust formatting!

src/lib.rs
```diff
@@ -680,26 +787,44 @@ fn encode_partition(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut ContextW
...
+
+                    encode_partition(fi, fs, cw, subsize, &offset, 
+                        &Some(RDOOutput { 
+                            rd_cost: mode.rd_cost, 
+                            part_type: PartitionType::PARTITION_NONE, 
+                            part_modes: vec![mode] }));
+                }
+            }
+            else {
+                encode_partition(fi, fs, cw, subsize, bo, &None);
```
ycho: As record or comments for readers, this block of splits is done when input partition has not come from rdo_partition_decision(), i.e. 64x64 -> 32->32 or all unconditional 4x4s on right/bottom frame border.

src/lib.rs
ycho: As a record or comments for readers, this block of no-split partition comes from the case which does not go through rdo_partition_decision(). For ex, unconditionally split 32x32 partition, or all unconditional 4x4s on right/bottom frame border.

rzumer: I have added more detailed comments in this function to clarify how the partitioning decisions are made.

src/lib.rs
rzumer: This should be addressed with the new comments added to this function.

src/lib.rs
rzumer: This line break was added for readability.

Issue #107, opened 2018-03-02
Read 10-bit y4m input.


tdaede:
This requires mostly adding support to the y4m crate:

https://github.com/PistonDevelopers/y4m

tdaede:
This is done. We don't code 10 bit but that's for later.

PR #213, opened 2018-05-30
Read 10-bit y4m input. #107

xiphmont:
Note this cannot yet encode high bitdepth; it reads it and then shifts
to 8 before encode.


src/plane.rs
```diff
@@ -57,11 +57,29 @@ impl Plane {
...
-            for (self_pixel, source_pixel) in self_row.iter_mut().zip(source_row.iter()) {
-                *self_pixel = *source_pixel as u16;
+            match source_bytewidth {
+                1 =>
+                    for (self_pixel, source_pixel) in self_row.iter_mut().zip(source_row.iter()) {
+                        *self_pixel = *source_pixel as u16;
+                    },
+                2 => {
+                    let mut source_iter = source_row.iter();
+                    for self_pixel in self_row.iter_mut() {
```
lu-zero: Maybe something like

``` rust
for (self_pixels, bytes) in self_row.iter_mut().zip(source_iter.chunks(2)) {
    *self_pixel = (bytes[0] as u16) << 8 | (bytes[1] as u16);
}
```
is nicer

xiphmont: yes indeed; updating

PR #214, opened 2018-05-30
Read 10-bit y4m input. #107

xiphmont:
Note this cannot yet encode high bitdepth; it reads it and then shifts
to 8 before encode.  Updated with [corrected and tested] version of lu_zero's pattern.


lu-zero:
Looks fine for me.

Issue #124, opened 2018-03-14
Should not save/restore all the block contexts in frame!


ycho:
Found serious problem in our save/restore (i.e. checkpoint/rollback) states, that we are saving "blocks: Vec<Vec>" of "struct BlockContext", which means all the MI blocks of a frame!

rzumer:
Fixed in #250.

PR #215, opened 2018-05-31
Should not save/restore all the block contexts in frame! #124

xiphmont:


lu-zero:
I wonder if that helps with @rzumer desync issues.

tdaede:
Please rebase.

tdaede:
I cherry-picked just the first commit and it speeds up the encoder by 2x with the default speed settings, and doesn't seem to cause any issues.

negge:
Do we have any performance numbers for this patch?

tdaede:
Replaced by #250

Issue #189, opened 2018-04-30
reorder intra enums


tdaede:
we need to implement libaom 866da798e0ef59be3f0d20835603765881db69ad
it's not visible right now because we only code DC_PRED

tdaede:
done

PR #218, opened 2018-06-01
In inter frame, only write intra mode if block is intra

xiphmont:
While looking at #189, I found that inter blocks in inter frames are
still writing an intra mode.  This doesn't get tripped yet as we're
only writing intra frames, but it will soon enough.


xiphmont:
Oops, there's more screwup there.  Let me look at this harder.

tdaede:
Can you squash? I can do it too but want to make sure I get the right thing.

PR #221, opened 2018-06-05
Add missing Paeth predictor test

rzumer:
This tests the Paeth predictor as part of the `pred_max()` function.

Some indentation and naming consistency fixes in `predict.rs` and `bench.rs` were added as well. These can be moved to a different PR if preferred.


tdaede:
Aha, that's important! It's okay to have the changes together for now.

PR #223, opened 2018-06-06
Add smooth predictors and tests

rzumer:
Smooth predictors require knowledge of bit depth to clamp predictions, so the native function includes the `bd` parameter.


rzumer:
Now with H and V variants.

PR #234, opened 2018-06-12
build.rs: watch for changes the full aom/ folder

### LINKED ISSUE #233: A change in the `aom_build/aom` folder may not trigger a rebuild
Perhaps one problem is that in `build.rs` the path watched for changes is `"cargo:rerun-if-changed=aom_build/aom/build"` instead of the root of the submodule.

PR coming.

naufraghi:
The aom/build folder watched before may not change for all changes.

Closes #233

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -67,7 +67,7 @@ fn main() {
     // Manually fix the comment so rustdoc won't try to pick them
     format_write(builder, "tests/aom.rs");
 
-    println!("cargo:rerun-if-changed=aom_build/aom/build");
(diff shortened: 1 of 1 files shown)


naufraghi:
@lu-zero, here is the first patch we were testing yesterday.

lu-zero:
Sounds good and thank you :)

Removing the build dir if present before running cmake might be a good idea, not sure if would make sense to add this option in the cmake crate instead of rav1e though.

PR #235, opened 2018-06-12
Only allocate coeff buffers up to 32x32.

tdaede:
64x64 transforms only need 32x32 coefficients at most.
This speeds up the encoder some due to not zeroing the memory.


ycho:
Right, same change is in my PR for enabling 64x64 tx. So, merge conflict will come to me!

PR #236, opened 2018-06-12
context: fix non-reference pattern used to match a reference

tmatth:
Fixes:
```
error[E0658]: non-reference pattern used to match a reference (see issue #42640)
    --> src/context.rs:1590:13
     |
1590 |         for (name, start, end) in &self.map {
     |             ^^^^^^^^^^^^^^^^^^ help: consider using a reference: `&(name, start, end)`
```

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1587,9 +1587,9 @@ struct FieldMap {
 impl FieldMap {
     /// Print the field the address belong to
     fn lookup(&self, addr: usize) {
-        for (name, start, end) in &self.map {
+        for &(name, start, end) in &self.map {
             // eprintln!("{} {} {} val {}", name, start, end, addr);
-            if addr >= *start && addr < *end {
+            if addr >= start && addr < end {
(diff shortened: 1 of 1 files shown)


tmatth:
Note: this was with cargo 0.26.0-nightly (1d6dfea44 2018-01-26)

tmatth:
Refs https://github.com/rust-lang/rust/issues/42640

rzumer:
Is this necessary? Latest stable rustc seemed to work.

tmatth:
@rzumer failed on the above version of nightly.

rzumer:
Yes, but this is an unsupported version, stable 0.26 should compile. I am not against fixing compilation on older versions of rustc if readability does not suffer, however.

tmatth:
@rzumer yeah I don't have strong opinions about it either.

tmatth:
Closing as we can assume that it's unlikely for people to hit this unless they are on old nightly.

Issue #178, opened 2018-04-17
cmake-building is unreliable


tdaede:
- [ ] a manual cargo clean is needed when the libaom code gets updated
- [ ] on some platforms there is no way to avoid the wxwindows dependency
- [ ] the encoding-decoding test is not present yet

Reverting #99 would be an improvement.

lu-zero:
I'd rather have a feature so who wants to just build the encoder won't have to deal with the additional overhead.

tdaede:
That's fine (we need some way to get it for the encoding-decoding test anyway)

tdaede:
I infodumped into this etherpad: https://public.etherpad-mozilla.org/p/rav1e-cmake

tdaede:
I think this is OK now.

PR #237, opened 2018-06-12
build.rs: remove CMakeCache.txt just after the build

naufraghi:
This may fix the first problem in issue #178, deleting the cache we
are forcing a libaom rebuild if the `CMakeLists.txt` is updated.

To be tested, because I'm not testing the build output but only the build times :)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,7 +8,7 @@ extern crate cc;
 use std::env;
 use std::path::Path;
 
-use std::fs::File;
+use std::fs::{self, File};
 use std::io::Write;
 
 fn format_write(builder: bindgen::Builder, output: &str) {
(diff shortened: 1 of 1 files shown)


naufraghi:
The old `cfg` variable was a bit miss-named, because `Config::new(...).build() -> PathBuf` and that path is the `out_dir` one can provide as argument or `OUT_DIR` from the environment.

Given that path it's easy to delete the cache file, I'm not sure what that imply exactly, but I measured:

Before:
- ~5s of build time changing an option value in `aom/CMakeLists.txt`,
- ~2s for a trivial change in `src/lib.rs`
- ~5s for a trivial change in `aom/args.c`

After:
- ~10s of build time changing an option value in `aom/CMakeLists.txt`,
- ~2s for a trivial change in `src/lib.rs`
- ~10s for a trivial change in `aom/args.c`

So I think the patch is triggering a full rebuild, and it is quite time consuming 2x build times.

I've a set of scripts in the lines o the https://github.com/github/scripts-to-rule-them-all, a good enough solution may be:

1. put the cache removal under an env var
2. let the `script/setup` call `env RAV1E_RESET_AOM_CACHE=1 cargo build`

So the the usual `cargo build` will be as fast as before, but there is an easy script to call if something strange happens.

PR #241, opened 2018-06-13
[wip] Play with Cow<>s to reduce the checkpoint impact 

lu-zero:
This way restoring the CDFContext is just a matter of dropping + overwriting pointers.

TODO:
- [ ] benchmark
- [ ] redo in a different way the tracing system Cow<>s make impossible to walk around the memory.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -767,9 +767,9 @@ static av1_nz_map_ctx_offset_32x64: [i8; 1024] = [
   21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21,
   21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21,
   21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21,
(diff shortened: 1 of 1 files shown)


tdaede:
innnnnnnnnnnnteresting. The first thing to keep in mind is that this may show bigger gains once our rollback does not involve a copy. Secondly, the CoW chunks are pretty large with the way symbols are split out right now.

lu-zero:
This is more or less a first working step before I make so we make `checkpoint()` return a normal context so we can save it, IF this first step works as intended, the same Cow approach can be used for the other fields so in the end the rdo inner loop would be more or less:

``` rust
for (mode, block, ...) in rdo_choices_for_partition(partition) {
    let wcw = cw.checkpoint(); // glorified clone() all the Cows are just a fat pointer copy
    wcw.compute_rd_stats(mode, block, ...); // we spend time allocating here only for what we change
    if wcw.rd_stats.better(best_candidates) {
        best_candidates.push(wcw); // we can keep those around or be greedy and keep only one
    }
}

for wcw in best_candidates {
    // more computation
}
...

cw.update(wcw); // assuming the rd estimation is a partial encoding and not a full encoding step
```

Assuming I'm not too tired and I'm overlooking something quite important.

tdaede:
Going to close this as we don't use checkpoints anymore.

PR #250, opened 2018-06-14
Only store the left and above contexts.

negge:


tdaede:
Looks good. I think if we converted the above and left contexts to a full frame rather than just the narrow strips, we could avoid copying them too (and the cost of additional ram)

PR #257, opened 2018-06-15
Set proper size for intra reference sample buffers

fbossen:
Increase size for intra reference sample buffers to include
above-left, above-right and below-left samples. Make size a
function of the maximum supported transform size.


tdaede:
needs rebase

PR #259, opened 2018-06-15
Panic if 420 is not used.

mbebenita:


ycho:
LGTM, thanks for fix!

lu-zero:
You might want to use [`unimplemented!()`](https://doc.rust-lang.org/std/macro.unimplemented.html)

PR #260, opened 2018-06-15
Drop some unneeded clones

lu-zero:
~ 10% faster encoding for -s 0 on nyan.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2157,7 +2157,7 @@ impl ContextWriter {
         }
     }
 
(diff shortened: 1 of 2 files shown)


rzumer:
Great, there are probably more that can be removed from the partition and mode RDO decisions.

ycho:
NIce- I have been thinking when .clone() is required and when not.

lu-zero:
I started profiling a little rav1e and looks like we are spending LOTS of time allocating and copying data.

PR #264, opened 2018-06-16
Do not initialize the scratch buffers in encode_tx_block

### LINKED ISSUE #242: Skip zero initialization of buffers in encode_tx_block
in particular, residual, coeffs, and rcoeffs

lu-zero:
We allocate the largest possible size of those buffers and
work only on a subset of it.

Fix #242.


ycho:
LGTM, Thanks for improving this!

PR #266, opened 2018-06-18
Update README.md

rzumer:
List new supported prediction modes and describe the current design of rdo.rs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,10 @@ Because AV1 is not yet frozen, it relies on an exact decoder version and configu
 
 * Intra frames
 * 64x64 superblocks
+* 4x4 to 32x32 RDO-selected square blocks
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ Because AV1 is not yet frozen, it relies on an exact decoder version and configu
 
 * Intra frames
 * 64x64 superblocks
-* H, V, and DC prediction modes
+* DC, H, V, Paeth, and smooth prediction modes
 * 4x4 DCT and ADST transforms
```
tdaede: You could update this to include the variable partition and transform sizes, too.

rzumer: I made some changes, please let me know if this is enough. Regarding transform sizes, have they actually changed? I'm not familiar with how they are picked currently, and I don't think I touched them.

Issue #246, opened 2018-06-14
Enable OBU Header for AV1 v1.0 aomdec


ycho:
Can be part of #103 but aiming to comply current master of libaom,
so that rav1e bitstream can be decodable by current version of aomdec.

rzumer:
Closed by #341.

PR #268, opened 2018-06-20
[WIP] Enable obu

ycho:
#246

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -338,6 +338,7 @@ pub enum ReferenceMode {
 }
 
 const REF_FRAMES: u32 = 8;
+const REF_FRAMES_LOG2: u32 = 3;
 
 /*const NONE_FRAME: isize = -1;
 const INTRA_FRAME: usize = 0;*/
@@ -749,10 +750,10 @@ impl<'a> UncompressedHeader for BitWriter<'a, BE> {
         }
       }
 
-      self.write_bit(fi.disable_cdf_update);
+      self.write_bit(fi.disable_cdf_update)?;
 
       if seq.force_screen_content_tools == 2 {
-        self.write_bit(fi.allow_screen_content_tools != 0);
+        self.write_bit(fi.allow_screen_content_tools != 0)?;
       } else {
         assert!(fi.allow_screen_content_tools ==
                 seq.force_screen_content_tools);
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -65,14 +65,111 @@ impl Frame {
...
 }
 
 impl Sequence {
-    pub fn new() -> Sequence {
+    pub fn new(width: usize, height: usize) -> Sequence {
+        let mut operating_point_idc = [0 as u16; MAX_NUM_OPERATING_POINTS];
+        let mut level = [[1, 2 as usize]; MAX_NUM_OPERATING_POINTS];
+        let mut tier = [0 as usize; MAX_NUM_OPERATING_POINTS];
+
+        for i in 0..MAX_NUM_OPERATING_POINTS {
```
lu-zero: Why setting them here again if you already set them just right above?

src/lib.rs
```diff
@@ -366,6 +474,57 @@ impl<'a> UncompressedHeader for BitWriter<'a, BE> {
...
+    fn write_sequence_header_obu(&mut self, seq: &Sequence, fi: &FrameInvariants)
+        -> Result<(usize), std::io::Error> {
+        let mut size: usize = 0;
+
+        self.write(3, seq.profile)?; // profile 0, 3 bits
+        self.write(1, 0)?; // still_picture
+        self.write(1, 0)?; // reduced_still_picture
+
+        if false {	// if reduced_still_picture == true
+            assert!(false);
```
lu-zero: what about 
``` rust
panic!("some informative message")
```
or using 
``` rust 
unimplemented!("some informative message")
```

src/lib.rs
ycho: Yeap- right. Some members of Sequence are set when it is constructed and others are set right before it is written to bitstream, i.e. after frame is finished for encoding.
Then, I haven't really decided where to init them, so duplicated. Thanks for catching them.

src/lib.rs
```diff
@@ -758,45 +758,75 @@ fn aom_uleb_size_in_bytes(mut value: u64) -> usize {
...
-        let obu_payload_size_uleb = aom_uleb_size_in_bytes(obu_payload_size);
-
-        /*const size_t length_field_size =
-            obu_memmove(obu_header_size, obu_payload_size, data);
-        if (write_uleb_obu_size(obu_header_size, obu_payload_size, data) !=
-            AOM_CODEC_OK) {
-        			return AOM_CODEC_ERROR;
-        }*/
+        packet.write(&buf2).unwrap();
+        buf2.clear();
```
ycho: @lu-zero : I'd like this checked by you. So, if I call 
packet.write(&buf1).unwrap(); and packet.write(&buf2).unwrap(); in order,
bu1 and bu2 is written to packet buffer in order, correct?

lu-zero: I think so.

ycho:
Fix already submitted by mbx.

tdaede:
Superseded by #341.

Issue #267, opened 2018-06-18
Transform type RDO


ycho:
Must be coded for all but intra mode chroma blocks.

rzumer:
As of #279 this should be complete for intra mode.

rzumer:
Closing this for now as we are done with this issue unless further work is required for inter mode.

PR #273, opened 2018-06-21
Clean up and refactor RDO-related areas

rzumer:
* Force a `cargo clean` when the submodule is out of date, as it can be necessary to link correctly  
* Remove assertion to support additional (non-DCT) transform types (as part of #267)  
* Rename some variables, simplify some operations
* Move `RAV1E_PARTITION_TYPES` to `partition.rs` (from `predict.rs`)
* Move RDO decision functions to `rdo.rs`
* Extract rate-distortion cost computation to a new function

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,9 +41,7 @@ use partition::*;
 use transform::*;
 use quantize::*;
 use plane::*;
-use predict::*;
 use rdo::*;
-use ec::*;
 use std::fmt;
 
 extern {
@@ -574,6 +572,7 @@ fn encode_block(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut ContextWrite
             }
         }
     }
+
     // these are only valid for 4:2:0
     let uv_tx_size = match bsize {
         BlockSize::BLOCK_4X4 | BlockSize::BLOCK_8X8 => TxSize::TX_4X4,
@@ -581,6 +580,7 @@ fn encode_block(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut ContextWrite
         BlockSize::BLOCK_32X32 => TxSize::TX_16X16,
         _ => TxSize::TX_32X32
     };
+
     let mut bw_uv = bw*tx_size.width_mi() >> xdec;
     let mut bh_uv = bh*tx_size.height_mi() >> ydec;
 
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -781,12 +627,10 @@ bsize: BlockSize, bo: &BlockOffset) -> f64 {
...
 
     let bs = mi_size_wide[bsize as usize];
-    let is_too_big =
-        bo.x as i32 + bs as i32 - fi.w_in_b as i32 > 0 ||
-        bo.y as i32 + bs as i32 - fi.h_in_b as i32 > 0;
+    let must_split = bo.x + bs as usize > fi.w_in_b || bo.y + bs as usize > fi.h_in_b;
 
     let is_splittable = bsize > fi.min_partition_size;
-    let is_codable = bsize < BlockSize::BLOCK_64X64 && !is_too_big;
+    let is_codable_unsplit = bsize < BlockSize::BLOCK_64X64 && !must_split;
```
tdaede: Is is_codable_unsplit the exact opposite of must_split? It would be nice to be consistent here if you're renaming.

rzumer: No, it just means that the partition is codable as a leaf node (e.g. as `PARTITION_NONE`).

ycho: I think the term "bsize < BlockSize::BLOCK_64X64" is required only because tx64x64 is not working now.
(Which, I know from my wip patch for enabling tx64x64)
So, if we add that term to line 630, like
let must_split = bo.x + bs as usize > fi.w_in_b || bo.y + bs as usize > fi.h_in_b || bsize > BlockSize::BLOCK_32X32
then
line 633 simply becomes :  let is_codable_unsplit = !must_split;
In that sense, what tdaede inquired can be true. I.e if it is not 'must split' (for whatever reason) then it 'is_codable' == true.
So, for two reasons, 1) future integration with "enabling tx64x64" and 
2) "bsize < BlockSize::BLOCK_64X64" is better placed in must_split = ...
i.e. if no_split (i.e. must_split == false), then always is_codable == true.
I'd suggest:
must_split = bo.x + bs as usize > fi.w_in_b || bo.y + bs as usize > fi.h_in_b || bsize > BlockSize::BLOCK_32X32;
is_codable == !must_split;

src/rdo.rs
```diff
@@ -42,3 +49,182 @@ pub fn sse_wxh(src1: &PlaneSlice, src2: &PlaneSlice, w: usize, h: usize) -> u64
     }
     sse
 }
+
+// Compute the rate-distortion cost for an encode
+fn compute_rd_cost(fi: &FrameInvariants, fs: &FrameState,
```
ycho: Thanks for moving this function to rdo.rs!

src/lib.rs
rzumer: This was already addressed in the last commit.

src/lib.rs
```diff
@@ -627,10 +627,14 @@ bsize: BlockSize, bo: &BlockOffset) -> f64 {
...
 
     let bs = mi_size_wide[bsize as usize];
-    let must_split = bo.x + bs as usize > fi.w_in_b || bo.y + bs as usize > fi.h_in_b;
 
-    let is_splittable = bsize > fi.min_partition_size;
-    let is_codable_unsplit = bsize < BlockSize::BLOCK_64X64 && !must_split;
+    // Always split if the current partition is too large
+    let must_split = bo.x + bs as usize > fi.w_in_b ||
+        bo.y + bs as usize > fi.h_in_b ||
+        bsize >= BlockSize::BLOCK_64X64;
```
ycho: Oh, yes.. the author already though what I thought and suggested! Great!

src/lib.rs
```diff
@@ -627,10 +627,14 @@ bsize: BlockSize, bo: &BlockOffset) -> f64 {
...
+
+    // must_split overrides the minimum partition size when applicable
+    let can_split = bsize > fi.min_partition_size || must_split;
```
ycho: And even this exact name 'can_split', which I was thinking it might be better than is_splittable!

ycho:
LGTM!

ycho:
In similar spirit, I guess you might consider write a separate patch that does:
adding condition "|| bsize >= BlockSize::BLOCK_64X64 " to the end of "must_split = ... "  in encode_partition_topdown()

rzumer:
@ycho Good point.

PR #276, opened 2018-06-21
Add support for coding global motion vectors

atomnuker:


src/ec.rs
```diff
@@ -347,6 +350,71 @@ impl Writer {
...
+        }
+    }
+    fn write_refsubexpfin(&mut self, n: u16, k: u16, r: i16, v: i16) -> Result<(), std::io::Error> {
+        let recentered_v = self.recenter_finite_nonneg(n, r as u16, v as u16);
+        return self.write_subexpfin(n, k, recentered_v);
+    }
+    fn write_s_refsubexpfin(&mut self, n: u16, k: u16, r: i16, v: i16) -> Result<(), std::io::Error> {
+        return self.write_refsubexpfin((n << 1) - 1, k, r + (n - 1) as i16, v + (n - 1) as i16);
+    }
+}
```
rzumer: Can you please add some comments, references, and/or more descriptive variable names to uncommented functions? For someone unfamiliar with the writer's behavior and abbreviations used in the function names it is difficult to parse, so at least summarizing the purpose of each function would be helpful.

src/lib.rs
```diff
@@ -440,9 +445,34 @@ fn write_uncompressed_header(packet: &mut Write, sequence: &Sequence,
...
+                    let bits = 12 - 6 + 3 - !fi.allow_high_precision_mv as u8;
+                    let bits_diff = 12 - 3 + fi.allow_high_precision_mv as u8;
+                    BCodeWriter::write_s_refsubexpfin(&mut uch, (1 << bits) + 1,
+                                                      3, mv_x_ref >> bits_diff,
+                                                      mv_x >> bits_diff)?;
+                    BCodeWriter::write_s_refsubexpfin(&mut uch, (1 << bits) + 1,
+                                                      3, mv_y_ref >> bits_diff,
+                                                      mv_y >> bits_diff)?;
+                }
+                GlobalMVMode::ROTZOOM => { /* TODO: Add support */ }
```
rzumer: Please call `unimplemented!();` for `ROTZOOM` and `AFFINE`.

ycho:
Was this PR tested other ways than Travis, to merge?

atomnuker:
Yes, I tested coding arbitrary MVs and reading them out in the decoder, everything worked fine except them doing anything. I'm still trying to figure out why. But the actual coding of values for any number of reference frames works.

ycho:
atomnuker: can you explain further what you meant with "them doing anything" please?
I believe MC (motion compensation) part is not implemented yet.

atomnuker:
They don't do anything. I can't explain it any further than this. If the encoder encodes a global motion vector the decoder should essentially shift all pixels by the offset coded, but nothing happens despite the decoder getting the values okay. I think its due to the fact that we always enable ER which afaik means the decoder tries to mark most data from previous frames as corrupt and won't use it.

tdaede:
No, the global motion doesn't affect all pixels coded, but rather affects the motion vector value provided by the GLOBALMV inter prediction mode. Unless you have a patch somewhere that is coding GLOBALMV mode, this won't do anything.

PR #284, opened 2018-06-28
Restrict scope of RDO above low speed levels

rzumer:
Transform types and prediction modes are currently all tested at all speed levels, which slows down every speed level considerably. With this change, speed levels 0-3 will test all prediction modes and transform types exhaustively, while higher levels will default to a smaller set of prediction modes and DCT transforms.


rzumer:
Encoding time (Q80): 1.205 -> 0.523

tdaede:
Please rebase.

PR #286, opened 2018-07-05
Factor out alignment math.

mbebenita:


src/lib.rs
```diff
@@ -136,12 +161,12 @@ impl FrameInvariants {
...
             height: height,
-            padded_w: ((width+7)>>3)<<3,
-            padded_h: ((height+7)>>3)<<3,
-            sb_width: (width+63)/64,
-            sb_height: (height+63)/64,
-            w_in_b: 2 * ((width+7)>>3) ,	// MiCols, ((width+7)/8)<<3 >> MI_SIZE_LOG2
-            h_in_b: 2 * ((height+7)>>3),	// MiRows, ((height+7)/8)<<3 >> MI_SIZE_LOG2
+            padded_w: width.align_power_of_two(3),
+            padded_h: height.align_power_of_two(3),
+            sb_width: width.align_power_of_two_and_shift(6),
```
tdaede: LGTM. Not sure if the _and_shift variants are much clearer than just shifting though...

PR #289, opened 2018-07-06
Fix formatting issues for a subset of the files.

mbebenita:
This PR incrementally re-formats some of the rust source files. I picked rust formatting rules and options in a way that minimizes the number of changes in this PR.

I've did not format `lib.rs` in this PR in order to avoid conflicts with other open PRs.

Files that pass `cargo +nightly fmt -- --check` are commented out in the `ignore` section of the `rustfmt.toml` file. Files that have been mostly formatted are ignored, but marked with a `# Partial` comment.


rzumer:
How does `cargo fmt` output differ between current stable and nightly toolchains? I would like to stick with stable if possible, and it is not possible to install the nightly `rustfmt` without the entire nightly toolchain.

mbebenita:
> differ between current stable and nightly toolchains?

Unfortunately, the ignore option in `rustfmt.toml` is only available in nightly. Not sure yet how to get things running under stable.

PR #293, opened 2018-07-11
Update to Feb 14.

tdaede:


ycho:
Looks like https://gitlab.xiph.org/xiph/aom-rav1e/tree/rav1e_14 does not exist.
Please create rav1e_14 and push your change there.

PR #296, opened 2018-07-13
Enable code formatting, with 2 space tabs.

mbebenita:


ycho:
Okay... the history gone with the 2 tabs! [LGTM]

mbebenita:
> history gone with the 2 tabs

unfortunate

robinst:
Hey, landed here from [this Reddit thread](https://www.reddit.com/r/rust/comments/8z5y3f/xiphrav1e_the_fastest_and_safest_av1_encoder) and excited about this project!

Out of curiosity, what was the reason for choosing 2 space indent?

The majority of Rust code out there use 4 spaces and it's also the standard style, see here: https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/guide.md#indentation-and-line-width

ycho:
Hi,

I think it is simply because some developers started feeling pressure of lack of columns, as nesting gets deeper.
Thanks for showing your interests in rav1e project!

Cheers,
Yushin

tdaede:
The more important part of this commit was at least cleaning up the files that had a mix of 2 indents. I would be ok switching back to 4 if it's more comfortable. But yeah, a lot of codec code is 2 space.

PR #297, opened 2018-07-13
Fix most Clippy warnings.

mbebenita:


src/context.rs
```diff
@@ -1000,9 +1003,7 @@ impl CDFContext {
       coeff_base_eob_cdf: av1_default_coeff_base_eob_multi,
       coeff_base_cdf: av1_default_coeff_base_multi,
       coeff_br_cdf: av1_default_coeff_lps_multi
-    };
-
-    c
```
ycho: oh you mean this... by removing the return.

ycho:
LGTM for the patch "Fix Clippy warnings.", btw I see that it has lots of changes beyond clippy warnings? Thx for so many fixes!

src/context.rs
mbebenita: Yes, there's no reason to create a new bindings.

PR #304, opened 2018-07-16
Use CMake-built objects for encoder.

tdaede:
Previously we built encoder objects directly in build.rs with cc,
but now build them with CMake just like the common and decoder
objects. This also removes the duplicate aom_build/*.h files.


lu-zero:
@rzumer might still need that for the `rdo_ml` branch though.

tdaede:
We can always bring libntr.a back. I don't think the ml stuff needs the aom_config.h etc. If it does we can bring those back too.

PR #306, opened 2018-07-16
Fix README.md command line.

tdaede:


ycho:
I remember that  CONFIG_KF_CTX is turned on by default when lv_map is commited and
probably same for CONFIG_INTRA_EDGE2?

ycho:
LGTM

PR #309, opened 2018-07-16
Use release mode in example instructions

steveklabnik:
This makes a *huge* difference in runtime performance, so you probably want to use it.

Then again, maybe not! No worries if you don't want this PR  :)


tdaede:
Indeed! Our test scripts use it.

PR #316, opened 2018-07-16
Mention cmake requirement in README.md

jmvalin:


tdaede:
Intentionally merging both commits.

PR #333, opened 2018-07-16
Use target_feature to fallback to C when sse2 is not available

lu-zero:


src/transform.rs
```diff
@@ -18,22 +18,49 @@ use partition::TxType;
...
+    tx_type: *const libc::c_int
+  );
+}
+
+#[cfg(target_feature = "sse2")]
+use self::av1_fht4x4_sse2 as av1_fht4x4;
+#[cfg(target_feature = "sse2")]
+use self::av1_fht8x8_sse2 as av1_fht8x8;
+
+#[cfg(not(target_feature = "sse2"))]
```
tdaede: Is there a way for this to be an #else or something? So we don't have to list each instruction support in the "not" in the future.

lu-zero: You have to use [cfg_if](https://github.com/alexcrichton/cfg-if) [here](
https://github.com/rust-lang-nursery/stdsimd/blob/master/examples/nbody.rs#L30) a working example of it.

PR #337, opened 2018-07-16
Clean up CDFs.

mbebenita:
Avoid passing the superfluous `nsymbs` value around, this is now always computed from the length of the cdf slice, i.e. `cdf.len() - 1`.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1791,6 +1791,7 @@ impl ContextWriter {
           av1_ext_tx_ind[tx_set_type as usize][tx_type as usize] as u32,
           &mut self.fc.inter_ext_tx_cdf[eset as usize]
             [square_tx_size as usize]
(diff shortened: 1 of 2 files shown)


src/context.rs
```diff
@@ -1804,19 +1802,19 @@ impl ContextWriter {
           self,
           av1_ext_tx_ind[tx_set_type as usize][tx_type as usize] as u32,
           &mut self.fc.intra_ext_tx_cdf[eset as usize]
-            [square_tx_size as usize][intra_dir as usize],
-          num_ext_tx_set[tx_set_type as usize]
+            [square_tx_size as usize][intra_dir as usize]
+            [..num_ext_tx_set[tx_set_type as usize] + 1]
```
tdaede: ... yeah, like this.

PR #341, opened 2018-07-17
Td obu 5

tdaede:


src/lib.rs
```diff
@@ -1137,74 +1140,81 @@ fn write_obus(packet: &mut Write, sequence: &mut Sequence,
             let mut bw1 = BitWriter::<BE>::new(&mut buf1);
             // uleb128()
             let mut coded_payload_length = [0 as u8; 8];
-            let leb_size = aom_uleb_encode(obu_payload_size, &mut coded_payload_length);
+            let leb_size = aom_uleb_encode(obu_payload_size + 1, &mut coded_payload_length);
```
ycho: Can you explain again why +1 for payload size?

ycho: Oh, one byte for the obu_header! (That's so with Feb version)

tdaede: It counts the header byte. If you had an OBU extension it would be higher, but later they will move the size later and not count it, so I didn't bother "properly" counting the obu header size.

ycho: Oh, it was indeed one byte obu header. The first header written above was OBU_TEMPORAL_DELIMITER. Confused.

src/lib.rs
```diff
@@ -1137,74 +1140,81 @@ fn write_obus(packet: &mut Write, sequence: &mut Sequence,
...
-    // TODO: If # of tiles > 1 or show_existing_frame is true,
-    // write Frame Header OBU here.
-    if write_frame_header {
-        write_frame_header_obu_helper(packet, sequence, fi, OBU_Type::OBU_FRAME_HEADER,
-            obu_extension);
+    {
+        let mut bw1 = BitWriter::<BE>::new(&mut buf1);
+        bw1.write_obu_header(OBU_Type::OBU_FRAME_HEADER, obu_extension);
     }
+    packet.write(&buf1).unwrap();
```
ycho: So, regardless whether it is show_existing_frame, writing OBU_FRAME_HEADER obu and sequence obu here.
In that case, if we enable show_existing_frame, it will write sequence obu, which should not.

tdaede: Yeah I didn't test show existing frame, we don't expose it in any way. I think the sequence header write is valid, just wasteful.

src/lib.rs
```diff
@@ -1708,44 +1719,28 @@ fn encode_frame(sequence: &mut Sequence, fi: &mut FrameInvariants, fs: &mut Fram
...
     } else {
-        let obu_extension = 0;
+        let tile = encode_tile(fi, fs); // actually tile group
+        let obu_payload_size = tile.len() as u64;
         let mut buf1 = Vec::new();
         {
             let mut bw1 = BitWriter::<BE>::new(&mut buf1);
-            bw1.write_obu_header(OBU_Type::OBU_FRAME, obu_extension).unwrap();
+            // uleb128()
+            let mut coded_payload_length = [0 as u8; 8];
```
ycho: I originally used 16 bytes here, like = [0 as u8; 16];
but 8 would be still fine, for 8 bytes, it is 7 x 8 = 56 bits, so 2 ^ 56 size in bytes.

tdaede: In the libaom spec it's actually not allowed to be more than 2^32-1 bytes.

ycho: Oh, right... uleb128() is really so.
The tile_size_minus_1 uses "le(TileSizeBytes)", which has no limit in # of bytes

src/lib.rs
```diff
@@ -1708,44 +1719,28 @@ fn encode_frame(sequence: &mut Sequence, fi: &mut FrameInvariants, fs: &mut Fram
             None => (),
         }
     } else {
-        let obu_extension = 0;
+        let tile = encode_tile(fi, fs); // actually tile group
+        let obu_payload_size = tile.len() as u64;
         let mut buf1 = Vec::new();
         {
             let mut bw1 = BitWriter::<BE>::new(&mut buf1);
-            bw1.write_obu_header(OBU_Type::OBU_FRAME, obu_extension).unwrap();
```
ycho: So, in Feb version, encoder does not write OBU_FRAME if there is one tile ony?
I guessed decoder can still decode even though OBU_FRAME_HEADER comes in (the switch..case flow is written that way), but if it is OBU_FRAME_HEADER, i.e "if (obu_header.type != OBU_FRAME) break;"  (or show_existing_frame == true,) then the obu decoding does not continue to OBU_TILE_GROUP case which actually decode a tile, i.e. inside aom_decode_frame_from_obus(), in current version of libaom I mean.

ycho:
Looks Great To Me. Thanks for HACKING! Please LAND!

src/lib.rs
tdaede: OK, then we'll need to patch this to support OBU_FRAME once that's added.

PR #342, opened 2018-07-17
Speedup Quantize

lu-zero:
We get all over some decent speedups:

  | Baseline | Cache+inline | mdivu | Pregen
-- | -- | -- | -- | --
TX_4X4, 20 | 0.00% | 1.32% | 0.52% | 0.66%
TX_4X4, 55 | 0.00% | 1.55% | 0.37% | 0.85%
TX_8X8, 20 | 0.00% | -0.08% | 2.90% | 3.51%
TX_8X8, 55 | 0.00% | 0.02% | 3.20% | 3.66%


tdaede:
I think the target_feature commit shouldn't be there?

tdaede:
Nit pick: Second commit title should be "Convert integer divide into equivalent multiply and add operations."

benches/bench.rs
```diff
@@ -401,6 +401,7 @@ fn write_b_bench(b: &mut Bencher, tx_size: TxSize, qindex: usize) {
         x: sbx,
         y: sby
       };
+      fs.qc.update(fi.qindex, tx_size);
```
tdaede: This could have been immutable instead, e.g. part of the FrameInvariants constructor.

src/lib.rs
```diff
@@ -948,9 +954,9 @@ pub fn process_frame(sequence: &Sequence, fi: &FrameInvariants,
     let y4m_bytes = y4m_dec.get_bytes_per_sample();
     let csp = y4m_dec.get_colorspace();
     match csp {
-        y4m::Colorspace::C420 |
```
tdaede: Unrelated change.

lu-zero: Oh, right, I moved it away.

PR #344, opened 2018-07-17
Implement Q_ADAPT_PROBS.

tdaede:


tdaede:
This is broken for quantizers > 20.

ycho:
LGTM.
Tested for four ranges of qindex, and worked.

PR #346, opened 2018-07-17
Implement cdef dist

jmvalin:


src/rdo.rs
```diff
@@ -43,6 +44,41 @@ pub struct RDOPartitionOutput {
...
+      sum_d += d;
+      sum_s2 += (s*s) as i64;
+      sum_d2 += (d*d) as i64;
+      sum_sd += (s*d) as i64;
+    }
+  }
+  let svar = (sum_s2 - ((sum_s as i64 * sum_s as i64 + 32) >> 6)) as f64;
+  let dvar = (sum_d2 - ((sum_d as i64 * sum_d as i64 + 32) >> 6)) as f64;
+  let sse = (sum_d2 + sum_s2 - 2 * sum_sd) as f64;
+  let ssim_boost = 0.5_f64 * (svar + dvar + (400 << 2 * coeff_shift) as f64) / f64::sqrt((20000 << 4 * coeff_shift) as f64 + svar * dvar);
```
tmatth: Can we get a comment or named constants to explain where the 400 and 20000 come from?

src/rdo.rs
```diff
@@ -43,6 +44,41 @@ pub struct RDOPartitionOutput {
 }
 
 // Sum of Squared Error for a wxh block
+fn cdef_dist_wxh_8x8(src1: &PlaneSlice, src2: &PlaneSlice) -> u64 {
```
smarter: I guess it's called like this because it's the distortion metric used in cdef, but here it's confusing since it's used for everything.

src/rdo.rs
```diff
@@ -43,6 +44,41 @@ pub struct RDOPartitionOutput {
 }
 
 // Sum of Squared Error for a wxh block
```
smarter: This comment should be attached to the next function.

jmvalin:
Updated with the SSE comment at the right place and some info about the constants.

src/rdo.rs
```diff
@@ -42,6 +43,42 @@ pub struct RDOPartitionOutput {
   pub skip: bool
 }
 
+fn cdef_dist_wxh_8x8(src1: &PlaneSlice, src2: &PlaneSlice) -> u64 {
+  let coeff_shift = 0;
```
smarter: Add a TODO for setting this properly based on the bitdepth

PR #347, opened 2018-07-17
Fix a rebase glitch in quantize

lu-zero:
The offsets computation got lost.


tdaede:
you have an accidental submodule commit

PR #354, opened 2018-07-18
Enable EXT_PARTITION_TYPES.

tdaede:


src/context.rs
```diff
@@ -1722,7 +1756,11 @@ impl ContextWriter {
     let has_rows = (bo.y + hbs) < self.bc.rows;
     let ctx = self.bc.partition_plane_context(&bo, bsize);
     assert!(ctx < PARTITION_CONTEXTS);
-    let partition_cdf = &mut self.fc.partition_cdf[ctx];
+    let partition_cdf = if bsize <= BlockSize::BLOCK_8X8 {
+      &mut self.fc.partition_cdf[ctx][..5]
```
ycho: I have carefully checked this part, i.e. the cdf size reduced to 5, 4 for # of symbols + 1 for counter, then verified correct to me. Instead of 5, I'd recommend to use PARTITION_TYPES + 1. Btw, PARTITION_TYPES should be 4 but wrong in our current codebase.

ycho:
Tested with fixed size partition, two alternating cases.
1) 1st SB row has all 4x4s, 2nd SB row has all 16x16s, ...
2) 1st SB columns has all 4x4s, 2nd SB columns has all 16x16s, ...
then, it works fine.

ycho:
LGTM now! Pls land.

PR #361, opened 2018-07-19
Cdef work

xiphmont:
CDEF is working with no mismatches on testing.... much more work to do, but we can begin reviewing.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -68,7 +68,7 @@ DEC_FILE="dec_file.y4m"
 export RUST_BACKTRACE=1
 
 # Build and run encoder
-cargo run --bin rav1e --release -- $SEQ -o $ENC_FILE -s 3 -r rec_file.y4m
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -68,7 +68,7 @@ DEC_FILE="dec_file.y4m"
 export RUST_BACKTRACE=1
 
 # Build and run encoder
-cargo run --bin rav1e --release -- $SEQ -o $ENC_FILE -s 3
+cargo run --bin rav1e --release -- $SEQ -o $ENC_FILE -s 3 -r rec_file.y4m
```
tdaede: unrelated change (though maybe useful)

tdaede:
I'll squash these to merge, unless you'd rather squash.

Have you tried an AWCY run? (or just looking at the images? does it look better?)

xiphmont:
just looking at the images thusfar; they do in fact look better.

Go ahead and squash.

Monty

PR #363, opened 2018-07-19
CDEF bounds error fix

xiphmont:
Correction to end-of-frame bounds checking on skip flags (can't get
skip flag for blocks of the endge of the frame)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2030,58 +2030,60 @@ fn cdef_frame(fi: &FrameInvariants, rec: &mut Frame, bc: &mut BlockContext) {
             // Each direction block is 8x8 in y, potentially smaller if subsampled in chroma
             for by in 0..8 {
                 for bx in 0..8 {
-                    let mut dir = 0;
-                    let mut var: i32 = 0;
-                    let skip = bc.at(&sbo.block_offset(bx, by)).skip;
-                    for p in 0..3 {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2030,58 +2030,60 @@ fn cdef_frame(fi: &FrameInvariants, rec: &mut Frame, bc: &mut BlockContext) {
...
-                        let mut local_sec_damping:i32 = cdef_sec_damping + coeff_shift;
-                        let mut local_dir:usize = 0;
+                    let block_offset = sbo.block_offset(bx, by);
+                    if block_offset.x < bc.cols && block_offset.y < bc.rows {
+                        let mut dir = 0;
+                        let mut var: i32 = 0;
+                        let skip = bc.at(&block_offset).skip;
+                        for p in 0..3 {
+                            let mut rec_plane = &mut rec.planes[p];
+                            let mut cdef_plane = &mut cdef_frame.planes[p];
```
lu-zero: The 3 lines above are the same as
``` rust
for (p, (rec_plane, cdef_plane)) in rec.planes.iter_mut()
  .zip(cdef_plane.planes.iter_mut())
  .enumerate() {
```

Since you need `p` the gain is just being sure that the compiler drops the bounds check when accessing `.planes`.

PR #365, opened 2018-07-20
Encode decode tests

### LINKED ISSUE #3: Add a mode to test encode/decode mismatch
Probably by relying on aomdec being on the PATH

lu-zero:
Here the initial implementation of two encode-decode test.
It cover all the speeds and some quantization.
- [x] Encodes random frame data
- [x] Decodes successfully the produced packet
- [x] Decodes unsuccessfully the produced packets
- [x] Detects packet corruption

In the process of using the test, I found the hard way a couple of issues that got fixed in the first commits.

Fixes #3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ target
 Cargo.lock
 *.y4m
 *.ivf
+src/lib/aom.rs
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,6 @@ script:
     - |
         cargo build --verbose &&
         cargo test --verbose &&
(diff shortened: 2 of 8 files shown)


tdaede:
It looks like the encode/decode test isn't actually running on Travis.

lu-zero:
we need to decide which one to run and how often, the speed test takes a
lot of time they are behind a feature and disabled by default.

tdaede:
Well, I want at least a basic reconstruction test on Travis... it can run on a non-default speedlevel, I guess.

lu-zero:
Ok, let me add that.

PR #370, opened 2018-07-25
Define reference frame buffer array

fbossen:
A reference frame buffer with 8 slots is defined. The buffer is
updated after encoding each frame according to the value of
refresh_frame_flags. The value of refresh_frame_flags is also
set and properly encoded.

### DIFF
--- a/src/bin/rav1e.rs
+++ b/src/bin/rav1e.rs
@@ -11,6 +11,7 @@ extern crate rav1e;
 extern crate y4m;
 
 use rav1e::*;
+use std::rc::Rc;
 
 fn main() {
   let (mut io, config) = EncoderConfig::from_cli();
@@ -34,12 +35,13 @@ fn main() {
     framerate.den
   );
 
-  let mut last_rec: Option<Frame> = None;
+  let mut rec_buffer: [Option<Rc<Frame>>; 8] = [None, None, None, None, None, None, None, None];
   loop {
     //fi.frame_type = FrameType::KEY;
     fi.frame_type =
       if fi.number % 30 == 0 { FrameType::KEY } else { FrameType::INTER };
 
+    fi.refresh_frame_flags = if fi.frame_type == FrameType::KEY { 0xff } else { 1 };
     fi.intra_only = fi.frame_type == FrameType::KEY
       || fi.frame_type == FrameType::INTRA_ONLY;
     fi.use_prev_frame_mvs = !(fi.intra_only || fi.error_resilient);
@@ -50,7 +52,7 @@ fn main() {
       &mut io.output,
       &mut y4m_dec,
       y4m_enc.as_mut(),
-      &mut last_rec
+      &mut rec_buffer
     ) {
       break;
     }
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -2216,7 +2217,13 @@ pub fn process_frame(sequence: &mut Sequence, fi: &mut FrameInvariants,
                 let rec_frame = y4m::Frame::new([&rec_y, &rec_u, &rec_v], None);
                 y4m_enc.write_frame(&rec_frame).unwrap();
             }
-            *last_rec = Some(fs.rec);
+
+            let rfs = Rc::new(fs.rec);
```
lu-zero: The test below probably needs the same code (and maybe it could be wrapped in a tiny inline function)

src/lib.rs
```diff
@@ -847,10 +848,10 @@ impl<'a> UncompressedHeader for BitWriter<'a, BE> {
         }
       } else { // Inter frame info goes here
         if fi.intra_only {
```
ycho: I think you'd like to do:
assert!(fi.refresh_frame_flags != 0xff);
given av1 spec that: 
"If frame_type is equal to INTRA_ONLY_FRAME, it is a requirement of bitstream conformance that refresh_frame_flags is not equal to 0xff."

src/bin/rav1e.rs
```diff
@@ -34,12 +35,13 @@ fn main() {
...
   );
 
-  let mut last_rec: Option<Frame> = None;
+  let mut rec_buffer: [Option<Rc<Frame>>; 8] = [None, None, None, None, None, None, None, None];
   loop {
     //fi.frame_type = FrameType::KEY;
     fi.frame_type =
       if fi.number % 30 == 0 { FrameType::KEY } else { FrameType::INTER };
 
+    fi.refresh_frame_flags = if fi.frame_type == FrameType::KEY { 0xff } else { 1 };
```
ycho: We have "const REF_FRAMES: u32 = 8;" in src/lib.rc,
Then,
let allFrames = (1 << REF_FRAMES) - 1;
Can we replace '0xff' with allFrames? (Sorry..... *_*)

ycho: Though the av1 spec used named constants, I think we we can keep as it is since we all knows what 8 is for.
For ex, when av1 spec calculates internal frame width and height, it uses numbers directly instead of named constants. So, pls skip my comments above.

ycho:
From spec, I think we'd also like consider adding:
if  show_existing_frame == 1 {
 fi.refresh_frame_flags = 0 }
at some place like 
1) rav1e.rs, after the commented
//fi.show_existing_frame = fi.number % 2 == 1;
(FYI, show_existing_frame might not work correctly after OBU has merged. We have not tested it yet)
then later assert whether its value is 0 at places with "if fi.show_existing_frame {".

fbossen:
@ycho you are right that refresh_frame_flags should be 0 when show_existing_frame is 1 but it may be better to deal with those issues in a separate pull request

Issue #374, opened 2018-07-26
be197cf580d causes corruptions at speed=0 or 1


ycho:
https://github.com/xiph/rav1e/commit/be197cf580d5e45d867d09d75d6bb7b2da2c77c9

rzumer:
Fixed in #375.

Issue #373, opened 2018-07-26
Move skip RDO inner loop than prediction RDO


ycho:
Otherwise, exactly the same prediction is done twice for skip and non-skip modes.
Best place  to compare skip/non-skip would be inside encode_block(), which will 
return the RDO decision for skip flag together with rd_cost.

rzumer:
I'm not sure what this means, but as mentioned before, testing skip after making a decision for the prediction mode results in lower quality than not even testing skip at all, with current RD modelling.

ycho:
I see, I will take care of this task.

ycho:
New skip RDO #474 has landed, https://github.com/xiph/rav1e/commit/c2d7b6af3bb081b85dbde6adccab5522802390a9 and it works flawlessly.

rzumer:
I assume this was fulfilled by #474, so closing.

PR #375, opened 2018-07-26
Quick fix the corruption with speed=0 or 1 (#374)

ycho:
The commit be197cf580d5e45d867d09d75d6bb7b2da2c77c9 has caused
the corrupted decoded video frames with speed=0 or 1, since its landing on
June 26.

This bug has been there around a month and believed to be the culprit of
breakage of ext_partition_types PR by TD-Linux as well.

Better fix will come, which is also coincidentally stated earlier in
https://github.com/xiph/rav1e/issues/373.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -190,58 +190,28 @@ pub fn rdo_mode_decision(
   let partition_start_x = (bo.x & LOCAL_BLOCK_MASK) >> xdec << MI_SIZE_LOG2;
   let partition_start_y = (bo.y & LOCAL_BLOCK_MASK) >> ydec << MI_SIZE_LOG2;
 
-  for &skip in &[false, true] {
-    // Don't test skipped blocks at higher speed levels
-    if fi.config.speed > 1 && skip {
-      continue;
+  let skip = false;
+
+  let checkpoint = cw.checkpoint();
+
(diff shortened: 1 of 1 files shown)


rzumer:
Considering this loop will be permanently removed (#373) I recommend removing the enclosing scope (braces) as well.

src/rdo.rs
```diff
@@ -190,12 +190,8 @@ pub fn rdo_mode_decision(
...
   let partition_start_y = (bo.y & LOCAL_BLOCK_MASK) >> ydec << MI_SIZE_LOG2;
 
-  for &skip in &[false, true] {
-    // Don't test skipped blocks at higher speed levels
-    if fi.config.speed > 1 && skip {
-      continue;
-    }
-
+  let skip = false;
+  {
```
rzumer: Considering this loop will be permanently removed (#373) I recommend removing the enclosing scope (braces) as well.

ycho: :) I initially removed the { .. } then git diff showed me lots of changes, so left the { }? :)

ycho: So, we prefer remove { }, right?

rzumer: Yes, the whole enclosing scope will show up in a `git diff` if the indentation changes as it will when the outer scope is removed, but that is fine.

PR #386, opened 2018-07-30
WIP - Align plane data to 16 bytes.

mbebenita:
The first step is to aligns plane strides to 16 bytes. This caused a bug in the CDEF code related to slices not matching. I think I fixed that issue, but multiplication overflow errors still remain and I'm not sure how to fix this. 

@xiphmont can you take a look and see if I'm doing something obviously wrong here?


tdaede:
The multiplication overflow errors are unrelated, they are due to missing clamping. My libaom update will fix this.

ycho:
This has been merged as part of #407 , thanks!

PR #387, opened 2018-07-30
Cdef rdosetup

xiphmont:
Implement all the bitstream infrastructure for full CDEF
Some abstraction setup work to get to the RDO work

### DIFF
--- a/src/bin/rav1e.rs
+++ b/src/bin/rav1e.rs
@@ -25,8 +25,8 @@ fn main() {
     None => None
   };
 
-  let mut fi = FrameInvariants::new(width, height, config);
   let mut sequence = Sequence::new(width, height);
+  let mut fi = FrameInvariants::new(width, height, config, &mut sequence);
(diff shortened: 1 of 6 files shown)


xiphmont:
We could.  Is it superior in any way?

lu-zero:
You won't need explicit lifetimes :)

tdaede:
You could also just do a straight-up copy into the fi. (maybe that's what lu-zero meant, I don't know how just putting two references in a struct would avoid the lifetimes)

lu-zero:
@tdaede exactly.

xiphmont:
...so it should be SequenceInvariants? :-)  It's worth noting the FrameInvariants aren't... they mutate in the course of encoding.

Are folks happy with a copy into fi?  There are two somewhat opposing suggestions here....

tdaede:
This was replaced by montywork right?

xiphmont:
yes.

PR #390, opened 2018-07-31
Enable INTRA_EDGE prediction

ycho:
This affects all intra prediction modes in rav1e except DC_PRED.


tdaede:
Can we hold off on this until later tonight when I land my libaom update? It adds a flag that turns this on and off, so after that we can know it works both ways.

ycho:
Sure!

ycho:
btw,  INTRA_EDGE should have already turned on around time level-map is enabled by default, i.e Feb?

tdaede:
Yeah, that bit of code is still disabled in my updated libaom. That's why I wanted to land that first.

tdaede:
needs rebase

you'll need to push your revert onto rav1e_15b

PR #393, opened 2018-07-31
Propagate skip to the transform type decision function

rzumer:
This should make skip decisions more accurate, but skip RDO remains disabled, as it currently reduces coding efficiency. No desynchronization is detected via the decoding tests with this change and skip RDO enabled. Related: #374

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1498,7 +1498,7 @@ fn encode_block(seq: &Sequence, fi: &FrameInvariants, fs: &mut FrameState, cw: &
 
     let tx_type = if tx_set > TxSet::TX_SET_DCTONLY && fi.config.speed <= 3 {
         // FIXME: there is one redundant transform type decision per encoded block
-        rdo_tx_type_decision(fi, fs, cw, luma_mode, bsize, bo, tx_size, tx_set)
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1498,7 +1498,7 @@ fn encode_block(seq: &Sequence, fi: &FrameInvariants, fs: &mut FrameState, cw: &
 
     let tx_type = if tx_set > TxSet::TX_SET_DCTONLY && fi.config.speed <= 3 {
         // FIXME: there is one redundant transform type decision per encoded block
-        rdo_tx_type_decision(fi, fs, cw, luma_mode, bsize, bo, tx_size, tx_set)
+        rdo_tx_type_decision(fi, fs, cw, luma_mode, bsize, bo, tx_size, tx_set, skip)
```
ycho: Because skip==true (this flag is defined for a partition) means coefficient are not coded, so the transform is NOT applied, the skip flag should be false when rdo_tx_type_decision() is called.
I think you might be confused with two different kinds of skip, one for tx block (which is coded a part of coefficient coding) and another for partition, which is coded once per each partition.

rzumer:
@ycho OK, this is wrong then.

PR #395, opened 2018-07-31
Remove 'Writer' from 'ContextWriter' structure

xiphmont:
This change is to allow multiple and throwaway writers to be used in
predictive and out-of-order coding.  A Writer must be passed in (and
checkpointed) separately/explicitly.


xiphmont:
As discussed in the channel.... first step for out-of-order and throwaway Writers.

PR #397, opened 2018-08-01
Rustify some nested loops.

mbebenita:
Removes some redundant code by using closures. This should not have an effect on performance as they ought to be trivially inlined.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1260,6 +1260,19 @@ impl BlockContext {
     }
   }
 
+  pub fn for_each<F>(&mut self, bo: &BlockOffset, bsize: BlockSize, f: F) -> ()
+  where
+    F: Fn(&mut Block) -> ()
+  {
+    let bw = bsize.width_mi();
+    let bh = bsize.height_mi();
+    for y in 0..bh {
+      for x in 0..bw {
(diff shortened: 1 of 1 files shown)


src/context.rs
```diff
@@ -1349,14 +1362,7 @@ impl BlockContext {
...
   ) {
-    let bw = bsize.width_mi();
-    let bh = bsize.height_mi();
-
-    for y in 0..bh {
-      for x in 0..bw {
-        self.blocks[bo.y + y as usize][bo.x + x as usize].mode = mode;
-      }
-    }
+    self.for_each(bo, bsize, |block| block.mode = mode);
```
ycho: Ok, you convert my code to better one! thx!

src/context.rs
```diff
@@ -1417,25 +1423,11 @@ impl BlockContext {
...
   pub fn set_skip(&mut self, bo: &BlockOffset, bsize: BlockSize, skip: bool) {
-    let bw = bsize.width_mi();
-    let bh = bsize.height_mi();
-
-    for y in 0..bh {
-      for x in 0..bw {
-        self.blocks[bo.y + y as usize][bo.x + x as usize].skip = skip;
-      }
-    }
+    self.for_each(bo, bsize, |block| block.skip = skip);
```
ycho: here as well.... nice~

PR #407, opened 2018-08-02
New test

ycho:


src/plane.rs
```diff
@@ -34,9 +36,10 @@ pub struct Plane {
 
 impl Plane {
   pub fn new(width: usize, height: usize, xdec: usize, ydec: usize) -> Plane {
+    let stride = width.align_power_of_two(4); // Force 16 byte alignment.
```
mbebenita: Because the stride is not in bytes, this actually aligns to 32 bytes.

ycho:
@mbebenita , hi could you rebase your commit
https://github.com/xiph/rav1e/pull/407/commits/d55390e6e83ed55a19551a77fbebd191b82fc20b
and push again please?
So that we can push your own PR as a separate commit.

ycho:
@mbebenita , I've just landed this, could not wait for you.

PR #408, opened 2018-08-02
Use zero MV for all blocks in inter frames

fbossen:
All blocks in inter frames are coded using a single reference (LAST_FRAME, which is set to previous frame) and a zero motion vector (GLOBALMV). Mixing of intra and inter blocks within a same frame is currently unsupported, as inter mode context derivation is incomplete.


tdaede:
Awesome! Going to squash and merge.

ycho:
Sadly, this merged patch has caused desync on 2nd frame of nyan, at -s 0 and 1.
Was fine for -s 2,3,4,5.
The test I've used for -s 0 :
./target/release/rav1e nyan.y4m -o test.ivf -r test_rec.y4m --quantizer 10 --speed=0 --limit=2

rzumer:
Related: #160

ycho:
Trying debugging this patch. At speed=5 and enforcing min_partition_size = 4X4, it cause same desync. So, I expect the cause is related to 4x4 inter modes, for ex there additional rules for inter 4x4 partitions compared to that of intra.

fbossen:
@ycho resubmitted as #416 Yes, there was an issue with 4x4 blocks and another one with transform types

ycho:
Oh, I checked disabling tx_type rdo did not change thre symptom, but there was sth I missed then.

PR #416, opened 2018-08-02
Use zero MV for all blocks in inter frames

fbossen:
Should now work with all speed settings


src/lib.rs
```diff
@@ -1680,7 +1681,7 @@ pub fn write_tx_tree(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut Context
     let plane_bsize = get_plane_block_size(bsize, xdec, ydec);
 
     if bw_uv > 0 && bh_uv > 0 {
-        let uv_tx_type = tx_type;   // if inter mode, uv_tx_type == tx_type
+        let uv_tx_type = if has_coeff {tx_type} else {TxType::DCT_DCT}; // if inter mode, uv_tx_type == tx_type
```
ycho: Oh... yes. For Inter mode UV, av1 always uses DCT. Sorry that I cannot recall this and write....! Nice catch!

src/lib.rs
```diff
@@ -1689,12 +1690,17 @@ pub fn write_tx_tree(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut Context
...
+                y: bo.y  - ((bh * tx_size.height_mi() == 1) as usize)
+            };
+
             let po = PlaneOffset {
                 x: sb_offset.x + partition_x,
                 y: sb_offset.y + partition_y
             };
 
-            encode_tx_block(fi, fs, cw, w, p, &bo, chroma_mode, uv_tx_size, uv_tx_type,
+            encode_tx_block(fi, fs, cw, w, p, &tx_bo, chroma_mode, uv_tx_size, uv_tx_type,
```
ycho: Ooops... for 4x4 chroma we needed this since the last 4x4 luma position cannot be used for chroma 4x4 ! 
Sorry- my thought didn't reach that far..

src/lib.rs
ycho: Oh, right- when luma tx-block has no coeffs, then tx_type for chroma should ne DCT_DCT.
I think I completely forgot this.

ycho:
1) Tested for -s 0 and 1, nyan 2 frames and worked fine. Also decodable and by aomdec and its decoded images are matched to rav1e encoder output images.
2) Tested with : "cargo test --release --features=decode_test -- --ignored" and passed all 3 tests
(The branch is fbossen:mc1, which does not have most recently added test, odd_size_frame_with_full_rdo().)

LGTM. Thanks for quick fix!

PR #422, opened 2018-08-03
Increase frame size and # of frames of odd_size test

ycho:


ycho:
@rzumer Come to me almost none in release mode of test.

PR #425, opened 2018-08-03
[WIP] Do the inter prediction for a whole partition

ycho:
Inter mode prediction can be done for a whole partition,
instead of each tx-block.

TODO: Fix the plane_offset() function of BlockOffset struct for chroma
planes, so that bo.plane_offset() can be directly and conveniently called
instead of using a block of long codes in several places.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1428,8 +1428,6 @@ pub fn encode_tx_block(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut Conte
 
     if mode.is_intra() {
       mode.predict_intra(&mut rec.mut_slice(po), tx_size);
-    } else {
-      mode.predict_inter(fi, p, po, &mut rec.mut_slice(po), tx_size);
     }
 
     if skip { return false; }
@@ -1497,7 +1495,7 @@ fn encode_block(seq: &Sequence, fi: &FrameInvariants, fs: &mut FrameState,
         cw.write_angle_delta(w, 0, luma_mode);
     }
 
-      if has_chroma(bo, bsize, xdec, ydec) && !is_inter {
+    if has_chroma(bo, bsize, xdec, ydec) && !is_inter {
         cw.write_intra_uv_mode(w, chroma_mode, luma_mode, bsize);
         if chroma_mode.is_directional() && bsize >= BlockSize::BLOCK_8X8 {
             cw.write_angle_delta(w, 0, chroma_mode);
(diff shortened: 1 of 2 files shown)


ycho:
@fbossen, can you review this pls? somehow your id does not show up in reviewer list.

ycho:
ooops-- pushed debug code for the other commit (TODO) as well!

ycho:
TODO is filed https://github.com/xiph/rav1e/issues/426
and its fix, basing on this patch, is ready for PR submission.

tdaede:
Is this mostly a speed benefit?

ycho:
Yeah- in many sense, since encoder an do MC for whole block at one time, regardless of the # of multipe tx blocks used, though we have only one tx with rav1e atm. The AV1 spec also says this, though it does not affect the syntax, in decode_block( r, c, subSize ),
compute_prediction( ) is done right before
residual( )

src/lib.rs
```diff
@@ -1531,10 +1529,29 @@ fn encode_block(seq: &Sequence, fi: &FrameInvariants, fs: &mut FrameState,
         TxType::DCT_DCT
     };
 
-    // if inter, predict here
-    //mode.predict(&mut rec.mut_slice(po), tx_size);
-
     if is_inter {
+        // Inter mode prediction can take place once for a whole partition,
+        // instead of each tx-block.
+        for p in 0..3 {
```
fbossen: In the AV1 spec, the compute_prediction() function has a loop from 0 to  1 + HasChroma * 2. Presumably there are cases where a small partition may not have associated chroma?

ycho:
Yes, I think only luma plane when the input is mono, but... we are not caring for it other than YUV 4:2:0!

fbossen:
Maybe I am misreading the spec, but HasChroma can also be 0 when coding in 4:2:0 mode. For instance when the luma partition is smaller than 8x8.

src/lib.rs
ycho: Yes, you are right. If luma is 4x4, then there is no chroma 4x4 until the last 4x4 luma.
Also, in that case, the offset for chroma planes should be derived from the 1st 4x4 luma's coordinate. Thanks for pointing this out! I will fix soon.

ycho:
As I replied the above comment, fbossen's comment is valid.

src/lib.rs
ycho: btw, I found that I don't need iterate over planes here, since write_tx_tree() take care of all planes by itself, and it also check whether there is chroma partition exist.

ycho: Oh- no, the  luma_mode.predict_inter() should be still called for each plane, just it sholud not be called if there is no chroma block.

ycho:
@fbossen , Revised with fix following your point.

PR #427, opened 2018-08-04
Restore pkg-config dependency for non-MSVC Windows targets

rzumer:
This aims to enable compilation on non-MSVC Windows targets.
This change is untested, so in order to pull it I would like confirmation that it works on at least one non-MSVC Windows target. If compilation fails, please provide the location of the `aom.pc` file in the build directory.


rzumer:
Need to document the `pkg-config` dependency that remains on non-MSVC Windows targets, if this strategy turns out to work as expected.

PR #428, opened 2018-08-04
Fix plane_offset() of BlockOffset struct for chroma planes

ycho:
It has been wrong when SB is partitioned, thus we were using a block of codes doing that.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1167,11 +1167,13 @@ impl BlockOffset {
   /// Offset of the top-left pixel of this block.
   pub fn plane_offset(&self, plane: &PlaneConfig) -> PlaneOffset {
     let po = self.sb_offset().plane_offset(plane);
+
     let x_offset = self.x & LOCAL_BLOCK_MASK;
     let y_offset = self.y & LOCAL_BLOCK_MASK;
+
     PlaneOffset {
-      x: po.x + (x_offset << BLOCK_TO_PLANE_SHIFT),
-      y: po.y + (y_offset << BLOCK_TO_PLANE_SHIFT)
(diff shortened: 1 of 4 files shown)


fbossen:
Looks like a nice cleanup

src/context.rs
```diff
@@ -1167,11 +1167,13 @@ impl BlockOffset {
...
   pub fn plane_offset(&self, plane: &PlaneConfig) -> PlaneOffset {
     let po = self.sb_offset().plane_offset(plane);
+
     let x_offset = self.x & LOCAL_BLOCK_MASK;
     let y_offset = self.y & LOCAL_BLOCK_MASK;
+
     PlaneOffset {
-      x: po.x + (x_offset << BLOCK_TO_PLANE_SHIFT),
-      y: po.y + (y_offset << BLOCK_TO_PLANE_SHIFT)
+        x: po.x + (x_offset >> plane.xdec << BLOCK_TO_PLANE_SHIFT),
```
smarter: Shouldn't this be << (BLOCK_TO_PLANE_SHIFT - plane.xdec), similar to the other plane_offset function ?

ycho:
@lu-zero 
I've revised one of bench function, write_b_bench(), since for ex 8x8 tx shoud be applied
to (bx,by) , where both bx and by are even number, i.e. on 8x8 grid points on 4x4 BlockOffset (i.e. MI) grids.

src/context.rs
ycho: @smarter, hey more essence of the reason that I think I missed when I originally replied Aug, 2018 is now here! https://github.com/xiph/rav1e/issues/1062#issuecomment-469812492

PR #431, opened 2018-08-06
Assert to check  cdef_dist is called for multiple of 8x8 pixels

ycho:
Currently, cdef_dist_wxh() is called for 4x4 as well, which is a bug.


tdaede:
Looks fine, though if this is actually called by 4x4 blocks, won't Travis fail now?

ycho:
Well, I guess you thought cdef_dist is turned on by default, but it is off by default!

rzumer:
Indeed... if you have tested this with it turned on, then I approve.

ycho:
Of course this assert can be only hit with the option "--tune=Psychovisual", which I've tried.
The asserts are put in the beginning of cdef_dist_wxh(), before it calls cdef_dist_wxh_8x8() for each pair of 8x8 pixels, one for the input and another for the reconstructed.

PR #435, opened 2018-08-07
Cdef splitencode2

xiphmont:
Last encoding structural work to allow CDEF RDO decisions to happen at end-of-partition;

Implements split partition encode with 'everything up to the CDEF strength' and 'everything after the CDEF strength' coded in parallel separate Writers, such that the CDEF strength can be determined at the end of partition encode.

Also corrects a serious algebra error int he WriterRecorder, now that it's in full use. Local range state was not being used properly during replay.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1471,8 +1471,16 @@ impl BlockContext {
     }
   }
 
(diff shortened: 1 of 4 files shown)


tdaede:
LGTM

PR #436, opened 2018-08-08
Improve block bounds checking in CDEF

xiphmont:
The block bounds calculation in CDEF was doing nothing because of a
calculation error; due to misunderstanding why, a second, more detailed
guard was added later that should have been unnecessary.

Correct the earlier/easier check to function properly, remove the more
complex and now unnecessary later guard.


xiphmont:
One more quickie that technically counts as 'boundaries' :-)
Basic CDEF strength is signaled and calculated by superblock; no need to repeat for each block

ycho:
LGTM. 👍

PR #437, opened 2018-08-08
Enable and test high bit depth input

### LINKED ISSUE #325: Support 10-bit video in 10-bit mode.
We currently use 16 bit data and support reading 10-bit y4m, but convert it to 8 bit. Instead we should leave it at 10 bit. This requires signaling 10 bit mode in the header, and then using the appropriate quantization table.

rzumer:
* Handle high bit depth across the encoder
* Allow 10- and 12-bit 4:2:0 y4m input
* Add encode-decode test for high bit depth
* Add operations on high bit depth sequences in the build script (commented out by default)
* Update README.md
* Move `context::clamp()` to `util.rs` and make it generic

Closes #325.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ On Windows, pkg-config is not required. A Perl distribution such as Strawberry P
 
 # Compressing video
 
-Input videos must be 8-bit 4:2:0, in y4m format.
+Input videos must be 8- or 10-bit 4:2:0, in y4m format.
 
 ```
 cargo run --release --bin rav1e -- input.y4m -o output.ivf
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ rav1e temporarily uses libaom's transforms and CDF initialization tables, but is
 * 4x4 to 32x32 RDO-selected square blocks
 * DC, H, V, Paeth, and smooth prediction modes
 * 4x4 DCT and ADST transforms
+* 8-, 10- and 12-bit depth color
 * Variable speed settings
 * ~10 fps encoding @ 480p
 
@@ -38,7 +39,7 @@ On Windows, pkg-config is not required. A Perl distribution such as Strawberry P
 
 # Compressing video
 
-Input videos must be 8- or 10-bit 4:2:0, in y4m format.
+Input videos must be in y4m format and have 4:2:0 chroma subsampling.
 
 ```
 cargo run --release --bin rav1e -- input.y4m -o output.ivf
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -191,9 +191,8 @@ fn adjust_strength(strength: i32, var: i32) -> i32 {
 // CDEF parameters are stored for each 64 by 64 block of pixels.
 // The CDEF filter is applied on each 8 by 8 block of pixels.
 // Reference: http://av1-spec.argondesign.com/av1-spec/av1-spec.html#cdef-process
-pub fn cdef_frame(fi: &FrameInvariants, rec: &mut Frame, bc: &mut BlockContext) {
-    let bit_depth = 8;
-    let coeff_shift = bit_depth - 8;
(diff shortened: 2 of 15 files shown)


src/cdef.rs
```diff
@@ -191,9 +192,8 @@ fn adjust_strength(strength: i32, var: i32) -> i32 {
 // CDEF parameters are stored for each 64 by 64 block of pixels.
 // The CDEF filter is applied on each 8 by 8 block of pixels.
 // Reference: http://av1-spec.argondesign.com/av1-spec/av1-spec.html#cdef-process
-pub fn cdef_frame(fi: &FrameInvariants, rec: &mut Frame, bc: &mut BlockContext) {
-    let bit_depth = 8;
-    let coeff_shift = bit_depth - 8;
+pub fn cdef_frame(fi: &FrameInvariants, rec: &mut Frame, bc: &mut BlockContext, bit_depth: usize) {
```
lu-zero: I'd fold it in the `FrameInvariants`.

src/lib.rs
```diff
@@ -676,9 +688,9 @@ impl<'a> UncompressedHeader for BitWriter<'a, BE> {
     }
 
     fn write_color_config(&mut self, seq: &mut Sequence) -> Result<(), std::io::Error> {
-        self.write_bit(false)?; // 8 bit video
-        self.write_bit(seq.monochrome)?; 	// monochrome?
-        self.write_bit(false)?;  					// No color description present
+        self.write_bit(seq.bit_depth > 8)?; // 8 bit video
```
lu-zero: change the comment while at it.

src/partition.rs
```diff
@@ -355,32 +355,32 @@ use plane::*;
...
-  pub fn predict_intra<'a>(self, dst: &'a mut PlaneMutSlice<'a>, tx_size: TxSize) {
+  pub fn predict_intra<'a>(self, dst: &'a mut PlaneMutSlice<'a>, tx_size: TxSize, bit_depth: usize) {
     assert!(self.is_intra());
 
     match tx_size {
-      TxSize::TX_4X4 => self.predict_intra_inner::<Block4x4>(dst),
-      TxSize::TX_8X8 => self.predict_intra_inner::<Block8x8>(dst),
-      TxSize::TX_16X16 => self.predict_intra_inner::<Block16x16>(dst),
-      TxSize::TX_32X32 => self.predict_intra_inner::<Block32x32>(dst),
+      TxSize::TX_4X4 => self.predict_intra_inner::<Block4x4>(dst, bit_depth),
```
lu-zero: Ideally we could const-propagate the bit_depth the same way we do with blocksize.

It can happen at a later time if you do not feel confident about it.

src/cdef.rs
rzumer: If you are suggesting to duplicate `bit_depth` between `Sequence` and `FrameInvariants`, I don't think that this is a good idea.

src/partition.rs
rzumer: I prefer to leave this for a future PR.

rzumer:
This is ready for 10-bit support. If I get a very fast response on the side of `y4m` I can enable 12-bit as well, or this can be merged beforehand.

src/cdef.rs
tdaede: We can wait with this until a future Sequence-inside-FrameInvariants patch.

src/partition.rs
tdaede: It's also less critical, it doesn't control branching.

README.md
```diff
@@ -38,7 +38,7 @@ On Windows, pkg-config is not required. A Perl distribution such as Strawberry P
 
 # Compressing video
 
-Input videos must be 8-bit 4:2:0, in y4m format.
+Input videos must be 8- or 10-bit 4:2:0, in y4m format.
```
tdaede: Please update readme to include 12-bit as well.

PR #438, opened 2018-08-08
Remove duplicate frame header code.

tdaede:


ycho:
Oh.... wanted to say, if you wanted to delete
//write_uncompressed_header(&mut packet, sequence, fi).unwrap();
in encode_frame()?

tdaede:
Ah, indeed. There is another round of cleanup I'm doing after this, I'll put it in that patch.

PR #439, opened 2018-08-09
Write level 31 instead of 0.

tdaede:
We can't guarantee we meet a level, so choose level 31 for now
which has no constraints.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -605,7 +605,8 @@ impl<'a> UncompressedHeader for BitWriter<'a, BE> {
         self.write_bit(false)?; // no timing info present
         self.write(5, 0)?; // one operating point
         self.write(12,0)?; // idc
-        self.write(5, 0)?; // level
+        self.write(5, 31)?; // level
+        self.write(1, 0)?; // tier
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -605,7 +605,8 @@ impl<'a> UncompressedHeader for BitWriter<'a, BE> {
         self.write_bit(false)?; // no timing info present
         self.write(5, 0)?; // one operating point
         self.write(12,0)?; // idc
-        self.write(5, 0)?; // level
+        self.write(5, 31)?; // level
+        self.write(1, 0)?; // tier
```
rzumer: Seems that tier is not written to the header.

tdaede: The condition is actually whether level is >7:
https://aomediacodec.github.io/av1-spec/#sequence-header-obu-syntax

rzumer: I see. I was looking at the reduced still picture header area, my bad.

rzumer: I see. I was looking at the reduced still picture block, which does not code tier values in any case. My bad.

PR #441, opened 2018-08-10
Cdef rdo

xiphmont:
Basic CDEF RDO functional.

Selects one of 8 hardwired strengths after all other superblock decisions.  Does not yet optimize the per-frame strength list, and is doing redundant find_dir calculations.  However, it functions as intended.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1473,9 +1473,9 @@ impl BlockContext {
 
   pub fn set_cdef(&mut self, sbo: &SuperBlockOffset, cdef_index: u8) {
     let bo = sbo.block_offset(0, 0);
-    let bw = cmp::min (bo.x + 8, self.blocks[bo.y as usize].len());
-    let bh = cmp::min (bo.y + 8, self.blocks.len());
-
+    // Checkme: Is 16 still the right block unit for 128x128 superblocks?
+    let bw = cmp::min (bo.x + MAX_MIB_SIZE, self.blocks[bo.y as usize].len());
+    let bh = cmp::min (bo.y + MAX_MIB_SIZE, self.blocks.len());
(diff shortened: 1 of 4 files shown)


src/context.rs
```diff
@@ -1464,9 +1464,9 @@ impl BlockContext {
 
   pub fn set_cdef(&mut self, sbo: &SuperBlockOffset, cdef_index: u8) {
     let bo = sbo.block_offset(0, 0);
-    let bw = cmp::min (bo.x + 8, self.blocks[bo.y as usize].len());
-    let bh = cmp::min (bo.y + 8, self.blocks.len());
-
+    // Checkme: Is 16 still the right block unit for 128x128 superblocks?
+    let bw = cmp::min (bo.x + MAX_MIB_SIZE, self.blocks[bo.y as usize].len());
+    let bh = cmp::min (bo.y + MAX_MIB_SIZE, self.blocks.len());
     for y in bo.y..bh {
```
ycho: Have you considered using for_each here?

xiphmont: can't; if the superblock is bigger than the frame it will overflow the edge and panic.

xiphmont: Sorry, I mean, if the superblock overlaps the edge of the frame.

ycho: I see, I guess this relates to what we have discussed while ago.
I understand that, unless for_each has boundary check itself, you cant use it.

xiphmont: [beta AWCY run incoming, should be fun....]

xiphmont: ycho: exactly

xiphmont: Lots of AWCY fails, only partial results.  Failed encodes succeed here, even with debugging builds. Hrm.

PR #442, opened 2018-08-10
Revert "Enable and test high bit depth input (#437)"

ycho:
This reverts commit 8c5453501423ff67f9d973a2a1cc1df7e24e02b2.
Since 8c5453501423ff67f9d973a2a1cc1df7e24e02b2 has broken encoder, causing desync (in 1st frame) btw enc and dec for the test:
./target/release/rav1e nyan.y4m -o test.ivf -r test_rec.y4m --quantizer=10 --speed=0 --limit=5

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "aom"]
 	path = aom_build/aom
 	url = https://gitlab.xiph.org/xiph/aom-rav1e.git
-	branch = rav1e_16b
+	branch = rav1e_15b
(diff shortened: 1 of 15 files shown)


xiphmont:
Travis is supposed to test for this, but it appears the Travis test is busted somehow.  Agreed with revert policy until it is fixed. r=me

tdaede:
Well, the travis test uses different content. Something about that content causes this not to trigger.

ycho:
Yes, the decode_test uses random images, not from input files?

PR #447, opened 2018-08-11
Revert "Revert "Enable and test high bit depth input (#437)" (#442)"

rzumer:
After investigating the errors caused by the original PR, it seems that the tools used for comparison are interpolating samples from Y4M files labelled as using the `C420jpeg` color space, which the libaom decoder always uses, regardless of the chroma positioning indicated in the header.

In order to support outputting Y4M reconstructions at high bit depths, a change in `rav1e.rs` set the Y4M encoder's color space to be the same as that of the input file's. This means that rav1e was now outputting `C420Mpeg2` reconstructions for input files using that color space, while the decoder output was always `C420jpeg` and therefore required interpolating chroma samples in order to be converted to a raw image. Of course, the result was that due to interpolation error, there was a difference in raw output between the reconstructed output and the AOM decoder output.

~~The change that causes this issue was reverted in order to maintain compatibility between the two output streams, and because it does not currently affect compatibility with high bit depth streams, as ~~the `y4m` crate does~~ we do not yet support encoding Y4M files at high bit depths.~~ However, the real issue that prompted a reversal of the original PR stems from:  
* the reference decoder seemingly not handling chroma placement correctly
* the decision to decode to raw images in order to detect errors rather than simply comparing Y4M output

I recommend testing encoder output by comparing the reconstructed and decoded Y4M files, excluding the header, in the future.

The changes to this PR compared to #437 are as follows:
* ~~the Y4M encoder used to output reconstruction is no longer initialized to a non-default color space~~
* ~~using the `-r` option at high bit depths will panic, as the `y4m` crate does not currently support high bit depth encoding~~
* the Y4M encoder used to output reconstruction is now working at high bit depth

Relevant libaom issue: https://bugs.chromium.org/p/aomedia/issues/detail?id=2096


rzumer:
The test failure in CI is due to CDEF changes introduced after #437.

rzumer:
@Kagami interesting, on my end it panicks with "bad input" above 8-bit. I will look into it next week.

rzumer:
Regarding the reconstruction output at high bit depths, here is the relevant issue in `y4m`: https://github.com/PistonDevelopers/y4m/issues/13

We can perform our own conversion for now using `from_raw_parts`.

src/lib.rs
```diff
@@ -1817,27 +1817,57 @@ pub fn process_frame(sequence: &mut Sequence, fi: &mut FrameInvariants,
...
+                    vec![128u8; pitch_y * height],
+                    vec![128u8; pitch_uv * (height / 2)],
+                    vec![128u8; pitch_uv * (height / 2)]);
+                
+                let (stride_y, stride_u, stride_v) = (
+                    fs.rec.planes[0].cfg.stride,
+                    fs.rec.planes[1].cfg.stride,
+                    fs.rec.planes[2].cfg.stride);
+
+                for (y, line) in fs.rec.planes[0].data.chunks_mut(stride_y).enumerate() {
```
lu-zero: Rec does not mutate and you can use `chunks_mut()` on `rec_y` and friends.

``` rust
for (line, line_out) in fs.rec.planes[0].data.chunks().zip(rec_y.chunks_mut(pitch_y))
```

rzumer: Thank you.

tdaede:
LGTM.

PR #448, opened 2018-08-11
Add an Ilog impl for primitive integers

lu-zero:
There is at least one place that would use it but on 32bit in a pending patch from @ycho


tdaede:
needs rebase

lu-zero:
Rebased :)

PR #451, opened 2018-08-13
Move cli-specific code where it belongs

lu-zero:
That's a preliminary to the actual user API patch.
This way we:
- can optionally build rav1e w/out the cli
- have 2 dependencies optional as well (`y4m` is sort of tiny, but `clap` is relatively big)

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -7,6 +7,115 @@
 // Media Patent License 1.0 was not distributed with this source code in the
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
+// Imported from clap, to avoid to depend directly to the crate
+macro_rules! _clap_count_exprs {
+    () => { 0 };
+    ($e:expr) => { 1 };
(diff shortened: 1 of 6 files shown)


tdaede:
After the revert revert, needs rebase.

src/util.rs
```diff
@@ -7,6 +7,115 @@
 // Media Patent License 1.0 was not distributed with this source code in the
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
+// Imported from clap, to avoid to depend directly to the crate
```
tdaede: Is there really no better way to do this?

lu-zero: that's the simplest way since `Display`, `FromStr` and a mean to iterate values would be needed. Incidentally it provides also the same enum iterator capability we use for other enums so we could use it there as well and remove another crate :)

tdaede: I see. I guess we could have a separate "external" settings struct or something too. I guess this is okay.

PR #454, opened 2018-08-14
switch to 2018 edition

gibix:
- add rust_2018_preview lang feature
- cargo fix --edition
- cargo fix --edition-idioms


lu-zero:
Since `rav1e` is trying to stay stable-ready, could you please split the stable-breaking changes from the stable-ok ones? (adding the dyn when needed should be acceptabled)

gibix:
ok, I kept only the idiomatic part

PR #456, opened 2018-08-14
Add more frame dimensions to the decode test

rzumer:
This ensures that frame sizes that are not divisible by 16 are working as expected.

Also re-enable CDEF at low bit depth, as it was erroneously always disabled.


xiphmont:
lgtm

PR #459, opened 2018-08-14
Fix HBD with CDEF, reenable CDEF

xiphmont:


ycho:
Nice- that was a quick fix!

PR #464, opened 2018-08-15
Implement idct 4x4 in rust.

KyleSiefring:
Before merging I want to

- Find out if this will disable any simd code and add note that in the commit message

- Add at least some comments

### DIFF
--- a/src/transform.rs
+++ b/src/transform.rs
@@ -84,7 +84,7 @@ fn inv_txfm2d_add_4x4_rs(
   {
     let mut temp_in: [i32; 4] = [0; 4];
     for (raw, clamped) in input_slice.iter().zip(temp_in.iter_mut()) {
-      *clamped = clamp_value(*raw, bd + 8);
+      *clamped = clamp_value(*raw, ranges[0]);
     }
     av1_idct4(temp_in, buffer_slice, ranges[0]);
   }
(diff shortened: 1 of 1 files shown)


src/transform.rs
```diff
@@ -12,6 +12,104 @@ extern crate libc;
...
+  // stage 0
+
+  // stage 1
+  bf1[0] = bf0[0];
+  bf1[1] = bf0[2];
+  bf1[2] = bf0[1];
+  bf1[3] = bf0[3];
+
+  // stage 2
+  bf0[0] = half_btf(COSPI_INV[32], bf1[0], COSPI_INV[32], bf1[1], cos_bit);
```
lu-zero: This could be a single assignment.
``` rust
let bf1 = [input[0], input[2], input[1], input[3]];

let bf0 = [half_btf(COSPI_INV[32], bf1[0], COSPI_INV[32], bf1[1], cos_bit),
           half_btf(COSPI_INV[32], bf1[0], -COSPI_INV[32], bf1[1], cos_bit),
...

File: src/transform.rs [hunk 2]
```diff
@@ -12,6 +12,104 @@ extern crate libc;
...
+fn inv_txfm2d_add_4x4_rs(
+  input: &[i32], output: &mut [u16], stride: usize, bd: usize
+) {
+  let ranges = get_ranges(bd);
+  let mut buffer = [0i32; 4 * 4];
+  for r in 0..4 {
+    let input_slice = &input[r * 4..(r + 1) * 4];
+    let buffer_slice = &mut buffer[r * 4..(r + 1) * 4];
+    let mut temp_in: [i32; 4] = [0; 4];
+    for c in 0..4 {
```
lu-zero: You should pass directly the slice.

src/transform.rs
```diff
@@ -12,6 +12,104 @@ extern crate libc;
...
+      temp_in[c] = input_slice[c];
+    }
+    av1_idct4(temp_in, buffer_slice, ranges[0]);
+  }
+
+  for c in 0..4 {
```
lu-zero: Same here:
- use chunk iterators,
- pass around slices

KyleSiefring: Would chunking work for this part? It works on 0, 4, 8 and 12 instead of 0, 1, 2 and 3.

src/transform.rs
KyleSiefring: I would like if this code did everything here in the future. https://aomedia.googlesource.com/aom/+/master/av1/common/av1_inv_txfm2d.c#292

I just noticed that I forgot two clamps...

src/transform.rs
lu-zero: Oh, right it is vertical so chunks are less useful (but you can use them since you are taking the c element of each line)

src/transform.rs
lu-zero: Ideally you could do the same we did for the prediction and have generic code over the blocksizes.

src/transform.rs
KyleSiefring: I'm guessing the goal is to get rid of [] accesses and use iterators instead?

Didn't really understand how to do this.

tdaede:
LGTM.

PR #465, opened 2018-08-15
[WIP]Fix for top-down RDO search not updating mode

ycho:
AWCY Results for subset1 at --speed=1

master_s1@2018-08-14T20:42:35.057Z -> fix_topdown_rdo_mode_update_s1@2018-08-15T20:06:31.213Z

-0.1532 | -0.2611.    |     N/A        |  -0.1801 | -0.1267 | -0.1137 |    -0.2634

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -269,6 +269,8 @@ pub fn rdo_mode_decision(
     }
   }
 
+  cw.bc.set_mode(bo, bsize, best_mode_luma);
+
   assert!(best_rd >= 0_f64);
 
   RDOOutput {


ycho:
we could add setmode() to inside topdown_rdo(), but I (and you as well) put it at the end of pred mode decision, just in case.
'just in case' means, the reason why bottom-up rdo (i.e, -s 0) is not affected is it re-encode after rdo_mode_decision() is called, then the re-encode calls setmode(). However, there is comment like this there. "// FIXME: every final block that has gone through the RDO decision process is encoded twice"

PR #469, opened 2018-08-16
Enable mixing intra and inter blocks in inter frames

fbossen:


src/context.rs
```diff
@@ -1091,29 +1091,47 @@ impl BlockOffset {
...
+      y: y as usize
+    }
+  }
 }
 
 #[derive(Copy, Clone)]
 pub struct Block {
   pub mode: PredictionMode,
-  pub bsize: BlockSize,
+  //pub bsize: BlockSize,
```
tdaede: drop these commented out bits

src/context.rs
```diff
@@ -1091,29 +1091,47 @@ impl BlockOffset {
...
+  pub n8_h: usize, /* block height in the unit of mode_info */
+  pub is_sec_rect: bool,
 }
 
 impl Block {
   pub fn default() -> Block {
     Block {
       mode: PredictionMode::DC_PRED,
-      bsize: BlockSize::BLOCK_64X64,
+      //bsize: BlockSize::BLOCK_64X64,
```
tdaede: same

ycho:
If possible, please don't squash into one commit.

src/context.rs
```diff
@@ -1091,29 +1091,47 @@ impl BlockOffset {
...
+      //sb_type: BlockSize::BLOCK_64X64,
       partition: PartitionType::PARTITION_NONE,
       skip: false,
       ref_frames: [INTRA_FRAME; 2],
       neighbors_ref_counts: [0; TOTAL_REFS_PER_FRAME],
-      cdef_index: 0
+      cdef_index: 0,
+      n8_w: BlockSize::MI_SIZE_WIDE[BLOCK_64X64 as usize],
```
tdaede: I think you can just to BLOCK_64X64.width_mi() here

src/context.rs
```diff
@@ -1748,6 +1776,345 @@ impl ContextWriter {
     symbol_with_update!(self, w, enable as u32, &mut self.fc.filter_intra_cdfs[block_size as usize]);
   }
 
+  fn get_mvref_ref_frames(&mut self, ref_frame: usize) -> ([usize; 2], usize) {
+    let ref_frame_map: [[usize; 2]; TOTAL_COMP_REFS] = [
+      [ LAST_FRAME,  BWDREF_FRAME  ], [ LAST2_FRAME,  BWDREF_FRAME  ],
```
tdaede: This is fine for now but I'd rather drop all the *_FRAME constants at some point, maybe just with a find/replace here.

src/context.rs
```diff
@@ -1097,23 +1097,29 @@ impl BlockOffset {
...
       bsize: BlockSize::BLOCK_64X64,
+      sb_type: BlockSize::BLOCK_64X64,
       partition: PartitionType::PARTITION_NONE,
       skip: false,
       ref_frames: [INTRA_FRAME; 2],
       neighbors_ref_counts: [0; TOTAL_REFS_PER_FRAME],
-      cdef_index: 0
+      cdef_index: 0,
+      n8_w: BlockSize::MI_SIZE_WIDE[BLOCK_64X64 as usize],
+      n8_h: BlockSize::MI_SIZE_HIGH[BLOCK_64X64 as usize],
```
ycho: Copying the comment I've left at atomnuker's branch.
"You probably know n8_w and n8_h means n4_w and n4_h?"
I've fixed most of these outdated names in aomenc, but missed many in aomdec."

src/context.rs
```diff
@@ -1097,23 +1097,29 @@ impl BlockOffset {
...
+  pub n8_w: usize, /* block width in the unit of mode_info */
+  pub n8_h: usize, /* block height in the unit of mode_info */
 }
 
 impl Block {
   pub fn default() -> Block {
     Block {
       mode: PredictionMode::DC_PRED,
       bsize: BlockSize::BLOCK_64X64,
+      sb_type: BlockSize::BLOCK_64X64,
```
ycho: Copying the comment I've left at atomnuker's branch.
"The existing 'bsize' is exactly the same to newly added sb_type."

src/context.rs
fbossen: Yes, these names should be changed. Even n4_w and n4_h may be confusing. Can we do this refactoring in a separate PR?

src/context.rs
fbossen: Yes. Both have been removed at this time since they are not used.

src/context.rs
ycho: Sure- let's correct them in separate patch. There are several more defined locally.

src/context.rs
tdaede: I agree

PR #474, opened 2018-08-17
Test skip on and off in RDO loop

fbossen:
TD, pls don't land this yet!

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -217,6 +217,9 @@ pub fn rdo_mode_decision(
     // Find the best chroma prediction mode for the current luma prediction mode
     for &chroma_mode in mode_set_chroma {
       for &skip in &[false, true] {
+        // Don't skip when using intra modes
+        if skip && luma_mode.is_intra() { continue; }
+
         let mut wr: &mut dyn Writer = &mut WriterCounter::new();
         let tell = wr.tell_frac();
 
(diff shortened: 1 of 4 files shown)


src/rdo.rs
```diff
@@ -217,33 +215,35 @@ pub fn rdo_mode_decision(
...
-        fs,
-        w,
-        h,
-        w_uv,
-        h_uv,
-        bo,
-        cost,
-        seq.bit_depth
-      );
+      for &skip in &[false, true] {
```
ycho: Yes, putting skip/non-skip inside the loop for prediction mode, we can avoid duplicated run of each prediction, and I've pointed out this to rzumer.
I think your method here works fine since it passed the Travis testing.
The previous skip RDO has been reverted mainly due to it introduced corruptions.

fbossen:
master@bf537 -> skip@14ce4

```
-0.7872 | -1.1801 | -1.2715 |  -0.1556 | -0.2205 |  0.4549 |    -0.3761
```

ycho:
Great- I expect more gains comes from lower rates?

fbossen:
@ycho Yes, on the sequences where this makes a difference, the gains are mostly at the lower rates

ycho:
Can you also share the change of encoding time with skip testing please?

fbossen:
@ycho The full data is at https://beta.arewecompressedyet.com/?job=master%40bf537&job=skip%4014ce4
Seems actually to run a bit faster (maybe because things can be skipped during final encode)

ycho:
Let me do final testing of comparing enc and dec videos for nyan at -s 0 and 1.

PR #476, opened 2018-08-19
Re-eliminate pkg-config dependency under MinGW

daddesio:
On MinGW, the "no_build_target(cfg!(windows))" line unfortunately
prevents out/lib/pkgconfig/aom.pc from being created, causing the
"// MSVC" code branch to be taken which results in a linker error
(-laom not found).

Rather than fixing pkg-config, we can pass an absolute path to
libaom.a like we already do to aom.lib for MSVC. This has the
benefit that rav1e can be compiled with MinGW directly from cmd,
where pkg-config usually isn't available.

Tested on:
* MinGW-Builds 8.1.0 (cmd, msys2 terminal)
* Visual Studio 2017 15.8.1 (cmd, msys2 terminal)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,7 @@
 // build.rs
 
 extern crate cmake;
+#[cfg(unix)]
 extern crate pkg_config;
 #[cfg(unix)]
 #[cfg(feature = "decode_test")]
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -38,20 +39,12 @@ fn main() {
...
-            let bin_dir = if debug {
-                "Debug"
-            } else {
-                "Release"
-            };
-            println!("cargo:rustc-link-search=native={}", dst.join("build").join(bin_dir).to_str().unwrap());
-            println!("cargo:rustc-link-lib=static=aom");
-        }
+        println!("cargo:rustc-link-search=native={}", dst.join("build").to_str().unwrap());
+        println!("cargo:rustc-link-search=native={}", dst.join("build/Debug").to_str().unwrap());
```
lu-zero: It is fine even if only one of the two will exist and have content (since it is built on the release o debug path for cargo already) and rustc does not seem to complain about non-existing paths.

rzumer:
Thanks for the fix.

PR #477, opened 2018-08-19
Clean up lib.rs

lu-zero:
This is another sort-of-impacting preliminary before I lay down the user API.


ycho:
Thanks for the work. I was planning letting level map coder out of lib.rs.

lu-zero:
If nobody is against it I'd push it and get to the next step: `src/api.rs` :)

tdaede:
no issues here

PR #478, opened 2018-08-20
Search more intra modes in inter frames

fbossen:
Reuse exisiting intra mode lists to determine which intra modes
to search in an inter frame

### DIFF
--- a/src/predict.rs
+++ b/src/predict.rs
@@ -38,7 +38,6 @@ pub static RAV1E_INTRA_MODES_MINIMAL: &'static [PredictionMode] = &[
 
 pub static RAV1E_INTER_MODES: &'static [PredictionMode] = &[
   PredictionMode::GLOBALMV,
-  PredictionMode::DC_PRED,
 ];
 
 // Weights are quadratic from '1' to '1 / block_size', scaled by 2^sm_weight_log2_scale.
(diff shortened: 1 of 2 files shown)


fbossen:
| -2.5792 |  1.3078 |  1.5778 |  -2.4461 | -2.6847 | -2.9884 |    -1.3680 |

src/predict.rs
```diff
@@ -38,7 +38,6 @@ pub static RAV1E_INTRA_MODES_MINIMAL: &'static [PredictionMode] = &[
 
 pub static RAV1E_INTER_MODES: &'static [PredictionMode] = &[
   PredictionMode::GLOBALMV,
-  PredictionMode::DC_PRED,
```
lu-zero: Maybe could be made a separate commit to tell why.

fbossen: In this PR the list of modes tested in inter frames is constructed from the combination of a list of inter modes and a list of intra modes. There is thus no need anymore to list intra modes in the list of inter modes.

ycho:
Is the awcy results above from default speed, i.e. 3?
If yes, has -s 0 and 1 test clear? :)

fbossen:
@ycho Results are for the default speed (3). I haven't tested at speeds 0 and 1 (I don't expect an issue specific to small blocks here), but can do.

ycho:
Just tested for -s 0 and 1, and all fine. Just it now is SUPER SLOWER!! Let me land..

PR #479, opened 2018-08-20
Fix chroma distortion computation

fbossen:
Chroma distortion was sometimes being computed on inexistant 2x2 blocks

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -153,8 +153,6 @@ fn compute_rd_cost(
   let mut w_uv = (w_y >> xdec) & mask;
   let mut h_uv = (h_y >> ydec) & mask;
 
-  let bo_chroma = BlockOffset { x: bo.x & !xdec, y: bo.y & !ydec };
-
   if (w_uv == 0 || h_uv == 0) && is_chroma_block {
     w_uv = MI_SIZE;
     h_uv = MI_SIZE;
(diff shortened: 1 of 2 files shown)


fbossen:
At speed 0, where small blocks are used:

| -1.2099 | -5.2414 | -5.0444 |  -1.0668 | -1.1727 | -1.1425 |    -2.2226 |

ycho:
Thanks for fixing this. 
I should have tested this by putting assert!( w >= 4) ... there or in sse_wxh(),
as I did similar thing for cdef_dist_wxh(), so that wxh is multiple of 8x8s.

src/rdo.rs
```diff
@@ -147,10 +147,23 @@ fn compute_rd_cost(
     unimplemented!();
   };
 
+  let PlaneConfig { xdec, ydec, .. } = fs.input.planes[1].cfg;
+
+  let mask = !(MI_SIZE - 1);
+  let mut w_uv = (w_y >> xdec) & mask;
+  let mut h_uv = (h_y >> ydec) & mask;
+
+  let bo_chroma = BlockOffset { x: bo.x & !xdec, y: bo.y & !ydec };
```
ycho: I did not get why bo_chroma need truncation if chroma is not sampled in either x or y.
If chroma is 444, xdec == ydec == 0, then bo_chroma.x|y can only locate the size 8x8 blocks when each 4x4 luma has its corresponding 4x4 chromas blocks?

src/rdo.rs
```diff
@@ -163,7 +161,7 @@ fn compute_rd_cost(
   // Add chroma distortion only when it is available
   if w_uv > 0 && h_uv > 0 {
     for p in 1..3 {
-      let po = bo_chroma.plane_offset(&fs.input.planes[p].cfg);
+      let po = bo.plane_offset(&fs.input.planes[p].cfg);
```
ycho: Oh.. deleted here! Sorry- ignore my previous question. :)

PR #480, opened 2018-08-20
Assert sse_wxh() inputs multiple of 4x4 blocks

ycho:


src/rdo.rs
```diff
@@ -102,6 +102,9 @@ fn cdef_dist_wxh(
 
 // Sum of Squared Error for a wxh block
 fn sse_wxh(src1: &PlaneSlice<'_>, src2: &PlaneSlice<'_>, w: usize, h: usize) -> u64 {
+  assert!(w & 0x3 == 0);
```
fbossen: Is the precedence of & and == different in Rust than C?
Should we use a 0x3 constant or something like (MI_SIZE - 1)?

ycho: I think the same precedence rule btw C and Rust, since above means (w & 0x3) == 0, did we need to ( and ) in C, otherwise C views it as w &( 0x3 == 0)? I think not.
Right, I think (MI_SIZE - 1) is better.

fbossen: In C, you need to add parentheses. Otherwise you may get this warning: & has lower precedence than ==; == will be evaluated first

fbossen: Looks like Rust has different precedence rules: https://doc.rust-lang.org/reference/expressions.html#expression-precedence

ycho: Oh, you are right. In C, I needed to use parenthesize (w & 0x3). Thanks for checking this up.
With rust, in addition to my actual probing of values in assert many times before,  I just checked that bitwise and boolean operators are higher than comparison, as in https://doc.rust-lang.org/1.22.1/book/second-edition/appendix-02-operators.html

fbossen:
LGTM

Issue #463, opened 2018-08-15
Redundant computations in rdo_mode_decision()


fbossen:
For intra partitions when chroma is present, there is a loop over all possible chroma modes. encode_block_b() is called for each chroma mode. The luma portion is the same regardless of the chroma mode. However the luma portion is processed/encoded every time.
The loop over chroma modes shouldn't reprocess the luma portion for each iteration.

ycho:
Also commented from: https://github.com/xiph/rav1e/pull/482#issuecomment-414481070

One possible (easier) way to resolve this issue would be to decouple the nested RDO loops of luma and chroma intra modes as in libaom. Meanwhile, it might require separation of luma and chroma processing in current encode_block_b(), and doing skip test indepedently for luma and chroma.

I expect harder way is to buffer the predicted/reconstructed pixels and their contexts for chroma prediction.

barrbrain:
This is still a major opportunity for improvement. With the current structure of luma-chroma RDO, there should be negligible loss in compression efficiency from moving to: best luma mode (luma-only RD), best non-CfL chroma mode (chroma-only RD for given luma mode), test CfL RD if eligible.

PR #482, opened 2018-08-20
Reduce the number of intra modes searched for chroma

fbossen:
Limit the intra modes searched for chroma to the colocated luma
mode plus the DC mode if different from the first mode

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -214,15 +214,14 @@ pub fn rdo_mode_decision(
   for &luma_mode in &mode_set {
     assert!(fi.frame_type == FrameType::INTER || luma_mode.is_intra());
 
-    let same_as_luma_modes = &[ luma_mode ];
-    let mode_set_chroma = if is_chroma_block && luma_mode.is_intra() {
-      intra_mode_set
-    } else {
-      same_as_luma_modes
-    };
+    let mut mode_set_chroma = vec![ luma_mode ];
+
+    if is_chroma_block && luma_mode.is_intra() && luma_mode != PredictionMode::DC_PRED {
+      mode_set_chroma.push(PredictionMode::DC_PRED);
+    }
 
     // Find the best chroma prediction mode for the current luma prediction mode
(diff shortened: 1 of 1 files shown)


src/rdo.rs
```diff
@@ -214,15 +214,14 @@ pub fn rdo_mode_decision(
...
-    };
+    let mut mode_set_chroma = vec![ luma_mode ];
+
+    if is_chroma_block && luma_mode.is_intra() && luma_mode != PredictionMode::DC_PRED {
+      mode_set_chroma.push(PredictionMode::DC_PRED);
+    }
 
     // Find the best chroma prediction mode for the current luma prediction mode
-    for &chroma_mode in mode_set_chroma {
+    for &chroma_mode in &mode_set_chroma {
```
ycho: For given luma_mode, it now checks only one chroma_mode, so it does not need for loop here?

fbossen: Sometimes there should be a second mode (see push() on line 220)

ycho:
Related to the goal of this patch (i.e. improving speed), I again figure out that if more than one chroma intra modes are tested as previously, luma prediction in encode_block_b() is called duplicated among different chroma intra modes though luma work as context for chroma?

src/rdo.rs
ycho: Oh right- you mentioned it in the commit message as well. If luma is not DC then chroma rdo loop test DC and the non-DC mode.

fbossen:
@ycho Yes, I recently opened #463 on this topic

fbossen:
At default speed setting (3):

| -0.5450 |  1.1756 |  1.2647 |  -0.7288 | -0.4745 | -0.5919 |    -0.0551 |

fbossen:
Runs about 40% at default speed setting

ycho:
That's nice speed up. Also, I am sure that improving Y psnr while losing small in UV will be much appreciated by @tterribe.

PR #487, opened 2018-08-21
[WIP] Fix rdo_tx_type_decision() to use chroma prediction mode if intra

ycho:
Change in AWCY, so doing this harm the PSNR!?
Anyway, since this patch affect the intra block only, let me run on awcy with intra frame only.

I've also realized that the recent change of chroma intra pred mode RDO only search
for luma's pred mode and DC_PRED, so mostly luma mode == chroma mode, which makes
this patch almost not useful. 

Default -s 3:
master-7fe6aecf37ad0a99d552113fd5de9a580f0872dd -> tx_type_rdo_use_chroma_pred_2fa0ce

0.0016 | -0.0641 |  0.0061 |   0.0117 | 0.0047 |  0.0039 |    -0.0594

-s 0:
master-7fe6aec_s0_subset1 -> tx_type_rdo_use_chroma_pred_2fa0ce_s0_subset1@2018-08-21T21:58:06.640Z

0.0028 |     N/A |  0.0343 |   0.0043 | 0.0056 |  0.0129 |    -0.0014

So, it seems not useful at all to me!

This signify me that including chroma tx distortion everr useful, so testing it.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1188,7 +1188,7 @@ pub fn encode_block_b(fi: &FrameInvariants, fs: &mut FrameState,
 
     let tx_type = if tx_set > TxSet::TX_SET_DCTONLY && fi.config.speed <= 3 && !skip {
         // FIXME: there is one redundant transform type decision per encoded block
-        rdo_tx_type_decision(fi, fs, cw, luma_mode, bsize, bo, tx_size, tx_set, bit_depth)
+        rdo_tx_type_decision(fi, fs, cw, luma_mode, chroma_mode, bsize, bo, tx_size, tx_set, bit_depth)
     } else {
         TxType::DCT_DCT
     };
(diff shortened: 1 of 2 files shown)


ycho:
While the idea is correct, AWCY results says it is never useful. So drop!

rzumer:
Is the BD rate result different now?

ycho:
Nope- didn't try again but just re-opened to illustrate what has been pre-condition before my another PR https://github.com/xiph/rav1e/pull/489 applies.

fbossen:
Do we understand why there is a BD rate difference? Shouldn't the chroma cost be the same regardless of the tx type selected for luma?

ycho:
@fbossen, @tterribe : YES. Franks's statement of "Shouldn't the chroma cost be the same regardless of the tx type selected for luma?" makes me clear about that tx_type rdo does not need chroma.
Given luma pred mode and chroma pred mode, we want to decide tx_type.
Them changes of tx_type only affects luma cost, not chroma. So, whether including chroma or not should not affect the decision of luma chroma pred modes. (So, tx_type RDO only need to consider luma)

ycho:
Thinking and seeing again the run flow in the code, the chroma cost can be (very) slightly changes because chroma coeffs are encoded right after luma coeffs are encoded. The all the level map coder CDFs  has been adpated to luma coeffs, when chroma coeffs start to be coded.

ycho:
Right, if they don't share cdfs, then what I just said is wrong again.

PR #489, opened 2018-08-22
tx_type_rdo_decision() uses luma only

ycho:
master-48326cdc05f9f042ae05e6d7066f1b89152daaba -> tx_type_rdo_use_luma_only@2018-08-22T00:10:46.630Z

-0.1634 | -0.2861 | -0.2602 |  -0.1993 | -0.2388 | -0.2718 |    -0.2315

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1401,9 +1401,9 @@ pub fn encode_block_b(seq: &Sequence, fi: &FrameInvariants, fs: &mut FrameState,
             }
         }
       }
-      write_tx_tree(fi, fs, cw, w, luma_mode, bo, bsize, tx_size, tx_type, skip, bit_depth); // i.e. var-tx if inter mode
+      write_tx_tree(fi, fs, cw, w, luma_mode, bo, bsize, tx_size, tx_type, skip, bit_depth, false); // i.e. var-tx if inter mode
     } else {
-      write_tx_blocks(fi, fs, cw, w, luma_mode, chroma_mode, bo, bsize, tx_size, tx_type, skip, bit_depth, cfl);
+      write_tx_blocks(fi, fs, cw, w, luma_mode, chroma_mode, bo, bsize, tx_size, tx_type, skip, bit_depth, cfl, false);
     }
 }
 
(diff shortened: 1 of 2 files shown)


tdaede:
Do you have any idea why this improves it?

ycho:
:) Let's figure it out!
Either there still is another bug in computing chroma distortion or I did something good I didn't intended.

ycho:
@fbossen 
The only reason I see is ekarlier stated in https://github.com/xiph/rav1e/pull/487

The rdo_tx_type_decision( ) does not use chroma_mode when it is called during mode decision.
i.e. rdo_tx_type_decision(fi, fs, cw, luma_mode, bsize, bo, tx_size, tx_set, bit_depth);

Then, write_tx_blocks() is using luma_mode for both luma and chroma intra modes, which is WRONG.
      i.e. write_tx_blocks(
        fi, fs, cw, wr, **mode, mode**, bo, bsize, tx_size, tx_type, false, bit_depth
      );
Because when encode_tx_block() encodes chroma, it should now the tx_type, which comes from
let uv_tx_type = uv_intra_mode_to_tx_type_context(chroma_mode);
but chroma_mode is WRONG.

The mysterious part is  https://github.com/xiph/rav1e/pull/487 tries to fix this but did not improve the coding efficiency as I reported.

ycho:
@fbossen, @tterribe 
I think we should land this patch, i.e. tx_type rdo only uses luma because the chroma cost (coeffs + chroma  pred if intra)  does not change for different luma tx_type.

src/encoder.rs
```diff
@@ -1207,16 +1207,17 @@ pub fn encode_block_b(fi: &FrameInvariants, fs: &mut FrameState,
 
             luma_mode.predict_inter(fi, p, &po, &mut rec.mut_slice(&po), plane_bsize);
         }
-        write_tx_tree(fi, fs, cw, w, luma_mode, bo, bsize, tx_size, tx_type, skip, bit_depth); // i.e. var-tx if inter mode
+        write_tx_tree(fi, fs, cw, w, luma_mode, bo, bsize, tx_size, tx_type, skip, bit_depth, false); // i.e. var-tx if inter mode
```
fbossen: Should this also be false in the inter case? The compute_tx_type() function in the spec returns the same tx type for luma and chroma when is_inter is true.

ycho: You mean true (that is luma_only), that is already done, but not here and inside tx_type_decision() in rdo.rs (Line 311).

fbossen: Sorry, commented on the wrong part of the code...

src/rdo.rs
```diff
@@ -306,11 +308,11 @@ pub fn rdo_tx_type_decision(
     let tell = wr.tell_frac();
     if is_inter {
       write_tx_tree(
-        fi, fs, cw, wr, mode, bo, bsize, tx_size, tx_type, false, bit_depth
+        fi, fs, cw, wr, mode, bo, bsize, tx_size, tx_type, false, bit_depth, true
```
fbossen: Should this also be true in the inter case? The compute_tx_type() function in the spec returns the same tx type for luma and chroma when is_inter is true.

PR #490, opened 2018-08-22
Deblock

xiphmont:
Setup for the loop filter.  This adds some infrastructure we might as well start using soon, and given the massive rebases and work going on in parallel in nearby code, no need to hold off for a completed loop filter megamerge.

As of now, no functional change.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1131,7 +1131,7 @@ pub fn encode_block_b(fi: &FrameInvariants, fs: &mut FrameState,
     let is_inter = !luma_mode.is_intra();
     if is_inter { assert!(luma_mode == chroma_mode); };
 
-    cw.bc.set_size(bo, bsize);
+    cw.bc.set_block_size(bo, bsize);
     cw.bc.set_mode(bo, bsize, luma_mode);
 
     if fi.frame_type == FrameType::INTER {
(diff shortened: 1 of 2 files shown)


src/encoder.rs
```diff
@@ -1166,17 +1166,19 @@ pub fn encode_block_b(fi: &FrameInvariants, fs: &mut FrameState,
...
-
     // these rules follow TX_MODE_LARGEST
     let tx_size = match bsize {
         BlockSize::BLOCK_4X4 => TxSize::TX_4X4,
         BlockSize::BLOCK_8X8 => TxSize::TX_8X8,
         BlockSize::BLOCK_16X16 => TxSize::TX_16X16,
         _ => TxSize::TX_32X32
     };
+    cw.bc.set_tx_size(bo, tx_size);
+    // Were we not hardcoded to TX_MODE_LARGEST, block tx size would be written here
```
ycho: This is very good point.

If TX_MODE_SELECT is signaled, for intra block, tx_size is required even if the block is skipped.
So, tx_size must be signaled and also written to tx_size mode book keeping.

xiphmont:
On Wed, Aug 22, 2018 at 12:14 PM Yushin Cho <notifications@github.com>
wrote:

> *@ycho* approved this pull request.
>
>    1.
>
>    Can you please consider renaming 'loop_filter' more specifically so
>    that they stand for deblock only and not for other kinds of loop filters,
>    such as CDEF and Loop Restoration?
>    The "DeblockState;" is nice example.
>
>
I had actually renamed absolutely everything, and the complete naming
disconnect with the spec (mostly what I'm implementing from) was becoming
progressively harder to follow.  But I think I will in fact rename things
more thoroughly as I continue.


>
>    1.
>    2.
>
>    You might also consider put all loop filter code in new separate file,
>    src/deblock.rs, as is with CDEF.
>
>
I am in fact doing that.  Existing pattern had this code go into lib.rs
er... encoder.rs instead.  I wanted that to land because the rest is mostly
standalone in deblock.rs.

PR #497, opened 2018-08-24
Search CfL parameters of unit magnitude

barrbrain:
Replace the fixed parameters for chroma-from luma with a search among the 8 alpha pairs combining -1, 0 and 1. Only search if the current best chroma mode is intra.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -49,6 +49,7 @@ pub struct RDOPartitionOutput {
   pub bo: BlockOffset,
   pub pred_mode_luma: PredictionMode,
   pub pred_mode_chroma: PredictionMode,
+  pub pred_cfl_params: CFLParams,
   pub ref_frame: usize,
   pub mv: MotionVector,
   pub skip: bool
@@ -190,6 +191,7 @@ pub fn rdo_mode_decision(
   bsize: BlockSize, bo: &BlockOffset) -> RDOOutput {
   let mut best_mode_luma = PredictionMode::DC_PRED;
   let mut best_mode_chroma = PredictionMode::DC_PRED;
+  let mut best_cfl_params = CFLParams::new();
   let mut best_skip = false;
   let mut best_rd = std::f64::MAX;
   let mut best_ref_frame = INTRA_FRAME;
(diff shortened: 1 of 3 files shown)


barrbrain:
[AWCY results](https://beta.arewecompressedyet.com/?job=master-b1381798de246737c8a7406a8e536cd97f04d85f&job=rav1e-cfl-pred-2%402018-08-24T09%3A19%3A33.749Z) at the default speed:

PSNR | PSNR Cb | PSNR Cr | PSNR HVS | SSIM | MS SSIM | CIEDE 2000
------|----------|----------|------------|------|----------|-------------
-0.1670 | -6.4655 | -10.8567 | -0.1342 | -0.1773 | -0.1380 | -2.2103

I also tidied up the parameter plumbing a little, now that it is complete.

src/rdo.rs
```diff
@@ -240,41 +242,51 @@ pub fn rdo_mode_decision(
...
-        );
-
-        if rd < best_rd {
-          best_rd = rd;
-          best_mode_luma = luma_mode;
-          best_mode_chroma = chroma_mode;
-          best_ref_frame = ref_frame;
-          best_mv = mv;
-          best_skip = skip;
+        for &[alpha_u, alpha_v] in &cfl_alpha_set {
```
barrbrain: The corresponding names in the spec are `CflAlphaU` and `CflAlphaV`.
I'm considering naming these `cfl_alpha_u` and `cfl_alpha_v` but in the spec those names refer to the positions in the  magnitude CDF.

ycho: I think this alpha parameter search sholud be done only when chroma_mode == CFL_PRED?
Also, you might want to add speed condition like, "&& fi.config.speed <= "?

ycho: I also see that UV_CFL_PRED is neither in RAV1E_INTRA_MODES_MINIMAL or RAV1E_INTRA_MODES, did you mean this?

src/rdo.rs
```diff
@@ -240,41 +242,51 @@ pub fn rdo_mode_decision(
     };
 
     // Find the best chroma prediction mode for the current luma prediction mode
-    let cfl = &CFLParams::new();
     for &chroma_mode in &mode_set_chroma {
+      let mut cfl_alpha_set = vec![ [-1, 0], [-1, 1], [-1, -1], [1, -1], [0, -1], [1, 0], [0, 1], [1, 1] ];
+      if chroma_mode != PredictionMode::UV_CFL_PRED {
+        cfl_alpha_set.truncate(1);
```
ycho: Oh, is this doing cfl_alpha_set become null, if current chroma_mode is not CFL? So that below 
 for &[alpha_u, alpha_v] in &cfl_alpha_set {.... } does not run?

barrbrain: Here we avoid iterating CfL parameters for modes other than `UV_CFL_PRED` by truncating the list to a single placeholder.

ycho: I checked that cfl_alpha_set.truncate(1);  makes the size  of cfl_alpha_set  as 1, so still have one entry then
the below for &[alpha_u, alpha_v] in &cfl_alpha_set { ... " is iterated once.

ycho: Yes, but it iterated once for other modes, for ex chroma DC_PRED?
Is that intended?

ycho: I tried cfl_alpha_set.truncate(0); then it skips the alpha search loop.

ycho: truncate(0) here?

src/rdo.rs
ycho: Can we put
 assert!(chroma_mode == PredictionMode::UV_CFL_PRED); inside this loop?

src/rdo.rs
ycho: Yes, I've initially misunderstood that the design where parameter search is located. It is put inner most of current nested rdo loops. So, setting cfl to the 1st entry of cfl_alpha_set[] and passing to encode_block_b() seems harmless for non CFL_PRED modes. Wouldn't it be helpful if cfl parameter to encode_block_b() does not have any info(like null or None) when choma mode is mot CFL_REPD?

src/rdo.rs
ycho: Also, is there any specific reason why the alpha search loop is inside skip test loop?
Because, we would move the skip testing into inside encode_block_b(), in order to fix the redundant call of predictions for non-skip and skip cases.

barrbrain:
After rebase, [AWCY results](https://beta.arewecompressedyet.com/?job=master-df5a83239c62d5517f1b3aef51f4128603b43d7a&job=rav1e-cfl-pred-5%402018-08-24T22%3A57%3A36.884Z) at default speed:

| -0.1601 | -6.0022 | -9.5223 |  -0.1626 | -0.1923 | -0.1940 |    -1.9919 |

PR #501, opened 2018-08-24
Fix shift amount in 2-d convolution loop

fbossen:


fbossen:
That's on top of previous commit:

|    PSNR |  PSNR Cb |  PSNR Cr | PSNR HVS |    SSIM | MS SSIM | CIEDE 2000 |
|    ---: |     ---: |     ---: |     ---: |    ---: |    ---: |       ---: |
| -8.6935 | -11.4644 | -11.9302 |  -9.1163 | -9.3540 | -9.6274 |    -9.7880 |

fbossen:
1-pel step ME gives the following as a whole:

|     PSNR |  PSNR Cb |  PSNR Cr | PSNR HVS |     SSIM |  MS SSIM | CIEDE 2000 |
|     ---: |     ---: |     ---: |     ---: |     ---: |     ---: |       ---: |
| -18.6012 | -18.3146 | -18.6150 | -19.3885 | -19.3896 | -20.1962 |   -17.7978 |

src/partition.rs
```diff
@@ -709,7 +709,7 @@ ref_frame: usize, mv: &MotionVector, bit_depth: usize) {
                 sum += intermediate[r + k][c] as i32 * SUBPEL_FILTERS[y_filter_idx][row_frac as usize][k];
               }
               let output_index = r * stride + c;
-              let val = ((sum + 8) >> 4).max(0).min(max_sample_val);
+              let val = ((sum + 1024) >> 11).max(0).min(max_sample_val);
```
ycho: So, its not shift right by 7 + 7 = 14 bits?

ycho: Oh, the 1st stage already did 3 bits above. So, 3 + 11 =14 bits.

ycho: If we use higher precision vars for sum and intermediate, can we even shift less at 1st stage?

fbossen: the spec tells you how much to shift. No choice here...

PR #503, opened 2018-08-25
Add 1/8th pel motion refinement step

fbossen:
Currently this feature yields an average loss of about 1 percent BD rate, so keep it disabled by default

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -340,7 +340,7 @@ impl FrameInvariants {
             showable_frame: true,
             error_resilient: true,
             intra_only: false,
-            allow_high_precision_mv: false,
+            allow_high_precision_mv: true,
             frame_type: FrameType::KEY,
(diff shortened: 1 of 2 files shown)


fbossen:
1/8 pel motion search currently yields an average BD rate loss (mainly at the lower bit rates)

| 0.9264 |  1.5328 |  1.6731 |   1.0290 | 1.1079 |  1.2343 |     1.4055 |

tdaede:
Looks like it helps on some of the clips, but I guess we need RD in the motion search to take advantage of that.

ycho:
RIght- we don't count axis of a mv_cost, and rd_cost for inter mode only comes from prediction error.

ycho:
Maybe other filters than the default 8-tab would help show the real benefits of 1/8 pel MVs, which will need another monster RDO over the subpel interpolation filters!

Issue #398, opened 2018-08-01
Convert to the bench built-in once it is ready


lu-zero:
Tracking issue: https://github.com/rust-lang/rust/issues/29553

lu-zero:
Related issue https://github.com/rust-lang/rust/issues/50297

djrenren:
Heads up, the most likely case is that following https://github.com/rust-lang/rust/issues/50297 bench will be made an external crate that consumes those  APIs (shouldn't really change the heart of this issue though)

lu-zero:
Hopefully the change would be still minimal :)

lu-zero:
I opened https://github.com/xiph/rav1e/pull/507 as an alternative path. Criterion offers quite a number of good features out of box and seems a good fit since we plan to give a try at native SIMD now that at least some of it is available on stable rust.

rzumer:
Closing this issue since the alternative solution was merged.

PR #507, opened 2018-08-26
Move the benchmarks to Criterion

lu-zero:
It produces more information including pretty graphs and immediate speedup/slowdown evaluation compared to previous runs.

Solves https://github.com/xiph/rav1e/issues/398 by other means.


rzumer:
Seems fine to me as a dev dependency, but that means there is no chance of removing external dependencies for benchmarking, so I would like to have @tdaede's opinion.

lu-zero:
[here](https://asciinema.org/a/198275) the full output.

tdaede:
That's actually really nice, ffmpeg and x264 internal benchmarks have similar functionality. I don't have any problem depending on criterion for this.

PR #512, opened 2018-08-27
Speed up probability update function

fbossen:
Provides an overall encoder run time reduction of about 1 percent


tdaede:
Very nice.

PR #514, opened 2018-08-27
Speed up diff() function

fbossen:
Reduces overall run time by about 2 percent

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1169,10 +1169,9 @@ fn diff(dst: &mut [i16], src1: &PlaneSlice<'_>, src2: &PlaneSlice<'_>, width: us
   let src1_stride = src1.plane.cfg.stride;
   let src2_stride = src2.plane.cfg.stride;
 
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -1166,11 +1166,17 @@ fn write_obus(packet: &mut dyn io::Write, sequence: &mut Sequence,
...
-        for i in 0..width {
-            dst[j*width + i] = (src1.p(i, j) as i16) - (src2.p(i, j) as i16);
-        }
+  let src1_stride = src1.plane.cfg.stride;
+  let src2_stride = src2.plane.cfg.stride;
+
+  for ((l, s1), s2) in dst.chunks_mut(width)
+                        .zip(src1.as_slice().chunks(src1_stride))
+                        .zip(src2.as_slice().chunks(src2_stride))
+                        .take(height) {
```
lu-zero: take before the first `zip()`, may lead to a further speedup.

PR #515, opened 2018-08-27
SSSE3 implementation of CfL predictor

barrbrain:
Not quite as fast as the C intrinsics implementation in libaom yet.
Already a lot faster than the current native rust implementation.

### DIFF
--- a/src/predict.rs
+++ b/src/predict.rs
@@ -443,8 +443,8 @@ pub trait Intra: Dim {
     let max = _mm_set1_epi16((1 << bit_depth) - 1);
 
     for j in 0..Self::H {
-      let luma = ac.as_ptr().offset(stride as isize * j as isize);
-      let line = output.as_mut_ptr().offset(32 * j as isize);
+      let luma = ac.as_ptr().offset((32 * j) as isize);
+      let line = output.as_mut_ptr().offset((stride * j) as isize);
 
       let mut i = 0isize;
       while (i as usize) < Self::W {
--- a/src/predict.rs
+++ b/src/predict.rs
@@ -18,6 +18,11 @@ use partition::*;
 use std::mem::*;
 use context::INTRA_MODES;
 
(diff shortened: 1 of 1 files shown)


src/predict.rs
```diff
@@ -426,11 +431,51 @@ pub trait Intra: Dim {
...
+    let alpha_sign = _mm_set1_epi16(alpha);
+    let alpha_q12 = _mm_slli_epi16(_mm_abs_epi16(alpha_sign), 9);
+    let dc_q0 = _mm_set1_epi16(output[0] as i16);
+    let max = _mm_set1_epi16((1 << bit_depth) - 1);
+
+    for (line, luma) in
+      output.chunks_mut(stride).take(Self::H).zip(ac.chunks(32))
+    {
+      let mut i = 0isize;
+      while (i as usize) < Self::W {
```
lu-zero: What happens if you unroll it 2x? (the fact that using the iterator does not unroll it might or might not be a bug.

barrbrain: This already fully unrolls for 4x4. I'll update this once more before we merge. Using pointer arithmetic for the outer loop eliminates branching.

src/predict.rs
```diff
@@ -469,11 +470,13 @@ pub trait Intra: Dim {
     bit_depth: usize
   ) {
     if alpha == 0 { return; }
+    assert!(32 >= Self::W);
+    assert!(ac.len() >= 32 * Self::H);
+    assert!(stride >= Self::W);
+    assert!(output.len() >= stride * Self::H);
```
lu-zero: If you move them in the function body, does the compiler elide the checks for you even if you use the normal iterators/array indexing? (It should)

barrbrain: I'll try that for the next change. The CI failures highlight that these assertions are imprecise. I'll update to permit CfL on the bottom row in positions other than the left-most column.

PR #516, opened 2018-08-27
[WIP] Select a reasonable MV predictor from MVs in stack

fbossen:
Select the MV predictor with the smallest L1 distance from the
MV to be coded

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,6 +4,8 @@
 set -e
 
 #SEQ=!!!!! ENTER YOUR FAVORITE Y4M HERE !!!!!
+#SEQ=/Users/bossen/Movies/yuv/objective-1-fast/ducks_take_off_1080p50_60f.y4m
+SEQ=/Users/bossen/Movies/yuv/objective-1-fast/blue_sky_360p_60f.y4m
 
 if [[ -z "${SEQ}" ]]; then
   SEQ=nyan.y4m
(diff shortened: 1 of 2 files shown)


fbossen:
This isn't working terribly well at the moment:

| -0.1738 | -0.2379 | -0.2976 |  -0.1795 | -0.1365 | -0.1248 |    -0.2665 |

tdaede:
I don't think this ever gave big gains in libaom either, assuming that whatever you were coding before was relatively sane.

ycho:
The description of this PR needs be updated since the code now considers logarithmic distance of MV in each dimension, which is far closer to actual bit rates than linear L1 distance.

fbossen:
This isn't getting better

| -0.0636 | -0.1560 | -0.1680 |  -0.0585 | -0.0120 | -0.0445 |    -0.0648 |

ycho:
Wouldn't it because MV cost is still negligible ? I.e. search range is not large yet....*_*

fbossen:
@ycho I should probably try lower speed settings. More bits should be spent on MVs when blocks are smaller

ycho:
Also, is mv_cost included in final rd_cost used by bottom_up and top_down partition search? I think last time your fixed that partitioning cost be included.

rzumer:
Since it is unlikely that this PR will be revived any time soon, I will close it now as suggested.

PR #518, opened 2018-08-28
Add search for NEARESTMV mode

fbossen:
Move MV stack construction out of encoding function such that the
value of the nearest MV can be determined before encoding

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2331,7 +2331,7 @@ impl ContextWriter {
     let (_rf, _rf_num) = self.get_mvref_ref_frames(INTRA_FRAME);
 
     let target_n4_h = bsize.height_mi();
-    let target_n4_w = bsize.height_mi();
+    let target_n4_w = bsize.width_mi();
 
     let mut max_row_offs = 0 as isize;
     let row_adj = (target_n4_h < BlockSize::MI_SIZE_HIGH[BLOCK_8X8 as usize]) && (bo.y & 0x01) != 0x0;
--- a/src/context.rs
+++ b/src/context.rs
@@ -2103,11 +2103,14 @@ impl ContextWriter {
     }
   }
 
-  fn has_tr(&mut self, bo: &BlockOffset) -> bool {
+  fn has_tr(&mut self, bo: &BlockOffset, bsize: BlockSize, is_sec_rect: bool) -> bool {
     let sb_mi_size = BlockSize::MI_SIZE_WIDE[BLOCK_64X64 as usize]; /* Assume 64x64 for now */
     let mask_row = bo.y & LOCAL_BLOCK_MASK;
     let mask_col = bo.x & LOCAL_BLOCK_MASK;
-    let mut bs = cmp::max(self.bc.at(bo).n4_w, self.bc.at(bo).n4_h);
+    let target_n4_w = bsize.width_mi();
+    let target_n4_h = bsize.height_mi();
+
+    let mut bs = target_n4_w.max(target_n4_h);
 
     if bs > BlockSize::MI_SIZE_WIDE[BLOCK_64X64 as usize] {
       return false;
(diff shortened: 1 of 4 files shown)


fbossen:
| -6.6882 | -7.3289 | -7.6921 |  -6.5564 | -6.5062 | -6.5456 |    -6.8485 |

ycho:
@atomnuker pls reviewer this patch

src/context.rs
```diff
@@ -2396,19 +2404,21 @@ impl ContextWriter {
...
         let found_match = self.scan_col_mbmi(bo, col_offset, max_col_offs, &mut processed_cols, ref_frame, mv_stack,
-                                             &mut far_newmv_count);
+                                             &mut far_newmv_count, bsize);
         col_match |= found_match;
       }
     }
 
     let total_match = if row_match { 1 } else { 0 } + if col_match { 1 } else { 0 };
 
+    assert!(total_match >= nearest_match);
```
atomnuker: Why? Doesn't seem like it'll be possible unless someone unintentionally breaks the code (at which point a desync should happen).

src/context.rs
```diff
@@ -2131,27 +2134,28 @@ impl ContextWriter {
...
      * to the right won't have been decoded) */
-    if (blk.n4_w > blk.n4_h) && blk.is_sec_rect {
+    if (target_n4_w > target_n4_h) && is_sec_rect {
       has_tr = false;
     }
 
     /* The bottom left square of a Vertical A (in the old format) does
      * not have a top right as it is decoded before the right hand
      * rectangle of the partition */
+/*
```
atomnuker: Why disable this? Isn't it required to exclude some edge cases?

src/context.rs
```diff
@@ -2428,7 +2438,7 @@ impl ContextWriter {
   }
 
   pub fn find_mvrefs(&mut self, bo: &BlockOffset, ref_frame: usize,
-                     mv_stack: &mut Vec<CandidateMV>) -> usize {
+                     mv_stack: &mut Vec<CandidateMV>, bsize: BlockSize, is_sec_rect: bool) -> usize {
```
atomnuker: Why did you change from using blk.is_sec_rect to needing it as a parameter? Its always set to false now when you call find_mvrefs.

src/context.rs
fbossen: Yes, this won't ever happen. The only reason it's there is to facilitate understanding of the lines below. In particular, the fact that total_match is equal to at least 1 when nearest_match is 1 may not be obvious. Alternatively this could me put in a comment

src/context.rs
fbossen: I commented it out for now to avoid relying on blk (current block). Down the road we'll have to decide whether to write things into blk before calling find_mvrefs() or pass values as parameters. For now I am passing values as parameters.

src/context.rs
fbossen: blk.is_sec_rect would always be false for now. Same issue as in previous comment: do we pass values as parameters or do we write them into the current blk structure and read them out?

src/rdo.rs
```diff
@@ -240,10 +243,10 @@ pub fn rdo_mode_decision(
     }
 
     let ref_frame = if luma_mode.is_intra() { INTRA_FRAME } else { LAST_FRAME };
-    let mv = if luma_mode != PredictionMode::NEWMV {
-      MotionVector { row: 0, col: 0 }
-    } else {
-      motion_estimation(fi, fs, bsize, bo, ref_frame)
+    let mv = match luma_mode {
```
ycho: Once mv is decided here, is "best_mv = mv;" at line 293 still required?

src/rdo.rs
```diff
@@ -297,6 +301,7 @@ pub fn rdo_mode_decision(
   }
 
   cw.bc.set_mode(bo, bsize, best_mode_luma);
+  cw.bc.set_motion_vector(bo, bsize, best_mv);
```
ycho: Also, unless best_mv is updated, we can do this right after "let mv = ...." ?

src/context.rs
```diff
@@ -2321,14 +2326,18 @@ impl ContextWriter {
     }
   }
 
-  fn setup_mvref_list(&mut self, bo: &BlockOffset, ref_frame: usize, mv_stack: &mut Vec<CandidateMV>) -> usize {
+  fn setup_mvref_list(&mut self, bo: &BlockOffset, ref_frame: usize, mv_stack: &mut Vec<CandidateMV>,
+                      bsize: BlockSize, is_sec_rect: bool) -> usize {
     let (_rf, _rf_num) = self.get_mvref_ref_frames(INTRA_FRAME);
 
+    let target_n4_h = bsize.height_mi();
+    let target_n4_w = bsize.height_mi();
```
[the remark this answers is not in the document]
fbossen: Nice catch! Will fix

src/rdo.rs
fbossen: No. Basically this is needed so that the RDO for the next block behaves correctly.

src/rdo.rs
fbossen: Yes, my understanding is that the best_mv needs to be captured in the RDOOutput so that it can be used in the final encode.

ycho: So, initially  best_mv = MotionVector { row: 0, col: 0 };, then assume intra mode is better than mv modes and no chance for doing "best_mv = mv;", then best_mv is still (0,0). Would this be expected?

fbossen:
Seems even better on a recent version of master:

| -7.7694 | -7.9452 | -8.2971 |  -7.7014 | -7.7181 | -7.9676 |    -7.7985 |

Issue #521, opened 2018-08-28
Pull tx_type RDO out of chroma RDO


ycho:
The goal is to improve speed by not calling redundant rdo_tx_type_decision() for entire chroma RDO.

PR #522, opened 2018-08-28
Pull rdo_tx_type_decision() out of encode_block_b()

ycho:
Also does pulling tx_type_rdo() out of chroma RDO, as requested in #521.

There should not be any change in coding performance but only speed improvements.

AWCY results for the default speed -s 3 will come at: 
https://beta.arewecompressedyet.com/?job=master-ea40e92a90aac92766cf9f89a6194d388c1917c9&job=pull_txtype_rdo%402018-08-28T20%3A21%3A59.599Z

For -s 1,
https://beta.arewecompressedyet.com/?job=master-ea40e92_s1%402018-08-28T18%3A16%3A46.433Z&job=pull_txtype_rdo_s1%402018-08-28T20%3A21%3A51.138Z

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -190,6 +190,37 @@ fn compute_rd_cost(
   (distortion as f64) + lambda * rate
 }
 
+pub fn rdo_tx_size_type(seq: &Sequence, fi: &FrameInvariants,
+  fs: &mut FrameState, cw: &mut ContextWriter, bsize: BlockSize,
+  bo: &BlockOffset, luma_mode: PredictionMode, skip: bool)
+  -> (TxSize, TxType) {
+  // these rules follow TX_MODE_LARGEST
+  let tx_size = match bsize {
+      BlockSize::BLOCK_4X4 => TxSize::TX_4X4,
+      BlockSize::BLOCK_8X8 => TxSize::TX_8X8,
+      BlockSize::BLOCK_16X16 => TxSize::TX_16X16,
+      _ => TxSize::TX_32X32
+  };
+  cw.bc.set_tx_size(bo, tx_size);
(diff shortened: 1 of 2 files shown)


src/rdo.rs
```diff
@@ -263,10 +295,11 @@ pub fn rdo_mode_decision(
 
           let mut wr: &mut dyn Writer = &mut WriterCounter::new();
           let tell = wr.tell_frac();
-
+          if skip { tx_type = TxType::DCT_DCT; };
```
fbossen: tx_type should be changed only for the current iteration, not all the following ones as well

ycho: Oops, correct. It should be the same for whole chroma loop.

ycho:
I think we first land https://github.com/xiph/rav1e/pull/489/commits, since now rdo_tx_type_decision() is indeed outside chroma rdo loop.

fbossen:
It's fine to land #489 first but it needs some attention (merge conflict)

src/context.rs
```diff
@@ -1557,6 +1557,7 @@ impl BlockContext {
     self.for_each(bo, txsize.block_size(), |block| { block.tx_w = tx_w; block.tx_h = tx_h } );
   }
 
+  //TODO: Don'w need set_tx_type()?
```
fbossen: I don't think we need to store tx_type since there are no block to block dependencies

ycho: Ok- I though some coding tools we haven't implemented yet would use tx_type of neighboring blocks as contexts to choose CDF.

ycho:
Confirm that this patch gives no changes in coding performance.
For -s 1, it gives almost x2 speed up. For -s 3, > x1 speedup.

fbossen:
LGTM

PR #523, opened 2018-08-29
Add test for NEARMV mode

fbossen:


fbossen:
On top of NEARESTMV:

| -2.1171 | -2.1000 | -2.1833 |  -2.0803 | -2.1274 | -2.1531 |    -2.0437 |

fbossen:
| -3.0453 | -2.6437 | -2.7548 |  -3.0513 | -3.0870 | -3.1340 |    -2.8649 |

fbossen:
At speed 1

| -2.6173 | -1.8835 | -1.8732 |  -2.5038 | -2.3420 | -2.3642 |    -2.1404 |

fbossen:
At speed 0

| -3.5690 | -2.8611 | -2.9673 |  -3.5090 | -3.3415 | -3.4368 |    -3.1690 |

src/rdo.rs
```diff
@@ -278,6 +280,7 @@ pub fn rdo_mode_decision(
     let mv = match luma_mode {
       PredictionMode::NEWMV => motion_estimation(fi, fs, bsize, bo, ref_frame),
       PredictionMode::NEARESTMV => if mv_stack.len() > 0 { mv_stack[0].this_mv } else { MotionVector { row: 0, col: 0 } },
+      PredictionMode::NEARMV => if mv_stack.len() > 1 { mv_stack[1].this_mv } else { MotionVector { row: 0, col: 0 } },
```
ycho: So, if NEARMV prediction mode is chosen, the 2nd nearest is used w/o additional info coded (that I vaguely imagined, sth like index in mv_stack or mv lists), correct?

fbossen: More recent commits in this PR have 3 NEARMV modes, where NEARMV mode is signalled and followed by an index into mv_stack.

src/partition.rs
```diff
@@ -340,7 +340,9 @@ pub enum PredictionMode {
   PAETH_PRED,
   UV_CFL_PRED,
   NEARESTMV,
-  NEARMV,
+  NEAR0MV,
```
ycho: Wow! Am I seeing AV2 before anyone else? :)
I thought all these prediction modes are strictly follows the av1 spec? So, no variants of NEARMV is possible?
Because the integer enum id for each mode is directly fed to symbol writing, so unless you parse again these new pred modes GLOBALMV will have three values shifted down?

fbossen: The values assigned to each mode are somewhat arbitrary. There is already a mismatch with libaom caused by UV_CFL_PRED I think. The generated bitstream still complies with the av1 spec.

tdaede: Unfortunately currently write_intra_mode_kf actually uses the numeric values. This only applies to the first modes (intra ones) of course, not to any past UV_CFL_PRED, and we could actually make a mapping function or something if we wanted.

ycho:
LGTM! Thanks for good work!

PR #526, opened 2018-08-30
Fix rdo_tx_size_type() missing calling inter_prediction()

ycho:


src/encoder.rs
```diff
@@ -1338,49 +1386,7 @@ pub fn encode_block_b(seq: &Sequence, fi: &FrameInvariants, fs: &mut FrameState,
...
-              }
-            }
-            else
-            {
-              luma_mode.predict_inter(fi, p, &po, &mut rec.mut_slice(&po), plane_bsize.width(),
-                                      plane_bsize.height(), ref_frame, mv, bit_depth);
-            }
-        }
-      }
+      inter_pred(fi, fs, cw, luma_mode, bsize, bo, bit_depth);
```
ycho: I think this is not enough, but also needs above lines 1315~1317.

ycho:
Closing this since fbossen's patch #527 is a better fix.

PR #527, opened 2018-08-30
Recompute inter prediction every time it's needed

fbossen:
It should be noted that encode_tx_block() overwrites prediction data stored in the reconstruction buffer. The prediction thus needs to be computed each time encode_tx_block() is called.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1229,6 +1229,52 @@ pub fn encode_tx_block(
     has_coeff
 }
 
+pub fn motion_compensate(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut ContextWriter,
+                         luma_mode: PredictionMode, ref_frame: usize, mv: MotionVector,
+                         bsize: BlockSize, bo: &BlockOffset, bit_depth: usize) {
+  if luma_mode.is_intra() { return; }
+
+  let PlaneConfig { xdec, ydec, .. } = fs.input.planes[1].cfg;
+
+  // Inter mode prediction can take place once for a whole partition,
+  // instead of each tx-block.
+  let num_planes = 1 + if has_chroma(bo, bsize, xdec, ydec) { 2 } else { 0 };
+  for p in 0..num_planes {
+    let plane_bsize = if p == 0 { bsize }
+    else { get_plane_block_size(bsize, xdec, ydec) };
+
+    let po = bo.plane_offset(&fs.input.planes[p].cfg);
+
+    let rec = &mut fs.rec.planes[p];
+
+    // TODO: make more generic to handle 2xN and Nx2 MC
+    if p > 0 && bsize == BlockSize::BLOCK_4X4 {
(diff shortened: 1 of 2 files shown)


fbossen:
This addresses the same issue as #526

fbossen:
Improvement at speed 0:

|     PSNR | PSNR Cb | PSNR Cr | PSNR HVS |     SSIM |  MS SSIM | CIEDE 2000 |
|     ---: |    ---: |    ---: |     ---: |     ---: |     ---: |       ---: |
| -11.8524 | -4.6238 | -4.7673 | -14.2262 | -11.0414 | -12.0377 |    -9.5923 |

fbossen:
Improvement at speed 1:

|     PSNR | PSNR Cb | PSNR Cr | PSNR HVS |    SSIM | MS SSIM | CIEDE 2000 |
|     ---: |    ---: |    ---: |     ---: |    ---: |    ---: |       ---: |
| -10.9594 | -5.8512 | -6.2345 | -11.8761 | -9.5286 | -9.9086 |    -8.4907 |

fbossen:
With this fix, it is beneficial to use tx type selection. Without this fix, it is detrimental.

ycho:
Oh, "encode_tx_block() overwrites prediction data", now I am seeing what's happening...
it is overwritten by idct an dequant, correct?!

ycho:
So, we have no way to avoid re-doing the same inter-prediction for each tx_type.

src/encoder.rs
```diff
@@ -1337,50 +1383,9 @@ pub fn encode_block_b(seq: &Sequence, fi: &FrameInvariants, fs: &mut FrameState,
...
-      {
-        let ref_frame = cw.bc.at(bo).ref_frames[0];
-        let mv = &cw.bc.at(bo).mv[0];
-        // Inter mode prediction can take place once for a whole partition,
-        // instead of each tx-block.
-        let num_planes = 1 + if has_chroma(bo, bsize, xdec, ydec) { 2 } else { 0 };
-        for p in 0..num_planes {
-            let plane_bsize = if p == 0 { bsize }
-            else { get_plane_block_size(bsize, xdec, ydec) };
+    motion_compensate(fi, fs, cw, luma_mode, ref_frame, mv, bsize, bo, bit_depth);
```
ycho: Shouldn't this be called only for inter mode block? i.e. inside the below if is_inter {...} ?

ycho:
I see that motion_compensate() is called for both intra and inter modes.
For ex, before calling rdo_tx_size_type() in several places. 
I thought motion_compensate() would be called for inter mode only.
Can you explain why ?

src/rdo.rs
```diff
@@ -430,6 +428,8 @@ pub fn rdo_tx_type_decision(
       continue;
     }
 
+    motion_compensate(fi, fs, cw, mode, ref_frame, mv, bsize, bo, bit_depth);
```
ycho: I see, now motion_compensate() is called here only.

PR #528, opened 2018-08-30
Return correct RDO cost when selecting no split

fbossen:


ycho:
Nice catch!
LGTM

ycho:
@rzumer , FYI.

PR #530, opened 2018-08-30
[WIP] Revert fbossen two commits aug30

ycho:
I am so sorry, I think I should have waited to see the AWCY results of these two patches.
There seems to have regressions by both patches, so trying to revert them.


ycho:
False Alarm!

PR #532, opened 2018-08-31
Fix RDO bug that was causing speed levels 1 and 2 to not work

fbossen:
Add assert statement to check that nearest MV is indeed the same
as the MV passed as argument to encode_block_b()

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1348,6 +1348,14 @@ pub fn encode_block_b(seq: &Sequence, fi: &FrameInvariants, fs: &mut FrameState,
                 MvSubpelPrecision::MV_SUBPEL_LOW_PRECISION
               };
               cw.write_mv(w, &mv, &ref_mv, mv_precision);
+            } else if luma_mode == PredictionMode::NEARESTMV {
+              if mv_stack.len() > 0 {
+                assert!(mv_stack[0].this_mv.row == mv.row);
+                assert!(mv_stack[0].this_mv.col == mv.col);
+              } else {
+                assert!(0 == mv.row);
+                assert!(0 == mv.col);
+              }
             }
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -1770,7 +1778,9 @@ fn encode_partition_topdown(seq: &Sequence, fi: &FrameInvariants, fs: &mut Frame
 
     match partition {
         PartitionType::PARTITION_NONE => {
-            let part_decision = if !rdo_output.part_modes.is_empty() {
+            // FIXME: For now always redo RDO since prior RDO may have relied on inaccurate
+            // neighbours
```
ycho: I am trying to understand which case redoing rdo_mode_decision()  is required.
I assume you have tested locally this patch stops the corruption.

fbossen: The case is where you have a first split. Then RDO is done on 4 smaller blocks. The best parameters for the last 3 smaller blocks assume that the previous smaller blocks were not split. But these blocks may actually end up being split. In such case the NEARESTMV may not be the same anymore, which causes the issue.

ycho: I think, in such case, even intra mode(s) of the 1st smaller block can change if it has split?

src/encoder.rs
```diff
@@ -1770,7 +1778,9 @@ fn encode_partition_topdown(seq: &Sequence, fi: &FrameInvariants, fs: &mut Frame
...
+            let part_decision = if !rdo_output.part_modes.is_empty() && false {
```
ycho: So, I am thinking that good chance of replacing 'false' here with sth better would be checking whether rdo_output.part_modes[0].bsize (does not exist currently) ==  current bsize.

PR #533, opened 2018-08-31
Change MV mode if NEAREST isn't NEAREST anymore

fbossen:


fbossen:
Addresses same issue as #532 but without speed penalty. RD penalty to be determined

fbossen:
| 1.8231 |  1.5903 |  2.0027 |   1.8724 | 1.8174 |  1.8226 |     1.7642 |

ycho:
Is above awcy results saying rergession? 
Otherwise,  just the comparison direction is this patch --> ref, and an improvement?

fbossen:
yes, it's an RD regression wrt to master at speed 1. But it runs 20-25% faster

PR #536, opened 2018-09-03
Trust the pointer arithmetic in cdef_filter_block

barrbrain:
This reduces the time for cdef_filter_frame by ~33%.
[AWCY results](https://beta.arewecompressedyet.com/?job=master-fd6d02d1d1b9f37d458105e3c70efede611da67c&job=rav1e-cdef-faster-unsafe%402018-09-03T15%3A04%3A33.425Z) at the default speed show a 6%~8% reduction in total encoding time.

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -116,9 +116,10 @@ fn constrain(diff: i32, threshold: i32, damping: i32) -> i32 {
 // Unlike the AOM code, our block addressing points to the UL corner
 // of the 2-pixel padding around the block, not the block itself.
 // The destination is unpadded.
-fn cdef_filter_block(dst: &mut [u16], dstride: i32, input: &[u16], istride: i32,
-                     pri_strength: i32, sec_strength: i32, dir: usize, pri_damping: i32,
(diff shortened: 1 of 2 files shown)


src/cdef.rs
```diff
@@ -116,9 +116,10 @@ fn constrain(diff: i32, threshold: i32, damping: i32) -> i32 {
 // Unlike the AOM code, our block addressing points to the UL corner
 // of the 2-pixel padding around the block, not the block itself.
 // The destination is unpadded.
-fn cdef_filter_block(dst: &mut [u16], dstride: i32, input: &[u16], istride: i32,
-                     pri_strength: i32, sec_strength: i32, dir: usize, pri_damping: i32,
-                     sec_damping: i32, xsize: i32, ysize: i32, coeff_shift: i32) {
+unsafe fn cdef_filter_block(dst: &mut [u16], dstride: isize, input: &[u16],
```
lu-zero: Would be sort of nicer if you mark only the loop unsafe.

barrbrain: The motivation for lifting it to the function level was to avoid adding another level of nesting.
This approach minimizes changed lines while admitting optimization of the inner loop.

PR #537, opened 2018-09-04
Use padded frames

fbossen:
Frames are extended to include padded samples.
Motion estimation and compensation functions are modified to take advantage of padding.


fbossen:
There is a small RD benefit because motion estimation is now unrestricted.

| -0.6387 | -0.4012 | -0.3765 |  -0.6075 | -0.6647 | -0.6308 |    -0.5219 |

PR #540, opened 2018-09-05
[WIP] Use Rayon to parallelize rdo

lu-zero:
It is more or less just an example on what we can do, it is currently ridiculously slow since I make some avoidable clones to not touch the encoder code.

- [x] refactor the code to use `for_each`
- [x] refactor the code to use `flat_map`
- [x] refactor the code to use `max_by`
- [x] use `par_iter()`
- [ ] rework the FrameState so you do not have to make that many clones of it
- [ ] make it actually fast.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -295,29 +295,40 @@ pub fn rdo_mode_decision(
       rdo_tx_size_type(seq, fi, fs, cw, bsize, bo, luma_mode, ref_frame, mv, false);
 
     // Find the best chroma prediction mode for the current luma prediction mode
-    mode_set_chroma.iter().for_each(|&chroma_mode| {
+    let best = mode_set_chroma.iter().flat_map(|&chroma_mode| {
+      let fs = &mut fs.clone();
+      let cw = &mut cw.clone();
+
       let mut cfl = CFLParams::new();
       if chroma_mode == PredictionMode::UV_CFL_PRED {
-        if !best_mode_chroma.is_intra() { return; }
-        let cw_checkpoint = cw.checkpoint();
-        let wr: &mut dyn Writer = &mut WriterCounter::new();
-        write_tx_blocks(
-          fi, fs, cw, wr, luma_mode, luma_mode, bo, bsize, tx_size, tx_type, false, seq.bit_depth, cfl, true
-        );
-        cw.rollback(&cw_checkpoint);
-        cfl = rdo_cfl_alpha(fs, bo, bsize, seq.bit_depth);
+        if best_mode_chroma.is_intra() {
+            let cw_checkpoint = cw.checkpoint();
+            let wr: &mut dyn Writer = &mut WriterCounter::new();
+            write_tx_blocks(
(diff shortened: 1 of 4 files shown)


src/rdo.rs
```diff
@@ -267,7 +267,9 @@ pub fn rdo_mode_decision(
   let mut mv_stack = Vec::new();
   let mode_context = cw.find_mvrefs(bo, LAST_FRAME, &mut mv_stack, bsize, false);
 
-  for &luma_mode in &mode_set {
+//  for &luma_mode in &mode_set {
+//  mode_set.par_iter().for_each(|&luma_mode| {
+  mode_set.iter().for_each(|&luma_mode| {
```
ycho: I think I see similar change in your other recent PR, which you explained as a 'intra' or 'prep'?

src/rdo.rs
```diff
@@ -328,20 +330,27 @@ pub fn rdo_mode_decision(
...
-          best_skip = skip;
-        }
-
         cw.rollback(&cw_checkpoint);
-      }
-    }
-  }
+
+        (rd, luma_mode, chroma_mode, cfl, ref_frame, mv, skip)
+      }).fold((best_rd, best_mode_luma, best_mode_chroma, best_cfl_params, best_ref_frame, best_mv, best_skip),
```
ycho: I read what 'fold' is for from https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold,
there it says fold accumulates the collection or sequence (thru the iteration) of values to the initial value.
Is initial value here a tuple? i.e. (best_rd, best_mode_luma, best_mode_chroma, best_cfl_params, best_ref_frame, best_mv, best_skip), where their values given at lines from 235 above?

src/rdo.rs
```diff
@@ -328,20 +330,27 @@ pub fn rdo_mode_decision(
...
+        |best, cur| {
```
ycho: And best and cur is a tuple?

src/rdo.rs
```diff
@@ -328,20 +330,27 @@ pub fn rdo_mode_decision(
...
+        if cur.0 < best.0 {
+          cur
+        } else {
+          best
+        }
+      });
+
+      best_rd = best.0;
```
ycho: Looks like my guess of best being a tuple or vector seems correct, since each field (best_rd, best_mode_luma, ... ) is indexed by .number...  What a rust!

src/rdo.rs
lu-zero: Yes, I'm trying to untangle bits so it is clearer, in the other set I used a proper struct instead of a tuple.

src/rdo.rs
lu-zero: yes :)

src/rdo.rs
lu-zero: exactly

src/rdo.rs
lu-zero: Notice that then you can spare all this noise using `min_by()` :)

rzumer:
Is this PR still needed or is it superseded by #984?

lu-zero:
It was an unrelated experiment, we can close it.

PR #546, opened 2018-09-06
Boost quantizer on key frames

fbossen:


fbossen:
| -4.3776 | -7.6207 | -8.3368 |  -4.3122 | -5.6248 | -5.2596 |    -6.6171 |

src/bin/rav1e.rs
```diff
@@ -61,10 +61,17 @@ fn main() {
   );
 
   loop {
-    //fi.frame_type = FrameType::KEY;
     fi.frame_type =
       if fi.number % 30 == 0 { FrameType::KEY } else { FrameType::INTER };
 
+    fi.base_q_idx =
+      if fi.frame_type == FrameType::KEY {
+        let q_boost = 15;
```
ycho: So, +15 (in QIndex) is the boosting factor for I frame.

ycho: Subtract it to frame QP if Key frame.

Issue #548, opened 2018-09-06
Can't compile custom rustfmt.


KyleSiefring:
nightly works now so I've been using that.

KyleSiefring:
I have a patch to change the Readme

PR #549, opened 2018-09-06
Update instructions for using rustfmt and clippy

KyleSiefring:
#548


mbebenita:
Looks good to me, assuming the patches we landed in rustfmt are part of nightly.

PR #559, opened 2018-09-08
Avoid searching out of bounds motion vectors in ME

### LINKED ISSUE #558: Failed to decode tile data
I use google translation.
When encoding with rav1e with the following command, it becomes a frame which can not be decoded correctly.

```rav1e --quantizer 100 pedestrian_area_1080p25.y4m --output pedestrian_area_1080p25_q100.ivf```

The error messages of libaom are as follows.

> Warning: Failed to decode frame 6: Corrupt frame detected
> Warning: Additional information: Failed to decode tile data

Input video
https://media.xiph.org/video/derf/
pedestrian_area_1080p25.y4m

Output video
https://www.dropbox.com/s/8behoitm2g6k10f/pedestrian_area_1080p25_q100.ivf?dl=0

rav1e b3283b8

### PR DESCRIPTION
**Pull request by fbossen**
Restrict motion vectors to lie within the bounds defined by the AV1 spec.

Fixes #558

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -9,6 +9,7 @@
 
 use context::BlockOffset;
 use context::BLOCK_TO_PLANE_SHIFT;
+use context::MI_SIZE;
 use partition::*;
 use plane::*;
 use FrameInvariants;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/me.rs [hunk 1]
```diff
@@ -51,10 +52,18 @@ pub fn motion_estimation(
       let range = 32 as isize;
       let blk_w = bsize.width();
       let blk_h = bsize.height();
-      let x_lo = po.x - range;
-      let x_hi = po.x + range;
-      let y_lo = po.y - range;
-      let y_hi = po.y + range;
+      let border_w = 128 + blk_w as isize * 8;
+      let border_h = 128 + blk_h as isize * 8;
+      let cols = (rec.frame.planes[0].cfg.width + MI_SIZE - 1) / MI_SIZE;
```
barrbrain: Perhaps the mutiplication and division by MI_SIZE could be replaced with shifts by MI_SIZE_LOG2.

fbossen: Compilers should be able to figure out that a division of an unsigned int by 4 (MI_SIZE) can be implemented with a shift.
I generally prefer using multiplications and divisions as there is no confusion about operator precedence

PR #561, opened 2018-09-10
Change limit in EncoderConfig from u64 to i64

ycho:


ycho:
Closed since #563 seems a better fix.

PR #562, opened 2018-09-10
Call is_intra() only once in mode decision rdo loop

ycho:


ycho:
If any, will make rav1e 0.01% faster.

src/rdo.rs
```diff
@@ -292,23 +292,22 @@ pub fn rdo_mode_decision(
...
-    if is_chroma_block
-      && luma_mode.is_intra()
-      && luma_mode != PredictionMode::DC_PRED
-    {
-      mode_set_chroma.push(PredictionMode::DC_PRED);
-    }
-
-    if is_chroma_block && luma_mode.is_intra() && bsize.cfl_allowed() {
-      mode_set_chroma.push(PredictionMode::UV_CFL_PRED);
+    if luma_mode_is_intra && is_chroma_block {
```
fbossen: Refactoring the if statements is a good change

src/rdo.rs
```diff
@@ -292,23 +292,22 @@ pub fn rdo_mode_decision(
     cw.find_mvrefs(bo, LAST_FRAME, &mut mv_stack, bsize, false);
 
   for &luma_mode in &mode_set {
-    assert!(fi.frame_type == FrameType::INTER || luma_mode.is_intra());
+    let luma_mode_is_intra = luma_mode.is_intra();
```
fbossen: Not sure there is any benefit in doing this. Do you have one in mind?

ycho: Yes, they are not for speed up, but simply meant to avoid calling .is_intra() functions, which gives the same info.

tdaede:
Dunno why Travis isn't updating on Github, but on travis it says it passed, so manually landing.

ycho:
:) We have to provide company who made Travis with this patch for use as fuzz testing!

ycho:
Thanks for merging!

PR #564, opened 2018-09-10
Add Display for Context

lu-zero:
And use it in rav1e to output the same stats it did before.


tdaede:
It's okay to land this even with Appveyor broken.

PR #568, opened 2018-09-11
Apply normative clamping to MV stack

fbossen:
See section 7.10.2.14 context and clamping process

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2430,6 +2430,20 @@ impl ContextWriter {
 
     /* TODO: Handle single reference frame extension */
 
+    // clamp mvs
+    for mv in mv_stack {
+      let blk_w = bsize.width();
+      let blk_h = bsize.height();
+      let border_w = 128 + blk_w as isize * 8;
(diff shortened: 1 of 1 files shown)


fbossen:
|   PSNR | PSNR Cb | PSNR Cr | PSNR HVS |    SSIM | MS SSIM | CIEDE 2000 |
|   ---: |    ---: |    ---: |     ---: |    ---: |    ---: |       ---: |
| 0.0001 | -0.0150 | -0.0001 |   0.0000 | -0.0003 | -0.0003 |    -0.0022 |

src/context.rs
```diff
@@ -2430,6 +2430,20 @@ impl ContextWriter {
 
     /* TODO: Handle single reference frame extension */
 
+    // clamp mvs
```
ycho: I think motion_estimation() can limit the mv range?

fbossen: Not sufficiently, because the range is not the same for all blocks. NEAREST may produce an MV that was ok for a previous block but not the current one.

PR #571, opened 2018-09-12
Implement segmentation

atomnuker:
Only works for now on keyframes and as long as the delta does not have to make the quantization cdfs change index.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -262,6 +262,25 @@ pub enum TxClass {
   TX_CLASS_VERT = 2
 }
 
+#[derive(Copy, Clone, PartialEq)]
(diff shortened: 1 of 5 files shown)


tdaede:
Looks fine, but how you are testing this?

atomnuker:
By setting fs.segmentation.features[0][SegLvl::SEG_LVL_ALT_Q as usize] = true; and adjusting the delta, of course.
Ready for merging, does not desync on inter frames anymore.

tdaede:
Needs rebase

tdaede:
I'd like more encode decode tests for this, but given how long this has been open, I think we should land it now.

PR #572, opened 2018-09-12
Fix the stats output

lu-zero:
The context status is updated on receive.

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -126,10 +126,9 @@ pub fn process_frame(ctx: &mut Context,
                 _ => panic! ("unknown input bit depth!"),
             }
 
-            eprintln!("{}", ctx);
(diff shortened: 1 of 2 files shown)


src/bin/common.rs
```diff
@@ -126,10 +126,9 @@ pub fn process_frame(ctx: &mut Context,
                 _ => panic! ("unknown input bit depth!"),
             }
 
-            eprintln!("{}", ctx);
             let _ = ctx.send_frame(input);
-
             let pkt = ctx.receive_packet().unwrap();
+            eprintln!("{}", ctx);
             write_ivf_frame(output_file, pkt.number as u64, pkt.data.as_ref());
```
barrbrain: At this point, ctx.fi.number and pkt.number differ by one but ctx.fi.frame_type matches the encoded packet.

PR #573, opened 2018-09-12
[WIP] Beautify the code for ME search range setup

ycho:
Rewrite some part of ME search range setup code easier to read.

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -52,18 +52,19 @@ pub fn motion_estimation(
       let range = 32 as isize;
       let blk_w = bsize.width();
       let blk_h = bsize.height();
-      let border_w = 128 + blk_w as isize * 8;
-      let border_h = 128 + blk_h as isize * 8;
-      let cols = (rec.frame.planes[0].cfg.width + MI_SIZE - 1) / MI_SIZE;
-      let rows = (rec.frame.planes[0].cfg.height + MI_SIZE - 1) / MI_SIZE;
-      let x_min = -(bo.x as isize) * (8 * MI_SIZE) as isize - border_w;
-      let x_max = (cols - bo.x - blk_w / MI_SIZE) as isize * (8 * MI_SIZE) as isize + border_w;
-      let y_min = -(bo.y as isize) * (8 * MI_SIZE) as isize - border_h;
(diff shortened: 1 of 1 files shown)


src/me.rs
```diff
@@ -93,6 +94,11 @@ pub fn motion_estimation(
         steps.push(1);
       }
 
+      let x_min = 8 * (-(bo.x as isize) * MI_SIZE as isize - border_w);
```
ycho: Would renaming these four variables to specifying the ME search range in four directions as: left_offset, right_offset, up_offset and down_offset ever confuse you? &_&

src/me.rs
```diff
@@ -52,18 +52,19 @@ pub fn motion_estimation(
...
+      let border_w = 16 + blk_w as isize; // in full pel unit
+      let border_h = 16 + blk_h as isize;
+
+      let x_lo = po.x - range;
+      let x_hi = po.x + range;
+      let y_lo = po.y - range;
+      let y_hi = po.y + range;
+
+      let x_lo = x_lo.max(-border_w);
+      let x_hi = x_hi.min((fi.w_in_b * MI_SIZE) as isize + border_w);
```
fbossen: This should include "- blk_w"?
I am not a fan of duplicating computations here. It should be best to define a MV range limit and then use this limit to determine the range for the initial ME

ycho: Please see that I directly clip on x_hi (i.e. last search pixel on right, which is frame size aligned in MI_size + border_w, correct?) instead of range (i.e. offset).
Yes, I rephrased it for easier read, but x_hi = .. are duplicated thus can waste cpu cycles. 
Let me try how to not do that. (btw, pls see the trade-in that I used fi.w_in_b! :) )

fbossen: x_hi + blk_w should be less than fi.w_in_b * MI_SIZE + border_w, thus blk_w should be in there somewhere. The use of fi.w_in_b is good (had missed that previously)

ycho: Oh, right- max ME search range should exclude block size itself.

ycho:
Closing this PR since PR #574 which does less invasive renaming replaces this.

PR #575, opened 2018-09-12
Rdo iterators

lu-zero:
Some propedeutics to the rayon experiments.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -332,15 +332,19 @@ pub fn rdo_mode_decision(
       seq, fi, fs, cw, bsize, bo, luma_mode, ref_frame, mv, false,
     );
 
+    let all_skips = [false, true];
+
+    // Don't skip when using intra modes
+    let skips = if luma_mode_is_intra {
+        &all_skips[..1]
+    } else {
+        &all_skips[..]
+    };
+
     // Find the best chroma prediction mode for the current luma prediction mode
     for &chroma_mode in &mode_set_chroma {
-      for &skip in &[false, true] {
-        // Don't skip when using intra modes
-        if skip && luma_mode_is_intra {
-          continue;
-        }
-
-        let mut wr: &mut dyn Writer = &mut WriterCounter::new();
+      skips.iter().for_each(|&skip| {
+        let wr: &mut dyn Writer = &mut WriterCounter::new();
         let tell = wr.tell_frac();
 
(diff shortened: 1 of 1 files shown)


src/rdo.rs
```diff
@@ -291,9 +291,9 @@ pub fn rdo_mode_decision(
   let mode_context =
     cw.find_mvrefs(bo, LAST_FRAME, &mut mv_stack, bsize, false);
 
-  for &luma_mode in &mode_set {
+  mode_set.iter().for_each(|&luma_mode| {
```
ycho: I'd like to learn what would be benefits by doing this way?

lu-zero: It is an intermediate step to get to use iter_par and have rayon spawn a thread for each iteration.

Tomorrow I'll re-read the code since sleep-rebasing did not work that well :)

lu-zero: Ideally from here we could go this way:

`for loop` -> `iter().for_each()` -> `iter().map().min_by()` -> `par_iter().map().min_by`.

Ideally evaluating each luma mode in parallel and then having a pass computing the best.

This would work better if we have the least amount of mutable state across and @barrbrain is already working on reducing it.

Using `flat_map()` so we could have each chroma evaluation going in parallel as well would require even more effort so I stopped thinking about it right now.

benches/bench.rs
```diff
@@ -13,6 +13,7 @@ extern crate rand;
 extern crate rav1e;
 
 mod predict;
+mod rdo;
```
barrbrain: This straggler breaks the check on Travis CI.

src/rdo.rs
```diff
@@ -362,17 +376,18 @@ pub fn rdo_mode_decision(
...
-          continue;
-        }
+      let decision: &[bool] = if luma_mode_is_intra {
+        &[false]
+      } else {
+        &[false, true]
+      };
 
-        let mut wr: &mut dyn Writer = &mut WriterCounter::new();
+      best = decision.iter().map(|&skip| {
```
barrbrain: Unconditionally setting `best` here always chooses the last luma mode.

lu-zero:
Rebased and reduced so hopefully we can land this so we do not have to rebase this part anymore :)

lu-zero:
Some more commits, now that I had time to test them. Few little steps further :)

PR #578, opened 2018-09-13
[WIP] If predicted me search center is used, reduce search range with full pel step

ycho:
Reduce the search range with full pel step size to full pel, if predicted me search center is used.

Compared to original predicted mv method, this change cause less regressions while keep helping high speed motion like MINECRAFT and wikipedia.

[AWCY for original pmv](https://beta.arewecompressedyet.com/?job=master-ff5bbc33b1cd825ba74947c89a0aa9b744ae20dd&job=use_pmv%402018-09-12T22%3A44%3A36.066Z)
vs
[AWCY for this patch, i.e. adjusted search space and step](https://beta.arewecompressedyet.com/?job=master-ff5bbc33b1cd825ba74947c89a0aa9b744ae20dd&job=use_pmv_exp1%402018-09-12T23%3A28%3A22.404Z)

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -49,7 +49,12 @@ pub fn motion_estimation(
         x: (bo.x as isize) << BLOCK_TO_PLANE_SHIFT,
         y: (bo.y as isize) << BLOCK_TO_PLANE_SHIFT
       };
-      let range = 32 as isize;
+      let mut range = 32 as isize;
+      let mut step = 2;
+      if bsize < BlockSize::BLOCK_32X32 {
+        range = 16;
+        step = 1;
+      }
       let blk_w = bsize.width();
       let blk_h = bsize.height();
       let border_w = 128 + blk_w as isize * 8;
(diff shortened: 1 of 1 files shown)


src/me.rs
```diff
@@ -49,7 +49,12 @@ pub fn motion_estimation(
         x: (bo.x as isize) << BLOCK_TO_PLANE_SHIFT,
         y: (bo.y as isize) << BLOCK_TO_PLANE_SHIFT
       };
-      let range = 32 as isize;
+      let mut range = 32 as isize;
+      let mut step = 2;
+      if bsize < BlockSize::BLOCK_32X32 {
```
lu-zero: maybe:

``` rust
let (range, step) = if bsize < BlockSize::BLOCK_32X32 { (16, 1) } else (32, 2) };
```

src/me.rs
```diff
@@ -85,8 +90,12 @@ pub fn motion_estimation(
 
       let mode = PredictionMode::NEWMV;
       let mut tmp_plane = Plane::new(blk_w, blk_h, 0, 0, 0, 0);
-
-      let mut steps = vec![8, 4, 2];
+      let mut steps = vec![];
```
lu-zero: We could do something like this:
``` rust
let steps = [8, 4, 2, 1];
let range = if bsize >= BlockSize::BLOCK_32X32 { 0 } else { 1 } .. if fi.allow_high_precision_mv { 4 } else 3;

for step in &steps[range] {
```
It is possibly more compact as long we are trimming a slice :)

src/me.rs
ycho: Thanks- I am not rustic enough... yet..

src/me.rs
ycho: This is very rustic.
So, the range will be one of four cases:
[8,4,2,1], [4,2,1], [8,4,2], [4,2].
I am now rustic.

src/me.rs
```diff
@@ -85,8 +90,12 @@ pub fn motion_estimation(
...
+      if bsize >= BlockSize::BLOCK_32X32 {
```
fbossen: should the condition be step >= 2?

PR #585, opened 2018-09-17
Add encoding of order_hint

fbossen:
For now a single bit is used for encoding order_hint

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -103,9 +103,11 @@ impl Context {
         deblock: Default::default(),
       };
 
-      self.fi.order_hint = (self.fi.number % 30) as u32;
+      let frame_number_in_segment = self.fi.number % 30;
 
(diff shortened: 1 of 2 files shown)


src/api.rs
```diff
@@ -103,7 +103,9 @@ impl Context {
         deblock: Default::default(),
       };
 
-      self.fi.frame_type = if self.fi.number % 30 == 0 {
+      self.fi.order_hint = (self.fi.number % 30) as u32;
+
+      self.fi.frame_type = if self.fi.order_hint == 0 {
```
tdaede: LGTM too. Seems a bit weird to derive frame_type from order_hint but I guess it doesn't really matter.

rzumer: I agree that from a logical/readability point of view it might be worth using an intermediary variable and setting both `order_hint` and `frame_type` based on its value.

Issue #418, opened 2018-08-02
Run entropymode.c through corrode


tdaede:

lu-zero:
I already have the c2rust version of it handy. What do you want to do with it :)

ycho:
What's this??

lu-zero:
**libaom** entropymode machine-converted to something `rustc` can digest, it is useful mainly to start moving away from linking to it.

PR #587, opened 2018-09-18
[WIP] Port entropymode.c #418

jgh-:
I'm opening this PR a bit early because it would be good to get some feedback on what the goal of #418 is as well as discussion on a few notes I had while putting this together.

The current state of this branch is that it builds and it seems to work and tests are passing.  So far it is simply replacing the tables present in entropymode.c and removing those as dependencies in context.rs.

There are several functions in entropymode.h and entropymode.c, a couple seem to be implemented already in context.rs and the largest function, `av1_get_palette_color_index_context`, does not appear to be referenced at all in the rust code at this time (I don't see much if any code related to palettes, color indices, or color maps).

Additionally, I found a few inconsistencies / things that might be made clearer:

- `BLOCK_SIZES_ALL` is `24` in Rust, but `22` in C and the AV1 spec.
    - There are two additional block sizes present in rust that aren't in C or the spec: `BLOCK_128X32` and `BLOCK_32X128`
- `TX_SIZES` is defined twice and has two different values. Once as `TxSize::TX_SIZES` and again as `TX_SIZES` in context.rs.  The second one appears to correspond to `EXT_TX_SIZES` in the C code. I've changed the name in the rust code to match the C code to avoid confusion.
- `REFMV_MODE_CONTEXTS` is `9` in the Rust code, but `6` in the C code and spec.  I've changed that value, but I'm not sure if it has side-effects.

Still to be done: the CDFs in token_cdfs.h


src/entropymode.rs
```diff
@@ -0,0 +1,707 @@
...
+const MAX_TX_CATS: usize = (TxSize::TX_SIZES - TX_SIZE_CTX_MIN);
+const MAX_TX_DEPTH: usize = 2;
+
+
+// The spec indicates that BLOCK_SIZES should be 22, but BLOCK_SIZES_ALL is defined as 24 in the rust code.
+// There are also two additional symbols that do not appear in the spec: BLOCK_128X32 and BLOCK_32X128.
+
+const BLOCK_SIZES:usize = BlockSize::BLOCK_SIZES_ALL - 2;
+
+//TODO: Nice to have (although I wasnt able to find a way to do it yet in rust): zero-fill arrays that are
```
rzumer: Indeed, manual array padding seems to be the best solution right now.

src/entropymode.rs
```diff
@@ -0,0 +1,704 @@
+#![allow(non_upper_case_globals)]
```
rzumer: Can you please add a header to match (most of) the other source files?

src/token_cdfs.rs
```diff
@@ -0,0 +1,3525 @@
+#![allow(non_upper_case_globals)]
```
rzumer: Here as well.

src/entropymode.rs
jgh-: Yep, good catch. Done.

rzumer: Just nitpicking, but there's an extra newline at the start of this file.

jgh-: ah so there is!

rzumer: Thanks. :+1:

src/context.rs
```diff
@@ -68,11 +68,14 @@ pub const DIRECTIONAL_MODES: usize = 8;
 pub const KF_MODE_CONTEXTS: usize = 5;
 
 pub const EXT_PARTITION_TYPES: usize = 10;
-pub const TX_SIZES: usize = 4;
+
+//pub const TX_SIZES: usize = TxSize::TX_SIZES;
+pub const EXT_TX_SIZES: usize = 4;
```
ycho: We used to use the original name as in libaom, 'EXT_TX_SIZES' but renamed it to TX_SIZES.
Rav1e don't need to use the same name, and further more av1 spec does not use any 'EXT_', 
simply because all those 'EXT_' is default in av1.
Similar changes have applied to other variable or function names having 'EXT' or 'ext', I remember.

src/partition.rs
```diff
@@ -427,7 +427,7 @@ pub struct MotionVector {
 
 pub const NEWMV_MODE_CONTEXTS: usize = 7;
 pub const GLOBALMV_MODE_CONTEXTS: usize = 2;
-pub const REFMV_MODE_CONTEXTS: usize = 9;
+pub const REFMV_MODE_CONTEXTS: usize = 6;
```
ycho: @fbossen, Can you check this pls?

src/context.rs
jgh-: Is having two constants called TX_SIZES with different values desirable, or would it make sense to put the second TX_SIZES in a more descriptive namespace (or something like that)?

ycho: Having two different definition for TX_SIZES is clearly wrong. Also, its value should be 5 as you probably stated. I briefly checked  that where TX_SIZES defined in context.rs is used (It is static so only can be used in that file), then several multi-dim cdf definition referenced it and puzzled how could they worked correctly so far (Because all those cdfs are imported from libaom as extern .h or .c!
Btw, there is one reason why TX_SIZES = 4 has been used. That is, Rav1e has not enabled TX_64X64. So the tx size index 4 has nowhere been used in the code, as index to any array or etc.

jgh-: In that case it seems like there should be more data for some of these cdf tables, since I pulled these straight from the C code and using a value of 5 does not work. The C code uses EXT_TX_SIZES which is 4 (and has the second TX_SIZES = 5 used in other places)  Does it make sense to 0-fill the tables as necessary in case at some point those values become available?

jgh-: I think I'm beginning to see how EXT_TX_SIZES is being used with relation to the intra/inter tx sets.  I will take another run at it in the morning.

src/context.rs
```diff
@@ -52,30 +54,39 @@ pub const MAX_TX_SIZE: usize = 32;
...
 
-const EXT_PARTITION_TYPES: usize = 10;
-const TX_SIZES: usize = 4;
-const TX_SETS: usize = 9;
-const TX_SETS_INTRA: usize = 3;
-const TX_SETS_INTER: usize = 4;
+pub const EXT_PARTITION_TYPES: usize = 10;
 
+//pub const TX_SIZES: usize = TxSize::TX_SIZES;
+pub const EXT_TX_SIZES: usize = 4;
```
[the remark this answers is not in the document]
jgh-: There's already a TX_SIZE_CONTEXTS and its value is 3.  The arrays that use this value seem to be combined arrays of the `Default_Intra_Tx_Type_Set` tables in the spec.  So maybe TX_TYPE_SET_CONTEXTS would work?

src/entropymode.rs
```diff
@@ -0,0 +1,703 @@
...
+const TX_SIZE_CTX_MIN: usize = (TX_SIZE_LUMA_MIN + 1);
+const MAX_TX_CATS: usize = (TxSize::TX_SIZES - TX_SIZE_CTX_MIN);
+const MAX_TX_DEPTH: usize = 2;
+
+// LUTS ---------------------
+
+// The spec indicates that BLOCK_SIZES should be 22, but BLOCK_SIZES_ALL is defined as 24 in the rust code.
+// There are also two additional symbols that do not appear in the spec: BLOCK_128X32 and BLOCK_32X128.
+
+const BLOCK_SIZES:usize = BlockSize::BLOCK_SIZES_ALL - 2;
```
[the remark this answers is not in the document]
jgh-: OK. I will raise an issue for that since it's probably out of scope for this PR.

src/context.rs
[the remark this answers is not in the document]
tdaede: Makes sense to me.

src/entropymode.rs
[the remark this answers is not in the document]
tdaede: OK. Even the EXT_TX_SIZES rename could be made separate, though it's pretty simple to include in this patch.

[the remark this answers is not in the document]
jgh-: Fair enough, that one is pretty straightforward though. Just renaming a variable in 6 or 7 spots in the code (most of them being in entropymode.rs).  I took a quick try at the BLOCK_SIZES_ALL change and it would be a bit more involved - there are some arrays and things that rely on the value being 24, for example.

jgh-:
ok i think this is ready to go pending any further change requests, i may continue hacking away at separating rav1e from libaom but will make more PRs as needed.  Some of them are bigger jobs (transforms, scans).

jgh-:
Yeah I can focus on relevant components first now that I'm getting more comfortable with navigating the codebase.  Maybe worth opening an issue for discussion on remaining porting work to separate rav1e from libaom.

ycho:
You may squash and merge this PR, and hopefully edit the commit message when doing so.
(You probably know that you have chance to edit it after pressing "Squash and merge" button down below)

PR #589, opened 2018-09-18
Predicted ME search center for bottom-up partition rdo

ycho:
[AWCY results for 5 frames at speed 0](https://beta.arewecompressedyet.com/?job=master-s0_5f%402018-09-18T20%3A00%3A40.807Z&job=predicted_me_center_bottom_up_rdo-s0_5f%402018-09-18T20%3A01%3A32.697Z)

| -1.8632 | -1.7386 | -1.7199 |  -1.8155 | -1.8189 | -1.7893 |    -1.8325 |

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1640,7 +1640,7 @@ pub fn write_tx_tree(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut Context
 
 fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut FrameState,
                              cw: &mut ContextWriter, w_pre_cdef: &mut dyn Writer, w_post_cdef: &mut dyn Writer,
-                             bsize: BlockSize, bo: &BlockOffset) -> f64 {
+                             bsize: BlockSize, bo: &BlockOffset, pmv: &MotionVector) -> f64 {
     let mut rd_cost = std::f64::MAX;
 
     if bo.x >= cw.bc.cols || bo.y >= cw.bc.rows {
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -1970,9 +1970,10 @@ fn encode_tile(sequence: &mut Sequence, fi: &FrameInvariants, fs: &mut FrameStat
 
             // Encode SuperBlock
             if fi.config.speed == 0 {
+                let pmv = MotionVector { row: 0, col: 0 };
```
tdaede: Why don't we do this for both directions?

ycho: Can you explain what do you mean by "both directions"? 
If so, two reference frames? It is not landed yet.

tdaede: Er nevermind, I didn't understand the patch.

ycho:
If anyone objects to landing this patch, pls let me know in 1 hour! Otherwise, it will land!

ycho:
Copying derf's response from #daala IRC, since he seems not using gihub and assume LGTM from him

1:34 PM <codeview> derf: It seems predicted ME search center also works fine for bottom-up (i.e. speed 0) partition search. Got 1.86% for 5 frames.
2:33 PM <+derf> codeview: Bottom-up propagation or top-down propagation in a bottom-up search?
2:34 PM <codeview> top-down propagate in bottom-up earch
2:34 PM <codeview> bottom-up propagate is what I expected would not work so deferred.
2:41 PM <+derf> codeview: Glad to hear it works.

src/encoder.rs
tterribe: It would be nice to predict from our superblock neighbors, e.g., median-of-three, but even just the left (or top) superblock would do. But no objections to landing this patch as-is and trying that in a separate patch.

tterribe:
Yes, r=me.

P.S., I block all e-mail from Github since it spams me way too much for projects I'm not even contributing to, so if you want me to see something, please say so.

src/encoder.rs
ycho: Thanks! Your suggestion of having SB neighbor's MVs as reference for ME search center is w/o doubts we want to have it. I (or someone who cannot wait my rust code writing speed) will write it as a separate patch.

PR #591, opened 2018-09-19
Deblock opt3

xiphmont:
optimization work that is also a test run for a pattern that's part of real RDO for deblocking.


src/deblock.rs
```diff
@@ -21,26 +21,31 @@ use FrameType;
...
     let idx = if pli == 0 { pass } else { pli+1 };
-    // By-block filter strength delta, if the feature is active.
-    let block_delta = if deblock.block_delta_multi {
-        block.deblock_deltas[ idx ] << deblock.block_delta_shift
-    } else {
-        block.deblock_deltas[ 0 ] << deblock.block_delta_shift
-    };
+    let mut level = deblock.levels[idx];
+
+    if deblock.block_deltas_enabled {
```
lu-zero: if you feel like chiseling the code:

``` rust
let level =  if deblock.block_deltas_enabled { 
...
} else {
  deblock_levels[idx]
};
```

Issue #592, opened 2018-09-19
BLOCK_SIZES_ALL should be 22, and BLOCK_128X32 and BLOCK_32X128 removed


jgh-:
See discussion in #587 .  These values don't match the spec, BLOCK_128X32 and BLOCK_32X128 don't exist in the spec so they should be removed and BLOCK_SIZES_ALL count should be 22 (not 24).

PR #593, opened 2018-09-19
Remove BLOCK_128X32 and BLOCK_32X128 #592

jgh-:
BLOCK_SIZES is 22, and BLOCK_128X32 and BLOCK_32X128 do not exist in the AV1 spec.  This PR removes them.  Tests pass and the nyancat video seems to work. Issue #592


ycho:
LGTM, thanks for catching and fixing this!

PR #600, opened 2018-09-22
[PoC] Add AVX2 version of get_sad

tux3:
While playing around with rav1e I noticed it was spending a lot of time in the `get_sad` function, and vTune said further vectorization might help, so here's a rough hack to help rustc a little.

This should produce the same output as get_sad, and I think I've handled non-multiples of 8 sizes correctly, but this is definitely unsafe and worth double-checking!

As for results, rav1e now runs ~10% faster on my machine (encoding 24 frames at 1080p):

```
$ time ./target/release/rav1e -l 24 input.y4m -o output_short.ivf
```

|   | User | System | Real time |
| ------------- | ------------- | -- | -- |
| Before  | 128,33s | 0,37s | 2:08,73  |
| After  | 113,17s | 0,30s | 1:53,52 |

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -15,17 +15,16 @@ use plane::*;
 use FrameInvariants;
 use FrameState;
 
-#[inline(always)]
-pub fn get_sad(
+fn get_sad_generic(
   plane_org: &mut PlaneSlice, plane_ref: &mut PlaneSlice, blk_h: usize,
-  blk_w: usize
+  blk_w: usize, x_offset: usize
 ) -> u32 {
   let mut sum = 0 as u32;
 
   for _r in 0..blk_h {
     {
-      let slice_org = plane_org.as_slice_w_width(blk_w);
-      let slice_ref = plane_ref.as_slice_w_width(blk_w);
+      let slice_org = &plane_org.as_slice_w_width(blk_w)[x_offset..];
+      let slice_ref = &plane_ref.as_slice_w_width(blk_w)[x_offset..];
       sum += slice_org
         .iter()
         .zip(slice_ref)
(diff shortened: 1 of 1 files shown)


src/me.rs
```diff
@@ -15,11 +15,87 @@ use plane::*;
...
+  blk_w: usize
+) -> u32 {
+  #[cfg(target_arch = "x86")]
+  use std::arch::x86::*;
+  #[cfg(target_arch = "x86_64")]
+  use std::arch::x86_64::*;
+
+  let mut sum = 0 as u32;
+
+  {
```
lu-zero: Why this block?

src/me.rs
```diff
@@ -15,11 +15,87 @@ use plane::*;
...
+    }
+
+    let v32x4lower = _mm256_castsi256_si128(v32x8sum);
+    let v32x4upper = _mm256_extracti128_si256(v32x8sum, 1);
+    let mut v32x4sum = _mm_hadd_epi32(v32x4lower, v32x4upper);
+    v32x4sum = _mm_hadd_epi32(v32x4sum, v32x4sum);
+    sum += _mm_extract_epi32(v32x4sum, 0) as u32;
+    sum += _mm_extract_epi32(v32x4sum, 1) as u32;
+
+    // Hopefully cold, unlikely, and rarely executed slow path
```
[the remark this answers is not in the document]
lu-zero: Probably we could refactor `motion_estimation` and have the motion_estimation generic over the block as we did for the prediction.

src/me.rs
tux3: The intent is to drop the borrows `slice_org/slice_ref` have on `plane_org/plane_ref`, for the `plane_*.y += blk_h` at the end.

tux3:
I've tried to factor out the common code in a `get_sad_generic`.

sse2 shouldn't be too hard to write, but rustc should already be able to autovectorize the iter/zip/map loop, it's just not the most efficient implementation.

src/me.rs
```diff
@@ -39,6 +38,79 @@ pub fn get_sad(
...
+    size = ::std::cmp::min(slice_org.len(), slice_ref.len()) as isize;
+
+    let mut ptr_org = slice_org.as_ptr();
+    let mut ptr_ref = slice_ref.as_ptr();
+
+    let mut v32x8sum = _mm256_set1_epi8(0);
+
+    for _r in 0..blk_h {
+      let mut pos = 0;
+      while pos <= size-8 {
```
tux3: (I think I could replace that by `for pos in (0..=size-8).step_by(8)`, but I'm not sure which is more readable)

lu-zero:
I prepare [this](https://github.com/xiph/rav1e/pull/610/commits/028900dbebf13101731d9aa6ec6e84f418810512) to make easier to reason about get_sad.

Probably we should see if making specialized version helps. I'll try to come up with something the next days.

src/me.rs
```diff
@@ -39,6 +38,79 @@ pub fn get_sad(
...
+        let v16x8org = _mm_load_si128(ptr_org.offset(pos) as *const _);
+        let v16x8ref = _mm_load_si128(ptr_ref.offset(pos) as *const _);
+        let v32x8org = _mm256_cvtepu16_epi32(v16x8org);
+        let v32x8ref = _mm256_cvtepu16_epi32(v16x8ref);
+        let v32x8diff = _mm256_sub_epi32(v32x8org, v32x8ref);
```
KyleSiefring: org and ref will contain either 12/10/8-bit values so it's safe to subtract and abs before converting to 32 bits. If you are working with 128-bit vectors, it's best to use madd to convert to 32-bits and add horizontally. That being said, it's faster to add a few lines as 16 bits.

src/me.rs
```diff
@@ -39,6 +38,79 @@ pub fn get_sad(
...
+        pos += 8;
+      }
+
+      ptr_org = ptr_org.add(plane_org.plane.cfg.stride);
+      ptr_ref = ptr_ref.add(plane_ref.plane.cfg.stride);
+    }
+
+    let v32x4lower = _mm256_castsi256_si128(v32x8sum);
+    let v32x4upper = _mm256_extracti128_si256(v32x8sum, 1);
+    let mut v32x4sum = _mm_hadd_epi32(v32x4lower, v32x4upper);
```
KyleSiefring: You can use a regular add here, they are faster...

src/me.rs
```diff
@@ -39,6 +38,79 @@ pub fn get_sad(
...
+    v32x4sum = _mm_hadd_epi32(v32x4sum, v32x4sum);
```
KyleSiefring: Much faster. It's actually faster to shift the whole vector (_mm_bsrli_si128/_mm_srli_si128 (they are the same)) and do a normal add than to use hadd.

tux3:
I don't think I'll have time to rebase this right now, feel free to close if it's too much of a bother to fix up.

lu-zero:
If nobody beats me at it I'll rebase it during the next weekend.

KyleSiefring:
SSE assembly was added and the plan now is to do everything in assembly so I'm closing this for now.

PR #603, opened 2018-09-23
Remove unnecessary unwrap() calls.

negge:


rzumer:
I thought that these were options by design.

PR #605, opened 2018-09-23
BlockSize Cleanup

mbebenita:


mbebenita:
I guess, let's land this as is, and I'll continue with more refactoring later.

PR #608, opened 2018-09-24
eliminate unneded clamp in narrow deblocking filters

xiphmont:
An exhaustive check indicates that over the given parameter range, one of the clamp operations in the narrow deblocking filters is unneeded.


ycho:
LGTM.
I think adding some asserts here and there can possibly help you ensure not to be thrilled to see more of same kinds?! :)

xiphmont:
Asserts are a good idea; added.

PR #610, opened 2018-09-25
Use an iterator in get_sad

lu-zero:
- [x] write a benchmark for it
- [x] write the simplest possible iterator
- [x] benchmark
- [x] write a proper unit test for `get_sad()` (#611)
- [x] add `size_hint()` and additional functions to speed it up further

(I tested by comparing the `rav1e` output)

### DIFF
--- a/src/plane.rs
+++ b/src/plane.rs
@@ -232,7 +232,6 @@ pub struct IterWidth<'a> {
     width: usize,
 }
 
-// TODO: Implement more methods
 impl<'a> Iterator for IterWidth<'a> {
     type Item = &'a [u16];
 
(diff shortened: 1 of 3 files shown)


src/plane.rs
```diff
@@ -220,12 +220,38 @@ impl Plane {
...
   pub x: isize,
   pub y: isize
 }
 
+pub struct IterWidth<'a> {
+    ps: PlaneSlice<'a>,
+    width: usize,
+}
+
+// TODO: Implement more methods
```
rzumer: Could you just clarify what methods you are referring to, here?

rzumer: OK, I was just wondering if it was worth to indicate the methods that would be useful to implement in this particular case within the comment. If we want to implement all remaining methods then I guess it is not so important.

barrbrain: If that means we no longer have a clear intent to add more methods then we should drop the TODO.

lu-zero: Done :)

PR #612, opened 2018-09-25
Update copyright line in LICENSE.

negge:


rzumer:
The exact text in my commit is detected correctly in a fresh repository, but Github seems to cache license information, and it is not always clear how to make it refresh. If the license is still not detected after a few days I would suggest to contact Github to do it.

PR #614, opened 2018-09-25
Clean up for adding tiles

negge:


ycho:
Oh- you are the man who is going to implement tiles! Amazing!

ycho:
Sure- I wanted to consider tiling + multi-threading before but was hold off because it was thought as less priority at that time.

negge:
Got it, thanks :)

src/bin/common.rs
```diff
@@ -58,6 +58,17 @@ impl FromCli for EncoderConfig {
                 .possible_values(&Tune::variants())
                 .default_value("psnr")
                 .case_insensitive(true))
+            // Uncomment when we support tiles
+            /*.arg(Arg::with_name("TILEROWS")
+                .help("Number of tile rows log2")
+                .long("tile-rows")
+                .takes_value(true)
+                .default_value("0"))
```
lu-zero: You may add a range check since the amount of tiles is limited by `MAX_TILE_ROWS` and `MAX_TILE_COLS` and the frame dimension (we should pick sane minimum tile IMHO).

We might want to provide a `use-tiles` option that takes a boolean and decides the cpu-optimal tiles later.

I guess we could write a blueprint issue about tiles instead using the patch review though...

negge: I have added validation at the bottom where config.quantizer and config.speed are validated.  I'll move to an Arg based validator in a subsequent commit.

tdaede:
please rebase

rzumer:
Tiling support has been merged in a separate set of changes, so closing this PR, as I believe this one is no longer relevant.

PR #616, opened 2018-09-26
More Benchmarks (transform and me)

lu-zero:
Adding more is left to the reader :)


lu-zero:
Now with the project coding style for the second patch.

PR #623, opened 2018-09-27
Fix inter for 12-bit input.

### LINKED ISSUE #622: decode error at 12 bit caused by 1d697a41fdf7ebb56549cc766a6597228917dac9
bisection reveals 1d697a41fdf7ebb56549cc766a6597228917dac9 introduces a decode desync (though not stream desync) at 12 bit -s3 nyancat (the test run in build.sh, though not run by default).

KyleSiefring:
12-bit is a special case that rounds differently.

See 7.11.3.2. Rounding variable derivation process in the spec.

Fixes #622

### DIFF
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -931,6 +931,12 @@ impl PredictionMode {
         let max_sample_val = ((1 << bit_depth) - 1) as i32;
         let y_filter_idx = if height <= 4 { 4 } else { 0 };
         let x_filter_idx = if width <= 4 { 4 } else { 0 };
+        let (shift_1, shift_2) = {
+          let shift_offset = if bit_depth == 12 { 2 } else { 0 };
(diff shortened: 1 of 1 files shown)


src/partition.rs
```diff
@@ -931,6 +931,12 @@ impl PredictionMode {
         let max_sample_val = ((1 << bit_depth) - 1) as i32;
         let y_filter_idx = if height <= 4 { 4 } else { 0 };
         let x_filter_idx = if width <= 4 { 4 } else { 0 };
+        let (shift_1, shift_2) = {
```
rzumer: Here you can assign a single variable name for the tuple and access them through `shift.0` and `shift.1`. It might be a bit cleaner.

KyleSiefring: Done

lu-zero:
Thank you a lot :)

PR #626, opened 2018-09-28
Format SUBSIZE_LOOKUP slightly then skip it

KyleSiefring:


KyleSiefring:
Or does it?

lu-zero:
Looks good enough to me and to travis :)

PR #629, opened 2018-09-30
Frame reordering

fbossen:
Group frames in sets of 4. Encode the far one first and then the other three using a forward and a backward reference.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -146,9 +146,11 @@ impl Context {
   pub fn frame_properties(&mut self, idx: u64) -> bool {
     let key_frame_interval: u64 = 30;
 
-    let pyramid_depth = 1;
+    let reorder = false;
+
+    let pyramid_depth = if reorder { 1 } else { 0 };
     let group_src_len = 1 << pyramid_depth;
-    let group_len = group_src_len + pyramid_depth;
+    let group_len = group_src_len + if reorder { pyramid_depth } else { 0 };
     let segment_len = 1 + (key_frame_interval - 1 + group_src_len - 1) / group_src_len * group_len;
 
     let idx_in_segment = idx % segment_len;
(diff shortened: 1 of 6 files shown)


src/api.rs
```diff
@@ -54,15 +54,17 @@ impl Config {
...
 }
 
 pub struct Context {
   fi: FrameInvariants,
   seq: Sequence,
   //    timebase: Ratio,
-  frame_q: VecDeque<Option<Arc<Frame>>> //    packet_q: VecDeque<Packet>
+  frame_count: u64,
+  idx: u64,
+  frame_q: VecDeque<(u64, Option<Arc<Frame>>)> //    packet_q: VecDeque<Packet>
```
lu-zero: You might want to use a map probably e.g. [btreemap](https://doc.rust-lang.org/stable/std/collections/struct.BTreeMap.html).

fbossen: You are right. A BTreeMap is much cleaner. Thanks for the tip!

src/api.rs
```diff
@@ -168,9 +168,9 @@ impl Context {
       self.fi.base_q_idx = (self.fi.config.quantizer.min(255 - q_drop) + q_drop) as u8;
 
       let first_ref_frame = LAST_FRAME;
-      let second_ref_frame = ALTREF_FRAME;
+      let second_ref_frame = if idx_in_group == 0 { LAST2_FRAME } else { ALTREF_FRAME };
 
-      self.fi.primary_ref_frame = (first_ref_frame - LAST_FRAME) as u32;
+      self.fi.primary_ref_frame = PRIMARY_REF_NONE; //(first_ref_frame - LAST_FRAME) as u32;
```
tdaede: From my limited experience in libaom, you generally want to set this to whatever reference frame was at the same pyramid level (and none ended up better than a random last frame, as you also seem to have found out)

tdaede:
LGTM, but needs a rebase.

PR #632, opened 2018-10-04
Fix map syntax

ycho:
Luca, can you check whether what I try to fix is correct?

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -29,7 +29,7 @@ pub fn get_sad(
       sum += slice_org
         .iter()
         .zip(slice_ref)
-        .map(|(&a, &b)| (a as i32 - b as i32).abs() as u32)
+        .map(|(a, b)| (*a as i32 - *b as i32).abs() as u32)
         .sum::<u32>();
   }
 
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -126,8 +126,8 @@ fn sse_wxh(
     let row_sse = s1
       .iter()
       .zip(s2)
-      .map(|(&a, &b)| {
-        let c = (a as i16 - b as i16) as i32;
+      .map(|(a, b)| {
+        let c = (*a as i16 - *b as i16) as i32;
         (c * c) as u32
       }).sum::<u32>();
     sse += row_sse as u64;


lu-zero:
It should not have any effect.

Here a trick to have rustc print the variable type:
``` diff
--- a/src/me.rs
+++ b/src/me.rs
@@ -29,7 +29,7 @@ pub fn get_sad(
       sum += slice_org
         .iter()
         .zip(slice_ref)
-        .map(|(&a, &b)| (a as i32 - b as i32).abs() as u32)
+        .map(|(&a, &b)| { let a: () = a; let b: () = b; (a as i32 - b as i32).abs() as u32 } )
         .sum::<u32>();
   }
```

The error messages tell you what type is what:

``` rustc
error[E0308]: mismatched types
  --> src/me.rs:32:39
   |
32 |         .map(|(&a, &b)| { let a: () = a; let b: () = b; (a as i32 - b as i32).abs() as u32 } )
   |                                       ^ expected (), found u16
   |
   = note: expected type `()`
              found type `u16`

error[E0308]: mismatched types
  --> src/me.rs:32:54
   |
32 |         .map(|(&a, &b)| { let a: () = a; let b: () = b; (a as i32 - b as i32).abs() as u32 } )
   |                                                      ^ expected (), found u16
   |
   = note: expected type `()`
              found type `u16`

error[E0605]: non-primitive cast: `()` as `i32`
  --> src/me.rs:32:58
   |
32 |         .map(|(&a, &b)| { let a: () = a; let b: () = b; (a as i32 - b as i32).abs() as u32 } )
   |                                                          ^^^^^^^^
   |
   = note: an `as` expression can only be used to convert between primitive types. Consider using the `From` trait

error[E0605]: non-primitive cast: `()` as `i32`
  --> src/me.rs:32:69
   |
32 |         .map(|(&a, &b)| { let a: () = a; let b: () = b; (a as i32 - b as i32).abs() as u32 } )
   |                                                                     ^^^^^^^^
   |
   = note: an `as` expression can only be used to convert between primitive types. Consider using the `From` trait
```

ycho:
Both before and after this fix are correct, so closing.

PR #636, opened 2018-10-07
Add a function to produce a container-compatible sequence header

lu-zero:
This is needed to make downstream developers life easier (it is also hopefully simpler than what `aom` offers).

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -140,6 +140,36 @@ impl Context {
     Ok(())
   }
 
+  pub fn container_sequence_header(&mut self) -> Vec<u8> {
+    use bitstream_io::*;
+    use std::io;
+    fn sequence_header_inner(seq: &Sequence) -> io::Result<Vec<u8>> {
+      let mut buf = Vec::new();
+      {
+      let mut bw = BitWriter::endian(&mut buf, BigEndian);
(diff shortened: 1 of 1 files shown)


lu-zero:
@dwbuiten could you please doublecheck that's what you wanted?

src/api.rs
```diff
@@ -140,6 +140,37 @@ impl Context {
     Ok(())
   }
 
+  pub fn container_sequence_header(&mut self) -> Vec<u8> {
+    use bitstream_io::*;
+    use std::io;
+    fn sequence_header_inner(seq: &Sequence) -> io::Result<Vec<u8>> {
```
ycho: I think I encounter the naming '_inner' second time here, with the first one found in intra prediction?
Could you explain what it would mean? I am just curious on what the name means here... :)

lu-zero: In general when I have some function I can model with some internal function called only from the actual API function I tend to use `_inner`.

In this code I'm hiding the fact bitstream-io focus is on writing on generic I/O, thus accounting for failures that cannot happen if you are writing to memory.

So instead of writing lots of `unwrap()` to each `write()` function I just use the `?` operator and unwrap the return value.

tdaede:
r+

PR #639, opened 2018-10-09
Add dav1d MC assembly (but don't use it yet).

tdaede:


rzumer:
What is the purpose of the `ext` directory?

tdaede:
In dav1d, it means that it's from a separate project (x264).

We can name it whatever we want, this just replicates the dav1d directory structure.

rzumer:
Shouldn't both files be in that directory, in that case?

src/config.asm
```diff
@@ -0,0 +1,7 @@
+	%define ARCH_X86_32 0
```
lu-zero: This should be generated (and put away from the src dir)

build.rs
```diff
@@ -6,12 +6,15 @@ extern crate pkg_config;
...
 #[cfg(feature = "decode_test")]
 extern crate bindgen;
+extern crate nasm_rs;
 
 use std::env;
 use std::fs;
 use std::path::Path;
 
 fn main() {
+    nasm_rs::compile_library_args("librav1easm.a", &["src/x86/mc.asm"], &["-Isrc/"]);
```
lu-zero: This should be guarded so only x86* would use it. add another `-I` to find the config.asm or pass its symbols directly.

build.rs
```diff
@@ -6,12 +6,15 @@ extern crate pkg_config;
 #[cfg(unix)]
 #[cfg(feature = "decode_test")]
 extern crate bindgen;
+extern crate nasm_rs;
```
lu-zero: it should be guarded with a `cfg()` gate

Cargo.toml
```diff
@@ -26,6 +26,7 @@ num-traits = "0.2"
 
 [build-dependencies]
 cmake = "0.1.32"
+nasm-rs = "0.1.3"
```
lu-zero: it should be a `target.` build dependency.

appveyor.yml
```diff
@@ -6,10 +6,14 @@ environment:
...
 
 install:
+    - call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat"
     - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
     - appveyor DownloadFile http://www.tortall.net/projects/yasm/releases/yasm-1.3.0-win64.exe -FileName yasm.exe
+    - appveyor DownloadFile https://www.nasm.us/pub/nasm/releasebuilds/2.13.03/win64/nasm-2.13.03-win64.zip -FileName nasm.zip
+    - 7z e -y nasm.zip
     - rustup-init -yv --default-toolchain %channel% --default-host %target%
     - set PATH=%PATH%;%USERPROFILE%\.cargo\bin;%APPVEYOR_BUILD_FOLDER%
+    - dir
```
lu-zero: Maybe should be dropped.

PR #641, opened 2018-10-09
Check for the y4m encoder presence

lu-zero:
Unbreak encoding without producing the reconstructed video.

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -96,7 +96,7 @@ pub fn parse_cli() -> (EncoderIO, EncoderConfig, usize) {
 pub fn process_frame(
   ctx: &mut Context, output_file: &mut dyn Write,
   y4m_dec: &mut y4m::Decoder<'_, Box<dyn Read>>,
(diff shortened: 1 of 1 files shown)


src/bin/common.rs
```diff
@@ -144,15 +144,14 @@ pub fn process_frame(
...
-  let y4m_enc_uw = y4m_enc.unwrap();
-
   let mut has_data = true;
   while has_data {
     let pkt_wrapped = ctx.receive_packet();
     match pkt_wrapped {
     Ok(pkt) => {
       eprintln!("{}", pkt);
       write_ivf_frame(output_file, pkt.number as u64, pkt.data.as_ref());
+      if let Some(y4m_enc_uw) = y4m_enc.as_mut() {
```
ycho: Indentation for inner block?

ycho:
+2

src/bin/common.rs
lu-zero: in another commit, later ^^

PR #646, opened 2018-10-11
Add asserts for ref frames update with frame reordering

ycho:


src/api.rs
```diff
@@ -257,8 +257,10 @@ impl Context {
         self.fi.ref_frames[i] = if i == second_ref_frame - LAST_FRAME {
           (slot_idx as u64 + if lvl == 0 { 6 * group_src_len } else { group_src_len >> lvl }) & 7
         } else if i == ref_in_previous_group - LAST_FRAME {
+          assert!((slot_idx as u64 - group_src_len) >= 0);
           (slot_idx as u64 - group_src_len) & 7
         } else {
+          assert!((slot_idx as u64 - (group_src_len >> lvl)) >= 0);
```
fbossen: Am seeing the following during build: "warning: comparison is useless due to type limits"
Note that the computation below uses &7 to deal with cases where wraparound occurs.

src/api.rs
```diff
@@ -257,8 +257,10 @@ impl Context {
         self.fi.ref_frames[i] = if i == second_ref_frame - LAST_FRAME {
           (slot_idx as u64 + if lvl == 0 { 6 * group_src_len } else { group_src_len >> lvl }) & 7
         } else if i == ref_in_previous_group - LAST_FRAME {
+          assert!((slot_idx as u64 - group_src_len) >= 0);
```
fbossen: Am seeing the following during build: "warning: comparison is useless due to type limits"
Note that the computation below uses &7 to deal with cases where wraparound occurs.

lu-zero: In that case you should explicitly use [wrapping_sub()](https://doc.rust-lang.org/std/primitive.u64.html#method.wrapping_sub).

ycho:
Since the issue has been resolved the other way by fbossen, #647 , closing this.

PR #663, opened 2018-10-18
Add support for compound mode

fbossen:
Compound mode is used in conjunction with frame reordering. It is limited to the case where one forward and one backward reference are used.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -179,7 +179,7 @@ impl Context {
     let reorder = true;
     let multiref = reorder || self.fi.config.speed <= 2;
 
-    let pyramid_depth = if reorder { 1 } else { 0 };
+    let pyramid_depth = if reorder { 2 } else { 0 };
     let group_src_len = 1 << pyramid_depth;
     let group_len = group_src_len + if reorder { pyramid_depth } else { 0 };
     let segment_len = 1 + (key_frame_interval - 1 + group_src_len - 1) / group_src_len * group_len;
--- a/src/api.rs
+++ b/src/api.rs
@@ -264,7 +264,7 @@ impl Context {
         } as u8) & 7;
       }
 
-      self.fi.reference_mode = if multiref && reorder && idx_in_group != 0 {
+      self.fi.reference_mode = if multiref && reorder && idx_in_group != 0 && false {
         ReferenceMode::SELECT
       } else {
         ReferenceMode::SINGLE
--- a/src/api.rs
+++ b/src/api.rs
@@ -270,6 +270,7 @@ impl Context {
         ReferenceMode::SINGLE
       };
       self.fi.number = segment_idx * key_frame_interval + self.fi.order_hint as u64;
+      self.fi.me_range_scale = (group_src_len >> lvl) as u8;
     }
 
     true
--- a/src/api.rs
+++ b/src/api.rs
@@ -176,7 +176,7 @@ impl Context {
   pub fn frame_properties(&mut self, idx: u64) -> bool {
     let key_frame_interval: u64 = 30;
 
-    let reorder = true;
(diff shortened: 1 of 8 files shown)


src/encoder.rs
```diff
@@ -1949,22 +1949,38 @@ fn encode_partition_topdown(seq: &Sequence, fi: &FrameInvariants, fs: &mut Frame
...
+                    }
+                    if mode_luma == PredictionMode::NEW_NEWMV && mvs[0].row == 0 && mvs[0].col == 0 &&
+                        mvs[1].row == 0 && mvs[1].col == 0 {
+                        mode_luma = PredictionMode::GLOBAL_GLOBALMV;
+                    }
+                    mode_chroma = mode_luma;
+                } else {
+                    mode_luma = PredictionMode::NEWMV;
+                    for (c, m) in mv_stack.iter().take(4)
+                    .zip([PredictionMode::NEARESTMV, PredictionMode::NEAR0MV,
```
lu-zero: You might use [find](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map) instead.

fbossen: Sounds like a good idea. Note that this code isn't really related to this PR (it's just being re-indented here). I think it would make more sense to do this in a separate PR.

src/api.rs
```diff
@@ -264,7 +264,13 @@ impl Context {
         } as u8) & 7;
       }
 
+      self.fi.reference_mode = if multiref && reorder && idx_in_group != 0 {
```
tdaede: Shouldn't just be "if multiref"?

src/context.rs
```diff
@@ -1426,24 +1462,26 @@ impl BlockContext {
...
-  pub fn set_ref_frame(&mut self, bo: &BlockOffset, bsize: BlockSize, r: usize) {
+  pub fn set_ref_frames(&mut self, bo: &BlockOffset, bsize: BlockSize, r: &[usize; 2]) {
     let bw = bsize.width_mi();
     let bh = bsize.height_mi();
 
     for y in 0..bh {
       for x in 0..bw {
-        self.blocks[bo.y + y as usize][bo.x + x as usize].ref_frames[0] = r;
+        self.blocks[bo.y + y as usize][bo.x + x as usize].ref_frames[0] = r[0];
+        self.blocks[bo.y + y as usize][bo.x + x as usize].ref_frames[1] = r[1];
```
tdaede: can this just be ref_frames = r

src/encoder.rs
```diff
@@ -252,6 +252,59 @@ impl Sequence {
...
+                backward_idx = i as isize;
+                backward_hint = ref_hint;
+              }
+            }
+          }
+        }
+        if forward_idx < 0 {
+          false
+        } else if backward_idx >= 0 {
+          // set skip_mode_frame
```
tdaede: is this still a todo?

src/api.rs
fbossen: No. reference_mode == ReferenceMode::SELECT is synonymous to compound mode can be enabled at the block level. Multiref is synonymous to one of multiple reference frames can be selected at the block level

src/context.rs
fbossen: yes. good point

src/api.rs
```diff
@@ -176,7 +176,7 @@ impl Context {
   pub fn frame_properties(&mut self, idx: u64) -> bool {
     let key_frame_interval: u64 = 30;
 
-    let reorder = true;
+    let reorder = false;
```
ycho: To enable compound mode, setting "reorder = true;" is the only thing to do, correct?

fbossen: yes. that will do

PR #664, opened 2018-10-19
Remove C dependency from transform.rs

### LINKED ISSUE #652: Remove the depencency on libaom in transform.rs
- [ ] av1_fwd_txfm2d_4x4
- [ ] av1_fwd_txfm2d_8x8
- [ ] av1_fwd_txfm2d_16x16
- [ ] av1_fwd_txfm2d_32x32

shssoichiro:
Includes:
- Rewrite fwd transform functions from C to Rust
- Rewrite 32x32 inv transform functions from C to Rust
- Refactor transform.rs into smaller modules

This commit temporarily removes the calls into the
SIMD optimized functions. These will be rewritten
into Rust and readded in an upcoming PR.

Closes #652


lu-zero:
Some tests are failing, you should update them as well.

src/transform/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+use partition::TxSize;
+use partition::TxType;
+use partition::TX_TYPES;
+
+use util::*;
+
+use self::forward::*;
```
lu-zero:
```suggestion
pub use self::forward::*;
```

src/transform/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+use self::inverse::*;
```
lu-zero:
```suggestion
pub use self::inverse::*;
```

This should fix the broken tests I guess.

Issue #653, opened 2018-10-16
Remove the depencency on libaom in ec.rs


lu-zero:
The tests use the libaom ec reader, functions to be replaced or reworked:
 - [x] od_ec_dec
 - [x] od_ec_decode_bool_q15
 - [x] od_ec_decode_cdf_q15

PR #669, opened 2018-10-22
Port the entropy decoder in libaom to rust

spersson:
Fixes issue #653.

Something to get me started with helping out on this project. Since the decoding is not used anywhere else I simply included it directly in the test module.


lu-zero:
Thank you a lot, github today is having problems so it will take a while for travis to pick up and execute the tests.

PR #670, opened 2018-10-23
Integrate intra-prediction assembly from dav1d

barrbrain:
See #675 for full list of predictors to map.
As a first step, map `pred_dc_128`.

### DIFF
--- a/benches/predict.rs
+++ b/benches/predict.rs
@@ -25,8 +25,19 @@ pub fn generate_block(rng: &mut ChaChaRng) -> (Vec<u16>, Vec<u16>, Vec<u16>) {
   (block, above_context, left_context)
 }
 
+pub fn generate_block_u8(rng: &mut ChaChaRng) -> (Vec<u8>, Vec<u8>, Vec<u8>) {
+  let block = vec![0u8; BLOCK_SIZE.width() * BLOCK_SIZE.height()];
+  let above_context: Vec<u8> =
+    (0..BLOCK_SIZE.height()).map(|_| rng.gen()).collect();
+  let left_context: Vec<u8> =
+    (0..BLOCK_SIZE.width()).map(|_| rng.gen()).collect();
+
+  (block, above_context, left_context)
+}
+
 pub fn pred_bench(c: &mut Criterion) {
   c.bench_function("intra_dc_4x4", |b| intra_dc_4x4(b));
+  c.bench_function("intra_dc_128_4x4_u8", |b| intra_dc_128_4x4_u8(b));
   c.bench_function("intra_dc_left_4x4", |b| intra_dc_left_4x4(b));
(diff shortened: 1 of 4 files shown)


benches/predict.rs
```diff
@@ -25,8 +25,19 @@ pub fn generate_block(rng: &mut ChaChaRng) -> (Vec<u16>, Vec<u16>, Vec<u16>) {
...
+    (0..BLOCK_SIZE.height()).map(|_| rng.gen()).collect();
+  let left_context: Vec<u8> =
+    (0..BLOCK_SIZE.width()).map(|_| rng.gen()).collect();
+
+  (block, above_context, left_context)
+}
+
 pub fn pred_bench(c: &mut Criterion) {
   c.bench_function("intra_dc_4x4", |b| intra_dc_4x4(b));
+  c.bench_function("intra_dc_128_4x4_8bpc", |b| intra_dc_128_4x4_8bpc(b));
```
lu-zero: would be nicer calling it `_u8`.

src/predict.rs
```diff
@@ -892,6 +915,16 @@ pub mod test {
       }
     }
 
+    let mut o8 = vec![0u8; 32 * 32];
+
+    Block4x4::pred_dc_128(&mut o8, 32, 8);
+
+    for l in o8.chunks(32).take(4) {
+      for v in l[..4].iter() {
```
lu-zero: You can probably use

``` rust
assert_eq(l, &[128u8; 4][..])
```

or put the reference in a separate variable and feed `assert_eq` the two slices.

build.rs
```diff
@@ -29,7 +30,8 @@ fn main() {
         let mut config_include_arg = String::from("-I");
         config_include_arg.push_str(&out_dir);
         config_include_arg.push('/');
-        nasm_rs::compile_library_args("rav1easm", &["src/x86/mc.asm"], &[&config_include_arg, "-Isrc/"]);
+        nasm_rs::compile_library_args("rav1easm", &["src/x86/mc.asm", "src/x86/ipred.asm"], &[&config_include_arg, "-Isrc/"]);
+        println!("cargo:rustc-link-lib=static=rav1easm");
```
lu-zero: We should have upstream fix this part sooner or later I guess.

tdaede: What's the deal with the println!, it seems that this is breaking appveyor?

src/predict.rs
barrbrain: I moved this into a test for u8 code and applied the suggestion.

build.rs
lu-zero: It should not and this line is needed in nasm_rs (cf cc-rs)

lu-zero:
Alternatively someone might help having [cc-rs](https://github.com/alexcrichton/cc-rs) address this directly (I opened https://github.com/alexcrichton/cc-rs/issues/356 to see if upstream is fine with this approach).

Issue #594, opened 2018-09-19
Predict ME search center from SB neighbors


ycho:
As Tim commented in #589, https://github.com/xiph/rav1e/pull/589#discussion_r218902242,

It would be nice to predict (ME search center) from our superblock neighbors, e.g., median-of-three, but even just the left (or top) superblock would do. But no objections to landing this patch as-is and trying that in a separate patch.

ycho:
#673 achieves this.

PR #673, opened 2018-10-23
Hierarchical motion estimation

fbossen:
Motion estimation is initially done on 64x64 blocks which are downsampled to 16x16. A refinement step is then done on 32x32 blocks which are downsampled to 16x16.
Hierarchical motion estimation provides better run times when using large search ranges

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -28,8 +28,6 @@ use std::io;
 use std::io::*;
 use std::rc::Rc;
 
-pub const MAX_SB_SIZE: usize = 128;
-
 extern {
     pub fn av1_rtcd();
     pub fn aom_dsp_rtcd();
@@ -340,8 +338,12 @@ pub struct FrameState {
 
 impl FrameState {
     pub fn new(fi: &FrameInvariants) -> FrameState {
+        FrameState::new_with_frame(fi, Arc::new(Frame::new(fi.padded_w, fi.padded_h)))
+    }
+
+    pub fn new_with_frame(fi: &FrameInvariants, frame: Arc<Frame>) -> FrameState {
         FrameState {
-            input: Arc::new(Frame::new(fi.padded_w, fi.padded_h)),
+            input: frame,
             input_hres: Plane::new(
                 fi.padded_w/2, fi.padded_h/2,
                 1, 1,
@@ -530,27 +532,6 @@ impl FrameInvariants {
             use_tx_domain_distortion: use_tx_domain_distortion,
         }
     }
-
-    pub fn new_frame_state(&self) -> FrameState {
-        FrameState {
-            input: Arc::new(Frame::new(self.padded_w, self.padded_h)),
-            input_hres: Plane::new(
-                self.padded_w/2, self.padded_h/2,
-                1, 1,
-                (MAX_SB_SIZE + SUBPEL_FILTER_SIZE ) / 2, (MAX_SB_SIZE + SUBPEL_FILTER_SIZE) / 2
-            ),
-            input_qres: Plane::new(
-                self.padded_w/4, self.padded_h/4,
-                2, 2,
-                (MAX_SB_SIZE + SUBPEL_FILTER_SIZE) / 4, (MAX_SB_SIZE + SUBPEL_FILTER_SIZE) / 4
-            ),
-            rec: Frame::new(self.padded_w, self.padded_h),
-            qc: Default::default(),
-            cdfs: CDFContext::new(0),
-            deblock: Default::default(),
-        }
-    }
-
 }
 
 impl fmt::Display for FrameInvariants{
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -28,6 +28,8 @@ use std::io;
 use std::io::*;
 use std::rc::Rc;
 
+pub const MAX_SB_SIZE: usize = 128;
+
 extern {
     pub fn av1_rtcd();
     pub fn aom_dsp_rtcd();
@@ -42,9 +44,21 @@ impl Frame {
     pub fn new(width: usize, height:usize) -> Frame {
         Frame {
             planes: [
-                Plane::new(width, height, 0, 0, 128+8, 128+8),
-                Plane::new(width/2, height/2, 1, 1, 64+8, 64+8),
-                Plane::new(width/2, height/2, 1, 1, 64+8, 64+8)
+                Plane::new(
+                    width, height,
+                    0, 0,
+                    MAX_SB_SIZE + SUBPEL_FILTER_SIZE, MAX_SB_SIZE + SUBPEL_FILTER_SIZE
+                ),
+                Plane::new(
+                    width/2, height/2,
+                    1, 1,
+                    MAX_SB_SIZE/2 + SUBPEL_FILTER_SIZE, MAX_SB_SIZE/2 + SUBPEL_FILTER_SIZE
+                ),
+                Plane::new(
+                    width/2, height/2,
+                    1, 1,
+                    MAX_SB_SIZE/2 + SUBPEL_FILTER_SIZE, MAX_SB_SIZE/2 + SUBPEL_FILTER_SIZE
+                )
             ]
         }
     }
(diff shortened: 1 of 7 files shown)


fbossen:
Without frame reordering: (about 10-15% faster)

| -0.1981 |  0.0448 | -0.1303 |  -0.2206 | -0.2614 | -0.2580 |    -0.1438 |

With frame reordering: (about 2x faster)
| 0.2216 | -0.0067 |  0.4253 |   0.1535 | 0.2277 |  0.1713 |     0.2783 |

src/encoder.rs
```diff
@@ -2131,17 +2135,25 @@ fn encode_tile(sequence: &mut Sequence, fi: &FrameInvariants, fs: &mut FrameStat
             cw.bc.cdef_coded = false;
             cw.bc.code_deltas = fi.delta_q_present;
 
+            // Do subsampled ME
+            let mut pmvs: [Option<MotionVector>; REF_FRAMES] = [None; REF_FRAMES];
+            for i in 0..INTER_REFS_PER_FRAME {
+                let r = fi.ref_frames[i] as usize;
+                if pmvs[r].is_none() {
+                    pmvs[r] = estimate_motion_ss4(fi, fs, r, &bo);
```
ycho: So, this is 1/4 resolution ME for a SB, I think?

src/me.rs
```diff
@@ -138,6 +138,54 @@ pub fn motion_estimation(
...
+pub fn estimate_motion_ss4(
+  fi: &FrameInvariants, fs: &FrameState, ref_idx: usize, bo: &BlockOffset
+) -> Option<MotionVector> {
+  if let Some(ref rec) = fi.rec_buffer.frames[ref_idx] {
+    let po = PlaneOffset {
+      x: (bo.x as isize) << BLOCK_TO_PLANE_SHIFT >> 2,
+      y: (bo.y as isize) << BLOCK_TO_PLANE_SHIFT >> 2
+    };
+    let range = 32 * fi.me_range_scale as isize;
+    let blk_w = 64 >> 2;
```
ycho: Ok, it is 1/4 resolution.

src/me.rs
```diff
@@ -138,6 +138,54 @@ pub fn motion_estimation(
   }
 }
 
+pub fn estimate_motion_ss4(
```
ycho: ss means 's'ub's'ampled? :) I am always curious about the shortened names in computer programs!

src/encoder.rs
ycho: Also, not for this patch but I'd like to remind use that this approach of hierarchical ME starting from SB level can be  still benefited from using the MVs from already encoded neighbors as filed in #594.

src/rdo.rs
```diff
@@ -755,18 +755,20 @@ pub fn rdo_partition_decision(
...
 
         child_modes.extend(
           partitions
-            .iter()
-            .map(|&offset| {
-              rdo_mode_decision(seq, fi, fs, cw, subsize, &offset, spmvs)
+            .iter().zip(pmv_idxs)
+            .map(|(&offset, pmv_idx)| {
+              rdo_mode_decision(seq, fi, fs, cw, subsize, &offset,
+                &pmvs[REF_FRAMES*pmv_idx..REF_FRAMES*(pmv_idx+1)])
```
ycho: Can we consider using 2D array for this? i.e. pmvs[pmv_idx], but that will need several changes.

src/encoder.rs
```diff
@@ -2155,12 +2174,33 @@ fn encode_tile(sequence: &mut Sequence, fi: &FrameInvariants, fs: &mut FrameStat
...
                 let r = fi.ref_frames[i] as usize;
                 if pmvs[r].is_none() {
-                    pmvs[r] = estimate_motion_ss4(fi, fs, r, &bo);
+                    pmvs[r] = frame_pmvs[sby * fi.sb_width + sbx][r];
                     if let Some(pmv) = pmvs[r] {
-                        pmvs[r + 1*REF_FRAMES] = estimate_motion_ss2(fi, fs, r, &sbo.block_offset(0, 0), &pmv);
-                        pmvs[r + 2*REF_FRAMES] = estimate_motion_ss2(fi, fs, r, &sbo.block_offset(8, 0), &pmv);
-                        pmvs[r + 3*REF_FRAMES] = estimate_motion_ss2(fi, fs, r, &sbo.block_offset(0, 8), &pmv);
-                        pmvs[r + 4*REF_FRAMES] = estimate_motion_ss2(fi, fs, r, &sbo.block_offset(8, 8), &pmv);
+                        let pmv_w = if sbx > 0 {
```
ycho: I guess this implements the #594? i.e. Referencing neighboring SB's MVs. That is great!

src/encoder.rs
```diff
@@ -2155,12 +2174,33 @@ fn encode_tile(sequence: &mut Sequence, fi: &FrameInvariants, fs: &mut FrameStat
...
+                        } else {
+                            None
+                        };
+                        let pmv_s = if sby < fi.sb_height - 1 {
+                            frame_pmvs[sby * fi.sb_width + sbx + fi.sb_width][r]
+                        } else {
+                            None
+                        };
+
+                        pmvs[r + 1*REF_FRAMES] = estimate_motion_ss2(fi, fs, r, &sbo.block_offset(0, 0), &[Some(pmv), pmv_w, pmv_n]);
```
ycho: As commented in previous commit, 2D array of pmvs[pmv_index_for_32x32_blk][r] can be considered.

src/me.rs
```diff
@@ -138,6 +126,27 @@ pub fn motion_estimation(
...
+    for x in (x_lo..x_hi).step_by(step) {
+      let plane_org = p_org.slice(po);
+      let plane_ref = p_ref.slice(&PlaneOffset { x, y });
+
+      let sad = get_sad(&plane_org, &plane_ref, blk_h, blk_w);
+
+      if sad < *lowest_sad {
+        *lowest_sad = sad;
+        *best_mv = MotionVector {
+          row: 8 * (y as i16 - po.y as i16),
```
ycho: Would "8 * " still  correct for estimate_motion_ss4() and estimate_motion_ss2() ?
Shouldn't they "32 * " and "16 * " for *_ss4() and _ss2()?

fbossen: Good point. Note that the MVs are scaled by 4x and 2x and at the end of  _ss4() and _ss2() functions, so using 8x here should be correct

src/rdo.rs
fbossen: That is a good suggestion. Pushed a revision with it.

src/encoder.rs
```diff
@@ -1816,7 +1816,15 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
...
-        let mode_decision = rdo_mode_decision(seq, fi, fs, cw, bsize, bo, pmvs).part_modes[0].clone();
+
+        let pmv_idx = if bsize > BlockSize::BLOCK_32X32 {
+            0
+        } else {
+            ((bo.x & 32) >> 5) + ((bo.y & 32) >> 4) + 1
+        };
+        let spmvs = &pmvs[REF_FRAMES*pmv_idx..REF_FRAMES*(pmv_idx+1)];
+
+        let mode_decision = rdo_mode_decision(seq, fi, fs, cw, bsize, bo, spmvs).part_modes[0].clone();
```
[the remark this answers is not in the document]
fbossen: Selected? Let me know if you have a better name suggestion.

src/me.rs
```diff
@@ -154,18 +154,27 @@ fn full_search(x_lo: isize, x_hi: isize, y_lo: isize, y_hi: isize, blk_h: usize,
...
+    x: (bo.x as isize).min(fi.w_in_b as isize - blk_w as isize / 4).max(0) as usize,
+    y: (bo.y as isize).min(fi.h_in_b as isize - blk_h as isize / 4).max(0) as usize
+  }
+}
+
 pub fn estimate_motion_ss4(
   fi: &FrameInvariants, fs: &FrameState, ref_idx: usize, bo: &BlockOffset
 ) -> Option<MotionVector> {
   if let Some(ref rec) = fi.rec_buffer.frames[ref_idx] {
+    let blk_w = 64;
```
ycho: With this block size, estimate_motion_ss4() will be always called for a 64x64 SB, correct?
If so, I think putting some asserts to detect such condition would lessen any misuse of this function?

src/encoder.rs
```diff
@@ -38,24 +38,26 @@ pub struct Frame {
     pub planes: [Plane; 3]
 }
 
+const FRAME_MARGIN: usize = 16 + SUBPEL_FILTER_SIZE;
```
ycho: I guessed that restoring of original frame marine would be related to the speed of encoding? Since it enables SIMD code an be used w/o checking the borders. If not, I think I am missing the idea behind how these padded pixels are used

src/encoder.rs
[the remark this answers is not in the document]
ycho: :) Thanks for your openness!

ycho:
LGTM. 
If you can, please add in the commit message that can be independent to the original goal of this patch, such as  "uses MVs from neighbours". I think that will help us to track the factors that has changed the coding efficiency as well as a speed performance.
I ideally, I would even consider having the idea of "using MV references from neighboring SB" as separate patch, but it is your preference.

src/encoder.rs
fbossen: yes, padded pixels are used to avoid doing any boundary checks during motion compensation

src/me.rs
fbossen: Removed the hardcoded 64x64 size in this function and using a function parameter instead.

ycho: Yes, I see that, thanks!

ycho: So, now _ss4() and _ss2() are for ME on 1/4 resolution and 1/2 resolution, resp., for any input block sizes.

fbossen: Yes, it should now work for any block size, but haven't tested block sizes other 64x64 for 1/4 resolution and 32x32 for 1/2 resolution

Issue #674, opened 2018-10-24
Problems with compound mode?


moisesmcardona:
rav1e successfully created .ivf files. However, when I try to concatenate them with ffmpeg, it gave the following messages repeatedly:

```
[concat @ 0000024b2482a040] DTS 2 < 4 out of order
[ivf @ 0000024b24309140] Non-monotonous DTS in output stream 0:0; previous: 4, current: 2; changing to 5. This may result in incorrect timestamps in the output file.
[ivf @ 0000024b24309140] Non-monotonous DTS in output stream 0:0; previous: 5, current: 1; changing to 6. This may result in incorrect timestamps in the output file.
[ivf @ 0000024b24309140] Non-monotonous DTS in output stream 0:0; previous: 6, current: 2; changing to 7. This may result in incorrect timestamps in the output file.
[ivf @ 0000024b24309140] Non-monotonous DTS in output stream 0:0; previous: 7, current: 3; changing to 8. This may result in incorrect timestamps in the output file.
[ivf @ 0000024b24309140] Non-monotonous DTS in output stream 0:0; previous: 8, current: 4; changing to 9. This may result in incorrect timestamps in the output file.
[ivf @ 0000024b24309140] Non-monotonous DTS in output stream 0:0; previous: 9, current: 8; changing to 10. This may result in incorrect timestamps in the output file.
[ivf @ 0000024b24309140] Non-monotonous DTS in output stream 0:0; previous: 10, current: 6; changing to 11. This may result in incorrect timestamps in the output file.
....
```

After updating libaom and compiling ffmpeg again, I now see the following error:
```
[libaom-av1 @ 000001f770a77880] 1.0.0-825-g673a055b0
[libaom-av1 @ 000001f770a77880] Failed to decode frame: Unrecognized error code
[libaom-av1 @ 000001f770a77880]   Additional information:
[libaom-av1 @ 000001f770a77880] Failed to decode frame: Corrupt frame detected
[libaom-av1 @ 000001f770a77880]   Additional information: Reference frame containing this frame's initial frame context is unavailable.
[libaom-av1 @ 000001f770a77880] Failed to decode frame: Corrupt frame detected
[libaom-av1 @ 000001f770a77880]   Additional information: Reference frame containing this frame's initial frame context is unavailable.
[libaom-av1 @ 000001f770a77880] Failed to decode frame: Corrupt frame detected
[...]

PR #676, opened 2018-10-24
Combine OBUs of reordered frames

fbossen:
OBUs of reordered frames are combined with OBUs of a non-reordered
frame into a single packet. May address #674


moisesmcardona:
Yes. This fixes the warning messages from #674. The other part seemed to be from a bad ffmpeg or libaom build which got solved by rebuilding those tools. Thanks!

PR #677, opened 2018-10-25
Adds --low_latency and --keyint command line options

moisesmcardona:
* adds --compound as boolean to enable reordering
* adds --keyframe_interval which I think could be useful to specify other value than 30, say, for when encoding 60fps videos.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -31,8 +31,8 @@ impl Ratio {
 
 #[derive(Copy, Clone, Debug)]
 pub struct EncoderConfig {
-  pub compound: bool,
   pub key_frame_interval: u64,
+  pub low_latency: bool,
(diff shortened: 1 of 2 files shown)


tdaede:
Looks good.

As a matter of naming, --compound doesn't tell you very much. The only reason you'd ever want to disable it is if you want very low intrinsic latency. Maybe a --low_latency flag would be better?

moisesmcardona:
Renamed --compund to --low_latency.

gnafuthegreat:
Another note, it was suggested to me that the GOP/KF switch be named similarly to x264 for familiarity.  That would look like --keyint/-I (capital i) if we wanted to match.

tdaede:
Thanks!

PR #678, opened 2018-10-25
Loop restoration filter beginnings

xiphmont:
No functional change, but some added infrastructure and a Writer flow change in shared code menas I probably want to get this in sooner rather than later to avoid surprises.

Aside from adding RestorationUnits and a RestorationContext, coding the loop restoration parameters requires yet one more temporary Writer in the tile encode loop, as for each LRU the filter params have to be decided last, but written first.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -2111,7 +2111,9 @@ fn encode_tile(sequence: &mut Sequence, fi: &FrameInvariants, fs: &mut FrameStat
         cw.bc.reset_left_contexts();
 
         for sbx in 0..fi.sb_width {
+            let mut w_pre_cdef = WriterRecorder::new();
             let mut w_post_cdef = WriterRecorder::new();
+            let mut cdef_index = 0;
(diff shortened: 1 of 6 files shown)


src/encoder.rs
```diff
@@ -2080,20 +2125,31 @@ fn encode_tile(sequence: &mut Sequence, fi: &FrameInvariants, fs: &mut FrameStat
...
                                           BlockSize::BLOCK_64X64, &bo, &pmv);
             }
             else {
                 encode_partition_topdown(sequence, fi, fs, &mut cw,
-                                         &mut w, &mut w_post_cdef,
+                                         &mut w_pre_cdef, &mut w_post_cdef,
                                          BlockSize::BLOCK_64X64, &bo, &None);
             }
 
+            // CDEF has to be decisded before loop restoration, but coded after
```
lu-zero: typo

xiphmont: will catch it in next patch, thanks

PR #680, opened 2018-10-25
Transform domain distortion for RDO-based mode decision

ycho:
Enabled when speed >= 1.

[For 5 frames AWCY](https://beta.arewecompressedyet.com/?job=master-e06321d_s3_5f%402018-10-23T18%3A53%3A09.943Z&job=test8_tx_dist_5f%402018-10-25T23%3A13%3A04.088Z), 

| 0.1717 |  0.2794 |  0.1181 |   0.1141 | 0.1624 |  0.1534 |     0.6118 |

Currently, it causes regression of 0.17% for PSNR bd rate, with 44 (unit is probably sec?) --> 36 speed improvement, ~20% down. It also hurts CIEDE 2000 by 0.6%.

For now, the code is hard-coded to always use tx-domain distortion.
[Full AWCY results](https://beta.arewecompressedyet.com/?job=master-e06321d177cc14b1aa09952a4bb26f7f61512f5e&job=test8_tx_dist%402018-10-25T23%3A26%3A00.243Z)

| 0.3438 |  0.4575 |  0.3697 |   0.2728 | 0.2712 |  0.2613 |     0.5521 |

584 --> 505 (sec), ~ 14% down.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1363,9 +1363,12 @@ pub fn encode_tx_block(
     // Reconstruct
     dequantize(fi.base_q_idx, qcoeffs, &mut rcoeffs.array, tx_size, bit_depth);
 
+    let mut tx_dist: u32 = 0;
+    let mut dist: u64 = 0;
+    let mut diff: i64 = 0;
     // Store tx-domain distortion of this block
     if fi.use_tx_domain_distortion {
-        let tx_dist = coeffs
+        tx_dist = coeffs
             .iter()
             .zip(&rcoeffs.array[..tx_size.area()])
             .map(|(&a, &b)| {
@@ -1374,6 +1377,23 @@ pub fn encode_tx_block(
             }).sum::<u32>();
     }
     inverse_transform_add(&rcoeffs.array, &mut rec.mut_slice(po).as_mut_slice(), stride, tx_size, tx_type, bit_depth);
+
+    // For debug only
+    if fi.use_tx_domain_distortion {
+        // Compute distortion
+ //use rdo::sse_wxh;
+        dist =
+            //TODO: Repalce this function for tx domain distortion
+            sse_wxh(
+            &fs.input.planes[p].slice(po),
+            &rec.slice(po),
+            tx_size.width(),
+            tx_size.height(),
+            );
+        assert!(tx_size.area() >= 16);
+        //debug: compare pixel-domain and tx-domain distortions
+        diff = ((dist  - (tx_dist >> 4) as u64) / tx_size.area() as u64) as i64;
+    }
     has_coeff
 }
 
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -424,6 +424,7 @@ pub struct FrameInvariants {
     pub rec_buffer: ReferenceFramesSet,
     pub base_q_idx: u8,
     pub me_range_scale: u8,
+    pub use_tx_domain_distortion: bool,
 }
 
 impl FrameInvariants {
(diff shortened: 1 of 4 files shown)


src/rdo.rs
```diff
@@ -498,21 +597,25 @@ pub fn rdo_mode_decision(
         best.tx_size,
         best.tx_type,
         0,
-        &Vec::new()
+        &Vec::new(),
+        false // For CFL, luma sholud be always reconstructed.
```
lu-zero: typo

lu-zero:
Needs a rebase

src/encoder.rs
```diff
@@ -1232,39 +1236,63 @@ pub fn encode_tx_block(
...
+) -> (bool, i64) {
     let rec = &mut fs.rec.planes[p];
     let PlaneConfig { stride, xdec, ydec, .. } = fs.input.planes[p].cfg;
 
     if mode.is_intra() {
       mode.predict_intra(&mut rec.mut_slice(po), tx_size, bit_depth, &ac, alpha);
     }
 
-    if skip { return false; }
+    if skip { return (false, 0); }
```
fbossen: Should this return an actual distortion instead of 0?

src/rdo.rs
```diff
@@ -209,6 +227,68 @@ fn compute_rd_cost(
...
+          distortion += sse_wxh(
+            &fs.input.planes[p].slice(&po),
+            &fs.rec.planes[p].slice(&po),
+            w_uv,
+            h_uv
+          );
+        }
+      }
+    } else {
+
```
fbossen: Nothing to do here?

src/rdo.rs
ycho: where?

ycho: oh, should! thx!

src/encoder.rs
ycho: As you know, if tx block is a part of skipped partition, then only pixel domain distortion is available hence later compute_tx_rd_cost() acknowledge it and compute distortion in pixel domain there. 
However, I will change it return to -1 here to make sure the distortion is not computed yet as has been signaled that way in other places in my patch. Thx for the comment!

src/encoder.rs
```diff
@@ -1232,39 +1236,63 @@ pub fn encode_tx_block(
...
+        inverse_transform_add(rcoeffs, &mut rec.mut_slice(po).as_mut_slice(), stride, tx_size, tx_type, bit_depth);
+    } else {
+        //inverse_transform_add(rcoeffs, &mut rec.mut_slice(po).as_mut_slice(), stride, tx_size, tx_type, bit_depth);
+
+        // Store tx-domain distortion of this block
+        tx_dist = coeffs
+            .iter()
+            .zip(rcoeffs)
+            .map(|(a, b)| {
+                let c = (*a as i16 - *b as i16) as i32;
```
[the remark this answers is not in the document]
ycho: You are right, each tx coeff with 9 bits residue input can need full 15 bits + 1 bit (sign), so add/subtract two of them need one more bit to store the result, correct?
Furthermore, for the 10/12 bits input, I think it will need more precision than 16  + 1 bits.

src/rdo.rs
ycho: If not luma only and not skipped, then tx_dist is assumed to have the distortion for both luma and chroma.
I have revised to combine "if !luma_only" and "if skip" as one condition.

ycho:
I plan to use tx_domain_distortion only for faster speeds, like speed >= 3, any opinions?

PR #681, opened 2018-10-26
Fix semantics of low_latency

fbossen:
Frame reordering should be applied when low_latency is false

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -40,7 +40,7 @@ pub struct EncoderConfig {
 
 impl Default for EncoderConfig {
   fn default() -> Self {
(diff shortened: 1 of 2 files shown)


lu-zero:
Sounds fine.

src/api.rs
```diff
@@ -181,7 +181,7 @@ impl Context {
   pub fn frame_properties(&mut self, idx: u64) -> bool {
     let key_frame_interval: u64 = self.fi.config.key_frame_interval;
 
-    let reorder = self.fi.config.low_latency;
+    let reorder = !self.fi.config.low_latency;
```
ycho: So, now low_latency == true means, encode in low latency and don't use frame reordering, correct?

fbossen: yes

PR #684, opened 2018-10-29
Add coveralls support

lu-zero:


.travis.yml
```diff
@@ -14,10 +16,19 @@ before_install:
...
+    - cmake .. && make && sudo make install
+    - kcov --version
+    - cargo install cargo-kcov
 script:
     - |
         cargo build --verbose &&
         cargo test --verbose &&
         cargo test --verbose --release --features=decode_test -- --ignored &&
         cargo bench --verbose &&
         cargo doc --verbose
```
barrbrain: Without `&&` chaining here, we might discard errors.
```suggestion
         cargo doc --verbose &&
```

lu-zero: Sounds a good idea :) Thank you!

lu-zero: (solved manually because of the habit ^^;

PR #685, opened 2018-10-29
Use macro for libaom bindings in intra pred tests

barrbrain:
Also, add simple reference tests for u8 prediction.

### DIFF
--- a/src/predict.rs
+++ b/src/predict.rs
@@ -110,68 +110,6 @@ pub static extend_modes: [u8; INTRA_MODES] = [
   NEED_LEFT | NEED_ABOVE | NEED_ABOVELEFT  // PAETH
 ];
 
-extern {
-  #[cfg(test)]
-  fn highbd_dc_predictor(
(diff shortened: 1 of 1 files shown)


barrbrain:
Test changes are mentioned in the commit message, but GitHub UI hides the body of the message by default.

src/predict.rs
```diff
@@ -581,158 +519,52 @@ pub mod test {
...
-        4,
-        4,
-        above.as_ptr(),
-        left.as_ptr(),
-        8
-      );
-    }
-  }
+  macro_rules! wrap_aom_pred_fn {
+    ($fn_4x4:ident, $aom_fn:ident) => {
```
ycho: Hi- Can you instruct me what 'ident's do here?

lu-zero: they are `identfiers`. macros are made so you cannot pass the wrong syntactical element.

[here](https://doc.rust-lang.org/reference/macros-by-example.html) the whole documentation :)

Issue #675, opened 2018-10-24
Integrate AVX2 low-bitdepth assembly from dav1d


barrbrain:
- [x] pred_dc (#687)
- [x] pred_dc_128 (#670)
- [x] pred_dc_left (#687)
- [x] pred_dc_top (#687)
- [x] pred_h (#687)
- [x] pred_v (#687)
- [x] pred_paeth (#692)
- [x] pred_smooth (#692)
- [x] pred_smooth_h (#692)
- [x] pred_smooth_v (#692)
- [x] pred_cfl (#792)

barrbrain:
The remaining dc/h/v predictors are linked in #687.

barrbrain:
The `paeth`, `smooth`, `smooth_h` and `smooth_v` predictors are linked in #692.

PR #687, opened 2018-10-30
Integrate dav1d ipred.asm (dc, h, v)

barrbrain:
Add 8-bit benchmarks for all existing intra predictors.
Call the remaining dc, h, and v predictors if AVX2 is available.


barrbrain:
This completed the first half of #675.

PR #688, opened 2018-10-30
Extend the code coverage in travis

lu-zero:


lu-zero:
Let's try to reduce the time spent running some tests twice.

lu-zero:
Now it takes about 37minutes down to 41minutes.

PR #691, opened 2018-10-30
Rename ratio to Rational

lu-zero:
One of the feedbacks from today's API review.


lu-zero:
It is a rebase artefact, fixed now.

PR #692, opened 2018-10-31
Integrate dav1d ipred.asm (paeth, smooth, smooth_h, smooth_v)

barrbrain:
Apply more asm patches from dav1d.
Call the `paeth`, `smooth`, `smooth_h` and `smooth_v` predictors if AVX2 is available.

### DIFF
--- a/src/predict.rs
+++ b/src/predict.rs
@@ -178,6 +178,14 @@ decl_angular_ipred_fn!(rav1e_ipred_dc_top_avx2);
 decl_angular_ipred_fn!(rav1e_ipred_h_avx2);
 #[cfg(all(target_arch = "x86_64", not(windows)))]
 decl_angular_ipred_fn!(rav1e_ipred_v_avx2);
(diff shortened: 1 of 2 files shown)


barrbrain:
This add everything except `cfl` for #675.

src/predict.rs
```diff
@@ -422,6 +460,21 @@ where
   fn pred_smooth_h(
     output: &mut [T], stride: usize, above: &[T], left: &[T]
   ) {
+    #[cfg(all(target_arch = "x86_64", not(windows)))]
+    {
+      if size_of::<T>() == 1 && is_x86_feature_detected!("avx2") {
+        return unsafe {
+          rav1e_ipred_smooth_h_avx2(
```
tdaede: Is there a way to shorten this given that most of the ipred functions have the same call signature?

barrbrain: We're planning to unify the signatures and/or rework the dispatch in the near future.
In the first case, this can be a single macro.

PR #694, opened 2018-11-01
Add nasm files to the rerun-if-changed list

### LINKED ISSUE #690: Rebuild assembly on file change
It doesn't seem to do that right now. Learning this was unpleasant.

lu-zero:
Fixes #690


src/api.rs
```diff
@@ -17,14 +17,14 @@ use std::sync::Arc;
 // TODO: use the num crate?
 #[derive(Clone, Copy, Debug)]
 #[repr(C)]
-pub struct Ratio {
```
KyleSiefring: I'm guessing you didn't mean to include this.

lu-zero: yes it is a leftover.

PR #695, opened 2018-11-01
Modify frame management to allow for deeper frame hierarchies

fbossen:
Without increasing depth, yields the following when using frame reordering:

| -1.7571 | -2.3703 | -2.5057 |  -1.7010 | -1.7046 | -1.6254 |    -1.9372 |

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -224,12 +224,16 @@ impl Context {
         return false;
       }
 
+      fn pos_to_lvl(pos: u64, pyramid_depth: u64) -> u64 {
+        pyramid_depth - (pos | (1 << pyramid_depth)).trailing_zeros() as u64
+      }
+
       let lvl = if !reorder {
         0
       } else if idx_in_group < pyramid_depth {
         idx_in_group
       } else {
-        pyramid_depth - (idx_in_group - pyramid_depth + 1).trailing_zeros() as u64
+        pos_to_lvl(idx_in_group - pyramid_depth + 1, pyramid_depth)
       };
 
       // Frames with lvl == 0 are stored in slots 0..4 and frames with higher values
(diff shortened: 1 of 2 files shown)


fbossen:
Increasing pyramid depth to 3:

|    PSNR |  PSNR Cb |  PSNR Cr | PSNR HVS |    SSIM | MS SSIM | CIEDE 2000 |
|    ---: |     ---: |     ---: |     ---: |    ---: |    ---: |       ---: |
| -4.1125 | -10.0352 | -10.8160 |  -3.3456 | -5.2321 | -4.7807 |    -6.9573 |

ycho:
With more frame delays would be introduced, how about adding an option (as separate patches) to rav1e to distinguish no-frame-delay (i.e. similar to --rt) and with-any-frame-delay?

ycho:
Yes, I was aware of the option but I meant wanted be more clear about whether there is absolutely any frame delay or not. :)

lu-zero:
Seems good, later this code could enjoy a little bit of refactoring since those nested ifs are getting that function a bit too busy.

the pyramid depth would be a configuration option?

src/api.rs
```diff
@@ -254,18 +261,42 @@ impl Context {
...
+            }
+          } else if i == ref_in_previous_group - LAST_FRAME {
+            if lvl == 0 {
+              (slot_idx + 4 - 1) as u8 % 4
+            } else {
+              slot_idx as u8
+            }
+          } else {
+            let oh = self.fi.order_hint - (group_src_len as u32 >> lvl);
+            let lvl1 = pyramid_depth - oh.trailing_zeros().min(pyramid_depth as u32) as u64;
```
lu-zero: maybe 
``` rust
            let lvl1 = pyramid_depth.saturating_sub(oh.trailing_zeros() as u64);
```
Looks nicer

src/api.rs
```diff
@@ -223,7 +223,21 @@ impl Context {
         return false;
       }
 
-      let slot_idx = self.fi.order_hint % REF_FRAMES as u32;
+      let lvl = if !reorder {
+        0
+      } else if idx_in_group < pyramid_depth {
+        idx_in_group
+      } else {
+        pyramid_depth - (idx_in_group - pyramid_depth + 1).trailing_zeros() as u64
```
lu-zero: Maybe you want to use [saturating_sub](https://doc.rust-lang.org/std/primitive.u64.html#method.saturating_sub) as you did below? 

Probably a comment on what is the logic behind would be nice :)

Issue #635, opened 2018-10-07
Write tests for the transforms


lu-zero:
in `src/transform.rs`

- [ ] av1_fwd_txfm2d_4x4
- [ ] av1_fwd_txfm2d_8x8
- [ ] av1_fwd_txfm2d_16x16
- [ ] av1_fwd_txfm2d_32x32

- [ ] av1_inv_txfm2d_add_4x4
- [ ] av1_inv_txfm2d_add_8x8
- [ ] av1_inv_txfm2d_add_16x16
- [ ] av1_inv_txfm2d_add_32x32

tmatth:
@lu-zero would something equivalent to these make sense?
https://aomedia.googlesource.com/aom/+/067e984254e658863b8c6dc335e26f92ecc32c1c/test/transform_test_base.h#47

Namely: accuracy tests and bit exactness tests.

lu-zero:
It feels relatively overkill for an easy task. 
We can have it with [proptest](https://docs.rs/crate/proptest/), but it is harder than just do what we have for `get_sad`.

lu-zero:
Solved by #697

rzumer:
Closing as resolved.

PR #697, opened 2018-11-03
Draft: add tests for transforms

spersson:
Addressing issue #635.
I was thinking the easiest is to add round trip tests. But I'm not sure if it's good enough? It's a start at least. Should I also hardcode some cases and expected results?
I was also not sure about generating different random numbers on every test run. So far it's working, but perhaps there's a policy.
I commented out the types and sizes that are not supported now. If there are no plans to add support I can just remove those lines.
Tolerance of +-1 is accepted now, tolerance of 0 didn't work for bigger block sizes.
This is my first time writing tests for software, despite having tested software for a living over ten years now... :)

### DIFF
--- a/src/transform/mod.rs
+++ b/src/transform/mod.rs
@@ -282,7 +282,7 @@ mod test {
   use super::*;
   use rand::random;
 
-  fn test_roundtrip(tx_size: TxSize, tx_type: TxType) {
+  fn test_roundtrip(tx_size: TxSize, tx_type: TxType, tolerance: i16) {
     let mut src_storage = [0u16; 64 * 64];
     let src = &mut src_storage[..tx_size.area()];
     let mut dst_storage = [0u16; 64 * 64];
@@ -300,7 +300,7 @@ mod test {
     inverse_transform_add(freq, dst, tx_size.width(), tx_size, tx_type, 8);
 
     for (s, d) in src.iter().zip(dst) {
-      assert!(i16::abs((*s as i16) - (*d as i16)) <= 1);
+      assert!(i16::abs((*s as i16) - (*d as i16)) <= tolerance);
     }
   }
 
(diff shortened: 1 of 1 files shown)


lu-zero:
Another option is to compare implementations.

The rust implementation can be compared with the libaom one and the optimized implementations compared with the rust one once validated.

Any new code relying on the libaom copy should be optional since we plan to move away from aom.

ycho:
I also observed that there are errors of +-1 for 32x32 tx, and identity tx shows errors at even smaller size tx. For 32x32, we view that it is expected since less # of fractional bits are used, i.e. 2 bits compared to 3 bits for 4x4, 8x8, and 16x16.

tdaede:
Roundtrip tests are good. While the inverse transforms are required to be bitexact, the forward ones aren't, so we have a bit of leeway in implementation detail, and the +1/-1 thresholds are good for that.

If you'd like to add more involved dct tests, there's good examples of them in daala's dct.c file: http://git.xiph.org/?p=daala.git;a=blob;f=src/dct.c;h=adb0faf549bc9d164ae218c7cb085c6d56b90dbd;hb=HEAD

src/transform/mod.rs
```diff
@@ -276,3 +276,88 @@ pub fn inverse_transform_add(
...
+    for (s, d) in src.iter().zip(dst) {
+      assert!(i16::abs((*s as i16) - (*d as i16)) <= 1);
+    }
+  }
+
+  #[test]
+  fn roundtrips() {
+    use partition::TxSize::*;
+    use partition::TxType::*;
+    let combinations = vec![
```
tdaede: This can just be a static array.

spersson: I had tried that but had problem with IntoIter trait not implemented for arrays larger than 32. I looked some more now and it seems that the iter method works for all sizes. I will push this change soon.

spersson:
Thanks for the feedback, all of you! I just made a change to use smallest possible tolerances that still passes the test. A little bit worrying is that by running the tests many times I found that DCT_DCT for blocksize 32 sometimes gave error -+2.

tdaede:
OK. I'll land it with the +/-2 limits.

PR #700, opened 2018-11-05
Wrap get_sad bench in black_box.

tdaede:


benches/me.rs
```diff
@@ -45,7 +45,7 @@ fn bench_get_sad(b: &mut Bencher, bs: &BlockSize) {
   let plane_ref = rec_plane.slice(&po);
 
   b.iter(|| {
-      let _ = me::get_sad(&plane_org, &plane_ref, bsw, bsh);
+      let _ = black_box(me::get_sad(&plane_org, &plane_ref, bsw, bsh));
```
lu-zero: I was thinking about:

``` rust
black_box(me::get_sad)(&plane_org, &plane_ref, bsw, bsh);
```

tdaede: I can test that to see if it measures any different.

tdaede: So doing that is totally ineffective at black boxing, it turns out. You have to use a value from the function with black_box.

lu-zero: Thank you for testing :)

PR #701, opened 2018-11-06
Add a Pixel trait which represents a u8 or u16 pixel.

tdaede:


tdaede:
Yeah, I want the where AsPrimitive to be there too. I don't know how to do that though. I'll land this and then we can add that.

PR #702, opened 2018-11-06
U8 transforms 2

tdaede:


src/transform/inverse.rs
```diff
@@ -810,10 +812,10 @@ trait InvTxfm2D: Dim {
     Self::inv_txfm2d_add_rs(input, output, stride, tx_type, bd);
   }
 
-  fn inv_txfm2d_add_rs(
-    input: &[i32], output: &mut [u16], stride: usize, tx_type: TxType,
+  fn inv_txfm2d_add_rs<T>(
```
ycho: Do you know where this function figure out the size of tx?  Would Self::W and Self::H be that info?

tdaede: Yes, exactly.

src/util.rs
```diff
@@ -207,9 +207,12 @@ pub fn clamp<T: PartialOrd>(input: T, min: T, max: T) -> T {
   }
 }
 
-use num_traits::PrimInt;
+use num_traits::*;
 use std::mem::size_of;
 
+pub trait Pixel: PrimInt + Into<u32> + Into<i32> + AsPrimitive<i32> + 'static {}
+impl<T> Pixel for T where T: PrimInt + Into<u32> + Into<i32> + AsPrimitive<i32> + 'static {}
```
lu-zero: Wouldn't be better 
``` rust
impl Pixel for u8 {}
impl Pixel for u16 {}
```

tdaede: Ah yeah it would.

tdaede:
OK. I'm going to rebase this on 64x64 tx and move them all over at once.

tdaede:
Can you look over it a gain? Should be no longer wip

PR #703, opened 2018-11-06
Fix many complaints from clippy

shssoichiro:
Did not fix ones which may negatively impact performance, are likely to
make code harder to read, or are difficult to implement, but this does
considerably decrease the number of lints which are triggered by clippy.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -79,7 +79,7 @@ impl Config {
     let fi = FrameInvariants::new(
       self.frame_info.width,
       self.frame_info.height,
-      self.enc.clone()
+      self.enc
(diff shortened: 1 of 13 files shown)


lu-zero:
Looks nice but it conflates too many changes at the same time, could you please split them per clippy lint? Dropping unneeded `.clone()`s and `return`s is surely ok, using `from()` might impact ongoing sets so might be postponed.

shssoichiro:
This makes sense. I'll close this PR and split it out into smaller ones, then.

lu-zero:
Thank you :)

PR #704, opened 2018-11-06
Cleanup unneeded return statements identified by clippy

shssoichiro:
It is more idiomatic in Rust to omit the `return` keyword when possible.
This is a purely stylistic lint and has no effect on performance.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -805,7 +805,7 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
 
           self.write_bit(true)?; // trailing bit
           self.byte_align()?;
-          return Ok((()));
+          return Ok(());
         }
         self.write_bit(false)?; // show_existing_frame=0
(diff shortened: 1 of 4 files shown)


src/encoder.rs
```diff
@@ -805,7 +805,7 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
 
           self.write_bit(true)?; // trailing bit
           self.byte_align()?;
-          return Ok((()));
+          return Ok(());
```
lu-zero:
```suggestion
          Ok(())
```
Maybe?

shssoichiro: In this case, since it's an early return, we need the `return` keyword. However, since there was only one instance of the "extra parens" lint being triggered, and it was in a return statement, it seemed reasonable to group it in with this PR, hence this line being changed.

lu-zero: I see, I didn't check the whole codeflow :)

PR #707, opened 2018-11-06
[WIP] Add horz/vert (2:1 only) partitions to top-down partition rdo

ycho:
Not complete yet.

TODO : 
- [x] Required to add rectangular transforms, i.e. 8x4, 4x8, 16x8, 8x16, 32x16, 16x32, 32x64, 64x32
- [x] Check the rectangular transforms work correctly with existing helper functions
- [x] Check currently enabled intra prediction code can also work for rectangular cases
- [x] Test top-down partition rdo with new partitions added
- [ ] Check with each of minimum block size
- [ ] Check both of intra and inter modes works correctly with rectangular block sizes/

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -21,6 +21,7 @@ use rdo::*;
 use segmentation::*;
 use transform::*;
 use util::*;
+use partition::PartitionType::*;
 
 use bitstream_io::{BitWriter, BigEndian, LittleEndian};
 use std;
@@ -2380,12 +2381,14 @@ fn encode_partition_topdown(seq: &Sequence, fi: &FrameInvariants, fs: &mut Frame
         return;
     }
 
-    let bs = bsize.width_mi();
+    let bsw = bsize.width_mi();
+    let bsh = bsize.height_mi();
+    let is_square = bsw == bsh;
 
     // Always split if the current partition is too large
-    let must_split = bo.x + bs as usize > fi.w_in_b ||
-        bo.y + bs as usize > fi.h_in_b ||
-        bsize > BlockSize::BLOCK_64X64;
+    let must_split = (bo.x + bsw as usize > fi.w_in_b ||
+        bo.y + bsh as usize > fi.h_in_b ||
+        bsize > BlockSize::BLOCK_64X64) && is_square;
 
     let mut rdo_output = block_output.clone().unwrap_or(RDOOutput {
         part_type: PartitionType::PARTITION_INVALID,
@@ -2397,7 +2400,7 @@ fn encode_partition_topdown(seq: &Sequence, fi: &FrameInvariants, fs: &mut Frame
     if must_split {
         // Oversized blocks are split automatically
         partition = PartitionType::PARTITION_SPLIT;
-    } else if bsize > fi.min_partition_size {
+    } else if bsize > fi.min_partition_size && is_square {
         // Blocks of sizes within the supported range are subjected to a partitioning decision
         rdo_output = rdo_partition_decision(seq, fi, fs, cw,
             w_pre_cdef, w_post_cdef, bsize, bo, &rdo_output, pmvs);
@@ -2407,11 +2410,9 @@ fn encode_partition_topdown(seq: &Sequence, fi: &FrameInvariants, fs: &mut Frame
         partition = PartitionType::PARTITION_NONE;
     }
 
-    assert!(bsize.width_mi() == bsize.height_mi());
     assert!(PartitionType::PARTITION_NONE <= partition &&
             partition < PartitionType::PARTITION_INVALID);
 
-    let hbs = bs >> 1; // Half the block size in blocks
     let subsize = bsize.subsize(partition);
 
     if bsize >= BlockSize::BLOCK_8X8 {
@@ -2501,8 +2502,13 @@ fn encode_partition_topdown(seq: &Sequence, fi: &FrameInvariants, fs: &mut Frame
                           mode_luma, mode_chroma, ref_frames, mvs, bsize, bo, skip, seq.bit_depth, cfl,
                           tx_size, tx_type, mode_context, &mv_stack, false);
         },
-        PartitionType::PARTITION_SPLIT => {
-            if rdo_output.part_modes.len() >= 4 {
+        PARTITION_SPLIT |
+        PARTITION_HORZ |
+        PARTITION_VERT => {
+            let num_modes = if partition == PARTITION_SPLIT { 4 }
+                            else { 2 };
+
+            if rdo_output.part_modes.len() >= num_modes {
                 // The optimal prediction modes for each split block is known from an rdo_partition_decision() call
                 assert!(subsize != BlockSize::BLOCK_INVALID);
 
@@ -2518,12 +2524,25 @@ fn encode_partition_topdown(seq: &Sequence, fi: &FrameInvariants, fs: &mut Frame
                 }
             }
             else {
-                let partitions = [
-                    bo,
-                    &BlockOffset{ x: bo.x + hbs as usize, y: bo.y },
-                    &BlockOffset{ x: bo.x, y: bo.y + hbs as usize },
-                    &BlockOffset{ x: bo.x + hbs as usize, y: bo.y + hbs as usize }
-                ];
+                let hbsw = subsize.width_mi(); // Half the block size width in blocks
+                let hbsh = subsize.height_mi(); // Half the block size height in blocks
+
+                let p1 = BlockOffset{ x: bo.x + hbsw as usize, y: bo.y };
+                let p2 = BlockOffset{ x: bo.x, y: bo.y + hbsh as usize };
+                let p3 = BlockOffset{ x: bo.x + hbsw as usize, y: bo.y + hbsh as usize };
+
+                let mut partitions = vec![ bo ];
+
+                if partition == PARTITION_VERT || partition == PARTITION_SPLIT {
+                partitions.push(&p1);
+                };
+                if partition == PARTITION_HORZ || partition == PARTITION_SPLIT {
+                partitions.push(&p2);
+                };
+                if partition == PARTITION_SPLIT {
+                partitions.push(&p3);
+                };
+
                 partitions.iter().for_each(|&offset| {
                         encode_partition_topdown(
                             seq,
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -2206,7 +2206,7 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
     // Always split if the current partition is too large
     let must_split = bo.x + bs as usize > fi.w_in_b ||
         bo.y + bs as usize > fi.h_in_b ||
-        bsize > BlockSize::BLOCK_64X64;
+        bsize.greater_than(BlockSize::BLOCK_64X64);
 
     // must_split overrides the minimum partition size when applicable
     let can_split = bsize > fi.min_partition_size || must_split;
@@ -2238,7 +2238,7 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
 
         let mut cost: f64 = 0.0;
 
-        if bsize >= BlockSize::BLOCK_8X8 {
+        if bsize.gte(BlockSize::BLOCK_8X8) {
             let w: &mut dyn Writer = if cw.bc.cdef_coded {w_post_cdef} else {w_pre_cdef};
             let tell = w.tell_frac();
             cw.write_partition(w, bo, partition, bsize);
@@ -2294,7 +2294,7 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
 
         rd_cost = 0.0;
 
-        if bsize >= BlockSize::BLOCK_8X8 {
+        if bsize.gte(BlockSize::BLOCK_8X8) {
             let w: &mut dyn Writer = if cw.bc.cdef_coded {w_post_cdef} else {w_pre_cdef};
             let tell = w.tell_frac();
             cw.write_partition(w, bo, partition, bsize);
@@ -2331,7 +2331,7 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
 
             partition = PartitionType::PARTITION_NONE;
 
-            if bsize >= BlockSize::BLOCK_8X8 {
+            if bsize.gte(BlockSize::BLOCK_8X8) {
                 let w: &mut dyn Writer = if cw.bc.cdef_coded {w_post_cdef} else {w_pre_cdef};
                 cw.write_partition(w, bo, partition, bsize);
             }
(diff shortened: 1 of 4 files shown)


src/partition.rs
```diff
@@ -662,7 +662,8 @@ pub const GLOBALMV_CTX_MASK: usize =
 pub const REFMV_CTX_MASK: usize = ((1 << (8 - REFMV_OFFSET)) - 1);
 
 pub static RAV1E_PARTITION_TYPES: &'static [PartitionType] =
-  &[PartitionType::PARTITION_NONE, PartitionType::PARTITION_SPLIT];
+  &[PartitionType::PARTITION_NONE, /*PartitionType::PARTITION_HORZ,
+    PartitionType::PARTITION_VERT,*/ PartitionType::PARTITION_SPLIT];
```
lu-zero: Why?

src/rdo.rs
```diff
@@ -853,14 +855,25 @@ pub fn rdo_partition_decision(
...
-        let bs = bsize.width_mi();
-        let hbs = bs >> 1; // Half the block size in blocks
-        let partitions = [
-          bo,
-          &BlockOffset{ x: bo.x + hbs as usize, y: bo.y },
-          &BlockOffset{ x: bo.x, y: bo.y + hbs as usize },
-          &BlockOffset{ x: bo.x + hbs as usize, y: bo.y + hbs as usize }
-        ];
+
+        let hbsw = subsize.width_mi(); // Half the block size width in blocks
```
lu-zero: This could be factored out probably?

src/partition.rs
ycho: As 'Prep' means, this patch has not enabled actual use of horz and vert yet!

src/rdo.rs
ycho: Not sure what you meant but I derived hbsw, since it is referenced more than twice down below for p1 and p3?

lu-zero: The whole block :)

ycho: Oh, you probably mean another identical block of code shown in encode_partition_topdown(), yes I believe so, like given decide_partitions(partition) { ... }.

tdaede:
What's the status of this? Waiting for review?

ycho:
There is a change (factor out a block of duplicate code)  requested by lu_zero, so propably I need to address on it. But, we can do it on separate patch as well.

src/rdo.rs
ycho: If I want to factor out line 878 to 885, say get_sub_partitions(partition, subsize, bo)  -> partitions, 
I need to create (allocate) then return some (or all) of p1,p2, p3, which are pointer to BlockOffset.
What would be best way to create those values (pointers to BlockOffset) then return them that can be still live even after get_sub_partitions() function call is finished? So, with not passing p1,p2,p3 to get_sub_partitions().

src/rdo.rs
```diff
@@ -823,6 +823,23 @@ pub fn rdo_tx_type_decision(
   best_type
 }
 
+pub fn get_sub_partitions<'a>(four_partitions: &[&'a BlockOffset; 4],
```
ycho: @lu-zero, Can you check my rust code to factor out the block of code seem correct?

ycho:
@lu_zero: Would you have other comments with this PR?

ycho:
Causing corrupted enc and dec side frames, when hor and ver partitions enabled. So, DON'T MERGE.

ycho:
@fbossen, Currently, it does not crash and generate bitstream for -s 4, but with regression.
For -s 3 .. -s 1, encoder fails to generater full bitstream.

ycho:
FYI; you can use GUI av1 stream analyzer by mbx like this:
(download aomanalyzer binary from [here](https://people.xiph.org/~mbebenita/analyzer/)
/Users/yushin/Downloads/AOMAnalyzer-darwin-x64/AOMAnalyzer.app/Contents/MacOS/AOMAnalyzer http//people.xiph.org/~mbebenita/analyzer/inspect.js  ../rav1e/test.ivf

ycho:
Even with one frame only at -s 4, i.e. intra frame, there is regression.
So, there would be issues in both intra and inter modes.
[AWCY link](https://beta.arewecompressedyet.com/?job=master_c27a56946f0af5_1f%402018-11-21T19%3A43%3A51.676Z&job=enable_hor_ver_par_types_with_top_down_partition_rdo_1f%402018-11-21T19%3A44%3A30.880Z)

src/rdo.rs
```diff
@@ -859,14 +897,16 @@ pub fn rdo_partition_decision(
...
-          &BlockOffset{ x: bo.x, y: bo.y + hbs as usize },
-          &BlockOffset{ x: bo.x + hbs as usize, y: bo.y + hbs as usize }
+          &BlockOffset{ x: bo.x + hbsw as usize, y: bo.y },
+          &BlockOffset{ x: bo.x, y: bo.y + hbsh as usize },
+          &BlockOffset{ x: bo.x + hbsw as usize, y: bo.y + hbsh as usize }
         ];
+        let partitions = get_sub_partitions(&four_partitions, partition);
 
         let pmv_idxs = partitions.iter().map(|&offset| {
           if subsize > BlockSize::BLOCK_32X32 {
```
ycho: For splitting into 64x32 or 32x64, I think this should be fixed as : subsize.larger_than(BlockSize::BLOCK_32X32).

src/encoder.rs
```diff
@@ -1830,10 +1833,18 @@ pub fn write_tx_blocks(fi: &FrameInvariants, fs: &mut FrameState,
 
     // TODO: these are only valid for 4:2:0
     let uv_tx_size = match bsize {
-        BlockSize::BLOCK_4X4 | BlockSize::BLOCK_8X8 => TxSize::TX_4X4,
+        BlockSize::BLOCK_4X4 | BlockSize::BLOCK_4X8 |
```
lu-zero: this and the similar block below (line 1931) should/would be always the same?

src/rdo.rs
lu-zero: `get_sub_partitions` ? It is fine

src/rdo.rs
lu-zero: I somehow missed this comment

src/encoder.rs
ycho: Yes, so can be done by lookup table but have different mapping for each chroma subsampling, i.e. 422 and 444 will have different. 
I expect libaom has tables for this case, sth like max_tx_size[chroma_sample_x][chroma_sample_y][block_size] --> tx_size.

ycho: Line 702 in src/rdo.rs seems have same mapping as well.

src/rdo.rs
ycho: Ok- thanks!

ycho:
We have gain for -s4 (i.e. 64x32 and 32x64 introduced to 32x32 or 64x64), but unfortunately it does not work yet for other speeds. For -s 3, only first few frames seem decoded correctly/

| -1.4423 |  0.0906 |  0.2046 |  -1.6500 | -2.3018 | -2.2945 |    -1.9633 |

[AWCY link for -s 4](https://beta.arewecompressedyet.com/?job=master_2e4055a_s4%402018-11-30T01%3A01%3A02.828Z&job=test_rect_s4%402018-11-30T01%3A01%3A38.247Z)

ycho:
btw, I see that there is desync for -s 1, i.e. 4x4.

ycho:
Sadly, with one frame encoding only (so intra mode only) on AWCY, regressions for all -s 1, 2, 3.

ycho:
The 8x4 and 4x8 issue has been fixed by #788 (was an CFL issue)

ycho:
So, confirm that rectangular(2:1) partitions are working for all current intra mode cases with top-down partition rdo.

fbossen:
For speed levels other than 4 it seems that modifying 2 lines in has_tr() does help:
<pre>
-    if (target_n4_w < target_n4_h) && !is_sec_rect {
+    if (target_n4_w < target_n4_h) && (bo.x & target_n4_w) == 0 {
</pre>
and
<pre>
-    if (target_n4_w > target_n4_h) && is_sec_rect {
+    if (target_n4_w > target_n4_h) && (bo.y & target_n4_h) != 0 {
</pre>

ycho:
Thanks!
Also, I think that motion_compensate() would need revision for 4x8 and 8x4 cases, which currently handles 4x4 only.

ycho:
With your suggested change, it works for -3 and 2, but desync for -s 1. But still BIG progress and almost done!

ycho:
I can try revise motion_compensate() for 4x8 and 8x4.
Btw, I wonder what 'sec' in is_sec_rect did mean? :)

fbossen:
I think is_sec_rect meant "is second rectangular partition"

fbossen:
For the motion_compensate() function, the following seems to help:
<pre>
-    if p > 0 && bsize == BlockSize::BLOCK_4X4 {
+    if p > 0 && bsize < BlockSize::BLOCK_8X8 {
+      let bo = BlockOffset { x: bo.x | 1, y: bo.y | 1 };
</pre>

ycho:
Yes, but some more changes like doing MC for only two of 2x2 chroma pixels if blk sizes are 4x8 or 8x4. There is desync w/o it.

src/partition.rs
```diff
@@ -214,6 +214,16 @@ impl BlockSize {
 
     (offset_x, offset_y)
   }
+  
+  pub fn larger_than(self, other: BlockSize) -> bool {
+    (self.width() > other.width() && self.height() >= other.height()) ||
+    (self.width() >= other.width() && self.height() > other.height())
+  }
+
+  pub fn lte(self, other: BlockSize) -> bool {
```
fbossen: Rename this to larger_than_or_equal_to? "lte" could be interpreted as "less than or equal to"

ycho: Indeed, thanks for pointing out. (I was also seeing that likely misleading). Done.

src/encoder.rs
ycho: This issue has been resolved by #784.

src/rdo.rs
ycho: We can test which way is better.

src/partition.rs
```diff
@@ -214,6 +214,16 @@ impl BlockSize {
 
     (offset_x, offset_y)
   }
+  
+  pub fn greather_than(self, other: BlockSize) -> bool {
```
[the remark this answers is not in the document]
ycho: The reason we cannot use those operators is that BlockSize enum is not ordered, so....

src/encoder.rs
```diff
@@ -2362,7 +2367,7 @@ fn encode_partition_topdown(seq: &Sequence, fi: &FrameInvariants, fs: &mut Frame
     if must_split {
         // Oversized blocks are split automatically
         partition = PartitionType::PARTITION_SPLIT;
```
fbossen: When must_split is true, there may be cases where it is ok to use a horizontal or a vertical split, which may be more effective

ycho: Indeed, I will add a patch for it and report the gain by it.

ycho: After further checking and trying some implement, I think this place is not where it considers rectangular partitions with topdown. Because, rect partition is the leaf nodes, so for ex, once 64x64 is given then say it straddle on the right frame border thus wants to use 32x64 but that is not possible if only 8 pixels are outside the right frame border.
Instead, as in current code intends, once 16x16 square block is given to topdown_partition_rdo(), then it calls rdo_partition_decision() can pick 8x16 partition, if it beats 8x8s. HOWEVER, that seems not happening now, unfortunately, which I should fix.

ycho: Also, by av1 spec, the small rec partition size, like 8x4 or 4x8 are encoded by way of series of PARTITION_SPLITS to reach its next larger sqr size, i.e. 8x8 then finally encode PARTITION_HORZ or PARTITION_VERT.

ycho:
Now, the gain is -0.75% with default speed -s 3, but with almost x2 encoding time.

| -0.7523 | -2.2584 | -2.3763 |  -0.8520 | -1.4396 | -1.4903 |    -2.5631 |

[AWCY link](https://beta.arewecompressedyet.com/?job=test_fix_rect_s3%402018-12-10T20%3A51%3A20.211Z&job=master_66b4cd97_s3%402018-12-10T22%3A10%3A21.769Z)

So, we hold landing this PR until we obtain more gain (by finding and fixing bugs), otherwise probably we enable rectangular partitions for frame borders and/or for slower speed if it helps.

ycho:
Already found one new bug!
commit as : ca96e67429623f886249d3b7071e0ae5548ccacb

ycho:
If not other undiscovered bugs, another possible reasons why rectangular partitions does not give more gains with ra1ve would be missing directional intra predictions.

fbossen:
@ycho Do we know know how much gain there is if using rectangular only on frame boundaries? The impact on run time should be much lower.

ycho:
I've just run it (i.e. w/o enabling rect block on frame border) on awcy under name "fix4_rect_s3". fix5_rect_* are with the last fix.

[AWCY result will come here later](https://beta.arewecompressedyet.com/?job=fix4_rect_s3%402018-12-10T23%3A54%3A08.212Z&job=fix5_rect_s3%402018-12-10T23%3A21%3A06.407Z)

ycho:
After fix, -0.85% for -s 3.

| -0.8540 | -2.3654 | -2.4245 |  -0.9664 | -1.5547 | -1.6135 |    -2.6606 |

[AWCY result for -s 3](https://beta.arewecompressedyet.com/?job=fix5_rect_s3%402018-12-10T23%3A21%3A06.407Z&job=master_66b4cd97_s3%402018-12-10T22%3A10%3A21.769Z)

ycho:
@fbossen, I've commited frame-border only rectangular partitions, and ran it on awcy. I will share the results later.
The yet another reason of smaller-than-expected gain would be the way partitioning is done, 'topdown', as we have discussed shortly before, right? :)

ycho:
Enabling rectangular partition on frame borders only, the gain is -0.83% and almost no change in encoding time!

| -0.8346 | -1.0556 | -1.1209 |  -0.8008 | -0.8536 | -0.9104 |    -1.0911 |

[AWCY result]( https://beta.arewecompressedyet.com/?job=rect_on_frame_borders_only_s3%402018-12-11T00%3A35%3A43.682Z&job=master_66b4cd97_s3%402018-12-10T22%3A10%3A21.769Z)

ycho:
For -s 2 :

| -1.1888 | -1.4412 | -1.2668 |  -1.1731 | -1.1844 | -1.2189 |    -1.4010 |

[AWCY link for -s 2](https://beta.arewecompressedyet.com/?job=rect_on_frame_borders_only_s2%402018-12-11T00%3A36%3A19.832Z&job=master_66b4cd97_s2%402018-12-10T22%3A10%3A40.763Z)

ycho:
At -s 1, -1.26% gain.

| -1.2633 | -1.4617 | -1.3863 |  -1.2539 | -1.2575 | -1.2742 |    -1.3602 |

[AWCY link for -s 1](https://beta.arewecompressedyet.com/?job=rect_on_frame_borders_only_s1%402018-12-11T00%3A36%3A31.767Z&job=master_66b4cd97_s1%402018-12-10T22%3A10%3A52.245Z)

PR #710, opened 2018-11-08
Pass small copy types by value

shssoichiro:
It is more performant to pass small types which implement `Copy` by
value, because they can be passed directly through registers. Clippy is
conservative with this lint and optimizes for 32-bit architectures. The
setting can be overridden in `clippy.toml` if we choose to further
optimize for 64-bit architectures.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1475,13 +1475,13 @@ impl BlockContext {
     }
   }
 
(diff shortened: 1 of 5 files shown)


tdaede:
Looks good! In fact I think optimizing for 64 bit would make the most sense, if you want to open a pr for that.

PR #711, opened 2018-11-08
Add support for coding 64x64 blocks

fbossen:
Inverse 64-point transform is added.
Forward 64-point transform computes all 4096 coefficients such
that distortion from zeroing high-frequency coefficients can
be easily computed

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -470,7 +470,8 @@ impl FrameInvariants {
         let mut min_partition_size = if config.speed <= 1 { BlockSize::BLOCK_4X4 }
                                  else if config.speed <= 2 { BlockSize::BLOCK_8X8 }
                                  else if config.speed <= 3 { BlockSize::BLOCK_16X16 }
-                                 else { BlockSize::BLOCK_32X32 };
+                                 else if config.speed <= 4 { BlockSize::BLOCK_32X32 }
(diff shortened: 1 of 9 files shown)


fbossen:
At default speed/settings:

| -2.5005 | -4.9616 | -6.4775 |  -2.5585 | -2.6400 | -2.7794 |    -5.4490 |

src/encoder.rs
```diff
@@ -470,7 +470,8 @@ impl FrameInvariants {
         let mut min_partition_size = if config.speed <= 1 { BlockSize::BLOCK_4X4 }
                                  else if config.speed <= 2 { BlockSize::BLOCK_8X8 }
                                  else if config.speed <= 3 { BlockSize::BLOCK_16X16 }
-                                 else { BlockSize::BLOCK_32X32 };
+                                 else if config.speed <= 3 { BlockSize::BLOCK_32X32 }
```
SmilingWolf: This change leaves the final source as
```
else if config.speed <= 3 { BlockSize::BLOCK_16X16 }
else if config.speed <= 3 { BlockSize::BLOCK_32X32 }
```
so mutually exclusive checks for the same condition with different outcomes, one of which is never reached.
Is this intended? Or was the second line meant to be for speed <= 4?

fbossen: It was meant to be <= 4. Nice catch!

PR #714, opened 2018-11-08
Refactor speed level to act as an initialization preset

### LINKED ISSUE #290: Mapping speed level control in one place
It would be more maintainable if we do map speed level to activation of modes in one place, 
instead of having "if fi.speed > x " conditions several places over the codebase.

Also, skip is currently disabled from level > 1, but I expect forthcoming inter mode will need it almost at all speed levels.

shssoichiro:
This opens the path to finer control by the user over the encoding
options, and centralizes the speed preset logic in one location.

Closes #290

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -33,14 +33,115 @@ pub struct EncoderConfig {
   pub key_frame_interval: u64,
   pub low_latency: bool,
   pub quantizer: usize,
-  pub speed: usize,
-  pub tune: Tune
+  pub tune: Tune,
+  pub speed_settings: SpeedSettings,
 }
 
 impl Default for EncoderConfig {
   fn default() -> Self {
-    EncoderConfig { key_frame_interval: 30, low_latency: true, quantizer: 100, speed: 0, tune: Tune::Psnr  }
+    const DEFAULT_SPEED: usize = 3;
+    Self::with_speed_preset(DEFAULT_SPEED)
   }
+}
+
+impl EncoderConfig {
+  pub fn with_speed_preset(speed: usize) -> Self {
+    EncoderConfig {
+      key_frame_interval: 30,
+      low_latency: true,
+      quantizer: 100,
+      tune: Tune::Psnr,
+      speed_settings: SpeedSettings::from_preset(speed)
+    }
+  }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct SpeedSettings {
+  pub min_block_size: BlockSize,
+  pub multiref: bool,
+  pub fast_deblock: bool,
+  pub reduced_tx_set: bool,
+  pub tx_domain_distortion: bool,
+  pub encode_bottomup: bool,
+  pub rdo_tx_decision: bool,
+  pub prediction_modes: PredictionModesSetting,
+  pub include_near_mvs: bool,
+}
+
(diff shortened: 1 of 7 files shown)


shssoichiro:
I'm not completely confident that this is the cleanest approach and am open to suggestions; however, this config struct seemed like the best location to store these flags, since this struct is already passed around everywhere.

lu-zero:
I like it but I know that some other people might have diverging opinions :) 

@tdaede and @tterribe : would be ok for you?

tdaede:
Something like this is going to be needed for calibrating speed levels better. Currently our choices are rather arbitrary, but what I'd actually like to do is an automated training, where the encoder is run enabling these features one at a time, and the relative impacts are assessed.

shssoichiro:
Seems I broke the tests... One second and I'll fix those.

src/api.rs
```diff
@@ -33,16 +33,110 @@ pub struct EncoderConfig {
...
+  pub rdo_tx_decision: bool,
+  pub prediction_modes: PredictionModesSetting,
+  pub include_near_mvs: bool,
 }
 
 impl Default for EncoderConfig {
   fn default() -> Self {
-    EncoderConfig { key_frame_interval: 30, low_latency: true, quantizer: 100, speed: 0, tune: Tune::Psnr  }
+    const DEFAULT_SPEED: usize = 3;
+    Self::with_speed_preset(DEFAULT_SPEED)
```
lu-zero: This would loop infinitely

src/api.rs
```diff
@@ -33,16 +33,110 @@ pub struct EncoderConfig {
...
+      low_latency: true,
+      quantizer: 100,
+      tune: Tune::Psnr,
+      ..Default::default()
+    };
+    cfg.apply_speed_preset(speed);
+    cfg
+  }
+
+  pub fn apply_speed_preset(&mut self, speed: usize) {
```
ycho: Yes! This is wonderful way to keep only once function to control various switch affecting the time complexity of encoder. Thanks for nice idea!

src/api.rs
```diff
@@ -33,16 +33,110 @@ pub struct EncoderConfig {
...
+      BlockSize::BLOCK_64X64
+    }
+  }
+
+  fn multiref_preset(speed: usize) -> bool {
+    speed <= 2
+  }
+
+  fn fast_deblock_preset(speed: usize) -> bool {
+    speed > 3
```
ycho: Just, as preference, can you keep using <= and >=, so that inclusive? I.e. speed >= 4
(Sorry- I could not be picky here)

src/api.rs
```diff
@@ -33,16 +33,110 @@ pub struct EncoderConfig {
...
+  }
+
+  fn reduced_tx_set_preset(speed: usize) -> bool {
+    speed > 1
```
ycho: Same here, speed >= 2.

src/api.rs
lu-zero: You may have a `SpeedSettings` struct with the standard default impl and embed it in the `EncoderConfig`.

ycho: AWESOME IDEA!

lu-zero:
It needs a rebase now, could you please update? :)

PR #718, opened 2018-11-09
Improve CLI progress indicator and print stats after encode

shssoichiro:
Adds a new `--verbose` flag to display frame-by-frame info similar to
the old behavior.

While encoding, displays info similar to this:
```
╰─ rav1e ~/bbb_480p.y4m -o /dev/null --limit 100
848x480 @ 60/1 fps
encoded 99/100 frames, 1.23 fps, 419.91 Kb/s, est. size: 0.09 MB
```

The final line will update after each frame encoded. At the end of encoding, will print stats as follows:
```
╰─ rav1e ~/bbb_480p.y4m -o /dev/null --limit 100
848x480 @ 60/1 fps
encoded 100/100 frames, 1.23 fps, 419.91 Kb/s, est. size: 0.09 MB
Key Frames:      4    avg size:    6485 B
Inter:          96    avg size:     662 B
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ autobins = false
 repl = ["rustyline", "binaries"]
 comparative_bench = ["aom"]
 decode_test = ["bindgen", "aom"]
-binaries = ["y4m", "clap"]
+binaries = ["y4m", "clap", "termion"]
 default = ["binaries"]
 aom = ["cmake"]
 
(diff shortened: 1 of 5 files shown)


shssoichiro:
I looked into `indicatif` but the current limitation is that the y4m parser doesn't know how many frames are in the file (until we've finished encoding). So we can only calculate percent complete if `--limit` has been specified. I thought it would be preferable to have a more consistent output to the screen, rather than a progress bar only if the encoder knows the frame count.

PR #723, opened 2018-11-11
Less boilerplate

lu-zero:
The current stable eventually made possible to paste identifiers together, here a example on how to leverage it.

Ideally also the match we use in PredictionMode could be simplified, but right now I have a keyboard problem and typing is a bit slow, so I send this example before going further :)

### DIFF
--- a/src/predict.rs
+++ b/src/predict.rs
@@ -122,6 +122,9 @@ macro_rules! block_dimension {
         const W: usize = $W;
         const H: usize = $H;
       }
+
+      impl Intra<u8> for [<Block $W x $H>] {}
+      impl Intra<u16> for [<Block $W x $H>] {}
     }
   };
 }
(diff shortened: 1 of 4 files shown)


src/predict.rs
```diff
@@ -113,137 +113,32 @@ pub trait Dim {
...
-}
-
-pub struct Block32x8;
-
-impl Dim for Block32x8 {
-  const W: usize = 32;
-  const H: usize = 8;
+macro_rules! block_dimension {
+  ($W:expr, $H:expr) => {
+    paste::item! {
```
ycho: Can you explain what would be "item" here? Is it everything inside { ... } next to "item!"?

src/predict.rs
```diff
@@ -113,137 +113,32 @@ pub trait Dim {
...
+    }
+  };
 }
 
-pub struct Block16x64;
-
-impl Dim for Block16x64 {
-  const W: usize = 16;
-  const H: usize = 64;
+macro_rules! blocks_dimension {
```
ycho: Thank you. This is indeed much better!
So, I understand that the rustc itself copies N of "block_dimension! { $W, $H }" here, correct?
Then, I wonder which syntax actually dose that N copying, i.e. iterating?  "$(" .. ")"?

lu-zero: The macros have the concept of [repetition](https://doc.rust-lang.org/stable/rust-by-example/macros/repeat.html), you see in `$( ... )*` below how I'm using it. In general you express with `$(some pattern),+` or `$(some pattern),*` in the macro argument position the repeating sequence and then you have it repeated by putting in the body `$(` + `something you want to repeat`  +  `)*`.

src/predict.rs
ycho: I see, so in our case we use 'paste' with 'item' for copying declarations, not an expression.
I plan to use 'paste' to add rectangular tx functions, at least for declaration part, with 'item' then.
Do you think we can try even more tricks to shorten pub fn predict_intra<'a>() as well?
If possible, same idea can be applied to rect tx functions dispatcher, i.e. forward_transform() and inverse_transform_add().

tdaede:
Is this going to bump minimum rust version?

PR #726, opened 2018-11-13
Require correct minimal syn version.

tdaede:


Cargo.toml
```diff
@@ -23,6 +23,7 @@ rand = "0.5"
 rustyline = { version = "1", optional = true }
 y4m = { version = "0.3", optional = true }
 backtrace = "0.3"
+syn = "^0.15.20"
```
ycho: '^' ?

tdaede: it allows all versions above 0.15.20 that are still 0.15

ycho: oh...

lu-zero: "Version at least"

PR #729, opened 2018-11-13
Fix 64x64 tx to copy only 32x32 coeffs to output buffer

ycho:
- Also fix tx domain distortion computation for tx sizes having 32
   as one dimension.
- Improve quantize() to use adjusted_tx_size.


tdaede:
Does this have an AWCY improvement?

src/context.rs
```diff
@@ -1764,6 +1764,20 @@ macro_rules! symbol_with_update {
   };
 }
 
+pub fn av1_get_adjusted_tx_size(tx_size: TxSize) -> TxSize {
```
tdaede: Is there a better name for this? Or a comment? Like "nonzero_coeffs_size" or something?

ycho: The adjusted size is what level map coding is seeing, so maybe coded_size then

ycho:
Let me run awcy to check regression.

src/context.rs
ycho: so, av1_get_coded_tx_size()

ycho:
[AWCY results for -s 4 (i.e. 64x64s), --limit=5](https://beta.arewecompressedyet.com/?job=master-34370492343f7d9f80445%402018-11-14T00%3A14%3A25.833Z&job=fix_64x64tx-b0a28781%402018-11-14T00%3A15%3A10.946Z) will come soon.

ycho:
I think that tx-domain distortion for tx_size > 32x32 needs original tx coeffs, which currently wiped out by this patch. Meanwhile, @fbossen has point that he preserves those not coded coeffs (but original) so that it can be used during tx-domain distortion.

src/context.rs
tdaede: Er, that's also confusing because it's not the tx size we write :/

ycho:
Wow, my last trial (don't use tx-domain distortion if some of tx coeffs are not coded) seems working but it loses coding efficiency say 0.54% and slower by ~10%.
[AWCY results](https://beta.arewecompressedyet.com/?job=master-34370492343f7d9f80445%402018-11-14T00%3A14%3A25.833Z&job=fix2_64x64tx-b0a28781%402018-11-14T01%3A07%3A59.458Z)
So, I don't think we can merge this PR, though nice experimentation! :)
(But I am happy that tx-domain distortion is more helping with 64x64 tx)

ycho:
Or, we can use other parts except the change I did on 64x64tx.

PR #730, opened 2018-11-14
Update README.md

### LINKED ISSUE #449: Add info about CPU used to measure performance of the encoder
The [README.md](../blob/d9dfe043c8d7642e3c24290c7a64d01298646a63/README.md#features) file says in the "features" section that the current performance of the encoder is "~10 fps encoding @ 480p". Since encoding depends on a CPU, and there's no info about which CPU was used, the info about 10 fps may be misleading.

Adding info about the CPU would solve this problem, and would help people to understand which CPU is needed to achieve such performance.

rzumer:
Remove specific frame rate number as it depends on hardware, update transforms.

Closes #449.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,10 +13,10 @@ rav1e is an experimental AV1 video encoder. It is designed to eventually cover a
 * 64x64 superblocks
 * 4x4 to 64x64 RDO-selected square blocks
 * DC, H, V, Paeth, and smooth prediction modes
-* 4x4 DCT and ADST transforms
+* DCT, ADST and identity transforms (up to 64x64, 16x16 and 32x32 respectively)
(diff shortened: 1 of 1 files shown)


rektide:
This seems like negative progress. I'd way rather have some general expectation & idea, than absolutely nothing at all. I'd like this reverted and/or a proper attempt to satisfy #449.

rzumer:
There is no point in reverting this change, since performance changes constantly as features are added, so accuracy of numbers cannot be guaranteed, while "near real-time" is a reasonable ballpark for expectations on consumer-level hardware. If/when performance is tested systematically, then more info will be available, but it is still early in development, so it is not a high priority.

PR #734, opened 2018-11-14
Do not write restoration parameters when the feature is disabled

fbossen:


rzumer:
Could you please just inverse the conditional statement to remove the block if nothing needs to be done when the feature is disabled?

fbossen:
@rzumer better now?

rzumer:
Perfect, thank you.

PR #737, opened 2018-11-14
Tidy me.rs

lu-zero:
Move the nasm code in a stand alone module and avoid some warnings when it is disabled.

(on top of the previous pr)


src/me.rs
```diff
@@ -12,111 +12,130 @@ use context::BLOCK_TO_PLANE_SHIFT;
...
-    dst_stride: libc::ptrdiff_t
-  ) -> u32;
-}
+#[cfg(all(target_arch = "x86_64", not(windows), feature = "nasm"))]
+mod nasm {
+  use libc;
+  use plane::*;
+  use util::*;
+
+  extern {
```
KyleSiefring: You could combine the two externs if you are going to remove the separate x86 section.

src/me.rs
```diff
@@ -116,6 +127,13 @@ pub fn get_sad(
 
   sum
 }
+}
+
+#[cfg(all(target_arch = "x86_64", not(windows), feature = "nasm"))]
+pub use self::nasm::get_sad;
+
+#[cfg(not(any(target_arch = "x86_64", not(windows), feature = "nasm")))]
```
KyleSiefring: I'm pretty sure this is why appveyor fails on a windows build.

tdaede:
Please rebase.

src/me.rs
lu-zero: You are absolutely right :) Thank you for spotting it.

PR #740, opened 2018-11-15
Add 14 rectangular tx functions

ycho:


ycho:
No regression by this PR, as shown in [AWCY results](https://beta.arewecompressedyet.com/?job=master_8d9e7453_5f%402018-11-15T00%3A20%3A46.604Z&job=add_rect_tx_5f%402018-11-15T00%3A27%3A03.816Z)

src/transform/forward.rs
```diff
@@ -2057,3 +2073,170 @@ pub fn fht64x64(
...
   }
 }
+
+pub fn fht4x8(
+  input: &[i16], output: &mut [i32], stride: usize, tx_type: TxType,
+  bit_depth: usize
+) {
+  // SIMD code may assert for transform types beyond TxType::IDTX.
+  if tx_type < TxType::IDTX {
+    Block4x8::fwd_txfm2d(input, output, stride, tx_type, bit_depth);
```
ycho: I expected that any specialized tx function for each different size will be called at this place.
Though, at the moment, different tx size function is fetched by way of : 
    let txfm_func_col = cfg.txfm_type_col.get_func();
    let txfm_func_row = cfg.txfm_type_row.get_func();
in fwd_txfm2d(), which I think I desire occur at earlier stage, such as inside forward_transform() or at elast before all these functions, like fht4x8().., with its function name having different tx size.

lu-zero: I guess this can be handled while refactoring, any volunteer to write more bechmarks for this part?

PR #748, opened 2018-11-15
Fix for 32x16, 32x8, etc. rect tx functions to have idtx tx_type as well

ycho:
32x16, 16x32, 32x8, 8x32 to have idtx in addition to dct_dct.


ycho:
Right, I was pondering whether previous approach of rust code is correct and which way I need extend for rectangular txs, entirely considering about where to put those simd code....

PR #750, opened 2018-11-16
Add missing license headers

rzumer:


src/bin/common.rs
```diff
@@ -1,3 +1,12 @@
+// Copyright (c) 2017-2018, The rav1e contributors. All rights reserved
```
tmatth: LGTM, as an aside I looked this up out of curiousity (granted we're not using the GPL but the copyright year advice generally holds):

> The copyright notice should include the year in which you finished preparing the release (so if you finished it in 1998 but didn't post it until 1999, use 1998). You should add the proper year for each past release; for example, “Copyright 1998, 1999 Terry Jones” if some releases were finished in 1998 and some were finished in 1999. If several people helped write the code, use all their names.
> 
> For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage.
[1] https://www.gnu.org/licenses/gpl-howto.en.html

PR #753, opened 2018-11-16
Add rate constraint to motion estimation

fbossen:


fbossen:
BD rate improvement is small but also seems to bring some run time reduction (possibly because fewer MVs are tested at lower rates)

| -0.2531 | -0.3199 | -0.1577 |  -0.3000 | -0.3222 | -0.3644 |    -0.3088 |

fbossen:
Results with -s 0 are slightly better: 0.5% rate reduction and 5-10% run time reduction

tdaede:
Have you tested with low_latency false yet? Also, does this allow longer MVs?

tdaede:
Is this still a WIP? Do you want me to review it to land it, or are you still trying to get a bigger gain?

fbossen:
@tdaede It's still in progress. I recently increased the precision of some computations and things got worse, for the wikipedia sequence in particular. Need to understand why

fbossen:
Updated RD numbers:

| -0.3812 | -0.1558 | -0.2531 |  -0.4089 | -0.3725 | -0.3843 |    -0.2428 |

With low_latency set to false:

| -0.1339 | -0.1195 | -0.2081 |  -0.1302 | -0.1567 | -0.1560 |    -0.1599 |

src/rdo.rs
```diff
@@ -404,6 +410,15 @@ pub fn rdo_mode_decision(
     mode_contexts.push(cw.find_mvrefs(bo, ref_frames, &mut mv_stack, bsize, false, fi, false));
 
     if fi.frame_type == FrameType::INTER {
+      let mut pmv = [MotionVector{ row: 0, col: 0 }; 2];
```
ycho: So, previously pmv[] was defined for seven different reference frame but now two only. I wonder what kinds impact by this change, if any?

ycho: Oh, I read it wrong. pmvs[] has still slots for seven ref. While pmv[] has had two, for two references, and no change.

src/me.rs
```diff
@@ -159,14 +160,16 @@ pub fn motion_estimation(
...
+      let x_lo = po.x + ((-range + (cmv.col / 8) as isize).max(mvx_min / 8));
+      let x_hi = po.x + ((range + (cmv.col / 8) as isize).min(mvx_max / 8));
+      let y_lo = po.y + ((-range + (cmv.row / 8) as isize).max(mvy_min / 8));
+      let y_hi = po.y + ((range + (cmv.row / 8) as isize).min(mvy_max / 8));
 
-      let mut lowest_sad = 128 * 128 * 4096 as u32;
+      let mut lowest_cost = std::u32::MAX;
       let mut best_mv = MotionVector { row: 0, col: 0 };
 
+      let lambda = (get_lambda_sqrt(fi, bit_depth) * 128.0) as u32;
```
tdaede: Lambda is Q7 here?

src/me.rs
```diff
@@ -261,12 +273,19 @@ fn full_search(
...
-        }
+      let mv = MotionVector {
+        row: 8 * (y as i16 - po.y as i16),
+        col: 8 * (x as i16 - po.x as i16)
+      };
+
+      let rate1 = get_mv_rate(mv, pmv[0], allow_high_precision_mv);
+      let rate2 = get_mv_rate(mv, pmv[1], allow_high_precision_mv);
+      let rate = rate1.min(rate2 + 1);
+      let cost = 256 * sad + rate * lambda;
```
tdaede: But 256 here? Am I missing a shift?

src/me.rs
fbossen: it's Q8, but a factor 0.5 is thrown in here. Maybe I could rewrite as 256 * 0.5 to clarify

src/rdo.rs
```diff
@@ -404,6 +410,15 @@ pub fn rdo_mode_decision(
...
+      if mv_stack.len() > 0 { pmv[0] = mv_stack[0].this_mv; }
+      if mv_stack.len() > 1 { pmv[1] = mv_stack[1].this_mv; }
+      let cmv = pmvs[ref_slot_set[i] as usize].unwrap();
+      mvs_from_me.push([
+        motion_estimation(fi, fs, bsize, bo, ref_frames[0], cmv, seq.bit_depth, &pmv),
```
ycho: So, the roles are exchanged now.
The new parameter 'cmv' is for 'spatial prediction', i.e. ME seed, and pmv is from neighboring blocks searched by find_mvrefs() as defined by av1 spec, correct?

src/me.rs
tdaede: Oh, is the factor due to the x/y min step?

fbossen: 0.5 is just a fudge factor. No deep theoretical connection here

src/me.rs
```diff
@@ -280,6 +299,19 @@ fn adjust_bo(bo: &BlockOffset, fi: &FrameInvariants, blk_w: usize, blk_h: usize)
   }
 }
 
+fn get_mv_rate(a: MotionVector, b: MotionVector, allow_high_precision_mv: bool) -> u32 {
+  fn diff_to_rate(diff: i16, allow_high_precision_mv: bool) -> u32 {
+    let d = if allow_high_precision_mv { diff } else { diff >> 1 };
+    if d == 0 {
+      0
+    } else {
+      2 * (16 - d.abs().leading_zeros())
```
ycho: No need to check whether the value >= 0?

src/me.rs
tdaede: Oh, okay. Yeah, then I'd like it separated out / commented as that, just so it can be messed with later.

src/rdo.rs
ycho: Oh, wrong again. The dimension of pmvs was [5][8]

src/rdo.rs
fbossen: cmv is now the search center/seed. pmvs are the vectors used to determine the bit cost

PR #757, opened 2018-11-19
get_log_tx_scale: remove mutable variable

tmatth:
No functional change.

### DIFF
--- a/src/quantize.rs
+++ b/src/quantize.rs
@@ -89,6 +89,7 @@ fn divu_pair(x: i32, d: (u32, u32, u32)) -> i32 {
 #[cfg(test)]
 mod test {
   use super::*;
+  use partition::TxSize::*;
   #[test]
   fn test_divu_pair() {
     for d in 1..1024 {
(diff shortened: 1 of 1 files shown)


tmatth:
@ycho for your consideration. 

There's probably a way of getting rid of the conditionals altogether with some clever log2 + bittwiddling.

src/quantize.rs
```diff
@@ -15,12 +15,8 @@ use std::mem;
 
 pub fn get_log_tx_scale(tx_size: TxSize) -> i32 {
   let num_pixels = tx_size.area();
-  let mut shift_bits = 0;
 
-  if num_pixels > 256 { shift_bits += 1; };
-  if num_pixels > 1024 { shift_bits += 1; };
-
-  shift_bits
+  (num_pixels > 256) as i32 + (num_pixels > 1024) as i32
```
ycho: I think I tried exactly the same code in the beginning, borrowing from libaom, then rustc cannot pass it, since it does not do automatic type conversion (i.e. casting) from boolean to integer. :) Hence the CI raised red flag!

ycho: Oh you did casting, "as i32", sorry missed it-- *)*

ycho:
Do we know why there is red x over above with travis CI...?

src/quantize.rs
ycho: When casted, does rustc generate similar code as C? I,e, ture/false --> 1/0?

lu-zero: You can use `as` or `i32::from()` like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=e5986e05f5837950a0c167f6ba5f0ade)

lu-zero: and yes, boolean -> signed maps to 0/1 and not 0/-1 ^^

ycho: Ok- thanks for confirm!

ycho: @lu-zero , And thanks for example code for type conversion above!

lu-zero:
network problems on their side, I restarted it

src/quantize.rs
lu-zero: When the type may change `i32::from()` is suggested by clippy since would make refactoring easier. I consider that debatable.

tmatth: @lu-zero I think using `i32::from()` is reasonable.

PR #760, opened 2018-11-20
Don't use tx-domain distortion with top-down partition

ycho:


ycho:
This patch reduces the loss of coding efficiency with most of speedup preserved.
(Previously the loss has grown 0.5%, and now 0.2% by this patch)
Comparison of master w/ tx-domain OFF. vs. current patch.

[AWCY](https://beta.arewecompressedyet.com/?job=test_tx_dist_off%402018-11-19T23%3A17%3A59.582Z&job=test3_tx_dist_enforce_rec%402018-11-20T03%3A45%3A51.704Z)

0.2550 |  0.3880 |  0.4347 |   0.2861 | 0.2923 |  0.2991 |     0.5982

But I think we still have the possible issue #761 with top-down rdo partition, and this patch only fixes to have reconstructed pixels back during rdo_mode_decision(). 
Then I think the rec pixels from last mode tested in rdo_mode_decision() is used as reference pixels of neighbor for split of rdo_partition_decision() called by top-down rdo partition.

PR #763, opened 2018-11-21
Adaptive keyframe selection

### LINKED ISSUE #717: Adaptive keyframe placement
rav1e appears to currently use a constant interval between keyframes. Many other encoders, including aomenc, set a configurable "min" and "max" distance between keyframes, and use scenecut detection to determine the optimal location for a keyframe. This can provide a considerable boost to compression.

shssoichiro:
Detects the level of difference between the current frame and the
previous frame, and if the difference is above a preset threshold,
encodes the current frame as a keyframe.

Introduces a new CLI option, --max-keyint (-i), to set the maximum
interval between two keyframes. Reuses the existing --keyint (-I) option
as the minimum interval between two keyframes. Adjusts the default
settings to be 15 frames minimum and 150 frames maximum. In theory, 300
frames may even be a safe default, but a conservative default was
selected given that decoder performance is still insufficient for higher
definition videos, and particularly long keyframe intervals negatively
impact seek time.

Compression improvements were up to 25% on test sources.

Before:
```
Frame 1499 - Inter frame - 756 bytes - encoded 1500 frames, 1.15 fps, 656.97 Kb/s

Key Frames:     50    avg size:   27489 B
Inter:        1450    avg size:    2676 B
```

After:
```
Frame 1499 - Inter frame - 725 bytes - encoded 1500 frames, 1.15 fps, 538.58 Kb/s

Key Frames:     13    avg size:   27876 B
Inter:        1487    avg size:    2653 B
```

Closes #717

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -10,9 +10,11 @@
 use encoder::*;
 use partition::*;
 
+use std::cmp;
 use std::collections::BTreeMap;
 use std::fmt;
 use std::sync::Arc;
+use scenechange::SceneChangeDetector;
 
 // TODO: use the num crate?
 #[derive(Clone, Copy, Debug)]
@@ -30,7 +32,10 @@ impl Rational {
 
 #[derive(Copy, Clone, Debug)]
 pub struct EncoderConfig {
-  pub key_frame_interval: u64,
+  /// The *minimum* interval between two keyframes
+  pub min_key_frame_interval: u64,
+  /// The *maximum* interval between two keyframes
+  pub max_key_frame_interval: u64,
   pub low_latency: bool,
   pub quantizer: usize,
   pub tune: Tune,
@@ -47,7 +52,8 @@ impl Default for EncoderConfig {
 impl EncoderConfig {
   pub fn with_speed_preset(speed: usize) -> Self {
     EncoderConfig {
-      key_frame_interval: 240,
+      min_key_frame_interval: 12,
+      max_key_frame_interval: 240,
       low_latency: true,
       quantizer: 100,
       tune: Tune::Psnr,
@@ -166,7 +172,8 @@ impl Config {
     use self::EncoderStatus::*;
     match key {
       "low_latency" => self.enc.low_latency = value.parse().map_err(|_e| ParseError)?,
-      "key_frame_interval" => self.enc.key_frame_interval = value.parse().map_err(|_e| ParseError)?,
+      "min_key_frame_interval" => self.enc.min_key_frame_interval = value.parse().map_err(|_e| ParseError)?,
+      "key_frame_interval" => self.enc.max_key_frame_interval = value.parse().map_err(|_e| ParseError)?,
       "quantizer" => self.enc.quantizer = value.parse().map_err(|_e| ParseError)?,
       "speed" => self.enc.speed_settings = SpeedSettings::from_preset(value.parse().map_err(|_e| ParseError)?),
       "tune" => self.enc.tune = value.parse().map_err(|_e| ParseError)?,
@@ -190,7 +197,19 @@ impl Config {
       aom_dsp_rtcd();
     }
 
-    Context { fi, seq, frame_count: 0, frames_to_be_coded: 0, idx: 0, frame_q: BTreeMap::new(), packet_data: Vec::new() }
+    Context {
+      fi,
+      seq,
+      frame_count: 0,
+      frames_to_be_coded: 0,
+      idx: 0,
+      frame_q: BTreeMap::new(),
+      packet_data: Vec::new(),
+      segment_start_idx: 0,
+      segment_start_frame: 0,
+      frame_types: BTreeMap::new(),
+      keyframe_detector: SceneChangeDetector::new(&self.frame_info),
+    }
   }
 }
 
@@ -202,7 +221,11 @@ pub struct Context {
   frames_to_be_coded: u64,
   idx: u64,
   frame_q: BTreeMap<u64, Option<Arc<Frame>>>, //    packet_q: VecDeque<Packet>
-  packet_data: Vec<u8>
+  packet_data: Vec<u8>,
+  segment_start_idx: u64,
+  segment_start_frame: u64,
+  frame_types: BTreeMap<u64, FrameType>,
+  keyframe_detector: SceneChangeDetector,
 }
 
 #[derive(Clone, Copy, Debug)]
@@ -236,6 +259,22 @@ impl fmt::Display for Packet {
   }
 }
 
+#[derive(Debug, Clone, Copy)]
+enum FramePropsStatus {
+  FramePastSegmentEnd,
+}
+
+#[derive(Debug, Clone, Copy)]
+struct InterPropsConfig {
+  reorder: bool,
+  multiref: bool,
+  pyramid_depth: u64,
+  group_src_len: u64,
+  group_len: u64,
+  idx_in_group: u64,
+  group_idx: u64,
+}
+
 impl Context {
   pub fn new_frame(&self) -> Arc<Frame> {
     Arc::new(Frame::new(self.fi.padded_w, self.fi.padded_h))
@@ -245,7 +284,9 @@ impl Context {
   where
     F: Into<Option<Arc<Frame>>>
   {
-    self.frame_q.insert(self.frame_count, frame.into());
+    let idx = self.frame_count;
+    self.frame_q.insert(idx, frame.into());
+    self.save_frame_type(idx);
     self.frame_count = self.frame_count + 1;
     Ok(())
   }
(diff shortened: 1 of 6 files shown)


tdaede:
You'll need to fix the tests They probably just need to enforce maximum keyframe distance.

shssoichiro:
Looks like tests were failing in debug mode due to subtraction overflow, which didn't affect behavior in release mode. Should be fixed now.

src/api.rs
```diff
@@ -167,6 +173,7 @@ impl Config {
     match key {
       "low_latency" => self.enc.low_latency = value.parse().map_err(|_e| ParseError)?,
       "key_frame_interval" => self.enc.key_frame_interval = value.parse().map_err(|_e| ParseError)?,
+      "max_key_frame_interval" => self.enc.max_key_frame_interval = value.parse().map_err(|_e| ParseError)?,
```
tdaede: This API is opposite of x264, which is kind of confusing. I'd rather "key frame interval" be the max, and having a separate min option. In particular, most users shouldn't need to adjust the minimum interval - in fact, if our detector is smart enough we should be able to set it to 1. (it probably isn't right now)

src/api.rs
```diff
@@ -292,164 +333,229 @@ impl Context {
     sequence_header_inner(&self.seq).unwrap()
   }
 
-  pub fn frame_properties(&mut self, idx: u64) -> bool {
-    let key_frame_interval: u64 = self.fi.config.key_frame_interval;
+  fn setup_key_frame_properties(&mut self) {
```
tdaede: This might even be nicer as a fi initializer (e.g. new_key_frame or new_inter_frame). But you don't have to change it for this patch.

src/scenechange/mod.rs
```diff
@@ -0,0 +1,76 @@
+use encoder::Frame;
+use api::FrameInfo;
+
+/// Detects fast cuts using changes in colour and intensity between frames.
+/// Since the difference between frames is used, only fast cuts are detected
+/// with this method. This is probably fine for the purpose of choosing keyframes.
+pub struct SceneChangeDetector {
+  /// Minimum difference between YUV deltas that will trigger a scene change.
```
tdaede: *average of YUV deltas

src/scenechange/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+  threshold: u8,
+  /// Frame number and YUV values of the last frame analyzed
+  last_frame: Option<(usize, Vec<(u8, u8, u8)>)>,
```
tdaede: Why does scene change detector need to store the last frame? We should already have it as a reference.

src/scenechange/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+  input_bit_depth: u8,
+}
+
+impl Default for SceneChangeDetector {
+  fn default() -> Self {
+    Self {
+      threshold: 12,
```
tdaede: How did you determine the threshold?

ycho:
Can you also post AWCY result please? If you cannot, we can do it as well.

src/api.rs
shssoichiro: Yeah, I chose to have a minimum in order to account for scenes that might have rapid motion and would falsely detect multiple consecutive keyframes, even though it would be less efficient. Xvid avoids the need for this by adjusting the threshold to be less likely to pick a keyframe if it found one recently, and more likely if it's close to the max distance. (Despite its age, Xvid's scenecut detection is pretty good.) Ours could do something similar in the future.

For now I can flip the settings to be more intuitive.

src/scenechange/mod.rs
shssoichiro: This is the YUV444 converted values for the previous frame, so it avoids having to perform that conversion multiple times. It could be argued that the conversion from YUV420 to YUV444 is sufficiently fast that this isn't really necessary, though. I chose to favor reduced computation over reduced memory usage in this case, since rav1e as an encoder has a pretty low memory profile.

src/scenechange/mod.rs
shssoichiro: Experimentation and comparison with the original implementation I was copying from at https://pyscenedetect.readthedocs.io/en/latest/reference/detection-methods/. The Python implementation uses HSV values and a threshold of 30. I found that comparing the YUV values was sufficient in the cases I tested, and avoided a more costly YUV->RGB->HSV conversion, but that the deltas needed to be scaled down. The deltas for keyframes in YUV were about 1/3 to 1/2 of what they were in HSV, but non-keyframes were very unlikely to have a delta greater than 3 in YUV, whereas they may reach into the double digits in HSV.

shssoichiro:
Unsure if I can run against AWCY. Seems I need an API key?

src/scenechange/mod.rs
tdaede: Why not just do the subtraction in YUV420 space?

rzumer:
For reference I checked that x265 does to select keyframes and it seems they encode low resolution frames with the 2-pass method suggested by @tdaede in #717 and insert keyframes based on the intra/inter costs and some combination of parameters to make it more or less aggressive. Computing plane-averaged SAD is better than nothing (i.e. what we have now), but far from ideal.

gnafuthegreat:
One thing to note: I know the current keyint was chosen to be similar to x264's command line options.  If we wanted to continue that, if I'm understanding this correctly, keyint would need to become min-keyint and the new max-keyint would actually be the new keyint (if that makes sense).

src/scenechange/mod.rs
sjohannes: > as long as you multiply the delta from the U and V planes by 4

On a related note, is it worth considering different weights for the chroma vs luma components?

shssoichiro:
@gnafuthegreat It should be that way now after PR changes.

shssoichiro:
Everything should be cleaned up now. I agree we should take another pass at this doing something more intelligent later, but this approach is good enough for now IMO. Please take another look, and if someone could run against AWCY (or teach me how to do it) that would be great.

ycho:
[AWCY results for default speed](https://beta.arewecompressedyet.com/?job=master%402018-11-23T19%3A28%3A56.273Z&job=scenechange0_27573330%402018-11-25T13%3A52%3A05.575Z)

| -2.9141 | -2.3300 | -2.3638 |  -2.9066 | -3.0821 | -2.7983 |    -2.3367 |

We will see results for other speed, -s 1 and -s 0, as well.
@shssoichiro  Please verify whether this results matches to what you have.

shssoichiro:
That looks about right, thanks

src/encoder.rs
```diff
@@ -79,27 +79,69 @@ impl Frame {
...
+impl<'a> Iterator for PixelIter<'a> {
+  type Item = (u16, u16, u16);
+
+  fn next(&mut self) -> Option<<Self as Iterator>::Item> {
+    if self.y == self.height() - 1 && self.x == self.width() - 1 {
+      return None;
+    }
+    let pixel = (
+      self.planes[0].p(self.x, self.y),
+      self.planes[0].p(self.x / 2, self.y / 2),
```
lu-zero: Looks strange, the previous code was using all the planes.

ycho:
Thank you very much for great experiment and new feature.
I haven't had a chance to read the code yet (sorry!), and I wonder how much you think the chroma info contribute to the accuracy (or coding efficiency gain).

lu-zero:
The result shows we get improvements by using luma-only, I expect we would really few have scenechanges when luma remains the same and only chroma changes...

src/encoder.rs
shssoichiro: Whoops, copy and paste error.

shssoichiro:
I agree, luma is probably where most of the changes will be detected, but I think there's no harm in keeping the chroma as part of the equation, given that the impact on encoding speed seems to be almost zero. It may catch a couple of scenechanges that wouldn't be detected otherwise, or might eliminate a couple of false positives (fast fades, for example).

shssoichiro:
I'm wondering if we should hold off on merging and revisit the detection mechanism. [This resultset](https://beta.arewecompressedyet.com/?job=master_gop_size_240%402018-11-26T19%3A54%3A21.900Z&job=scenechange0_keyint_240_27573330%402018-11-27T00%3A32%3A28.994Z) seems to indicate, on average, minor regressions from this patch. There are probably too many false positives.

shssoichiro:
Also, would it be possible to get an API key for AWCY so I can run the test suite without needing to ask someone? Unsure what the process is for that.

tdaede:
OK, sent. That said, we don't have any particularly good test sets for this on AWCY (yet). But yeah, that test set does seem to indicate less sensitivity would be better - the vc-360p-1 set has basically no scenecuts.

tdaede:
To elaborate a bit, the problem you're most likely hitting in vc-360p-1 is lighting changes. The "real" fix to this is to use SATD instead of SAD (or a first pass encode), but that's out of scope for this patch.

tdaede:
I'll land it now, but could you maybe try a run like @ycho did on vc-360p-1, and pick a number that doesn't regress it? Then we can follow up with a PR to set that number. I sent you AWCY info. (If you don't have time, we can land it now and play with the number later)

shssoichiro:
Got them queued up. I'll see what the results end up looking like.

shssoichiro:
[150](https://beta.arewecompressedyet.com/?job=master_240_gop%402018-11-27T06%3A16%3A38.981Z&job=master_150_gop%402018-11-27T14%3A16%3A50.456Z) seems to provide the best results for the vc-360p-1 test set. I'm unsure how well this generalizes, however. I think changing the max interval to 150 is fine for short-term, but I'd like to revisit after we can improve the detection algorithm.

Is there a way we can get a test set with more variance? Is the limitation that most of the test sets are HD and therefore too slow at the moment?

Issue #606, opened 2018-09-23
use 64 bits in btf that is used by transforms


KyleSiefring:
a*x + b*y can get into the signed range.

https://bugs.chromium.org/p/aomedia/issues/detail?id=2151

32-bit will be fine for simd.

"In terms of hardware and vectorized implementations, this shouldn't cause any issues - you'll still get the correct result as long as you calculate everything using wrapping 32-bit arithmetic. But the assertion should be modified or removed, and an appropriate comment added."

lu-zero:
We can just use the explicit [overflow behaviour](https://doc.rust-lang.org/std/num/struct.Wrapping.html) as well. Willing to go his way?

- [ ] write the benchmarks
- [ ] see how using wrapping overflow compares
- [ ] see how using 64bit compares

(I can help if you feel that it is too confusing)

PR #768, opened 2018-11-25
Add explicit wrapping behaviour to txfm butterfly

KyleSiefring:
Addresses #606

### DIFF
--- a/src/transform/mod.rs
+++ b/src/transform/mod.rs
@@ -136,8 +136,16 @@ fn get_rect_tx_log_ratio(col: usize, row: usize) -> i8 {
 // performs half a butterfly
 #[inline]
 fn half_btf(w0: i32, in0: i32, w1: i32, in1: i32, bit: usize) -> i32 {
-  let result = (w0 * in0) + (w1 * in1);
-  round_shift(result, bit)
+  use std::num::Wrapping;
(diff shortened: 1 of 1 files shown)


src/transform/mod.rs
```diff
@@ -136,8 +136,16 @@ fn get_rect_tx_log_ratio(col: usize, row: usize) -> i8 {
...
 #[inline]
 fn half_btf(w0: i32, in0: i32, w1: i32, in1: i32, bit: usize) -> i32 {
-  let result = (w0 * in0) + (w1 * in1);
-  round_shift(result, bit)
+  use std::num::Wrapping;
+  // Ensure defined behaviour for when w0*in0 + w1*in1 is negative and
+  //   overflows, but w0*in0 + w1*in1 + rounding isn't.
+  let result = Wrapping(w0 * in0) + Wrapping(w1 * in1);
+  // Implement a version of round_shift with wrapping
+  if bit <= 0 {
```
lu-zero: You could use the explicit add and use the rotate operator or am I missing something?

lu-zero: I misread the code regarding round_shift, you can still use [wrapping_add](https://doc.rust-lang.org/nightly/std/primitive.i32.html#method.wrapping_add) though.

PR #770, opened 2018-11-27
Write RU-level restoration filter parameters

xiphmont:


xiphmont:
On Thu, Nov 29, 2018 at 1:06 PM Yushin Cho <notifications@github.com> wrote:

> *@ycho* commented on this pull request.
>
> Wondering what 'RU' means? Restoration Unit?
>
Yes.  BTW, it turns out I need to mark this one WIP.  I thought I could
start out by coding a restoration setup that enforced 1SB == 1RU, but it
turns out this just isn't possible in any configuration.  I have to be able
to handle fully asynchronous/on-demand RU parameter writes in the stream,
even if I'm not performing asynchronous RDO.  I have a design for that
that's not too complicated after all, but I have to scrap a lot of the
existing 770.

src/encoder.rs
```diff
@@ -1475,34 +1475,30 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
...
-          if seq.use_128x128_superblock {
-            self.write(1,0)?; // do not double the restoration unit from 128x128
-          } else {
-            self.write(1,0)?; // do not double the restoration unit from 64x64
+          // The Y shift value written here indicates shift up from superblock size
+          if !seq.use_128x128_superblock {
+            self.write(1, if rs.unit_size[0] > 64 {1} else {0})?;
+          }
+          if rs.unit_size[0] > 64 {
+            self.write(1, if rs.unit_size[0] > 128 {1} else {0})?;
```
lu-zero: Is it possible to have both conditions above true at the same time?

xiphmont:
On Mon, Dec 10, 2018 at 7:23 AM Luca Barbato <notifications@github.com>
wrote:

> *@lu-zero* commented on this pull request.
> ------------------------------
>
> In src/encoder.rs
> <https://github.com/xiph/rav1e/pull/770#discussion_r240189500>:
>
> >              use_lrf = true;
>              if i > 0 { use_chroma_lrf = true; }
>            }
>          }
>          if use_lrf {
> -          // At present, we're locked to a restoration unit size equal to superblock size.
> -          // Signal as such.
> -          if seq.use_128x128_superblock {
> -            self.write(1,0)?; // do not double the restoration unit from 128x128
> -          } else {
> -            self.write(1,0)?; // do not double the restoration unit from 64x64
> +          // The Y shift value written here indicates shift up from superblock size
> +          if !seq.use_128x128_superblock {
> +            self.write(1, if rs.unit_size[0] > 64 {1} else {0})?;
> +          }
> +          if rs.unit_size[0] > 64 {
> +            self.write(1, if rs.unit_size[0] > 128 {1} else {0})?;
>
> Is it possible to have both conditions above true at the same time?
>
Yes.

Monty

src/encoder.rs
xiphmont: Er, I should clarify-- yes, it's possible.  Previous code was simply hardwiring the case we were using, I decided I to replace it with the general case.

xiphmont:
Not needed after current frame is written out; no reason to keep the storage around in references.

tdaede:
Ah okay. Wonder if FrameState already has other data like this lying around in it...

PR #775, opened 2018-11-30
Fix missing inv sqrt2 factor for rect inv tx

ycho:
The forward rect tx has sqrt 2 term but inv rect tx has been missing 1/sqrt2 term.

### DIFF
--- a/src/transform/mod.rs
+++ b/src/transform/mod.rs
@@ -23,7 +23,8 @@ mod forward;
 mod inverse;
 
 static SQRT2_BITS: usize = 12;
-static SQRT2: i32 = 5793;
+static SQRT2: i32 = 5793;       // 2^12 * sqrt(2)
+static INV_SQRT2: i32 = 2896;   // 2^12 / sqrt(2)
 
 static COS_BIT_MIN: usize = 10;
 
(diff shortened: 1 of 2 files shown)


src/transform/mod.rs
```diff
@@ -111,23 +112,23 @@ fn sinpi_arr(n: usize) -> &'static [i32; 5] {
 
 /// Utility function that returns the log of the ratio of the col and row sizes.
 #[inline]
-fn get_rect_tx_log_ratio(col: usize, row: usize) -> i8 {
+pub fn get_rect_tx_log_ratio(col: usize, row: usize) -> i8 {
   if col == row {
-    return 0;
+    return 0
```
fbossen: why remove the semicolon?

ycho: I think I learned that rust syntax don't need it?
If you don't like it, let me restore them! :)

rzumer: I suggest keeping semicolons for explicit return statements, however I'm not sure that you need the `return` keyword at all in Rust, and we already use this shortcut in a few other places, so this might be better for consistency.

shssoichiro: For early returns, I believe you need the `return` keyword. If the return is part of the last expression in a function, you can omit the `return` and semicolon.

I'm a little surprised that this compiles successfully without the semicolon; from what I've seen, it seems to be the common practice to include the semicolon if you're using the `return` keyword.

ycho:
We have gain for -s4 (i.e. 64x32 and 32x64 introduced to 32x32 or 64x64), but unfortunately it does not work yet for other speeds. For -s 3, only first few frames seem decoded correctly/

| -1.4423 |  0.0906 |  0.2046 |  -1.6500 | -2.3018 | -2.2945 |    -1.9633 |

[AWCY link for -s 4](https://beta.arewecompressedyet.com/?job=master_2e4055a_s4%402018-11-30T01%3A01%3A02.828Z&job=test_rect_s4%402018-11-30T01%3A01%3A38.247Z)

src/transform/mod.rs
ycho: @lu_zero: Yes, I also think it needs 'return' to  exit the function there. Regarding whether ; is recommended and/or required, I'd ask lu_zero. In my understanding of how rustc parses, it can be omitted.

ycho: Let me land this for now. I think that the reason why it can compile w/o ; is 1) there is only one statement, and the ending } is parsed as end of a complete statement, so not needing ;.
(Believe it or not).
If you ever feel not comfortable, please revise as you like as a separate patch. :)
I recall ',' is sometimes added at the end of last element of array, sometimes not.

lu-zero: It works but it is a fun by-product, once I have a working keyboard I'll explain :)

The expression `return foo` has return type `!` (never).

PR #776, opened 2018-11-30
Refactor frame props setup

shssoichiro:
Moves the key frame props and inter frame props setup to be static
initializers on FrameInvariants, which is more cohesive. This also
prepares for the potential addition of a lookahead capability.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -264,22 +264,6 @@ impl fmt::Display for Packet {
   }
 }
 
-#[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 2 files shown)


lu-zero:
Seems ok, but maybe it is time to move this code on a different file, `encoder.rs` is quite busy already.

shssoichiro:
I can make a followup PR to split up the encoder.rs into separate modules. I agree that's a good change, but I'd rather do it separate from this.

PR #778, opened 2018-12-04
Refactor encoder module into smaller files

shssoichiro:


lu-zero:
It looks good, but doing it this way would impact a lot the pending works ongoing, @fbossen and @atomnuker have some large features impacted by your change (and to a smaller degree also @ycho).

If they are fine with rebasing on top of this it could land.

src/encoder/transform.rs
```diff
@@ -0,0 +1,362 @@
+use context::av1_get_coded_tx_size;
```
ycho: I don't think functions here, like encode_tx_block() can fall under transform.rs. Instead, block.rs would fit.

src/encoder/partition.rs
```diff
@@ -0,0 +1,557 @@
...
+    && (bsize == BlockSize::BLOCK_8X8
+      || partition != PartitionType::PARTITION_SPLIT)
+  {
+    cw.bc.update_partition_context(bo, subsize, bsize);
+  }
+
+  rd_cost
+}
+
+pub fn encode_partition_topdown(
```
ycho: These two partition rdo functions are core of rdo based mode decision. So, I think that it is better in the previous file name rdo.rs.

ycho:
Thanks for contribution, however I'd like ask defer doing major refactoring works esp. at file level, i.e. name changes, creation, and deletion, after several crucial features are finished.

shssoichiro:
Thanks for the feedback. This was also a concern as I was doing this refactor, so I agree with postponing and PR'ing file by file. If there are any files I can try to split out without interfering with open PRs, I may attempt to, but it might be more reasonable to hold off until the open features are finished.

ycho:
Hi Josh, thanks for understanding generously!

PR #781, opened 2018-12-04
Set the chroma sample position based on Y4M color space

rzumer:
No desync based on a run of the build script (using the default sequence, i.e. `CSP_VERTICAL`).


rzumer:
@lu-zero Now compiles.

PR #786, opened 2018-12-05
Unignore decode tests

rzumer:
The feature guard should be enough, and run time is not so long that we should ignore them by default while testing.

Related: https://github.com/xiph/rav1e/commit/b7a0378a748843e9ff9f8360335a7a4774d4ac98

### DIFF
--- a/src/test_encode_decode.rs
+++ b/src/test_encode_decode.rs
@@ -100,7 +100,6 @@ static DIMENSION_OFFSETS: &[(usize, usize)] =
   &[(0, 0), (4, 4), (8, 8), (16, 16)];
 
 #[test]
-#[ignore]
 fn speed() {
   let quantizer = 100;
   let limit = 5;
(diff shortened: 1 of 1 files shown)


rzumer:
I see that they were ignored by default because Travis is unable to run them. If anyone has a solution for running the decoding tests by default on the user side at least, please advise.

tdaede:
Oh, I see. When I reviewed the README I assumed we had no ignored tests left, I guess that wasn't the case. Yeah I guess we will have to revert the other one.

rzumer:
I assumed the same thing before it was merged, sorry about that.

shssoichiro:
If we can't fix the tests in travis, the README should probably state that you need to run both with and without `--ignored`. If you add `--ignored` it _only_ runs the ignored tests.

lu-zero:
I'll try to split the tests in smaller chunks so the 10min timeout won't happen, hopefully.

lu-zero:
I split some tests, but probably I should tweak the kcov invocation if we really want to unignore all of them.

rzumer:
Going to close this at least for now.

PR #791, opened 2018-12-06
Revise motion_compensate() for 4x8 and 8x4 partition sizes

ycho:


ycho:
@fbossen Yes, I made it messier. But I believe that was the best I could do .... :)

src/encoder.rs
```diff
@@ -1748,30 +1748,50 @@ pub fn motion_compensate(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut Con
     let po = bo.plane_offset(&fs.input.planes[p].cfg);
     let rec = &mut fs.rec.planes[p];
     // TODO: make more generic to handle 2xN and Nx2 MC
```
fbossen: This TODO line can probably be removed

src/encoder.rs
```diff
@@ -1748,30 +1748,50 @@ pub fn motion_compensate(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut Con
...
-        let mv2 = cw.bc.at(&bo.with_offset(-1,0)).mv;
-        let rf2 = cw.bc.at(&bo.with_offset(-1,0)).ref_frames;
-        let po2 = PlaneOffset { x: po.x, y: po.y+2 };
-        let po3 = PlaneOffset { x: po.x+2, y: po.y+2 };
-        luma_mode.predict_inter(fi, p, &po, &mut rec.mut_slice(&po), 2, 2, rf0, mv0, bit_depth);
-        luma_mode.predict_inter(fi, p, &po1, &mut rec.mut_slice(&po1), 2, 2, rf1, mv1, bit_depth);
-        luma_mode.predict_inter(fi, p, &po2, &mut rec.mut_slice(&po2), 2, 2, rf2, mv2, bit_depth);
-        luma_mode.predict_inter(fi, p, &po3, &mut rec.mut_slice(&po3), 2, 2, ref_frames, mvs, bit_depth);
+        // TODO: these are absolutely only valid for 4:2:0
+        if bsize == BlockSize::BLOCK_4X4 {
```
fbossen: There is quite a bit of code duplication in the following. Maybe this can be written as generic loops (over rows and columns) with a single call to predict_inter in the innermost loop?

src/encoder.rs
```diff
@@ -1748,30 +1748,50 @@ pub fn motion_compensate(fi: &FrameInvariants, fs: &mut FrameState, cw: &mut Con
...
+            let mv0 = cw.bc.at(&bo.with_offset(-1,-1)).mv;
+            let rf0 = cw.bc.at(&bo.with_offset(-1,-1)).ref_frames;
+            let mv1 = cw.bc.at(&bo.with_offset(0,-1)).mv;
```
ycho: I tried to factor out mv1,rf1,mv2,rf2 so that they can be shared by 4x4,4x8,8x4, but 
mv1 and rf1 is only available for 8x4, and vice versa for 4x8.
Also the size of MC block is all different for each sub8x8 block sizes, which made factoring out not straighford.

PR #801, opened 2018-12-06
Use variable-depth quantization coefficients

rzumer:
This adds a `Coefficient` trait to `quantize.rs` implemented for `i16` and `i32` (similar to `Pixel`), and changes the `quantize()` function coefficient arguments to allow arrays of that trait.


tdaede:
maaaaaaaan as_() is ugly :(

PR #803, opened 2018-12-07
Remove unnecessary checkpoints and variables

fbossen:


fbossen:
@ycho No change in AWCY test

Issue #72, opened 2018-02-12
Support chroma sampling other than 4:2:0


yushincho:
Would be happy if we can do this by 2020.

rzumer:
Started in #453.

rzumer:
More progress in #908, #911, #914. Remaining issue seems related to motion compensation, which causes blocking in chroma on inter frames in the restoration buffer. Inter frames are also not decodable, perhaps for the same reason.

rzumer:
Encoding is supported as of #947 with features disabled (inter frames, restoration, CfL prediction, <8X8 blocks).

PR #808, opened 2018-12-07
Fix undersized frame buffers with non-4:2:0 chroma

rzumer:
More work for #72.

This set allows to output the reconstruction for 4:4:4 chroma without crashing (4:2:2 is untested). The output `IVF` file for the default sequence is still undecodable. There are still likely many things to fix for higher sampling support, but this should help with debugging.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -44,9 +44,12 @@ const FRAME_MARGIN: usize = 16 + SUBPEL_FILTER_SIZE;
 impl Frame {
     pub fn new(width: usize, height: usize, chroma_sampling: ChromaSampling) -> Frame {
         let chroma_sampling_period = chroma_sampling.sampling_period();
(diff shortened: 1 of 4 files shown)


src/encoder.rs
```diff
@@ -42,7 +42,13 @@ pub struct Frame {
 const FRAME_MARGIN: usize = 16 + SUBPEL_FILTER_SIZE;
 
 impl Frame {
-    pub fn new(width: usize, height:usize) -> Frame {
+    pub fn new(width: usize, height: usize, chroma_sampling: ChromaSampling) -> Frame {
+        let (chroma_width, chroma_height) = match chroma_sampling {
```
shssoichiro: This match should probably be extracted into a method on `ChromaSampling`, since you're using the same logic in two places.

PR #809, opened 2018-12-08
Make sure full ME tests at least one motion vector

fbossen:


src/me.rs
```diff
@@ -264,8 +264,8 @@ fn full_search(
   lowest_cost: &mut u32, po: &PlaneOffset, step: usize, bit_depth: usize,
   lambda: u32, pmv: &[MotionVector; 2], allow_high_precision_mv: bool
 ) {
-  for y in (y_lo..y_hi).step_by(step) {
-    for x in (x_lo..x_hi).step_by(step) {
+  for y in (y_lo..y_hi+1).step_by(step) {
+    for x in (x_lo..x_hi+1).step_by(step) {
```
shssoichiro: It would be slightly cleaner to use `..=` (inclusive range operator)

fbossen: That's a good point

PR #811, opened 2018-12-08
Parallel tests

lu-zero:


lu-zero:
Ok, this way seems to make travis happy, I'd squash and merge.

src/test_encode_decode.rs
```diff
@@ -105,22 +105,47 @@ fn setup_encoder(
...
+            #[test]
+            fn [<speed_ $S>]() {
+                speed($S)
+            }
+        }
+    )*
+  }
+}
+
+test_speeds!{ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
```
barrbrain: Whitespace is a little odd here.
```suggestion
test_speeds!{ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }
```

tdaede:
squash please

src/test_encode_decode.rs
lu-zero: Folded in the rework.

lu-zero:
I'd push it today if nobody is against it.

tdaede:
Please do

PR #812, opened 2018-12-11
Increase horizontal search range of initial ME

fbossen:
Increase helps capture large MVs such as in the Minecraft sequence


lu-zero:
Seems fine, making the vertical search range twice isn't helping as well?

fbossen:
@lu-zero Increasing the vertical search doesn't seem to help, at least not on the initial frames of Minecraft where I was seeing the issue with the horizontal search range

PR #816, opened 2018-12-12
Enable rectangular partitions with bottomup rdo partition search

ycho:
2:1 rectangular partitinos only.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -2242,11 +2242,13 @@ pub fn encode_block_with_modes(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fr
 fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut FrameState,
                              cw: &mut ContextWriter, w_pre_cdef: &mut dyn Writer, w_post_cdef: &mut dyn Writer,
                              bsize: BlockSize, bo: &BlockOffset, pmvs: &[[Option<MotionVector>; REF_FRAMES]; 5]
-) -> f64 {
+) -> (f64, Option<RDOPartitionOutput>) {
     let mut rd_cost = std::f64::MAX;
+    let mut best_rd = std::f64::MAX;
+    let mut best_pred_modes: Vec<RDOPartitionOutput> = Vec::new();
 
     if bo.x >= cw.bc.cols || bo.y >= cw.bc.rows {
-        return rd_cost;
+        return (rd_cost, None);
     }
 
     let bsw = bsize.width_mi();
@@ -2261,10 +2263,12 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
     // must_split overrides the minimum partition size when applicable
     let can_split = (bsize > fi.min_partition_size && is_square) || must_split;
 
-    let mut partition = PartitionType::PARTITION_NONE;
+    let mut partition = PartitionType::PARTITION_INVALID;
+    let mut best_partition = partition;
     let mut best_decision = RDOPartitionOutput {
         rd_cost,
         bo: bo.clone(),
+        bsize: bsize,
         pred_mode_luma: PredictionMode::DC_PRED,
         pred_mode_chroma: PredictionMode::DC_PRED,
         pred_cfl_params: CFLParams::new(),
@@ -2281,12 +2285,14 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
     let mut subsize: BlockSize;
 
     // Code the whole block
+    // TODO(yushin): Try move PARTITION_NONE to below partition loop
     if !must_split {
         partition = PartitionType::PARTITION_NONE;
+        best_partition = partition;
 
         let mut cost: f64 = 0.0;
 
-        if bsize.gte(BlockSize::BLOCK_8X8) {
+        if bsize.gte(BlockSize::BLOCK_8X8) && is_square {
             let w: &mut dyn Writer = if cw.bc.cdef_coded {w_post_cdef} else {w_pre_cdef};
             let tell = w.tell_frac();
             cw.write_partition(w, bo, partition, bsize);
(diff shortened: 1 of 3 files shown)


src/encoder.rs
```diff
@@ -2281,12 +2285,14 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
     let mut subsize: BlockSize;
 
     // Code the whole block
+    // TODO(yushin): Try move PARTITION_NONE to below partition loop
```
ycho: As separate patch, I plan to try include PARTITION_NONE case in the partition loop below.

ycho:
[AWCY results for -s 0](https://beta.arewecompressedyet.com/?job=rect_part_bottomup_s0%402018-12-12T23%3A48%3A43.822Z&job=master_38bb357_s0%402018-12-12T23%3A48%3A10.228Z)

| -3.4508 | -4.7504 | -5.0264 |  -3.7131 | -3.6711 | -3.7920 |    -4.7571 |

src/encoder.rs
```diff
@@ -2261,10 +2263,12 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
     // must_split overrides the minimum partition size when applicable
     let can_split = (bsize > fi.min_partition_size && is_square) || must_split;
 
-    let mut partition = PartitionType::PARTITION_NONE;
+    let mut partition = PartitionType::PARTITION_INVALID;
+    let mut best_partition = partition;
```
rzumer: Here and below with `PARTITION_NONE`, I think you can eliminate the `let mut partition...` lines and assign the values to `best_partition` directly.

src/encoder.rs
```diff
@@ -2303,44 +2309,63 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
...
-        partition = PartitionType::PARTITION_SPLIT;
-        {
-        subsize = bsize.subsize(partition);
-        let hbsw = subsize.width_mi(); // Half the block size width in blocks
-        let hbsh = subsize.height_mi(); // Half the block size height in blocks
+        for &partition in RAV1E_PARTITION_TYPES {
+            if partition == PartitionType::PARTITION_NONE {
+                continue;
+            }
+            assert!(bsw == bsh);
```
rzumer: Please add some spacing around inner scopes in this area, for ease of reading.

src/encoder.rs
```diff
@@ -2303,44 +2309,63 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
...
+            if must_split {
+                let cbw = (fi.w_in_b - bo.x).min(bsw); // clipped block width, i.e. having effective pixels
+                let cbh = (fi.h_in_b - bo.y).min(bsh);
+                let mut split_vert = false;
+                let mut split_horz = false;
+                if cbw == bsw/2 && cbh == bsh { split_vert = true; }
```
rzumer: A comment might be good for this branch.

src/partition.rs
```diff
@@ -718,6 +718,7 @@ pub const GLOBALMV_CTX_MASK: usize =
   ((1 << (REFMV_OFFSET - GLOBALMV_OFFSET)) - 1);
 pub const REFMV_CTX_MASK: usize = ((1 << (8 - REFMV_OFFSET)) - 1);
 
+// Caution: PARTITION_SPLIT must be the last entry!
```
rzumer: A unit test would be good to assert this.

src/encoder.rs
```diff
@@ -2281,12 +2285,14 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
     let mut subsize: BlockSize;
 
     // Code the whole block
+    // TODO(yushin): Try move PARTITION_NONE to below partition loop
     if !must_split {
```
rzumer: I wonder if this would be better moved inside the partition type loop.

src/encoder.rs
```diff
@@ -2303,44 +2309,63 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
...
+        best_rd = rd_cost;
 
         encode_block_with_modes(seq, fi, fs, cw, w_pre_cdef, w_post_cdef, bsize, bo,
                             &mode_decision);
+
         best_decision = mode_decision;
+        best_pred_modes.push(best_decision.clone());
     }
 
     // Code a split partition and compare RD costs
```
rzumer: This comment needs an update ("code all partition splits and compare RD costs", for example).

src/encoder.rs
```diff
@@ -2303,44 +2309,63 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
...
-        let hbsh = subsize.height_mi(); // Half the block size height in blocks
+        for &partition in RAV1E_PARTITION_TYPES {
+            if partition == PartitionType::PARTITION_NONE {
+                continue;
+            }
+            assert!(bsw == bsh);
+            if must_split {
+                let cbw = (fi.w_in_b - bo.x).min(bsw); // clipped block width, i.e. having effective pixels
+                let cbh = (fi.h_in_b - bo.y).min(bsh);
+                let mut split_vert = false;
```
rzumer: My suggestion for this: set the value outside of the loop, near `can_split` and `must_split`, and call it `can_split_vert` (same with `split_horz`). It should simplify the content of the loop, which is quite complex now.

src/encoder.rs
```diff
@@ -2303,44 +2309,63 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
...
+                let mut split_horz = false;
+                if cbw == bsw/2 && cbh == bsh { split_vert = true; }
+                if cbh == bsh/2 && cbw == bsw { split_horz = true; }
+                if !split_horz {
+                    if partition == PartitionType::PARTITION_HORZ { continue; };
```
rzumer: Very minor, but I would combine the conditions into one if you are not expecting more code to come between the two conditions.

src/encoder.rs
```diff
@@ -2303,44 +2309,63 @@ fn encode_partition_bottomup(seq: &Sequence, fi: &FrameInvariants, fs: &mut Fram
...
+                &BlockOffset{ x: bo.x + hbsw as usize, y: bo.y },
+                &BlockOffset{ x: bo.x, y: bo.y + hbsh as usize },
+                &BlockOffset{ x: bo.x + hbsw as usize, y: bo.y + hbsh as usize }
+            ];
+            let partitions = get_sub_partitions(&four_partitions, partition);
 
-        rd_cost += partitions.iter().map(|&offset| {
-                encode_partition_bottomup(
+            partitions.iter().for_each(|&offset| {
+                if let (cost, Some(mode_decision)) = encode_partition_bottomup(
```
rzumer: Is this called only when `partition == PartitionType::PARTITION_SPLIT`? I am not seeing the condition, so at first glance it seems like we test this four-way split many times. If this is needed could you please add a comment explaining this part?

src/encoder.rs
ycho: sure-.

src/encoder.rs
ycho: As I commented in https://github.com/xiph/rav1e/pull/816/files/b8ea05b8aa72cd66d4711dad2ab07b0d0d449401#r241243259,
I plan to think about that, but the reason why I keep the block of code for PARTITION_NONE case is that only that case actually calls rdo_mode_decision() because it is the terminal (i.e. not splitable literally) in terms of partitioning. Also, pmv (predicted ME centers) are used for only that case, and preparing the partitions are NOT used for that case.

src/encoder.rs
ycho: I understand that, but would you still recommend moving those to outside the loop when those variables are only used if must_split is true?
The split_vert and split_horz are only used when must_split == true, when the input bsize straddles on the right or bottom frame borders, because there can be cases where horz and/or vert split is not possible so disallowed.

src/encoder.rs
ycho: Yeap- good point.

src/encoder.rs
rzumer: OK, it does make more sense like this.

src/encoder.rs
rzumer: Oops, the comment was even in the view. Sorry I missed it.

src/encoder.rs
ycho: No, it is called for other partition types, like horz and vert as well.
Previously before this patch, i,e. w/o horz and vert,  encode_partition_bottomup() is given the four quadrants of current block. However, with hor and vert partition types enabled, it can fed with, two rectangles of current block, for ex 16x8 and 8x16 for the block 16x16, for the case PARTITION_HORZ. But note that in that case,  encode_partition_bottomup( ) will set can_split false because it is not square and skip the loop for each partition types. Also, if the rectangle straddles on the frame border, it should not have passed to encode_partition_bottomup( ) by the above checking of split_horz or split_vert.

src/partition.rs
ycho: Right, I will do that-

src/encoder.rs
ycho: Done.

src/encoder.rs
rzumer: Right, it works as expected.

src/partition.rs
rzumer: This comment is probably no longer needed, since the unit test documents the need for `PARTITION_SPLIT` to be last.

ycho: Would others agree on rzumer's comment? I'd prefer keep it there since many of us not checking what's in the test.

PR #817, opened 2018-12-13
Add directional intra modes

fbossen:


src/partition.rs
```diff
@@ -108,6 +108,34 @@ impl BlockSize {
...
+      (64, 128) => BLOCK_64X128,
+      (128, 64) => BLOCK_128X64,
+      (128, 128) => BLOCK_128X128,
+      (4, 16) => BLOCK_4X16,
+      (16, 4) => BLOCK_16X4,
+      (8, 32) => BLOCK_8X32,
+      (32, 8) => BLOCK_32X8,
+      (16, 64) => BLOCK_16X64,
+      (64, 16) => BLOCK_64X16,
+      _ => BLOCK_INVALID
```
lu-zero: when it can happen? I'd mark it as `unreachable!()`

lu-zero:
(and travis finds the problems)

src/partition.rs
fbossen: Good point. Still learning rust :-)

fbossen:
At default speed (without significant increase in run time):

| -0.7923 | -1.1463 | -0.7454 |  -0.7838 | -0.8455 | -0.8588 |    -0.6728 |

fbossen:
At speed 1:

| -1.0674 | -1.3018 | -1.2524 |  -1.0905 | -1.2387 | -1.1538 |    -1.0946 |

src/rdo.rs
```diff
@@ -603,7 +593,64 @@ pub fn rdo_mode_decision(
   });
 
   if !best.skip {
-    intra_mode_set.iter().for_each(|&luma_mode| {
+    let tx_size = match bsize {
```
rzumer: Here I believe you can use `bsize.tx_size()` to avoid this large block.

src/rdo.rs
```diff
@@ -603,7 +593,64 @@ pub fn rdo_mode_decision(
...
+      let plane_org = fs.input.planes[0].slice(&po);
+      let plane_ref = rec.slice(&po);
+
+      (luma_mode, get_sad(&plane_org, &plane_ref, tx_size.height(), tx_size.width(), seq.bit_depth))
+    }).collect::<Vec<_>>();
+
+    sads.sort_by_key(|a| a.1);
+
+    // Find mode with lowest rate cost
+    let mut z = 32768;
```
rzumer: Is this meant to be a signed 16-bit value initialized to the maximum? In that case I would suggest making the type explicit and using `i16::MAX()` as the initialized value for clarity.

fbossen: This is an unsigned value. The value 32768 is the same as in ec.rs

src/rdo.rs
fbossen: Good point

PR #820, opened 2018-12-19
Refactor motion_compensate caller

tmatth:


tmatth:
I could go either way on whether or not that conditional should really be changed to an assertion. My rationale was that this is only used internally so it should be easy to make this guarantee.

tmatth:
> > I could go either way on whether or not that conditional should really be changed to an assertion. My rationale was that this is only used internally so it should be easy to make this guarantee.
> 
> probably it could live as a `debug_assert!` (and so many other)

Done.

Issue #631, opened 2018-10-02
Add tiles support


lu-zero:
- [ ] Implement the configuration settings
  - Context structure
  - Expose CLI options for it
- [ ] have the Frame provide a tile iterator [example](https://play.rust-lang.org/?gist=c14aa4c804dd8019b41bc5d21cb68a56&version=stable&mode=debug&edition=2015)
  - [ ] Have a FrameTileIterator
  - [ ] Have a FrameTile
  - [ ] Make the FrameTile produce PlaneSlice
- [ ] Use the tile elements in the codebase

rom1v:
Currently, `Plane(Mut)Slice<'_>` borrows the whole `Plane`, and just stores an offset `(x, y)`. It exposes the whole underlying data (e.g. via [`as_mut_slice()`](https://github.com/xiph/rav1e/blob/f0e759175f52a1dae4233ff528557fcf0e3c8319/src/plane.rs#L400-L405)), the user is responsible to use the [stride](https://github.com/xiph/rav1e/blob/f0e759175f52a1dae4233ff528557fcf0e3c8319/src/plane.rs#L19) information to access to different rows.

To support tiling, we must not be able to borrow memory outside the relevant region of the `Plane` (currently, if [`base == 0`](https://github.com/xiph/rav1e/blob/f0e759175f52a1dae4233ff528557fcf0e3c8319/src/plane.rs#L404) for example, the method returns a mutable reference to the whole data).

To avoid this problem, I suggest to create a `PlaneRegion<'_>`, which is basically a "rectangle view" of a `Plane`. Several non-overlapping `PlaneRegion<'_>` could be created via an iterator (like in the [example](https://play.rust-lang.org/?gist=c14aa4c804dd8019b41bc5d21cb68a56&version=stable&mode=debug&edition=2015)).

We could then make `Plane(Mut)Slice<'_>` (which is basically a cursor/view) use a `PlaneRegion<'_>` instead of a `Plane` (contrary to `PlaneRegion<'_>`, the cursor could be moved, expanded, etc. inside the region).

As a consequence, it will not be possible to provide a slice covering the whole surface (multiple rows) and let the user use the stride information anymore, since this would borrow data possibly belonging to other regions (tiles). Therefore, all code using `Plane(Mut)Slice<'_>` must be adapted to access the data (and also to take additional region offset into account). _Currently, I don't know the amount of work to adapt the existing code._

[...]

PR #821, opened 2018-12-19
[WIP] Tiling

rom1v:
https://github.com/xiph/rav1e/issues/631

For now, this provides an iterator on a Frame to retrieve non-overlapping tiles, which can be used to read or write the underlying Frame.

This paves the way to be able to encode tiles in parallel.


rom1v:
See #1126. I'm closing this one.

PR #822, opened 2018-12-20
Remove unnecessary clone()

rom1v:


rom1v:
However, this [other one](https://github.com/xiph/rav1e/blob/04d6c53f9b2e3c6aa86b99a36d450e470e28ac01/src/api.rs#L522) cannot be removed the same way.

PR #823, opened 2018-12-21
Fix some Clippy errors and warnings

vladimir-kazakov:
Changes that improve readability and make the code more Rusty (idiomatic).

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -379,7 +379,7 @@ impl Context {
     let idx = self.frame_count;
     self.frame_q.insert(idx, frame.into());
     self.save_frame_type(idx);
-    self.frame_count = self.frame_count + 1;
+    self.frame_count += 1;
     Ok(())
   }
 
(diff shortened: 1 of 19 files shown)


vladimir-kazakov:
A few questions:
1. Why some files cannot be changed? Changes in this PR are not that big or global to cause any trouble.
1. How do I know which files should be avoided right now?
1. What would be the correct condition to split this PR? Should it be done "by files that are safe to modify", or "by types of warnings and errors"? If by files, then how do I know which files are safe to touch?

lu-zero:
There are some areas that are actively being reworked, so I'd rather avoid collisions, I can poke the people I know that they are touching something and see if they are fine with it.

And that reminds me we also should discuss about moving to 2018.

vladimir-kazakov:
A smarter approach was taken to achieve the same goal. There's no need in this PR anymore. Closing it.

PR #829, opened 2018-12-29
transform: simplify log ratio helper

tmatth:


KyleSiefring:
-1 Use ilog or msb from util.rs.

tmatth:
@KyleSiefring yeah I realized that after pushing ;)

tmatth:
Assuming I'm using `cargo bench` correctly, `msb` is faster.

PR #831, opened 2018-12-31
Move retrieval of edges for intra prediction into a separate function

fbossen:
When testing multiples intra prediction modes for a same block, edges
are now retrieved only once instead of repeatedly for each mode


lu-zero:
rebased, now let see if it passes the tests.

lu-zero:
Rebased and merged.

Issue #832, opened 2019-01-01
Get mc.asm to compile


KyleSiefring:
mc.asm is not hooked and doesn't even compile up so inter prediction is slow (like 30% of the time spent). When added to the list of compiled asm files, I get an error complaining about undefined references to rav1e_mc_subpel_filters.

In dav1d, these tables are defined as such.

```
const int8_t ALIGN(dav1d_mc_subpel_filters[5][15][8], 8) = {
    [DAV1D_FILTER_8TAP_REGULAR] = {
        {   0,   1,  -3,  63,   4,  -1,   0,   0 },
...
```

This is basically the same thing as SUBPEL_FILTERS in partition.rs.
I cannot find any way to align an static array in rust. I also have no idea if a static rust array will have defined behavior when imported into assembly. It doesn't seem like it should.

I can think of three ways to work around this. 1st and 2nd are by defining the arrays a second time in C or asm. The 3rd is by defining the array only in C and removing the rust version.

I don't want to put this in mc.asm so the we can just copy over new versions. Maybe we use a separate file.

KyleSiefring:
I'm thinking of adding it to assembly. Should this be a permanent solution? Probably not.

tdaede:
Extern "C" should make it work in assembly. The assembly hardcodes C ABI in for various platforms, so as long as the rust uses C ABI, it's fine. That said, it seems that all the static alignment stuff only works on structs, not on arrays or other bare types :( maybe @lu-zero has a fix?

lu-zero:
For static tables maybe [this](https://play.rust-lang.org/?version=nightly&mode=release&edition=2018&gist=42b6a1c52274b0b3d331f946468c38dc) is enough

PR #834, opened 2019-01-02
Align the tables we want to use from asm

lu-zero:
- [x] export subpel_filters
- [ ] refactor the concept in a macro for further usage

Should help on #832

### DIFF
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -246,7 +246,7 @@ impl BlockSize {
 
     (offset_x, offset_y)
   }
-  
+
   pub fn greater_than(self, other: BlockSize) -> bool {
     (self.width() > other.width() && self.height() >= other.height()) ||
     (self.width() >= other.width() && self.height() > other.height())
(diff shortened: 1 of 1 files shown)


src/partition.rs
```diff
@@ -775,9 +775,13 @@ pub enum MvSubpelPrecision {
   MV_SUBPEL_HIGH_PRECISION
 }
 
+mod subpel_filters {
 pub const SUBPEL_FILTER_SIZE: usize = 8;
 
-const SUBPEL_FILTERS: [[[i32; SUBPEL_FILTER_SIZE]; 16]; 6] = [
+#[repr(C, align(256))]
```
tdaede: Looks fine, but do you really need 256 byte alignment?

lu-zero:
32 would be enough for all our current usage, I think?

KyleSiefring:
There seems to already be an aligned type in util.rs. https://github.com/xiph/rav1e/blob/master/src/util.rs#L133

lu-zero:
@KyleSiefring yes, we can rework the code to use it in the macro, but I'd rather have you confirm that this approach works while I'm flying back home :)

KyleSiefring:
Does not work. SUBPEL_FILTERS : Aligned = Aligned([ does work. I would like it if we could find some documentation that confirms that this behavior is defined.

lu-zero:
Do you have the code somewhere?

On Jan 2, 2019 16:20, "Kyle Siefring" <notifications@github.com> wrote:

Does not work. SUBPEL_FILTERS : Aligned = Aligned([ does work. I would like
it if we could find some documentation that confirms that this behavior is
defined.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/xiph/rav1e/pull/834#issuecomment-450890855>, or mute
the thread

KyleSiefring:
https://github.com/KyleSiefring/rav1e/tree/asm_data_in_rust_WIP_1

KyleSiefring:
FYI The variable is rav1e_mc_subpel_filters and not SUBPEL_FILTERS in my code.

lu-zero:
https://github.com/rust-av/rav1e/pull/new/asm_data_in_rust_WIP_1

this builds

tdaede:
Do we still need this?

PR #833, opened 2019-01-02
Add and Enable Loop Restoration Filters

xiphmont:
Work in progress, currently here for the automated testing.

PR #836, opened 2019-01-04
Store RestorationState into FrameState

rom1v:
## Context

In order to prepare for the possibility to encode tiles in parallel, I implemented an iterator providing non-overlapping tiles (`TileMut<'_>`), which are a "view" on the underlying frame (see https://github.com/xiph/rav1e/pull/821).

But _frames_ are only part of the data to be split.

To encode a tile, we will also need one `TileState` (similar to `FrameState`) per tile, each providing access to the tile (`TileMut<'_>`).

In addition, currently, the `RestorationState` is used both at "tile level" (accessing only data for a specific tile, independently of other tiles) and at "frame level" (crossing tile borders). To allow tile parallelization, the same way we created `TileMut<'_>` view from a `Frame`, we can create a `TileRestorationStateMut<'_>` (or whatever the name) view from a `RestorationState`. Concretely, such a view would expose only the `RestorationUnit`s (stored in the `RestorationPlane`s) belonging to the tile.

## Objective

Instead of managing several views (`TileMut<'_>`, `TileStateMut<'_>`, `TileRestorationStateMut<'_>`), the idea is to only use one (top-level) `TileStateMut<'_>` containing all tile-specific data and views.

Concretely, `FrameState` would expose an iterator providing `TileStateMut<'_>` instances, each one used to encode a tile (possibly in its own thread).

In `TileStateMut<'_>`:
 - `tile_state.tile` will expose the `TileMut<'_>`;
 - `tile_state.restoration` will expose the `TileRestorationStateMut<'_>`.

Once all the tiles for a frame have been encoded, we can then access the full `FrameState` (which has been written through the tile views) to execute "frame level" stuff (deblocking, restoration, …).

## PR

The first step to get a "unified" view for a tile, containing all necessary states, is to move `RestorationState` into `FrameState`.

Here is a summary of the commits (see commit messages for more details):
 - commit 1 removes `FrameState::window()`, which is unused and add unnecessary restrictions.
 - commit 2 avoids to borrow `FrameState` from a non-const `RestorationState` method (this would break borrowing rules once we move it).
 - commit 3 make `RestorationState` derive `Debug`.
 - commit 4 actually moves `RestorationState` into `FrameState`.
 - optional commits 5 and 6 narrow the borrowing to `FrameState` fields only (for consistency with what we do with `RestorationState`).

This PR is independent of tile encoding support (it just paves the way).

## Impact

This will conflicts with https://github.com/xiph/rav1e/pull/833, but should be easy to resolve.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -895,7 +895,7 @@ trait UncompressedHeader {
     fn write_deblock_filter_b(&mut self, fi: &FrameInvariants, fs: &FrameState) -> io::Result<()>;
     fn write_frame_cdef(&mut self, fi: &FrameInvariants) -> io::Result<()>;
     fn write_frame_lrf(&mut self, fi: &FrameInvariants, rs: &RestorationState) -> io::Result<()>;
-    fn write_segment_data(&mut self, fi: &FrameInvariants, fs: &FrameState) -> io::Result<()>;
+    fn write_segment_data(&mut self, fi: &FrameInvariants, segmentation: &SegmentationState) -> io::Result<()>;
     fn write_delta_q(&mut self, delta_q: i8) -> io::Result<()>;
 }
 #[allow(unused)]
@@ -1304,7 +1304,7 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
       self.write_bit(false)?; // no qm
 
       // segmentation
-      self.write_segment_data(fi, fs)?;
+      self.write_segment_data(fi, &fs.segmentation)?;
 
       // delta_q
       self.write_bit(false)?; // delta_q_present_flag: no delta q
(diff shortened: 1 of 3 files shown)


rom1v:
> you could shuffle away the Debug decoration in a separate commit

Done. :wink:

rom1v:
Yes. However, I'm not happy with `FrameState::new()` requiring a `&Sequence`, just to pass it to `RestorationState::new()`, which just uses it to read the boolean `use_128x128_superblock`.

src/api.rs
```diff
@@ -491,7 +491,7 @@ impl Context {
     if self.fi.show_existing_frame {
       self.idx += 1;
 
-      let mut fs = FrameState::new(&self.fi, self.seq.chroma_sampling);
+      let mut fs = FrameState::new(&self.seq, &self.fi, self.seq.chroma_sampling);
```
ycho: Maybe already answered in irc, but can you pls explain why fi info is required in fs?

rom1v: > why fi info is required
You mean `seq`?

ycho: Oh, right, the seq.

rom1v:
I updated the PR after #833 and #838 have been merged.

The new version is simpler since we don't need an additional `&Sequence` parameter everywhere.

PR #838, opened 2019-01-06
Move Sequence into FrameInvariants

rzumer:
This simplifies several function signatures.

I think that a reference to a `Sequence` would be better, but currently the header writing functions require a mutable `Sequence` and `FrameInvariants`, so there are some complications with mutable borrows. It is relatively simple to transition to a reference from there.

Will probably require changes to [crav1e](https://github.com/lu-zero/crav1e/).

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -235,7 +235,7 @@ impl Default for ChromaSamplePosition {
     }
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub struct Sequence {
     // OBU Sequence header of AV1
     pub profile: u8,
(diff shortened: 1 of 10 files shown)


src/encoder.rs
```diff
@@ -960,25 +962,27 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
...
 
         Ok(())
     }
 
-    fn write_sequence_header(&mut self, seq: &mut Sequence, fi: &FrameInvariants)
+    fn write_sequence_header(&mut self, fi: &mut FrameInvariants)
         -> io::Result<()> {
         self.write_frame_size(fi)?;
 
+        let mut seq = fi.sequence;
```
rom1v: You copy the sequence here (and modify the copy). Is it what you want?

rom1v:
```suggestion
        let seq = &mut fi.sequence;
```

rom1v: :+1:

I've never used the GitHub "merge suggestion" feature, but it produces awful commits :scream: Maybe you could just squash/fixup it with the previous one :)

lu-zero:
I guess it is time to fold this in and see how bad we broke everybody else sets :)

PR #841, opened 2019-01-08
Sub-pixel diamond search motion estimation

magwyz:
Currently high precision motion vectors are enabled by default.

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -258,7 +258,8 @@ pub fn motion_estimation<T: Pixel>(
           pmv, lambda,
           mvx_min, mvx_max, mvy_min, mvy_max,
           blk_w, blk_h,
-          &mut best_mv, &mut lowest_cost);
+          &mut best_mv, &mut lowest_cost, &mut None, ref_frame
+        );
       } else {
         let range = 16;
         let x_lo = po.x + ((-range + (cmv.col / 8) as isize).max(mvx_min / 8).min(mvx_max / 8));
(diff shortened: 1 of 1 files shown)


src/me.rs
```diff
@@ -148,113 +148,144 @@ pub fn motion_estimation(
...
-            }
-          }
+    for p in diamond_pattern.iter() {
+
+        let cand_mv = MotionVector {
+          row: center_mv.row + diamond_radius * p.0,
+          col: center_mv.col + diamond_radius * p.1
+        };
+
+        let cost = get_mv_cost(
```
ycho: The 'cost' here includes sad since get_mv_cost() now does sad calculation.
Comment about it would help readers, because traditionally 'mv_cost' means bit cost for the mv only in many video codec codebase and academic papers.

magwyz: Thank you for your comment.
I can also rename the 'get_mv_cost' function and the 'cost' variable if their name are confusing.

magwyz:
Here is the impact on the file size and the compression time computed with my own scripts.
I'm waiting for an access to AWCY to confirm these figures.
http://people.videolan.org/~magsoft/rav1e/diamond_me2_size_impact.png
http://people.videolan.org/~magsoft/rav1e/diamond_me2_time_impact.png

src/me.rs
ycho: I'd recommend a very verbose get_mv_and_sad_cost() here! :)

magwyz:
Here are the AWCY runs:
https://beta.arewecompressedyet.com/?job=magsoft_diamond_me2_ref%402019-01-10T08%3A03%3A15.473Z&job=magsoft-diamond_me2

The gain are still confirmed for Pierside and the shields at high rates.

src/me.rs
ycho: As we have discussed in the last daala meeting,
since get_mv_cost() involved lambda, "return 256 * sad + rate * lambda;", the function is essentially return the final RD cost. So, to me it would be best to name the function simply as "get_rd_cost()" I think.
Hence, I'd recommend: let rd_cost = get_rd_cost(...);

tdaede:
I definitely want this too, but can you rebase it on the fullpel only version?

magwyz:
My graphs were made at the default quantizer. Actually, the significant gain for some samples comes from the activation of the high precision motion vector inside the patches.
This is validated by these AWCY runs where I have enabled the high precision motion vectors:
https://beta.arewecompressedyet.com/?job=magsoft_test_high_precision_mv%402019-02-12T16%3A35%3A20.056Z&job=magsoft_diamond_me4_ref%402019-02-12T12%3A15%3A09.323Z

I do not think this pull request has a real interest.

magwyz:
Well, I probably went a bit too quick reanalysing the results.
I'll test what was suggested: adding the subpixel diamond motion estimation over the full pixel one.

ycho:
Can you disable the high precision motion vector in this patch to isolate its effect?
You can always turn it on and/or add and enable corresponding code for that mode in separate patch.

magwyz:
Sure, I can do this. That would be interesting.
However, this patch conflicts with #873, which will probably be merged before. That is why it may be interesting to take from here the idea of subpixel diamond motion estimation and to rebase it above #873 as @tdaede suggested.

magwyz:
So, I updated this branch with sub-pixel diamond motion estimation above the full-pixel and this seems interesting regarding the encoding time.
Here are the new AWCY runs: https://beta.arewecompressedyet.com/?job=diamond_me4_test%402019-02-18T15%3A32%3A45.972Z&job=subpel_diamond_me2_high_prec%402019-02-18T14%3A32%3A07.187Z&job=subpel_diamond_me_ref%402019-02-18T14%3A29%3A36.872Z

- subpel_diamond_me2_high_prec is the branch of this MR
- diamond_me4_test is the #873 branch
- subpel_diamond_me_ref is the master branch

magwyz:
I rebased the branch.
Here are the new AWCY runs:
https://beta.arewecompressedyet.com/?job=diamond_me2_ref%402019-02-20T10%3A29%3A40.010Z&job=diamond_me2_test%402019-02-20T12%3A53%3A26.595Z

magwyz:
Here are some updated AWCY runs:
https://beta.arewecompressedyet.com/?job=diamond_me2_test%402019-03-04T18%3A34%3A12.581Z&job=diamond_me2_ref%402019-03-04T18%3A34%3A42.544Z

PR #842, opened 2019-01-08
Remove duplicate fields

rom1v:
`lrf_type` and `unit_size` are per-plane values, there is no need to duplicate them as arrays in `RestorationState`.

(we could also rename `plane` to `planes`)

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -3036,7 +3036,7 @@ impl ContextWriter {
               }
             }
             RestorationFilter::Sgrproj{set, xqd} => {
-              match rs.lrf_type[pli] {
+              match rs.plane[pli].lrf_type {
                 RESTORE_SGRPROJ => {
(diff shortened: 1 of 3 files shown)


ycho:
I think the location matters, not a duplication and no change of size for the struct 'RestorationState' ? Now I see that you moved it under plane, one step deeper.

rom1v:
> I think the location matters

What do you mean?

> not a duplication and no change of size for the struct 'RestorationState'?

The `RestorationState` is now smaller, since I removed 2 field (but I think I missed your point).

> Now I see that you moved it under plane, one step deeper.

They were already (also) under plane.

ycho:
Oh, were they already existing under each of planes? Then, what intended and said is correct!
I misunderstood, thinking that you added new field for each plane.

PR #844, opened 2019-01-09
Revert xorigin and yorigin to usize

rom1v:
The type of fields xorigin and yorigin was changed from usize to isize
only for Plane::window(), which has now been removed.

Plane origins may not be negative, so revert their type to usize.

This partially reverts 1bd94a2d4b782446ae4be8645c46136cc988af1f.

### DIFF
--- a/src/plane.rs
+++ b/src/plane.rs
@@ -22,8 +22,8 @@ pub struct PlaneConfig {
   pub height: usize,
   pub xdec: usize,
   pub ydec: usize,
-  pub xorigin: isize,
-  pub yorigin: isize
+  pub xorigin: usize,
+  pub yorigin: usize
 }
 
 /// Absolute offset in pixels inside a plane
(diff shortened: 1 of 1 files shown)


lu-zero:
I do not have a specific opinion, those values are used in computations that mix usize and isize no matter what because of the padded data.

@barrbrain do you have a stronger opinion than mine? :)

rom1v:
> those values are used in computations that mix usize and isize no matter what because of the padded data

I also use them in tiles, where I use [unsigned values](https://github.com/xiph/rav1e/pull/821/files#diff-44243c4dc2bc086c7b55ac4feb98bb3eR20) (because they may not be negative). and I have to [cast them unnecessarily](https://github.com/xiph/rav1e/pull/821/files#diff-44243c4dc2bc086c7b55ac4feb98bb3eR163).

PR #846, opened 2019-01-09
Fix over restrictive borrow rules for plane slice

KyleSiefring:
Previous implementation prevented uses like
let newslice = oldslice.go_left().go_up();

### DIFF
--- a/src/plane.rs
+++ b/src/plane.rs
@@ -126,11 +126,11 @@ impl Plane {
     }
   }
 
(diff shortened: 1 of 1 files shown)


rom1v:
I disagree with your last commit.

```rust
struct X<'a> {
    v: &'a mut u32,
}

impl<'a> X<'a> {
    fn f(&mut self) -> &'a mut u32 {  // you don't want 'a here!
        self.v
    }
}

fn main() {
    let mut v = 32;
    let mut x = X { v: &mut v };
    {
        let _ = x.f();
    }
    let _ = x.f();
}
```

PR #848, opened 2019-01-10
Simplify RestorationState (prepare for tiling)

rom1v:
## Objective

As I explained in #836, I want to create `TileState<'_>` views from a `FrameState`, so that we can encode tiles in parallel (several threads could work on different `TileState<'_>` simultaneously).

`FrameState` now contains (since #836) a `RestorationState`, so `TileState<'_>` will contain a `TileRestorationState<'_>`, a tiled-view of the restoration state.

However, `RestorationState` contains `RestorationPlane`s, which each in turn contain a `PlaneConfig` named `clipped_cfg`. As a preliminary step, I would like to **remove this field entirely**, to simplify the future `TileRestorationState<'_>`.

## Details

AFAIU, this _clipped_ `PlaneConfig` was used for two reasons.

### Non-padded frame

`Frame`s are created with [padded width and height](https://github.com/xiph/rav1e/blob/969549044933efcdc4e98d900f63fc4fa69b3f40/src/encoder.rs#L431) ([rounded up to the next multiple of 8](
https://github.com/xiph/rav1e/blob/969549044933efcdc4e98d900f63fc4fa69b3f40/src/encoder.rs#L594-L595)).

Contrary to the other loop filters, the restoration filter is applied on the "cropped" (i.e. non-padded) frame. The `width` and `height` fields in `PlaneConfig` contain the padded values, only `FrameInvariants` knows the original values. _(thanks @xiphmont!)_

Therefore, on construction,`RestorationState` created a [clipped version](https://github.com/xiph/rav1e/blob/969549044933efcdc4e98d900f63fc4fa69b3f40/src/lrf.rs#L447-L461) from the frame `PlaneConfig`. It then used it in `lrf_filter_frame()` to get non-padded dimensions.

But we can achieve the same result by simply passing `&FrameInvariants` to `lrf_filter_frame()`. **This is what the first commit changes.**

### Mapping superblock → restoration unit

`RestorationPlane` exposes methods to retrieve the `RestorationUnit` associated to a `SuperBlockOffset` (currently, it [only called from `write_lrf()`](https://github.com/xiph/rav1e/blob/969549044933efcdc4e98d900f63fc4fa69b3f40/src/context.rs#L3017)).

For that purpose, it converted the offset to [plane coordinates](https://github.com/xiph/rav1e/blob/969549044933efcdc4e98d900f63fc4fa69b3f40/src/lrf.rs#L416-L422). But then, this made the implementation dependant of chroma subsampling, while in theory the mapping [superblock → restoration unit] is independant of subsampling. To handle this, the `xdec` and `ydec` fields of the clipped `PlaneConfig` were used to convert properly.

But we can retrieve the restoration unit directly from the superblock offset, without passing by plane coordinates. That way, we don't need `clipped_cfg` anymore. **This is what the second commit changes.**

### DIFF
--- a/src/lrf.rs
+++ b/src/lrf.rs
@@ -16,7 +16,6 @@ use context::SuperBlockOffset;
 use context::PLANES;
 use context::MAX_SB_SIZE;
 use plane::Plane;
-use plane::PlaneConfig;
 use plane::PlaneOffset;
 use std::cmp;
 use util::clamp;
@@ -376,10 +375,11 @@ impl RestorationUnit {
 
 #[derive(Clone, Debug)]
 pub struct RestorationPlane {
-  // all size units are subsampled if the plane is subsampled
-  pub clipped_cfg: PlaneConfig,
   pub lrf_type: u8,
   pub unit_size: usize,
+  // (1 << sb_log2) gives the number of superblocks (having size (1 << SUPERBLOCK_TO_PLANE_SHIFT))
+  // both horizontally and vertically in a restoration unit
+  pub sb_log2: usize,
   pub cols: usize,
   pub rows: usize,
   pub wiener_ref: [[i8; 3]; 2],
@@ -394,15 +394,11 @@ pub struct RestorationPlaneOffset {
 }
 
 impl RestorationPlane {
-  pub fn new(clipped_cfg: &PlaneConfig, lrf_type: u8, unit_size: usize) -> RestorationPlane {
-    let PlaneConfig { width, height, .. } = clipped_cfg;
-    // bolted to superblock size for now
-    let cols = cmp::max((width + (unit_size>>1)) / unit_size, 1);
-    let rows = cmp::max((height + (unit_size>>1)) / unit_size, 1);
+  pub fn new(lrf_type: u8, unit_size: usize, sb_log2: usize, cols: usize, rows: usize) -> RestorationPlane {
     RestorationPlane {
-      clipped_cfg: clipped_cfg.clone(),
       lrf_type,
       unit_size,
+      sb_log2,
       cols,
       rows,
       wiener_ref: [WIENER_TAPS_MID; 2],
@@ -411,29 +407,21 @@ impl RestorationPlane {
     }
   }
 
-  /// find the restoration unit offset corresponding to the this superblock offset
-  /// This encapsulates some minor weirdness due to RU stretching at the frame boundary.
-  pub fn restoration_plane_offset(&self, sbo: &SuperBlockOffset) -> RestorationPlaneOffset {
-    let po = sbo.plane_offset(&self.clipped_cfg);
-    RestorationPlaneOffset {
-      row: cmp::min(po.y / self.unit_size as isize, self.rows as isize - 1) as usize,
-      col: cmp::min(po.x / self.unit_size as isize, self.cols as isize - 1) as usize
-    }
+  fn restoration_unit_index(&self, sbo: &SuperBlockOffset) -> (usize, usize) {
+    (
+      (sbo.x >> self.sb_log2).min(self.cols - 1),
+      (sbo.y >> self.sb_log2).min(self.rows - 1),
+    )
   }
 
   pub fn restoration_unit(&self, sbo: &SuperBlockOffset) -> &RestorationUnit {
-    let rpo = self.restoration_plane_offset(sbo);
-    &self.units[rpo.row][rpo.col]
+    let (x, y) = self.restoration_unit_index(sbo);
+    &self.units[y][x]
   }
 
   pub fn restoration_unit_as_mut(&mut self, sbo: &SuperBlockOffset) -> &mut RestorationUnit {
-    let rpo = self.restoration_plane_offset(sbo);
-    &mut self.units[rpo.row][rpo.col]
-  }
-
-  pub fn restoration_unit_by_stripe(&self, stripenum: usize, rux: usize) -> &RestorationUnit {
-    &self.units[cmp::min((stripenum * 64 >> self.clipped_cfg.ydec) / self.unit_size, self.rows - 1)]
-      [cmp::min(rux, self.cols - 1)]
+    let (x, y) = self.restoration_unit_index(sbo);
+    &mut self.units[y][x]
   }
 }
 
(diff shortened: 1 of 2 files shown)


rom1v:
Fixed :tada:

So the half-rounding for cols/rows was on purpose, and the min/max were there for a reason :smile:

```diff
diff --git a/src/lrf.rs b/src/lrf.rs
index 119825f..b2ddcfb 100644
--- a/src/lrf.rs
+++ b/src/lrf.rs
@@ -408,7 +408,10 @@ impl RestorationPlane {
   }
 
   fn restoration_unit_index(&self, sbo: &SuperBlockOffset) -> (usize, usize) {
-    (sbo.x >> self.sb_log2, sbo.y >> self.sb_log2)
+    (
+      (sbo.x >> self.sb_log2).min(self.cols - 1),
+      (sbo.y >> self.sb_log2).min(self.rows - 1),
+    )
   }
 
   pub fn restoration_unit(&self, sbo: &SuperBlockOffset) -> &RestorationUnit {
@@ -438,8 +441,8 @@ impl RestorationState {
     let lr_uv_shift = if input.planes[1].cfg.xdec > 0 && input.planes[1].cfg.ydec > 0 {1} else {0};
     let y_unit_size = RESTORATION_TILESIZE_MAX >> (2 - sb_log2);
     let uv_unit_size = y_unit_size >> lr_uv_shift;
-    let cols = (fi.width + y_unit_size - 1) / y_unit_size;
-    let rows = (fi.height + y_unit_size - 1) / y_unit_size;
+    let cols = ((fi.width + (y_unit_size >> 1)) / y_unit_size).max(1);
+    let rows = ((fi.height + (y_unit_size >> 1)) / y_unit_size).max(1);
     RestorationState {
       plane: [RestorationPlane::new(RESTORE_SWITCHABLE, y_unit_size, sb_log2, cols, rows),
               RestorationPlane::new(RESTORE_SWITCHABLE, uv_unit_size, sb_log2, cols, rows),
@@ -489,7 +492,9 @@ impl RestorationState {
           } else {
             rp.unit_size
           };
-          let ru = &rp.units[si >> rp.sb_log2][rux];
+          // there may be more stripes than rows, due to how stripe_n is initialized
+          let ruy = (si >> rp.sb_log2).min(rp.rows - 1);
+          let ru = &rp.units[ruy][rux];
           match ru.filter {
             RestorationFilter::Wiener{coeffs} => {          
               wiener_stripe_rdu(coeffs, fi,
```

ycho:
Hi-, the 1st patch "Use non-padded dimensions from FrameInvariants " is LGTM.

src/lrf.rs
```diff
@@ -444,46 +432,31 @@ pub struct RestorationState {
...
-    // Currrently opt for smallest possible restoration unit size
-    let lrf_y_shift = if fi.sequence.use_128x128_superblock {1} else {2};
-    let lrf_uv_shift = lrf_y_shift + if xdec>0 && ydec>0 {1} else {0};
-    let lrf_type: [u8; PLANES] = [RESTORE_SWITCHABLE, RESTORE_SWITCHABLE, RESTORE_SWITCHABLE];
-    let unit_size: [usize; PLANES] = [RESTORATION_TILESIZE_MAX >> lrf_y_shift,
-                                      RESTORATION_TILESIZE_MAX >> lrf_uv_shift,
-                                      RESTORATION_TILESIZE_MAX >> lrf_uv_shift];
+    // if 128x128 superblock is enabled, we need a shift because currently
+    // SuperBlockOffset is always defined in terms of 64x64 superblocks
+    let sb_log2 = if fi.sequence.use_128x128_superblock {1} else {0};
```
ycho: @xiphmont, Do you think this is correct?
( (1 << sb_log2) gives # of SBs in a restoration unit, both horizontally and vertically)

ycho: So, if use_128x128_superblock is true, restoration unit size = 256x256, and it has 2 SBs in hor and ver?

rom1v: > So, if use_128x128_superblock is true, restoration unit size = 256x256, and it has 2 SBs in hor and ver?

Currently, the restoration unit size is always exactly 1 superblock:
 - 128×128 if `use_128x128_superblock`;
 - 64×64 otherwise.

`(1 << sb_log2)` gives the size of the restoration unit always expressed in "superblocks" of size 64×64, even if `use_128x128_superblock` (the caller always gives superblock offsets expressed in [64x64 blocks](https://github.com/xiph/rav1e/blob/07032e6b9702f05446629422750337a49794499f/src/context.rs#L1129-L1153)).

Concretely, if `use_128x128_superblock`, then `sb_log2 = 1`. `1 << sb_log2 == 2`, so the restoration unit is a square with side length 2 (a square 2×2 "superblocks-64x64"). So the restoration unit is 128×128.

Otherwise, `sb_log2 = 0`. `1 << sb_log2 == 1`, so the restoration unit is a square with side length 1 (a single "superblock-64x64"). So the restoration unit is 64×64.

(I kept the existing behavior.)

xiphmont:
> RestorationPlane exposes methods to retrieve the RestorationUnit associated to a SuperBlockOffset 
>(currently, it only called from write_lrf()).

It is used in more places in upcoming code.

> But we can retrieve the restoration unit directly from the superblock offset, without passing by plane 
> coordinates. That way, we don't need clipped_cfg anymore. This is what the second commit changes.

Clipped_cfg is needed for more than that; I need the actual clipped pixel location is the filtering (though, of course, I can get that from fi).  However, conversion is still dependent on subsampling depending on the flags that configure LRF; right now we're only using flags that just happen to work out with this change, but won't work once LRF is no longer working in 'simplest possible' mode.

No reason to stress about this, it's kind of vague, I'll point out specifics in the code.

src/lrf.rs
ycho: I think "always expressed in "superblocks" of size 64×64, even if use_128x128_superblock" confused me, and can do possibly some others as well.
Adding '64' somewhere with 'sb_log2', such as 'sb64_log2', would help readers.

In addition, the comment for sb_log2 member,  "// (1 << sb_log2) gives the number of superblocks (having size (1 << SUPERBLOCK_TO_PLANE_SHIFT))" will needs update once SUPERBLOCK_TO_PLANE_SHIFT became 7.

ycho: I think I am still correct.
If it says : "// (1 << sb_log2) gives the number of **64x64** superblocks (having size (1 << **6**))" ... ", then I agree it is correct.

However, with the current comment, as you commented above like "if use_128x128_superblock, then sb_log2 = 1. 1 << sb_log2 == 2", which is 2 of 64 pels, i.e. 1 << 6. While "(1 << SUPERBLOCK_TO_PLANE_SHIFT) is 128 if use_128x128_superblock is true.

rom1v: > If it says : "// (1 << sb_log2) gives the number of **64x64** superblocks (having size (1 << **6**))" ... ", then I agree it is correct.

I think this would become incorrect then, because if `SUPERBLOCK_TO_PLANE_SHIFT` becomes `7`, then `SuperBlockOffset` will be expressed in terms of superblocks 128×128, and `(1 << sb_log2)` will gives the number of superblocks 128×128.

However, this one would need to be updated:

    // SuperBlockOffset is always defined in terms of 64x64 superblocks

(but it would make the comment less clear)

xiphmont: > Currently, the restoration unit size is always exactly 1 superblock:

And that's temporary only as a bootstrap to get the code working for now.  It's a mode we will almost never use in practice once LRF is complete.

Getting all this sizing logic correct was a minor pain; stripping it out is a mistake.

rom1v: > as you commented above like "if use_128x128_superblock, then sb_log2 = 1. 1 << sb_log2 == 2", which is 2 of 64 pels, i.e. 1 << 6. While "(1 << SUPERBLOCK_TO_PLANE_SHIFT) is 128 if use_128x128_superblock is true.

Yes, the comment on this PR would need to be updated in that case (but not the comment in the code) :wink:

rom1v: >> Currently, the restoration unit size is always exactly 1 superblock:
> And that's temporary only as a bootstrap to get the code working for now. It's a mode we will almost never use in practice once LRF is complete.

The purpose of `sb_log2` is to encode the number of superblocks per restoration unit. Currently it is always 0 (or 1 due to the "superblock" VS "superblock-64x64" quirk), but the idea was to change it. Isn't it sufficient?

PR #849, opened 2019-01-10
Rewrite subpel code to look more like dav1d

KyleSiefring:


src/mc.rs
```diff
@@ -0,0 +1,326 @@
...
+    [0, 0, 10, 46, 56, 16, 0, 0],
+    [0, 0, 8, 44, 58, 18, 0, 0],
+    [0, 0, 6, 42, 60, 20, 0, 0],
+    [0, 0, 4, 40, 62, 22, 0, 0],
+    [0, 0, 4, 36, 62, 26, 0, 0],
+    [0, 0, 2, 34, 62, 30, 0, 0]
+  ]
+];
+
+fn round_shift(val: i32, shift: i32) -> i32 {
```
lu-zero: Probably we have something like that somewhere already.

src/util.rs
```diff
@@ -145,21 +147,21 @@ struct Align16;
 /// let mut x: AlignedArray<[i16; 64 * 64]> = UninitializedAlignedArray();
 /// assert!(x.array.as_ptr() as usize % 16 == 0);
 /// ```
-pub struct AlignedArray<ARRAY>
+pub struct AlignedArray<ARRAY, AlignType = Align16>
```
lu-zero: This part might require a little discussion, but the alignment is a property that should change depending on the platform, not on the usage in the code.

Ideally you could (and should) bump the alignment to 32 for now and hide that detail from the rest of the code.

src/mc.rs
KyleSiefring: Yeah, I'm pretty sure it's in the transform code. I was thinking of consolidating that in another patch. I can do it in this one though.

src/util.rs
KyleSiefring: I don't think making it platform specific is perfect since some avx2 code will only need 128 bit alignment.

PR #850, opened 2019-01-11
Store restoration units contiguously

rom1v:
The "matrix" of restoration units was stored in a vector of vectors.

To avoids an additional indirection for each row, store them in a
contiguous array instead. Besides the (theoretical) performance benerit
for every access, this will simplify the implementation of tiled-views
which have to expose non-overlapping parts of the "matrix" via unsafe
code.

Also, since its size is constant (although not known at compile-time),
replace the vector by a boxed slice.

### DIFF
--- a/src/lrf.rs
+++ b/src/lrf.rs
@@ -384,7 +384,7 @@ pub struct RestorationPlane {
   pub rows: usize,
   pub wiener_ref: [[i8; 3]; 2],
   pub sgrproj_ref: [i8; 2],
-  pub units: Vec<Vec<RestorationUnit>>
+  pub units: Box<[RestorationUnit]>,
 }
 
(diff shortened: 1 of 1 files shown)


lu-zero:
if you run the tests on your local machine what happens?

rom1v:
cargo test --features=decode_test

works (but some of them are ignored, like `test_encode_decode_aom::odd_size_frame_with_full_rdo` or all `test_encode_decode_aom::speed_XXX`.

I will amend the commit to get a new sha1 and push-force, so that the CI is run again.

lu-zero:
`cargo test --feature=decode_test --ignored`

rom1v:
OK, this works:

    cargo test --features=decode_test -- --ignored

:+1:

PR #856, opened 2019-01-15
Bypass filtering of strengh=0 cdef blocks for speed.

tdaede:


tdaede:
Yeah I tested it and it seems to work OK.

PR #859, opened 2019-01-16
Early termination for topdown partition search

ycho:


ycho:
There is very small speed gain at -s 1 with no change in bdrate.

[AWCY result for -s 1](https://beta.arewecompressedyet.com/?job=test0_early_exit_topdown_s1%402019-01-23T22%3A24%3A04.026Z&job=master_525f981_s1%402019-01-23T21%3A05%3A44.244Z)

rzumer:
Is this a simplification of the early exit method I had implemented for it earlier? If not how does it differ? Sorry I have not kept up to date with the latest changes.

ycho:
Sorry, I don't know which work you are referring to, you mean some earlier PR which has not merged?
But, the idea of "stop breaking blocks down if the RD cost is higher" is exactly right and same as in this patch.

rzumer:
I think this may explain the small change in performance; you changes may allow it to run more efficiently, but the early exit was already implemented in this partitioning scheme. I guess I should have documented it better.

Issue #794, opened 2018-12-06
Switch adaptive keyframe placement algorithm to a cost-based one + tweaks


dwbuiten:
Filing a bug as encouraged by peeps on IRC.

Realistically, this will (should) yeild much, much better results than just a basic diff-like algorithm, and is what most reasonable encoders for various codecs seem to use. And, at least to me, most cost-efficient placement makes more sense for an encoder than just 'different', or what a human may call a scene change.

x264's placement decision code is pretty short (by encoder standards), easy to read, and a good reference, IMO: The [scenecut](https://git.videolan.org/?p=x264.git;a=blob;f=encoder/slicetype.c;h=8c3ef3ba172c1283c727398e6b8b795b8a7c0b7a;hb=HEAD#l1423) and [scenecut_internal](https://git.videolan.org/?p=x264.git;a=blob;f=encoder/slicetype.c;h=8c3ef3ba172c1283c727398e6b8b795b8a7c0b7a;hb=HEAD#l1377) functions are the relevant ones.

Also of note in those are some simple tweaks that can improve perf quite a lot, I think, like [avoiding problems with flashes/strobes](https://git.videolan.org/?p=x264.git;a=blob;f=encoder/slicetype.c;h=8c3ef3ba172c1283c727398e6b8b795b8a7c0b7a;hb=HEAD#l1429).

Of course, I don't know how hard this is to integrate into rav1e with its current code structure, or if it falls under "wait for trellis / aq / etc", etc.

Apologies if this is obvious or not useful to file as a bug.

tdaede:
It's not dependent on aq, it's purely a frame level decision. The existing code importantly added the support for variable GOP structures, but there are definitely much better detection methods.

Basically, we need a lookahead or 2 pass with cost estimates for both a P and I frame.

dwbuiten:
Lookahead would be better (and what x264 uses), no? (So as to support e.g. streaming.) Unless by 2-pass, you mean something more akin to a lookahead-y small 2-pass rather than full stream?

tdaede:
Lookahead and 2-pass are basically equivalent (one stores the results in memory, one stores the results in a file), I would like to support them with the same code.

dwbuiten:
Yeah, that's fair.

shssoichiro:
[...]

PR #869, opened 2019-01-19
Implement lookahead queue

shssoichiro:
Allows rav1e to read frames into the frame queue ahead of the current
one, to allow for future features to be added that depend on a lookahead
pass. Currently the lookahead distance is hardcoded to 10. In the
future, this would probably be configurable as a CLI option. Increasing
the lookahead distance primarily comes with a memory cost. This lookahead
implementation aims to minimize the memory cost by removing previous frames
from the queue after they are processed.

Among other items, this is necessary to implement #794.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -17,6 +17,10 @@ use self::EncoderStatus::*;
 use std::{cmp, fmt, io};
 use std::collections::BTreeMap;
 use std::sync::Arc;
+use util::Fixed;
+use std::collections::BTreeSet;
+
+const LOOKAHEAD_FRAMES: u64 = 10;
 
 // TODO: use the num crate?
 #[derive(Clone, Copy, Debug)]
@@ -291,14 +295,6 @@ impl Config {
   }
 
   pub fn new_context(&self) -> Context {
-    let seq = Sequence::new(&self.frame_info);
-    let fi = FrameInvariants::new(
-      self.frame_info.width,
-      self.frame_info.height,
-      self.enc,
-      seq,
-    );
-
     #[cfg(feature = "aom")]
     unsafe {
       av1_rtcd();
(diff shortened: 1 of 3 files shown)


src/api.rs
```diff
@@ -369,16 +373,16 @@ impl fmt::Display for Packet {
...
+      self.config.frame_info.height.align_power_of_two(3),
+      self.config.frame_info.chroma_sampling
+    ))
   }
 
-  pub fn send_frame<F>(&mut self, frame: F) -> Result<(), EncoderStatus>
-  where
-    F: Into<Option<Arc<Frame>>>
-  {
+  pub fn send_frame(&mut self, frame: Option<Arc<Frame>>) -> Result<(), EncoderStatus> {
```
lu-zero: Please leave the `Into<Option>` in place.

shssoichiro: Fixed

shssoichiro:
Thanks. I tried to make the PR as small as possible--hence why this is separate from implementing the lookahead-based keyframe detection. I'm not sure if there would have been a reasonable point to split it even further.

src/api.rs
```diff
@@ -291,47 +295,47 @@ impl Config {
...
-  fi: FrameInvariants,
   //    timebase: Rational,
   frame_count: u64,
   frames_to_be_coded: u64,
   idx: u64,
+  frames_processed: u64,
+  /// Maps frame *number* to frames
   frame_q: BTreeMap<u64, Option<Arc<Frame>>>, //    packet_q: VecDeque<Packet>
+  /// Maps frame *idx* to frame data
+  frame_data: BTreeMap<u64, FrameInvariants>,
```
tdaede: Thanks for converting FrameInvariants to an array. Would it make sense to convert both frame_q and frame_data to both use either number or idx?

src/bin/common.rs
```diff
@@ -257,149 +258,175 @@ impl fmt::Display for FrameSummary {
   }
 }
 
+fn read_lookahead(ctx: &mut Context, y4m_dec: &mut y4m::Decoder<'_, Box<dyn Read>>, y4m_details: Y4MDetails) {
```
tdaede: Technically this function gets called even if we have 0 lookahead, right?

src/api.rs
shssoichiro: I had hoped to accomplish that, however I struggled to get that to work due to how frame reordering works--the fact that we don't know which idx links to which frame number until we've computed the invariants for that frame.

src/bin/common.rs
shssoichiro: This will get called but won't load any frames. I can add logic to the `needs_more_lookahead` method to handle that case. (Technically, the min lookahead should be 1.)

shssoichiro: Renamed to "read_frame_batch", since it may read one frame or it may read as many as it needs to fill the lookahead queue.

Issue #871, opened 2019-01-21
decode_test and decode_test_dav1d cannot be built at the same time


jbeich:
```
$ cargo build --features decode_test,decode_test_dav1d
    Updating crates.io index
error: multiple packages link to native library `clang`, but a native library can be linked only once

package `clang-sys v0.26.4`
    ... which is depended on by `bindgen v0.45.0`
    ... which is depended on by `dav1d-sys v0.1.1`
    ... which is depended on by `rav1e v0.1.0 (/path/to/rav1e)`
links to native library `clang`

package `clang-sys v0.22.0`
    ... which is depended on by `bindgen v0.37.0`
    ... which is depended on by `rav1e v0.1.0 (/path/to/rav1e)`
also links to native library `clang`
```

lu-zero:
Thank you for reporting, if you could test the pr I just sent would be great :)

lu-zero:
Perfect, thank you a lot!

PR #879, opened 2019-01-24
Revert 2c2777dc462e6385678a6e699fa04ffeea336371 -- breaks decode tests

xiphmont:
The fix to #871 breaks decode unit tests, that is:

cargo test --release --features=decode_test

fails either with a segfault or an indefinite spin.  I got no useful error output, but TD gets:

error: failed to run custom build command for `rav1e v0.1.0 (/home/thomas/sandbox/xav1)`
process didn't exit successfully: `/home/thomas/sandbox/xav1/target/release/build/rav1e-50bde2c63e94b5ba/build-script-build` (signal: 11, SIGSEGV: invalid memory reference)

Looks like a bug i new bindgen?


Cargo.toml
```diff
@@ -39,7 +39,7 @@ nasm-rs = { version = "0.1.4", optional = true }
 
 [target.'cfg(unix)'.build-dependencies]
 pkg-config = "0.3.12"
-bindgen = { version = "0.47", optional = true }
+bindgen = { version = "0.q37", optional = true }
```
tdaede: drop the q

xiphmont: maaaan

lu-zero:
I tried to reproduce the problem and seems somehow related to the nasm bump, disabling nasm make the tests pass, enabling it gets me a sigill.

PR #882, opened 2019-01-24
fix decoding test using dav1d

lu-zero:
Upstream managed to fix the discrepancy and I managed to fix a remaining mistake on our side.


tdaede:
Didn't we already merge the first commit?

PR #885, opened 2019-01-26
Add a enable_early_exit flag under fi

ycho:
Its value is set as true by default.


ycho:
Thanks a lot for review at late weekend night!

PR #887, opened 2019-01-27
Add Clippy to CI

vladimir-kazakov:
This PR adds Clippy (a linter for code quality verification) to the CI pipeline.

Currently, in order to prevent build failures and minimize the amount of noise in the build output, all errors and warnings are disabled. The idea is to start enabling different lints and lint categories slowly, one by one, fixing all found warnings and errors, so that such changes would be minimal and wouldn't affect anyone else.

Some files with `cfg_attr` attributes for Clippy had to be modified, because otherwise Clippy would treat them as warnings, since they're deprecated. Clippy is configured to treat warnings as errors, so this would cause build failures. I didn't find a way to ignore such issues, so I decided to simply fix them, because changes that they cause are quite small.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: rust
+rust:
+  - stable
 addons:
   apt:
     packages: binutils-dev libcurl4-openssl-dev zlib1g-dev libdw-dev libiberty-dev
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -24,6 +26,9 @@ before_install:
     - cmake .. && make && sudo make install
     - kcov --version
     - cargo install cargo-kcov
+before_script:
+  - rustup component add clippy
```
shssoichiro: Is there a way to only run this for the Clippy job itself? It's not that big of a deal, because this is generally a pretty quick procedure, but we don't need it in the other jobs.

EwoutH: You can remove the whole `before_script:`, and add `rustup component add clippy && cargo clippy --version && ` after `script:` in the Clippy job below

vladimir-kazakov: Done.

ycho:
Thx, I wanna see how smart it can detect.

PR #889, opened 2019-01-28
The rdo_mode_decision() only  returns RDOPartitionOutput struct

ycho:
This also makes it possiblt to remove all .clone() calls to
the return value of rdo_mode_decision().

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -2348,7 +2348,7 @@ fn encode_partition_bottomup(
         };
         let spmvs = &pmvs[pmv_idx];
 
-        let mode_decision = rdo_mode_decision(fi, fs, cw, bsize, bo, spmvs, false).part_modes[0].clone();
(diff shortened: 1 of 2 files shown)


ycho:
@rzumer, Erk, it got slower for -s 3.
@lu_zero, can you explain why?! :)

ycho:
With -s 1, at some QIndex (including the smallest one, 80), it very^100 slightly reduces the encoding time.

tdaede:
How small of changes are you trying to measure with AWCY here? If it's less than 5% then it's probably not accurate.

ycho:
There you go, I believe it was less than 0.5%! :)

tdaede:
Regardless of speed, this looks way cleaner.

ycho:
@tdaede, thanks for +1!

ycho:
@rzumer? Further comments?

PR #894, opened 2019-01-30
add updates for dependencies for default build

basilgohar:
Added cargo and mpv to the list of dependencies to install and build on Linux, as the default build.sh assumes their presence.  Also added the instructions for installing these dependencies on Fedora.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,16 @@ git submodule update --init
 
 This is also required every time you switch branches or pull a submodule change.
 
-In order to build, test and link to the codec on UNIX, you need Perl, NASM, CMake, Clang and pkg-config. To install this on Ubuntu or Linux Mint, run:
+In order to build, test and link to the codec on UNIX, you need Cargo, Perl, NASM, CMake, Clang, pkg-config and mpv. To install this on Ubuntu or Linux Mint, run:
 
 ```
-sudo apt install perl nasm cmake clang pkg-config
+sudo apt install cargo perl nasm cmake clang pkg-config mpv
+```
+
+To install this on Fedora, run:
+
+```
+sudo dnf install cargo perl nasm cmake clang pkg-config mpv
 ```
 
 On Windows, pkg-config is not required. A Perl distribution such as Strawberry Perl, CMake, and a NASM binary in your system PATH are required.


moisesmcardona:
A few days ago I was unable to compile rav1e because the cargo version in the repo was 1.30.0 and gave an error that Clippy is experimental. Hence, I had to uninstall cargo with `apt purge cargo` and install rust from their sh script, which installed version 1.32.0.

Just writing this here just in case the readme need to be updated.

rzumer:
I don't think that mpv is actually needed. A call is added at the tail of `build.sh` for testing, but that sample script is not required for building or testing.

basilgohar:
I understand. I was just going for completion to make it as easy and
unjarring for the newcomers to get started with it.

I could add a new, small section mentioning that it's optional, but I felt
that was a slippery slope of rewriting more-and-more of the README.

tdaede:
Yeah, I'll take this update but it should drop mpv as it's not needed to "build or test".

basilgohar:
> Yeah, I'll take this update but it should drop mpv as it's not needed to "build or test".

I understand, but then I would recommend that build.sh have the call to `mpv` removed, else a user will encounter an error if they do not have `mpv` installed.  I don't think that is, otherwise, a good user experience.  Should I add that to this pull request or make another one?

tdaede:
Whoops, I guess our readme doesn't even reference build.sh anymore.

Yeah, can you either remove the mpv call or just make it || true?

basilgohar:
I feel like taking a step back and rethinking this.  I may delete this pull request and approach this from a wider angle, if it's okay with you.  I want to fix the README a bit more, maybe be a bit more bold, and then also fix the build.sh script.

rzumer:
@basilgohar Do you still want this PR to stay opened? If you are not planning on updating it soon, I would like to close it in the interest of limiting the number of open PRs that are inactive. You are welcome to submit a broader set of changes in a new PR any time if you are still interested.

basilgohar:
> @basilgohar Do you want this PR to stay opened? If you are not planning on updating it soon, I would like to close it in the interest of limiting the number of open PRs that are inactive. You are welcome to submit a broader set of changes in a new PR any time if you are still interested.

@rzumer Thanks for bringing this back up.  I'll close it and reopen with the broader changers if I feel so inclined.

PR #896, opened 2019-01-30
Bump the dav1d and bindgen versions again

lu-zero:
I tested it with fedora 29 and seems working fine, could you please try and confirm?


tdaede:
Seems to work now.

PR #900, opened 2019-02-02
cdef: simplify best_cost/dir search

tmatth:


src/cdef.rs
```diff
@@ -84,12 +82,7 @@ fn cdef_find_dir(img: &[u16], stride: usize, var: &mut i32, coeff_shift: i32) ->
         }
         i+=2;
     }
-    for i in 0..8 {
-        if cost[i] > best_cost {
-            best_cost = cost[i];
-            best_dir = i;
-        }
-    }
+    let (best_cost, best_dir) = cost.iter().enumerate().map(|(x, y)| (y, x)).max().unwrap();
```
lu-zero: You should use [max_by_key](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max_by_key) the `max()` for a tuple is lexicographical.

PR #902, opened 2019-02-03
Do not depend on y4m in the rav1e library

lu-zero:
y4m should be used only by the cli tools.


shssoichiro:
It seems like this PR reverted some of the changes from #895, particularly using the `Decoder` trait within the encode/decode tests. Was this intentional?

lu-zero:
Yes, the decoder abstraction should live only for the cli, the library shouldn't need it at all

PR #903, opened 2019-02-03
Cleanups

lu-zero:
Try to remove some lingering bits of code.


lu-zero:
Ok, the patch was completely wrong :)

PR #904, opened 2019-02-03
Fix warnings caused by non-idiomatic returns

vladimir-kazakov:
This PR enables Clippy's [let_and_return](https://rust-lang.github.io/rust-clippy/master/index.html#let_and_return) and [needless_return](https://rust-lang.github.io/rust-clippy/master/index.html#needless_return) lints, and fixes related warnings.

P.S. I also noticed something strange, which is not related to this PR, but I think it's worth mentioning. Some blocks of code are marked with `#[cfg(debug)]` attributes ([for example](https://github.com/xiph/rav1e/blob/31bf6a6c91d323402ba81aa3a75862ca2a3c49cc/src/context.rs#L1810)). The intent is clear - to have additional functionality in debug builds, which must not present in the release ones. The problem is that there's no such configuration in Rust. So, currently, all code that is marked with such attribute is completely ignored. The closest value of this attribute with the desired behavior is [debug_assertions](https://doc.rust-lang.org/reference/conditional-compilation.html#debug_assertions). Is there really something wrong, or there's a way to make `#[cfg(debug)]` working?

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1812,6 +1812,7 @@ pub struct ContextWriter {
 }
 
 impl ContextWriter {
+  #[allow(clippy::let_and_return)]
   pub fn new(fc: CDFContext, bc: BlockContext) -> Self {
     #[allow(unused_mut)]
     let mut cw = ContextWriter {
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,4 +42,4 @@ jobs:
          - cargo doc --verbose
          - rustup component add clippy
          - cargo clippy --version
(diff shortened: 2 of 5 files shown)


src/context.rs
```diff
@@ -1811,6 +1811,7 @@ pub struct ContextWriter {
   fc_map: Option<FieldMap> // For debugging purposes
 }
 
+#[allow(clippy::let_and_return)]
```
shssoichiro: Would it make more sense to move this down one line so it applies to only the `new` method, instead of the whole `impl ContextWriter`?

vladimir-kazakov: Yes, you're right. Good catch. Fixed.

PR #912, opened 2019-02-04
Rename Context field frames_to_encode to limit.

negge:


rzumer:
`frames_to_be_coded` is not exactly an elegant name, but I think it is easier to understand at a glance compared to `limit`.

negge:
> `frames_to_be_coded` is not exactly an elegant name, but I think it is easier to understand at a glance compared to `limit`. Is the name changed to match the spec? If not, maybe `frame_limit` would be good.

The name was changed to match the command line parameter:
```
    -l, --limit <LIMIT>                          Maximum number of frames to encode [default: 0]
```

PR #916, opened 2019-02-04
cdef: use left-most direction in case of ties

tmatth:
Fixes desync

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -83,7 +83,8 @@ fn cdef_find_dir(img: &[u16], stride: usize, var: &mut i32, coeff_shift: i32) ->
         i+=2;
     }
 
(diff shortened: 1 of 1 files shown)


src/cdef.rs
```diff
@@ -83,7 +83,8 @@ fn cdef_find_dir(img: &[u16], stride: usize, var: &mut i32, coeff_shift: i32) ->
         i+=2;
     }
 
-    let (best_dir, best_cost) = cost.iter().enumerate().max_by_key(|(_i, &v)| v).unwrap();
+    // In case of a tie, the first direction must be selected.
+    let (best_dir, best_cost) = cost.iter().enumerate().max_by_key(|&(i, v)| (v, -(i as isize))).unwrap();
```
lu-zero: I let @tdaede decide :)

PR #919, opened 2019-02-04
Fix infinite loop for very short clips with frame reordering

### LINKED ISSUE #890: Never Ending task with a very low frame count .y4m file
Hi everyone,

I've been encoding videos in my own distributed computing project. I've been doing this by splitting a video into 1-second segments with ffmpeg and then submitting each clip to the project.

While I'm able to encode most of my clips, this one which seems to have just a couple of frames is not working: https://drive.google.com/file/d/1dvewc7SwdE7BRFQUa2VsvcredixJuJ9_/view?usp=sharing

rav1e never seems to finish but uses 100% of the CPU thread, so it is doing something.

Parameters used are:
`--quantizer 175 -s 0 --low_latency false`

Is it stuck in a loop which seems to never finish and doesn't exit? I'm using rav1e 774b0d1, which is the latest commit as of this issue.

std output just stays as follows:
```
3840x2160 @ 30000/1001 fps

encoded 1 frames, 0.00 fps, 8493.22 Kb/s                    
```

shssoichiro:
Fixes #890. Also fixes the tests to test frame reordering and lookahead correctly.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -506,6 +506,11 @@ impl Context {
       }
     }
 
+    match self.frame_q.get(&fi.number) {
+      Some(Some(_)) => {},
+      _ => { return Err(fi); }
+    }
+
     // Now that we know the frame number, look up the correct frame type
     let frame_type = self.determine_frame_type(fi.number);
     if frame_type == FrameType::KEY {
--- a/src/test_encode_decode_aom.rs
+++ b/src/test_encode_decode_aom.rs
@@ -33,6 +33,18 @@ fn fill_frame(ra: &mut ChaChaRng, frame: &mut Frame) {
   }
 }
 
+fn read_frame_batch(ctx: &mut Context, ra: &mut ChaChaRng) {
+  while ctx.needs_more_lookahead() {
+    let mut input = ctx.new_frame();
+    fill_frame(ra, Arc::get_mut(&mut input).unwrap());
+
+    let _ = ctx.send_frame(Some(input));
+  }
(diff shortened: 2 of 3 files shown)


src/test_encode_decode_aom.rs
```diff
@@ -207,6 +222,18 @@ fn reordering() {
   }
 }
 
+#[test]
+fn issue_890() {
```
lu-zero: Might be nicer to have a self-descriptive name and a comment referring to the issue, as @rzumer pointed out

src/test_encode_decode_aom.rs
```diff
@@ -33,6 +33,21 @@ fn fill_frame(ra: &mut ChaChaRng, frame: &mut Frame) {
   }
 }
 
+fn read_frame_batch(ctx: &mut Context, ra: &mut ChaChaRng) {
+  loop {
```
lu-zero: maybe using a plain while loop could be more straightforward.

src/api.rs
```diff
@@ -506,6 +506,11 @@ impl Context {
       }
     }
 
+    match self.frame_q.get(&fi.number) {
```
lu-zero: This could be something along the lines of
``` rust
if self.frame_q.get().map_or(false, |v| v.is_some()) { return Err(fi) }
```
It is obviously debatable what's more readable :)

shssoichiro: This caused the infinite loop to return. It could probably be fixed to have the same logic as the match statement, but given that it would require extra effort to do so, I think the `match` is clearer at showing the logic we want.

PR #920, opened 2019-02-04
Color preparations

kodabb:
A couple of minor things related to the big set later.


src/api.rs
```diff
@@ -463,13 +463,16 @@ impl Context {
 
   fn get_frame_properties(&mut self, idx: u64) -> Result<FrameInvariants, FrameInvariants> {
     if idx == 0 {
+      let mut seq = Sequence::new(&self.config.video_info);
```
tdaede: Would slightly prefer this as a parameter to new(), but can be changed separately.

PR #921, opened 2019-02-04
Color and hdr

kodabb:
* add color range support
* add hdr metadata support

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ repl = ["rustyline", "binaries"]
 comparative_bench = ["aom"]
 decode_test = ["bindgen", "aom"]
 decode_test_dav1d = ["dav1d-sys"]
-binaries = ["y4m", "clap"]
+binaries = ["y4m", "clap", "scan_fmt"]
 default = ["binaries", "nasm"]
 aom = ["cmake"]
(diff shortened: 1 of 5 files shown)


kodabb:
The input syntax could be maybe implemented a little better.

I'm not sure about exposing floats through the api: while they are ok on the cli, they might be some random obscure configurations that cause a small precision error due to the fixed point conversion. x265 does not have this problem because requires fields to be premultiplied to a known constant, and libvpx requires to use fixed point values as input (iirc). Libaom does not implement this part but it will likely behave as libvpx when it does, I am open to ideas

tdaede:
please rebase

Cargo.toml
```diff
@@ -30,6 +30,7 @@ quote = "^0.6.10" # hack for proc-macro-hack
 num-traits = "0.2"
 paste = "0.1"
 dav1d-sys = { version = "0.1.2", optional = true }
+scan_fmt = "0.1.3"
```
tdaede: @lu-zero this is only needed while building binaries right?

kodabb:
updated with @rzumer comment's, i'll check about `scan_fmt` and see what to do wrt fixed points

kodabb:
updated with the fixed point change and scan_fmt fix

PR #922, opened 2019-02-04
Simplify get_frame_properties/set_frame_properties

lu-zero:
Follow the same pattern used in FrameInvariants::new_inter_frame()

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -450,18 +450,14 @@ impl Context {
     cmp::min(next_detected.unwrap(), next_limit)
   }
 
-  fn set_frame_properties(&mut self, idx: u64) -> Result<(), ()> {
-    let props = self.get_frame_properties(idx);
-    let result = props.as_ref().map(|_| ()).map_err(|_| ());
-    match props {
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -450,18 +450,14 @@ impl Context {
...
-      Ok(props) | Err(props) => {
-        self.frame_data.insert(idx, props);
-      }
-    }
-    result
+  fn set_frame_properties(&mut self, idx: u64) -> bool {
+    let (fi, success) = self.get_frame_properties(idx);
+    self.frame_data.insert(idx, fi);
+
+    success
```
tdaede: Naming this "success" is weird as it really implies it's error handling and that the existing code would make more sense. Isn't this more like an EOF flag?

shssoichiro: There are situations other than EOF that could cause this flag to be false, mostly around frame reordering. Given that, I'm not sure what a better name for this variable might be.

lu-zero: We decided for `end_of_subgop`, next set will be renaming the functions to something a bit more descriptive.

PR #923, opened 2019-02-04
Consistently use 2-spaces indent

smarter:
Instead of a random mix of 2 spaces and 4 spaces.


KyleSiefring:
There are two unrelated patches here.

PR #925, opened 2019-02-05
cdef: refactor max element search into helper function with tests

tmatth:


src/cdef.rs
```diff
@@ -30,6 +30,13 @@ const CDEF_SEC_STRENGTHS: u8 = 4;
 // the max. */
 const CDEF_DIV_TABLE: [i32; 9] = [ 0, 840, 420, 280, 210, 168, 140, 120, 105 ];
 
+#[inline]
+fn find_max_element(elems: &[i32]) -> (usize, i32) {
```
smarter: Add a documentation comment explaining what this function does, and especially mentioning that it assumes that `elems` is not empty as a precondition. And maybe use a more descriptive name like `first_max_element`

PR #928, opened 2019-02-05
Fix failing build

shssoichiro:
PRs #912, #919, and #922 contained changes which conflicted with one
another, and needed to be reconciled in order to get the build to pass
with all of these changes together


shssoichiro:
There's also an option in the project settings that can avoid this in the future by requiring branches be up to date with master before merging, although it comes at the cost of having to wait for the build to complete again before merging subsequent PRs. It's probably worth discussing whether or not to enable it.

Issue #910, opened 2019-02-04
Unify/simplify encoder configuration


kodabb:
Right now there are two public fields that expose some part of the encoder configuration, `VideoInfo and `EncoderConfig`. Internally they get translated in `FrameInvariants` and `Sequence` for fixed parameters and gop-valid parameters.

This can create some unfun interactions between user-facing structures and internal ones, like the fact that `color_description` is set in `EncoderConfig`, but used in `Sequence` and right now nothing can initialize `Sequence.color_description`, since `Sequence` uses a `VideoInfo` structure.

Ideally I wouldn't mind having a single `config` variable where everything can be accessed, but if we want to limit scope depending on the validity of the parameter we could do that too.

PR #934, opened 2019-02-05
Simple config

kodabb:
as mentioned in #910 this should simplify the api a bit and other portions of the codebase

relevant commit message:
>    VideoDetails is mainly used while decoding y4m, but it is carried
>    around in the configuration, initializing various internal structures.
>
>   Having all encoding-related fields in a single structure simplifies
>  setting up and configuring the encoder, as well as offering
> an easier-to-use and more expandable API.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -45,12 +45,14 @@ fn write_b_bench(b: &mut Bencher, tx_size: TxSize, qindex: usize) {
     aom_dsp_rtcd();
   }
   let config = EncoderConfig {
+    width: 1024,
+    height: 1024,
     quantizer: qindex,
     speed_settings: SpeedSettings::from_preset(10),
     ..Default::default()
   };
   let sequence = Sequence::new(&Default::default());
-  let mut fi = FrameInvariants::new(1024, 1024, config, sequence);
+  let mut fi = FrameInvariants::new(config, sequence);
   let mut w = ec::WriterEncoder::new();
   let fc = CDFContext::new(fi.base_q_idx);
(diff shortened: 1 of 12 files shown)


benches/bench.rs
```diff
@@ -112,10 +114,15 @@ fn cdef_frame(c: &mut Criterion) {
 }
 
 fn cdef_frame_bench(b: &mut Bencher, w: usize, h: usize) {
-  let config =
-    EncoderConfig { quantizer: 100, speed_settings: SpeedSettings::from_preset(10), ..Default::default() };
+  let config = EncoderConfig {
+    width: w,
```
tdaede: Maybe name the parameters width and height so you can just do width, height, here?

src/api.rs
```diff
@@ -22,31 +22,6 @@ use std::collections::BTreeSet;
 
 const LOOKAHEAD_FRAMES: u64 = 10;
 
-#[derive(Debug, Clone, Copy)]
```
tdaede: @lu-zero is this going to break crav1e?

kodabb: yes but i'll send a pr to fix it

src/encoder.rs
```diff
@@ -208,7 +209,8 @@ impl ChromaSampling {
     match self {
       Cs420 => (2, 2),
       Cs422 => (2, 1),
-      Cs444 => (1, 1)
+      Cs444 => (1, 1),
+      Cs400 => (1, 1),
```
rzumer: This should be (2, 2) based on the spec (pg 119).

Issue #935, opened 2019-02-06
AppVeyor access denied


EwoutH:
For regular users, https://ci.appveyor.com/project/xiph/rav1e displays the message `Project not found or access denied.`

![screenshot_384](https://user-images.githubusercontent.com/15776622/52327652-24a08b80-29ed-11e9-9407-c87491746fab.png)

tdaede:
I can't figure out how to make this public :(

rzumer:
The history page with artifacts is accessible, so that seems normal to me.

EwoutH:
@tdaede No problem, let’s not waste too much time on it, PR #937 fixes it for now.

PR #937, opened 2019-02-06
Update AppVeyor link in README.md

EwoutH:
Switch AppVeyor badge link to https://ci.appveyor.com/project/tdaede/rav1e/history

Migrates #935


rzumer:
I think only `xiph` needs to be changed to `tdaede`.

EwoutH:
> I think only `xiph` needs to be changed to `tdaede`.

I think landing on the history page is better, the current page has the risk of being a failing PR.

Issue #938, opened 2019-02-06
CLI rework


lu-zero:
- [x] remove the repl
- [x] move the logic in `common.rs` in `rav1e.rs`
- [ ] reorder the cli options and group `I/O`-specific options and encoder-specific options

PR #939, opened 2019-02-06
Remove rav1repl command line encoding.

negge:
Part of #938.


lu-zero:
Looks fine to me.

rzumer:
Why is there code moved from `common.rs` to `rav1e.rs`? Is `common.rs` even needed if there is only the `rav1e` binary?

lu-zero:
`common.rs` could go away or we could actually spin it to `muxer.rs` assuming we'll have time to add the mkv input and output this week.

PR #940, opened 2019-02-06
Update the compilation profiles

lu-zero:
- Make the dev profile a little more optimized
- Make the other profiles have the debug information

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ jobs:
         script:
          - cargo install cargo-kcov
          - kcov --version
-         - cargo build --release --verbose
+         - cargo build
          - cargo kcov -v --coveralls --features=decode_test -- --verify --exclude-pattern=$HOME/.cargo,aom_build,.h,test
       - name: "Tests"
         script: cargo test --verbose --release --features=decode_test -- --ignored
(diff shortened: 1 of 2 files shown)


rzumer:
Seems good, but I just have a few questions:

* What is the difference in run time for debug mode?
* What is the rationale for adding debug information in release mode?
* Does setting debug to on in release mode impact performance?

lu-zero:
>     * What is the difference in run time for debug mode?

None, for the platforms we can target (ELF, MACH-O, PE). The debug information lives in a separate sections and could even be split in a separate file to keep the actual binary lean.

>     * What is the rationale for adding debug information in release mode?
It can be used to see what is going on on a per-function granularity and it is much better than landing in the disasm world at least on mac :)
>     * Does setting debug to on in release mode impact performance?
Luckily not, our drives are fast enough that the additional size is negligible :)

tdaede:
If debug is fast enough, we wouldn't need symbols in release. But fine for now. Not having symbols is super obnoxious.

lu-zero:
Let see if the code coverage works on release this way.

lu-zero:
The build time seems growing enough to have travis timeouts though...

lu-zero:
It needs more experimentation, the time usage grow too much and using dav1d requires a new suppression map.

PR #947, opened 2019-02-07
Support degraded 4:2:2/4:4:4 encoding

rzumer:
This reverts an erroneous commit that caused intra frames in 4:4:4 to no longer be decodable, and fixes some additional issues. 4:4:4 and 4:2:2 encoding now work down to `-s 0` with keyframes only (enforced on the API end) and restoration disabled. CfL prediction is also disabled based on a comment from @barrbrain.


tdaede:
from a readme point of view "limited" might sound better than "degraded"

rzumer:
@tdaede Agreed, changed it.

PR #951, opened 2019-02-07
Fully use tx-domain distortion (when chosen) with topdown partition search

ycho:
Fully use tx-domain distortion with topdown partition search
    
    The 'needs_rec' parameter for one of the two rdo_mode_decision() calls by
    topdown parition search has been passed with wrong value.
    It should be false, since re-encoding step follows once rdo_mode_decision()
    decided the mode.
    In fact, we don't need 'needs_rec' parameter in rdo_mode_decision(), so remove it.
    
    This patch will cause loss of coding efficiency (0.1250% at speed 1)
    with a small speedup.
    
    However, if there is more than one intra prediction blocks in a
    partition, i.e. tx_size < partition size, rdo_mode_decision() will need
    generate reconstrtucted pixels for intra prediction modes. Sine tx
    partition is not provided by rav1e yet, this condition can be
    implemented in the future.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -2433,7 +2433,7 @@ fn encode_partition_bottomup(
     };
     let spmvs = &pmvs[pmv_idx];
 
-    let mode_decision = rdo_mode_decision(fi, fs, cw, bsize, bo, spmvs, false);
+    let mode_decision = rdo_mode_decision(fi, fs, cw, bsize, bo, spmvs);
 
     rd_cost = mode_decision.rd_cost + cost;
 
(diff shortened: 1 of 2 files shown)


ycho:
[AWCY -s 3 result will come soon](https://beta.arewecompressedyet.com/?job=master-a45a2ac3b4cd035ca78febff6e3d56e83acaa486&job=test_fully_use_tx_dist_%23951%402019-02-08T00%3A47%3A14.601Z)
[AWCY -s 1 result also come later](https://beta.arewecompressedyet.com/?job=master-a45a2a_s1%402019-02-08T00%3A47%3A51.327Z&job=test_fully_use_tx_dist_%23951_s1%402019-02-08T00%3A48%3A15.052Z)

tdaede:
OK. Do you want to still leave the flag for when we add tx_size < partition_size? Or do you want to implement that by reencoding more at the end?

rzumer:
@tdaede If I understand correctly it is recalculated in `rdo_mode_decision()` based on the difference in block and transform sizes.

tdaede:
Ah I see, I read it as that was going to be added later. (line 505 accomplishes it)

rzumer:
Reopening to trigger a new Travis build, since it seems stuck.

ycho:
Thanks for reviews and re-trigering CI!

PR #952, opened 2019-02-07
Make receive_packet to report we are at the end of a flushing command

lu-zero:


src/bin/rav1e.rs
```diff
@@ -29,8 +29,6 @@ use decoder::Decoder;
 use decoder::VideoDetails;
 
 fn read_frame_batch<D: Decoder>(ctx: &mut Context, decoder: &mut D, video_info: VideoDetails) {
-  loop {
-    if ctx.needs_more_lookahead() {
```
shssoichiro: Doesn't this break the frame readahead? Not that we use lookahead for anything yet, but we likely will in the future.

lu-zero: Ideally we want to not leak the lookahead logic to the outside world.

A loop such as 
``` rust
loop {
  let f = get_frame();
  ctx.send_frame(&f);

  while let Ok(pkt) = ctx.receive_packet() {
    ...
  }
}

Should work just fine.

PR #953, opened 2019-02-08
Include `build.rs` in the set of files

jsgf:
Without it, a vendored version of the crate won't build.


jsgf:
Yeah, this resolves https://github.com/alexcrichton/cargo-vendor/issues/185

PR #955, opened 2019-02-08
Refactor api

lu-zero:
Some experiments


lu-zero:
https://github.com/xiph/rav1e/pull/1117 is the current approach to address the problem

PR #958, opened 2019-02-08
Write temporal delimiters only at the beginning of a packet.

### LINKED ISSUE #954: Temporal Delimiter OBUs are being inserted before every frame
TD OBUs are meant to be present at the beginning of a Temporal Unit, precisely to delimit them, and not before every frame, showable or not, within a given Temporal Unit. All these extra TDs are pointless overhead in the bitstream.

A simple "rav1e INPUT.y4m --output OUTPUT.ivf" seems to be enough to reproduce this. In my tests i used https://0x0.st/sfRh.y4m.

tdaede:
Fixes #954

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -339,6 +339,9 @@ impl Config {
       aom_dsp_rtcd();
     }
 
+    // initialize with temporal delimiter
+    let packet_data = TEMPORAL_DELIMITER.to_vec();
+
     Context {
       frame_count: 0,
       limit: 0,
@@ -347,7 +350,7 @@ impl Config {
       frame_q: BTreeMap::new(),
       frame_data: BTreeMap::new(),
       keyframes: BTreeSet::new(),
-      packet_data: Vec::new(),
+      packet_data,
       segment_start_idx: 0,
       segment_start_frame: 0,
       keyframe_detector: SceneChangeDetector::new(self.enc.bit_depth),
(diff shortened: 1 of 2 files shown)


src/api.rs
```diff
@@ -339,6 +339,9 @@ impl Config {
       aom_dsp_rtcd();
     }
 
+    let mut packet_data = Vec::new();
+    write_temporal_delimiter(&mut packet_data).unwrap();
```
negge: Please fail gracefully.

src/api.rs
```diff
@@ -593,7 +596,11 @@ impl Context {
 
         let mut fs = FrameState::new(fi);
 
-        let data = encode_frame(fi, &mut fs);
+        let sef_data = encode_frame(fi, &mut fs);
+        self.packet_data.extend(sef_data);
+        let data = self.packet_data.clone();
+        self.packet_data.clear();
+        write_temporal_delimiter(&mut self.packet_data).unwrap();
```
negge: Same thing here.

src/api.rs
```diff
@@ -635,6 +642,7 @@ impl Context {
             if fi.show_frame {
               let data = self.packet_data.clone();
               self.packet_data.clear();
+              write_temporal_delimiter(&mut self.packet_data).unwrap();
```
negge: Ditto.

src/api.rs
tdaede: This one is to something I just alloc'd so I know it can't fail, can I leave it?

src/encoder.rs
```diff
@@ -1729,19 +1729,19 @@ fn aom_uleb_encode(mut value: u64, coded_value: &mut [u8]) -> usize {
   leb_size
 }
 
+pub fn write_temporal_delimiter(
+  packet: &mut dyn io::Write
+) -> io::Result<()> {
+  packet.write(&[0x12, 0x00])?;
```
negge: Can you use the OBU_Type::OBU_TEMPORAL_DELIMITER enum?

src/api.rs
```diff
@@ -339,6 +339,9 @@ impl Config {
       aom_dsp_rtcd();
     }
 
+    // initialize with temporal delimiter
+    let packet_data = vec![0x12, 0x00];
```
negge: You could hide this inside a constant temporal OBU header somewhere, so you define it once and use it twice.

src/encoder.rs
tdaede: Yes, but not for the second byte which is the length.

PR #961, opened 2019-02-08
Remove the public Context::new_frame() API call.

negge:
The only place this is used is by the test_encode_decode_aom and
 test_encode_decode_dav1d client implementations using the public API.
Instead of passing the frame width and height to the client side
 implementation of read_frame_batch() through the public API, it simply
 passes the information as a parameter.

### DIFF
--- a/src/test_encode_decode_aom.rs
+++ b/src/test_encode_decode_aom.rs
@@ -20,6 +20,9 @@ use std::{mem, ptr, slice};
 use std::collections::VecDeque;
 use std::ffi::CStr;
 use std::sync::Arc;
+use util::Fixed;
+
+// TODO: merge this file with test_encode_decode_dav1d.rs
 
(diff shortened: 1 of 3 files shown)


src/test_encode_decode_aom.rs
```diff
@@ -324,7 +329,7 @@ fn encode_decode(
 
   let mut rec_fifo = VecDeque::new();
   for _ in 0..limit {
-    read_frame_batch(&mut ctx, &mut ra);
+    read_frame_batch(&mut ctx, &mut ra, w, h, ChromaSampling::Cs420);
```
rzumer: I think you need to rebase this onto the current master, because `encode_decode()` on both the AOM and dav1d ends now take a `chroma_sampling` argument and this will need to match.

lu-zero:
I'd rather not remove that function.

tdaede:
OK.

PR #962, opened 2019-02-08
Migrate to Rust 2018

### LINKED ISSUE #957: Migrate to Rust 2018
I tested the automated migration with Cargo shortly after release, but it did not work at all, so it will take some work to get it working.

rzumer:
Closes #957.

The original Rust 2018 migration page was advising to run `cargo fix`, which did not perform migration correctly, but `cargo fix --edition` did the trick.


tdaede:
Does compilation break between commits? If so it should just be 1.

rzumer:
@tdaede Squashed.

PR #963, opened 2019-02-08
Use named chroma_sampling_period variables.

negge:


src/encoder.rs
```diff
@@ -45,21 +45,23 @@ const FRAME_MARGIN: usize = 16 + SUBPEL_FILTER_SIZE;
 
 impl Frame {
   pub fn new(width: usize, height: usize, chroma_sampling: ChromaSampling) -> Frame {
-    let chroma_sampling_period = chroma_sampling.sampling_period();
+    let (chroma_sampling_period_h, chroma_sampling_period_v) =
+     chroma_sampling.sampling_period();
+    let luma_padding = MAX_SB_SIZE + FRAME_MARGIN;
     let (chroma_width, chroma_height, chroma_padding, chroma_xdec, chroma_ydec) = (
```
smarter: I don't see the point of doing let (a, b, ...) = { x, y, ...}; instead of let a = x; let b = y; ...

rzumer: It is more compact and those fields are all related to chroma, so it seems fine to me. I don't have a strong opinion on it though.

smarter: I think it makes it easier to match the name with the value.

src/encoder.rs
```diff
@@ -45,21 +45,26 @@ const FRAME_MARGIN: usize = 16 + SUBPEL_FILTER_SIZE;
 
 impl Frame {
   pub fn new(width: usize, height: usize, chroma_sampling: ChromaSampling) -> Frame {
-    let chroma_sampling_period = chroma_sampling.sampling_period();
+    let (luma_width, luma_height, luma_padding) = (
```
smarter: Plz no.

src/encoder.rs
rzumer: Then I would suggest grouping e.g. `chroma_width` and `chroma_height`, `chroma_xdec` and `chroma_ydec`, etc. which is easy to read and also saves space.

PR #965, opened 2019-02-08
Add submit_awcy tool

tdaede:


tools/submit_awcy.py
```diff
@@ -0,0 +1,71 @@
...
+        if retcode:
+            cmd = kwargs.get("args")
+            if cmd is None:
+                cmd = popenargs[0]
+            raise subprocess.CalledProcessError(retcode, cmd)
+        return output
+    subprocess.check_output = f
+
+if 'RAV1E_ROOT' not in os.environ:
+    print(GetTime(), "Please specify the DAALA_ROOT environment variable to use this tool.")
```
negge: Print RAV1E_ROOT instead of DAALA_ROOT.

tdaede: dropped cd

PR #966, opened 2019-02-08
Switch out compute_rd_cost functions for compute_distortion

tdaede:


src/rdo.rs
```diff
@@ -159,13 +159,12 @@ pub fn get_lambda_sqrt(fi: &FrameInvariants) -> f64 {
   q0 * (std::f64::consts::LN_2 / 6.0).sqrt()
 }
 
-// Compute the rate-distortion cost for an encode
-fn compute_rd_cost(
+// Compute the distortion cost for an encode
```
smarter: We should only use the word "cost" to refer to the RD cost. "for an encode" in this comment doesn't mean anything, so no comment at all would be better. (And even better would be a comment explaining what all the parameters of this function do).

src/rdo.rs
```diff
@@ -159,13 +159,12 @@ pub fn get_lambda_sqrt(fi: &FrameInvariants) -> f64 {
...
+fn compute_distortion(
   fi: &FrameInvariants, fs: &FrameState, w_y: usize, h_y: usize,
-  is_chroma_block: bool, bo: &BlockOffset, bit_cost: u32,
+  is_chroma_block: bool, bo: &BlockOffset,
   luma_only: bool
-) -> f64 {
-  let lambda = get_lambda(fi);
+) -> u64 {
 
   // Compute distortion
```
smarter: This comment is now useless.

src/rdo.rs
```diff
@@ -159,13 +159,12 @@ pub fn get_lambda_sqrt(fi: &FrameInvariants) -> f64 {
...
   let po = bo.plane_offset(&fs.input.planes[0].cfg);
```
smarter: The if/elseif/else block starting on the next line should really be replaced by a match.

src/rdo.rs
```diff
@@ -223,22 +222,23 @@ fn compute_rd_cost(
       }
     };
   }
-  // Compute rate
-  let rate = (bit_cost as f64) / ((1 << OD_BITRES) as f64);
+  distortion
+}
 
+fn compute_rd(fi: &FrameInvariants, bit_cost: u32, distortion: u64) -> f64 {
```
smarter: cost should only be used to refer to the RD cost:
```suggestion
fn compute_rd(fi: &FrameInvariants, rate: u32, distortion: u64) -> f64 {
```

src/rdo.rs
tdaede: Yes. I renamed some of them to rate but missed some.

src/rdo.rs
```diff
@@ -280,10 +273,13 @@ fn compute_tx_rd_cost(
       }
     }
   }
-  // Compute rate
-  let rate = (bit_cost as f64) / ((1 << OD_BITRES) as f64);
+  distortion
+}
 
-  (distortion as f64) + lambda * rate
+fn compute_rd(fi: &FrameInvariants, rate: u32, distortion: u64) -> f64 {
```
ycho: Ok, let be brave to ask full names for someone's sakes (including me)
compute_rd --> compute_rd_cost

src/rdo.rs
```diff
@@ -280,10 +273,13 @@ fn compute_tx_rd_cost(
...
+  let lambda = get_lambda(fi);
+  let rate_bits = (rate as f64) / ((1 << OD_BITRES) as f64);
```
ycho: rate_bits --> rate_in_bits

src/rdo.rs
```diff
@@ -524,32 +520,31 @@ pub fn rdo_mode_decision(fi: &FrameInvariants, fs: &mut FrameState,
...
             w,
             h,
             is_chroma_block,
             bo,
-            cost,
             false
           )
         };
+        let rd = compute_rd(fi, rate, distortion);
         if rd < best.rd {
```
ycho: I wanted : best.rd --> best.rd_cost, but it would be out of the scope of this patch....

PR #968, opened 2019-02-08
Only motion compensate once for each tx type search

tdaede:


tdaede:
Gonna hold off on this for the Pixel type stuff, it'll be easier once that's landed.

PR #969, opened 2019-02-08
Disable coveralls to speed up ci.

tdaede:


.travis.yml
```diff
@@ -34,7 +34,6 @@ jobs:
          - cargo install cargo-kcov
          - kcov --version
```
rzumer: What about these two lines?

lu-zero: They should go as well as the kcov installation above.

lu-zero:
We can do w/out kcov, hopefully something like https://github.com/xd009642/tarpaulin will get ready for use later in the year.

EwoutH:
Maybe we could do this a little more refined? Like only running Coveralls when merging into master?

lu-zero:
I'd give tarpaulin a go today, if it works could you please prepare a nightly-only task @EwoutH ?

tdaede:
Closing in favor of tarpaulin.

PR #973, opened 2019-02-08
Use Rust 2018 edition idioms

shssoichiro:
Automated by running `cargo fix --edition-idioms`


build.rs
```diff
@@ -3,8 +3,6 @@
 #[cfg(feature = "aom")]
 extern crate cmake;
 #[cfg(unix)]
-extern crate pkg_config;
-#[cfg(unix)]
```
lu-zero: This looks spurious

PR #979, opened 2019-02-10
Output JSON stats file with frame types

### LINKED ISSUE #898: Add keyframe numbers to (CLI) ouput
Would be nice to get a CLI option that outputs the frame-numbers of the keyframes. A comma-seperated string would probably be most usefull (`1,82,196,245` i.e.).

Then you can do a turbo turbo pass, (speed 10, quarter resolution, etc.) to get the optimal keyframe placement, which can be used to split the video at the correct points for optimal encoding in chunks.

Maybe I have some time to implement this later, for now I will leave it here as an issue.

shssoichiro:
More data will likely be added to this stats file in the future.

Adds in `--pass 1` or `--pass 2` CLI options to prepare for two-pass encoding.
Currently the only difference between these, or omitting the option (which assumes
one-pass encoding) is that using `--pass 1` outputs the stats file. This will be added to
and optimized later.

Closes #898

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ autobins = false
 comparative_bench = ["aom"]
 decode_test = ["bindgen", "aom"]
 decode_test_dav1d = ["dav1d-sys"]
-binaries = ["ivf", "y4m", "clap", "scan_fmt"]
+binaries = ["ivf", "y4m", "clap", "scan_fmt", "serde_json"]
 default = ["binaries", "nasm"]
 aom = ["cmake"]
 nasm = ["nasm-rs"]
@@ -29,6 +29,9 @@ syn = "^0.15.20"
 quote = "^0.6.10" # hack for proc-macro-hack
 num-traits = "0.2"
 paste = "0.1"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = { version = "1.0", optional = true }
 dav1d-sys = { version = "0.1.2", optional = true }
 scan_fmt = { version = "0.1.3", optional = true }
 ivf = { path = "ivf/", optional = true }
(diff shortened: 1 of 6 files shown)


tdaede:
Can you give an example of what the JSON looks like in the commit message? Or even better, some docs on it.

This puts the serialization in the rav1e binary. That's an OK place for it, but maybe it belongs in the library instead?

Cargo.toml
```diff
@@ -28,6 +28,9 @@ syn = "^0.15.20"
 quote = "^0.6.10" # hack for proc-macro-hack
 num-traits = "0.2"
 paste = "0.1"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
```
lu-zero: I'd make it optional.

src/api.rs
```diff
@@ -602,37 +619,15 @@ impl Context {
...
-        self.frames_processed += 1;
-        Ok(Packet {
-          data,
-          rec,
-          number: fi.number,
-          frame_type: fi.frame_type,
-          psnr
-        })
+        let fi = fi.clone();
+        self.finalize_frame(&*fs.input, rec, &fi)
```
lu-zero: this would be nicer in a separate patch.

shssoichiro: Opened https://github.com/xiph/rav1e/pull/982

lu-zero: Nice, now it needs just a rebase :)

shssoichiro:
Rebased. Here is a [sample of the stats file](https://pastebin.com/B318nLx0) as it is currently output. It is optimized for size, rather than human readability.

tdaede:
That looks OK, though I'm not sure of the utility of the config being stored?

@EwoutH @moisesmcardona what do you think of this format?

shssoichiro:
I'm not entirely sure when the config would be used, but x264 stats file records it, so we might use it at some point. Maybe it's not worth storing it until we actually have a reason to need it.

shssoichiro:
Removed "config" data from the output JSON since we don't need it at this time, and made `serde_json` optional since it's only used in the binary.

Output: https://pastebin.com/X14dLtwD

PR #983, opened 2019-02-11
Move full_search to an iterator

lu-zero:
The speed does not seem to change and probably the workload is still small enough that rayon won't be useful by using it directly.

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -264,8 +264,11 @@ fn full_search(
   lowest_cost: &mut u32, po: &PlaneOffset, step: usize, bit_depth: usize,
   lambda: u32, pmv: &[MotionVector; 2], allow_high_precision_mv: bool
 ) {
-  for y in (y_lo..=y_hi).step_by(step) {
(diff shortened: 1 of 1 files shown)


src/me.rs
```diff
@@ -264,8 +264,14 @@ fn full_search(
   lowest_cost: &mut u32, po: &PlaneOffset, step: usize, bit_depth: usize,
   lambda: u32, pmv: &[MotionVector; 2], allow_high_precision_mv: bool
 ) {
+    let search_range_y = (y_lo..=y_hi).step_by(step);
+    let search_range_x = (x_lo..=x_hi).step_by(step);
+    let search_area = search_range_y.flat_map(|y| { search_range_x.clone().map(move |x| (y, x)) });
+/*
```
shssoichiro: Not a fan of having commented code in the codebase. We have git history if we need to revert it. :slightly_smiling_face:

lu-zero: I left it as note on what the iterator was doing, but you are right, patch amended :)

PR #986, opened 2019-02-12
Download nasm archives from mirrors

barrbrain:
Working around an outage of https://www.nasm.us/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ before_install:
     - hash -r
     - which cmake
     - cmake --version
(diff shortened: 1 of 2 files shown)


tmatth:
@barrbrain the videolan mirror may be faster: http://download.videolan.org/contrib/nasm/

rzumer:
I'm not sure that just changing the URL is an improvement, what if that one goes down next?

tdaede:
I'm okay landing this now (with either mirror, I trust videolan more tho).

I think what we want to actually implement is https://docs.travis-ci.com/user/caching/

PR #995, opened 2019-02-13
Split encoder header

kodabb:
as requested by @smarter


kodabb:
maybe some more functions can be moved to header? also what about `FrameType`?

PR #997, opened 2019-02-14
Add RDOType parameter, which currently supports either TxDist or Full.

tdaede:


tdaede:
Considering dropping the for_rdo boolean and passing RDOType::Full in when needed, thoughts?

src/rdo.rs
```diff
@@ -813,6 +837,12 @@ pub fn rdo_tx_type_decision(
 
   let cw_checkpoint = cw.checkpoint();
 
+  let rdo_type = if fi.use_tx_domain_distortion {
+    RDOType::TxDist
+  } else {
+    RDOType::Full
```
ycho: So, 'Full' means pixel domain not both tx and pixel, correct?

tdaede: It means Pixel Domain + Full Rate. I can come up with different names or comments. Two other types I want to add include:

TxDistRate = tx_dist plus tx_dist_rate
MetricEst = SATD based rate and distortion estimate

I also need a training mode. I could put that as a RDOType or just have a separate flag for it.

tdaede:
BTW the reason I moved this to a parameter instead of fi is so that we can do different RDO types on different blocks in the future.

lu-zero:
unbreak the benchmarks please :)

src/rdo.rs
```diff
@@ -34,6 +34,23 @@ use std;
 use std::vec::Vec;
 use crate::partition::PartitionType::*;
 
+#[derive(Copy,Clone)]
+pub enum RDOType {
+  Full,
```
ycho: Yes, rephrasing the name so that we can understand from it will help us to use it correctly!

tdaede: Do you have a suggestion for names?

ycho: Maybe, 
Pixel Domain + Full Rate ==> PixelDistRealRate?
TxDistRate = tx_dist plus tx_dist_rate ==> TxDistRealRate
MetricEst = SATD based rate and distortion estimate ==> SatdDistSatdRate
(And you can make them all cap and boldfaced!)

ycho: Or, to save space,
RD_PIXEL_REAL
RD_TX_REAL
RD_SATD_SATD
?

tdaede: How's this?

tdaede:
Next patch will be the TxDistRate one. Then after that, SATD.

PR #1002, opened 2019-02-15
Make plane generic over pixel component type

rom1v:
In order to support both `u8` and `u16` for plane components, make the `Plane` structure generic over the component type. As a consequence, many other structures and functions also become generic.

Some functions are not `u8`-compatible yet, although they have been make generic over the component type to make the compilation work. They assert that the size of the generic parameter is 16 bits wide.

For this reason, the root context structure is unconditionally created as `Context<u16>` for now.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -49,7 +49,7 @@ fn write_b_bench(b: &mut Bencher, tx_size: TxSize, qindex: usize) {
     ..Default::default()
   };
   let sequence = Sequence::new(&Default::default());
-  let mut fi = FrameInvariants::new(config, sequence);
+  let mut fi = FrameInvariants::<u16>::new(config, sequence);
   let mut w = ec::WriterEncoder::new();
   let fc = CDFContext::new(fi.base_q_idx);
   let bc = BlockContext::new(fi.sb_width * 16, fi.sb_height * 16);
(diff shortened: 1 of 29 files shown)


tdaede:
FrameInvariants becoming generic really adds a lot of noise :/ this is because of the reference frame structure correct?

rom1v:
> this is because of the reference frame structure correct?

Yes, just because of its field:
```
pub rec_buffer: ReferenceFramesSet<T>,
```

tdaede:
Do you think it would make sense to do some dynamic dispatch over that field? Or does that make it even more complicated?

rom1v:
> Do you think it would make sense to do some dynamic dispatch over that field?

That would mean erasing its concrete type behind a trait:

```rust
pub trait RFS {}

impl<T: Pixel> RFS for ReferenceFramesSet<T> {}

pub struct FrameInvariants {
    // …
    pub rec_buffer: Box<dyn RFS>,
    // …
}
```

But then, from a `FrameInvariants`, we could not expose the `Frames` of the `rec_buffer` (because its concrete type would be lost), so every action on the `ReferenceFrameSet<T>` would need to be implemented through methods on the `RFS` trait.

It would be impractical IMO.

In fact, `ReferenceFrameSet` needs to [updated](https://github.com/xiph/rav1e/blob/master/src/encoder.rs#L2148) with a `FrameState<T>`, which is not possible if concrete type of the `ReferenceFrameSet` is not known (unless we also erase the `FrameState` parameter…).

tdaede:
I see. I guess we could also split out ReferenceFrameSet to be its own parameter. But that doesn't make many things better.

tdaede:
lu_zero can you check how the api.rs changes interact with crav1e?

lu-zero:
They would break everything, but I can fix after it lands.

lu-zero:
I need to check what works better compared to `dyn` dispatch, but we could also at a certain point of the chain use just a proxy like this 
```
enum Something {
   Something16(SomethingInner<u16>),
   Something8(SomethingInner<u8>
}
```

With `From and Deref` made to unwrap to the actual type.

PR #1006, opened 2019-02-16
Fix misc. errtas and outdated comments.

ycho:
Also, change some assert!() to debug_assert!() in bottomup partition.


src/encoder.rs
```diff
@@ -2017,7 +2017,7 @@ fn encode_tile(fi: &FrameInvariants, fs: &mut FrameState) -> Vec<u8> {
                                  BlockSize::BLOCK_64X64, &bo, &None, &pmvs);
       }
 
-      // CDEF has to be decisded before loop restoration, but coded after
+      // CDEF has to be decided before loop restoration, but coded after
```
ycho: @xiphmont, I finally did this!

PR #1007, opened 2019-02-16
Add coding style.

tdaede:


shssoichiro:
Semi-related: Is there a plan to run rustfmt over the whole codebase? There are many files that don't follow the conventions, simply because rustfmt hasn't been run.

vladimir-kazakov:
Perhaps it makes sense to add [EditorConfig](https://editorconfig.org) as part of this PR? It would make it easier to contribute to the project, removing the need to reconfigure a code editor, to read coding guidelines, and correctly understanding them.

lu-zero:
@vladimir-kazakov if you have time to prepare one it would be much appreciated :)

(or we could just revert to the default style for the sake of ecosystem compatibility)

vladimir-kazakov:
No problem, I could come up with something in the near future, if there's no desire to do it as part of this PR.

PR #1008, opened 2019-02-16
Add --speed-test option for testing speed settings in isolation

shssoichiro:
This setting is hidden from the CLI help message, since it is intended
to be for benchmarking purposes only.


shssoichiro:
Updated to make a few settings (quantizer, tuning, show_psnr) apply in speed-test mode as well.

PR #1011, opened 2019-02-16
Add a threads option

lu-zero:


src/bin/common.rs
```diff
@@ -212,6 +220,10 @@ pub fn parse_cli() -> CliOptions {
     )
     .get_matches();
 
+  if let Some(threads) = matches.value_of("THREADS").map(|v| v.parse().expect("Invalid number of threads")) {
```
tdaede: This should be "Threads must be an integer", it doesn't validate teh actual number as far as I know.

lu-zero: Updating accodingly (`parse()` would also make sure it fits the range, but it is relatively moot since we are talking about usize :))

PR #1013, opened 2019-02-17
Fixes for speed-test option

shssoichiro:
Defaults block size to 16x16
Allows specifying multiple settings at once (space separated)
Uses low_latency setting specified on CLI during speed-test

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -196,7 +196,7 @@ pub fn parse_cli() -> CliOptions {
     )
     .arg(
       Arg::with_name("SPEED_TEST")
-        .help("Run an encode using default encoding settings, adjusting only one; allows benchmarking settings in isolation")
+        .help("Run an encode using default encoding settings, manually adjusting only the settings specified; allows benchmarking settings in isolation")
         .hidden(true)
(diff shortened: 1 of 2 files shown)


lu-zero:
`16x16` or `64x64` ?

shssoichiro:
It should have been 16x16. Looks like some of my changes didn't get pushed.

src/bin/common.rs
```diff
@@ -368,12 +371,12 @@ fn load_speed_test_cfg(setting: &str) -> EncoderConfig {
     "min_block_size_8x8" => {
       cfg.speed_settings.min_block_size = BlockSize::BLOCK_8X8;
     },
-    "min_block_size_16x16" => {
```
lu-zero: Breaking script compatibility isn't great :) in this case is probably not important, but keep it in mind in the future.

shssoichiro: Yeah, in this case it's only because I'm reasonably sure nothing is using this yet.

shssoichiro: More specifically, these changes came out of trying to create a script that uses this feature, and finding the limitations.

PR #1014, opened 2019-02-17
Add EditorConfig

vladimir-kazakov:
This PR adds EditorConfig. It automatically reconfigures a code editor, which supports EditorConfig (https://editorconfig.org), to use style conventions used by the project.

This topic has started in #1007. The config is created based on the style I noticed in the project, but I may not know everything.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


lu-zero:
Looks good. Thank you a lot!

PR #1017, opened 2019-02-19
Fix miscalculation of PSNR with frame reordering

shssoichiro:
With frame reordering enabled, every other frame would be compared to a
blank frame, generating a very low PSNR calculation. Also fixes an issue
where the PlaneIter ends one pixel too early, which currently is only
used by PSNR calculation.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -552,6 +552,11 @@ impl Context {
     Ok(())
   }
 
+  pub fn get_frame(&self, frame_number: u64) -> Arc<Frame> {
+    // Clones only the arc, so low cost overhead
+    self.frame_q.get(&frame_number).as_ref().unwrap().as_ref().unwrap().clone()
+  }
+
   pub fn get_frame_count(&self) -> u64 {
     self.frame_count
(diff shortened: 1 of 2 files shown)


src/api.rs
```diff
@@ -792,6 +792,7 @@ impl Context {
     let mut psnr = None;
     if self.config.enc.show_psnr {
       if let Some(ref rec) = rec {
+        let original_frame = self.frame_q.get(&fi.number).as_ref().unwrap().as_ref().unwrap().clone();
```
shssoichiro: Only clones the `Arc`, so not costly.

src/api.rs
```diff
@@ -792,6 +792,7 @@ impl Context {
     let mut psnr = None;
     if self.config.enc.show_psnr {
       if let Some(ref rec) = rec {
+        let original_frame = self.frame_q.get(&fi.number).as_ref().unwrap().as_ref().unwrap();
```
lu-zero: This should be probably abstracted.

PR #1021, opened 2019-02-20
Add Shell auto-completion support

lu-zero:
It is implemented as an advanced command line option.

Run `rav1e advanced --help` for the list of supported shells.

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -226,6 +226,7 @@ pub fn parse_cli() -> CliOptions {
         ])
     )
     .subcommand(SubCommand::with_name("advanced")
+                .setting(AppSettings::Hidden)
                 .about("Advanced features")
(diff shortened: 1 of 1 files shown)


src/bin/common.rs
```diff
@@ -224,12 +225,30 @@ pub fn parse_cli() -> CliOptions {
           "no_scene_detection",
         ])
     )
-    .get_matches();
+    .subcommand(SubCommand::with_name("advanced")
```
shssoichiro: It might make sense to hide this from the CLI help text.
```suggestion
    .subcommand(SubCommand::with_name("advanced")
                .setting(AppSettings::Hidden)
```

tdaede: +1 on hiding it

lu-zero: So be it, I wonder if there is a mean to add `--long-help` or such.

PR #1022, opened 2019-02-20
Move lookahead logic into API

shssoichiro:


tdaede:
I've assigned @negge to review as I think he was looking at writing a similar patch last week.

shssoichiro:
Yes, the intent of this PR is to unblock https://github.com/xiph/rav1e/pull/955

PR #1023, opened 2019-02-21
[WIP] Avoid stride vs pitch confusion

rom1v:
The functions `deblock_sizeXXX()` used both `stride` and `pitch` variables as a trick to implement both vertical and horizontal filtering.

To avoid confusion, only expose `stride`, with an additional parameter to indicate the direction.

---

#### ~Pull request visualization~

~This PR is written over my "make plane generic" PR, so please ignore the first commit (which is #1002).~

#### Horizontal + vertical

This change only makes the "main pitch" / "secondary pitch", currently exposed as `stride` and `pitch` in the function signature, internal to the `deblock_sizeXXX()` functions. Instead, a single boolean is exposed as parameter in addition to the stride (like in `deblock_size()` on `master`: https://github.com/xiph/rav1e/blob/8057ee7cd6d007a316f66eb531736debf2c576dd/src/deblock.rs#L113-L115).

Several people on IRC (including myself) would prefer to split each `deblock_sizeXXX()` function into `deblock_h_sizeXXX()` and `deblock_v_sizeXXX()`. That way, the horizontal/vertical direction would be decided at compile time without an additional check at runtime, and for the vertical direction, we could use `copy_from_slice()` instead of copying item by item.

However, this is not easily done without either:
 - code duplication;
 - an additional copy of the vertical segment into contiguous slice;
 - an additional abstraction to access horizontal and vertical segments the same way…

Therefore, I would prefer to just change the function signatures without splitting the functions for now, so that I can continue my slice+stride to PlaneSlice refactor (that this change would help, because then, there is no two `stride` and `pitch` variables which could take any value anymore).

#### Possible error?

In current `master`, the last call in the loop of `deblock_size4()` is:

```rust
stride_copy(p, &x, sec_pitch);
```

For all others (`deblock_size6()`, `deblock_size8()`, `deblock_size14()`), it is:

```rust
stride_copy(&mut p[sec_pitch..], &x, sec_pitch);
```

@xiphmont could you confirm the difference is on purpose, please?

### DIFF
--- a/src/deblock.rs
+++ b/src/deblock.rs
@@ -357,24 +357,23 @@ fn mask4(p1: i32, p0: i32, q0: i32, q1: i32, shift: usize) -> usize {
 
 // Assumes rec[0] is set 2 taps back from the edge
 fn deblock_size4<T: Pixel>(
-  rec: &mut [T], pitch: usize, stride: usize, level: usize, bd: usize
+  rec: &mut [T], stride: usize, vertical: bool, level: usize, bd: usize
 ) {
-  let mut s = 0;
-  for _i in 0..4 {
-    let p = &mut rec[s..];
+  let (main_pitch, sec_pitch) = if vertical { (stride, 1) } else { (1, stride) };
+  for y in 0..4 {
+    let p = &mut rec[y * main_pitch..];
     let p1: i32 = p[0].as_();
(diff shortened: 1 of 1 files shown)


rom1v:
Replaced by #1024.

PR #1028, opened 2019-02-22
Correctly use Psy tuning as default via API

shssoichiro:


barrbrain:
The test failure for `test_encode_decode_aom::high_bd` is due to a silent overflow at [src/rdo.rs:104](https://github.com/xiph/rav1e/blob/8a62a98f/src/rdo.rs#L104).
It cab be resolved by prepending changing `16_265_089` to `16_265_089u64`.

lu-zero:
do you have a patch to land before this pr?

barrbrain:
Please see #1030 for the fix.

shssoichiro:
This seemed weird, so after looking a bit more, the CLI would correctly default to Psy tuning

`[src/bin/common.rs:397] matches.value_of("TUNE").unwrap().parse().unwrap() = Psychovisual`

However, the tests or anyone using the API directly would not be, which is why this change discovered a test failure.

I'll rebase with #1030 now in master.

PR #1031, opened 2019-02-22
Enable extreme comparisons lint

shssoichiro:
https://rust-lang.github.io/rust-clippy/master/index.html#absurd_extreme_comparisons

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2565,9 +2565,6 @@ impl ContextWriter {
       /* TODO: Set zeromv ref to the converted global motion vector */
     } else {
       /* TODO: Set the zeromv ref to 0 */
-    }
-
-    if ref_frames[0] <= INTRA_FRAME {
       return 0;
     }
 
--- a/src/deblock.rs
+++ b/src/deblock.rs
@@ -121,8 +121,8 @@ fn deblock_size<T: Pixel>(
   if !(block_edge
     || !block.skip
     || !prev_block.skip
-    || block.ref_frames[0] <= INTRA_FRAME
-    || prev_block.ref_frames[0] <= INTRA_FRAME)
+    || block.ref_frames[0] == INTRA_FRAME
+    || prev_block.ref_frames[0] == INTRA_FRAME)
   {
     0
   } else {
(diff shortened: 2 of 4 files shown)


lu-zero:
This lint feels wrong to me. @tdaede am I alone in that? :)

shssoichiro:
Why is that? If the variable is a `usize`, it can't ever be less than 0. Unless the concern is that it might change to an `isize` in the future?

src/context.rs
```diff
@@ -2559,9 +2559,6 @@ impl ContextWriter {
       /* TODO: Set zeromv ref to the converted global motion vector */
     } else {
       /* TODO: Set the zeromv ref to 0 */
-    }
-
-    if ref_frames[0] <= INTRA_FRAME {
```
tdaede: Would be interesting to see why this condition was there.

src/deblock.rs
```diff
@@ -121,8 +121,8 @@ fn deblock_size<T: Pixel>(
   if !(block_edge
     || !block.skip
     || !prev_block.skip
-    || block.ref_frames[0] <= INTRA_FRAME
-    || prev_block.ref_frames[0] <= INTRA_FRAME)
+    || block.ref_frames[0] == INTRA_FRAME
```
tdaede: This change is actually clearer.

src/transform/mod.rs
```diff
@@ -37,7 +37,7 @@ fn half_btf(w0: i32, in0: i32, w1: i32, in1: i32, bit: usize) -> i32 {
   //   overflows, but w0*in0 + w1*in1 + rounding isn't.
   let result = (w0 * in0).wrapping_add(w1 * in1);
   // Implement a version of round_shift with wrapping
-  if bit <= 0 {
+  if bit == 0 {
```
tdaede: meh on this one

tdaede:
I would say this is a less useful lint, but it did find some rather odd code.

src/context.rs
tdaede: Ohh, I see, it's collapsing this condition into the one above. That's fancy.

PR #1035, opened 2019-02-23
Expose separate rows from PlaneSlice

rom1v:
Currently, many functions access rectangular regions of planes (spanning multiple rows) via a primitive slice to the whole plane along with the stride information.

This strategy is not compatible with tiling, since this borrows the memory belonging to other tiles.

This is basically what I explained in https://github.com/xiph/rav1e/issues/631#issuecomment-456088577.

Therefore, only expose separate rows in `PlaneSlice`.

Concretely, here are the changes from the caller point of view:

```rust
// get a specific row

// before
let row = plane_slice.as_slice()[y * stride..(y + 1) * stride];
// or if we don't care about the end
let row = plane_slice.as_slice()[y * stride..];

// after
let row = &plane_slice[y];
```

```rust
// get a component value

// before
let value = plane_slice.as_slice()[y * stride + x];

// after
let value = plane_slice[y][x];
```

```rust
// iterate over rows

// before
for row in plane_slice.as_slice().chunks(stride).take(height) {
    let _first_four_values = &row[..4];
}

// after
for row in plane_slice.rows_iter() {
    let _first_four_values = &row[..4];
}
```

Note that in itself, it does not solve the incompatibility with tiles (`PlaneSlice` itself still borrows the whole plane primitive slice). This is only the first step to remove all usages of raw slice + stride information, so that we can then replace `PlaneSlice` by a tile-specific "plane region" structure which does not borrow memory outside the tile (cf #821).

For now, I only submit changes for `PlaneSlice`. I will submit similar changes for `PlaneMutSlice` once I get feedbacks on this one (and when it's finished).

### DIFF
--- a/src/plane.rs
+++ b/src/plane.rs
@@ -10,7 +10,7 @@
 use std::iter::FusedIterator;
 use std::fmt::{Debug, Display, Formatter};
 use std::mem;
-use std::ops::Range;
+use std::ops::{Index, Range};
 
 use crate::util::*;
 
@@ -410,6 +410,13 @@ impl<'a, T: Pixel> PlaneSlice<'a, T> {
     let new_x =
       (self.x + add_x as isize + self.plane.cfg.xorigin as isize) as usize;
     self.plane.data[new_y * self.plane.cfg.stride + new_x]
+  }
+}
+
+impl<'a, T: Pixel> Index<usize> for PlaneSlice<'a, T> {
+  type Output = [T];
+  fn index(&self, index: usize) -> &Self::Output {
+    self.row(index)
   }
 }
 
--- a/src/plane.rs
+++ b/src/plane.rs
@@ -375,24 +375,6 @@ impl<'a, T: Pixel> PlaneSlice<'a, T> {
     }
   }
 
-  pub fn as_slice(&self) -> &'a [T] {
-    let stride = self.plane.cfg.stride;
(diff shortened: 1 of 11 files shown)


tdaede:
This looks excellent!

src/plane.rs
```diff
@@ -306,23 +315,64 @@ impl<'a, T: Pixel> ExactSizeIterator for IterWidth<'a, T> { }
...
+pub struct RowsIter<'a, T: Pixel> {
+  ps: &'a PlaneSlice<'a, T>,
+  next_row: usize,
+}
+
+impl<'a, T: Pixel> Iterator for RowsIter<'a, T> {
+  type Item = &'a [T];
+
+  fn next(&mut self) -> Option<Self::Item> {
+    let remaining = self.ps.plane.cfg.height as isize - self.ps.y + self.next_row as isize;
```
rom1v: Oops, parentheses missing:

```rust
let remaining = self.ps.plane.cfg.height as isize - (self.ps.y + self.next_row as isize);
```

It's fixed in my `PlaneMutSlice` pull request #1040.

rom1v: Oops, parentheses missing:

```rust
let remaining = self.ps.plane.cfg.height as isize - (self.ps.y + self.next_row as isize);
```

It's fixed in my `PlaneMutSlice` pull request #1043.

tdaede: Weird that it didn't get picked up in tests.

PR #1037, opened 2019-02-24
Fix issues with reading lookahead

### LINKED ISSUE #1036: Encoding gets stuck in the last frame. Never ending
The encoding will get stuck in the last frame and will still consume CPU cycles. I guess it's stuck in a loop or something.

Related to #1010 ?

shssoichiro:
Fixes #1036 where encoder would hang at end of video if `--limit` not
used
Fixes issue where encoder would read too many frames, causing high
memory usage

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -510,8 +510,10 @@ pub struct Context<T: Pixel> {
 
 #[derive(Clone, Copy, Debug)]
 pub enum EncoderStatus {
-  /// The encoder needs more Frames to produce an output Packet
+  /// The encoder needs more data to produce an output Packet--used with frame reordering
   NeedMoreData,
+  /// The encoder needs more Frames to analyze lookahead
+  NeedMoreFrames,
   /// There are enough Frames queue
   EnoughData,
   ///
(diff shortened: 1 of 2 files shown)


src/api.rs
```diff
@@ -510,8 +510,10 @@ pub struct Context<T: Pixel> {
 
 #[derive(Clone, Copy, Debug)]
 pub enum EncoderStatus {
-  /// The encoder needs more Frames to produce an output Packet
+  /// The encoder needs more data to produce an output Packet--used with frame reordering
   NeedMoreData,
+  /// The encoder needs more Frames to analyze lookahead
+  NeedMoreFrames,
```
lu-zero: They are the same condition: The encoding cannot continue until more data is provided.

shssoichiro: Technically, not quite. In one case we have the data in the frame queue, and just need to increment the index to read it. In the other case, the data is not yet in the frame queue.

lu-zero: That implementation detail should not leak to the user IMHO.

shssoichiro: I agree, but I don't know of a better way to implement this without making the Context aware of the decoder. We've explicitly chosen to keep the Decoder trait and anything implementing it as part of the CLI.

I'll see if I can think over it and come up with a better approach.

lu-zero: Let's unbreak at least the experience for the user first. Land this patch.

lu-zero: I'll try to see what happens when you use only `NeedMoreData` myself later.

shssoichiro:
Going to merge this for now, to unbreak master. We will follow up on this PR with a better solution.

PR #1040, opened 2019-02-26
Use raw pointers for cdef_filter_block() params

rom1v:
Once all slice+stride have been replaced by `Plane(Mut)Slice`, we will not be able to provide multirows primitive slices anymore.

Therefore, expose raw pointers for the (unsafe) function `cdef_filter_block()`.

---

This is an alternative to #1039, which caused performances regression (cf https://github.com/xiph/rav1e/commit/fa65c3ab3fa6c14c6a7a007190434d0592067017).

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -128,7 +128,7 @@ fn constrain(diff: i32, threshold: i32, damping: i32) -> i32 {
 // of the 2-pixel padding around the block, not the block itself.
 // The destination is unpadded.
 unsafe fn cdef_filter_block<T: Pixel>(
-  dst: &mut [T], dstride: isize, input: &[T],
+  dst: *mut T, dstride: isize, input: *const T,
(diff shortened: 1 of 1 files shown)


rom1v:
> I measured a small regression

When I tested yesterday, it improved by ~2% (but I couldn't explain why).

barrbrain:
> > I measured a small regression
>
> When I tested yesterday, it improved by ~2% (but I couldn't explain why).

In essence, it is compiler noise. Let's not worry about it now. Looks good to me.

tdaede:
I tested on my throttling-disabled box and got:
```
cdef_frame(128, 128)    time:   [1.2557 ms 1.2558 ms 1.2558 ms]                                  
                        change: [-0.3673% -0.3115% -0.2614%] (p = 0.00 < 0.05)
                        Change within noise threshold.
  5 (5.00%) high severe
```

PR #1043, opened 2019-02-27
Expose separate rows from PlaneMutSlice

rom1v:
This PR is to `PlaneMutSlice` what #1035 was to `PlaneSlice`.

~It is written over #1040 (so ignore the first commit).~

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -55,12 +55,12 @@ fn first_max_element(elems: &[i32]) -> (usize, i32) {
 // in a particular direction. Since each direction have the same sum(x^2) term,
 // that term is never computed. See Section 2, step 2, of:
 // http://jmvalin.ca/notes/intra_paint.pdf
-fn cdef_find_dir<T: Pixel>(img: &[T], stride: usize, var: &mut i32, coeff_shift: usize) -> i32 {
+fn cdef_find_dir<'a, T: Pixel>(img: &PlaneSlice<'a, T>, var: &mut i32, coeff_shift: usize) -> i32 {
   let mut cost: [i32; 8] = [0; 8];
   let mut partial: [[i32; 15]; 8] = [[0; 15]; 8];
   for i in 0..8 {
     for j in 0..8 {
-      let p: i32 = img[i * stride + j].as_();
+      let p: i32 = img[i][j].as_();
       // We subtract 128 here to reduce the maximum range of the squared
       // partial sums.
       debug_assert!(p >> coeff_shift <= 255);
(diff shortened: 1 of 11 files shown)


rom1v:
Since I didn't write the commits in order (the git history is not the chronological order), github displays commits in an incorrect order in the pull request.

tdaede:
Landed the cdef fix, needs one more rebase sorry ;_;

src/plane.rs
```diff
@@ -316,28 +316,27 @@ impl<'a, T: Pixel> ExactSizeIterator for IterWidth<'a, T> { }
...
   fn next(&mut self) -> Option<Self::Item> {
-    let remaining = self.ps.plane.cfg.height as isize - self.ps.y + self.next_row as isize;
-    if remaining > 0 {
-      let row = self.next_row;
-      self.next_row += 1;
+    if self.plane.cfg.height as isize > self.y {
       // cannot directly return self.ps.row(row) due to lifetime issue
-      let range = self.ps.slice_range(row);
-      Some(&self.ps.plane.data[range])
+      let range = PlaneSlice::slice_range(&self.plane, self.x, self.y);
```
tdaede: Would it make more sense for slice_range to be a method on plane?

src/cdef.rs
```diff
@@ -404,13 +404,13 @@ pub fn cdef_filter_superblock<T: Pixel>(
             unsafe {
               let xsize = 8 >> xdec;
               let ysize = 8 >> ydec;
-              let dst = out_slice.offset_as_mutable(8 * bx >> xdec, 8 * by >> ydec);
-              let input = in_slice.offset(8 * bx >> xdec, 8 * by >> ydec);
-              assert!(dst.len() >= (ysize - 1) * out_stride + xsize);
-              assert!(input.len() >= ((ysize + 3) * in_stride + xsize + 4) as usize);
-              cdef_filter_block(dst.as_mut_ptr(),
+              assert!(out_slice.rows_iter().len() >= (8 * by >> ydec) + ysize);
```
tdaede: Is there a more direct way to get len() than producing an iterator first?

src/plane.rs
rom1v: Yes, you're right. Now that it is refactored that way, it is not related to `PlaneSlice` anymore.

src/cdef.rs
rom1v: Currently, no, but we can add `rows_count()` (or similar) on `PlaneMutSlice`.

I did not do that because it's only used in `assert!()`.

tdaede: OK, we can skip it.

lu-zero: The assert could be a `debug_assert!()` in case.

rom1v: When should we call `assert!()` vs `debug_assert!()` in rav1e?

Here, we call an unsafe code using raw pointers, which is undefined if this assertion fails.

lu-zero: Ideally we can move to `debug_assert!()` all the cases in which:
- We want to document the behavior, but
- We have tests to make sure we do not break it
- The code isn't user-facing. When it is, I'd keep the assert and document when it could panic.

PR #1044, opened 2019-02-28
Optimize debug builds

rom1v:
Unoptimized builds are unusable for video encoding.

Always building in release mode is not a good solution, because we still want to distinguish debug builds, having `debug_assert!()` enabled, from release builds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,9 @@ bench = false
 name = "bench"
 harness = false
 
+[profile.dev]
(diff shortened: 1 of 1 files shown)


tdaede:
Looks OK. This will still be slower than a release build because of multiple codegen units

rom1v:
Implemented by 3062fb0065ed464c6ab523cb99687ef40a6861ef.

PR #1046, opened 2019-03-01
No coveralls

tdaede:


barrbrain:
Let's enure that the tests that were run under kcov are still being run. It's not obvious to me on first reading that they are.

shssoichiro:
@barrbrain is correct. The tests phase has `--ignored` which _only_ runs the ignored tests. We never run the rest of the tests.

lu-zero:
I'd keep coveralls, I wonder if the non-kcov tools can do a better job at it.

lu-zero:
Now I'm trying to see if we could get the same coverage with less time.

lu-zero:
We managed to make it much more bearable, we can keep it I think.

PR #1050, opened 2019-03-01
Diamond search for half resolution motion estimation

magwyz:
It currently does not use previous frames computed motion.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1979,6 +1979,12 @@ fn build_coarse_pmvs<T: Pixel>(fi: &FrameInvariants<T>, fs: &FrameState<T>) -> V
 fn encode_tile<T: Pixel>(fi: &FrameInvariants<T>, fs: &mut FrameState<T>) -> Vec<u8> {
   let mut w = WriterEncoder::new();
 
+  let estimate_motion_ss2 = if fi.config.speed_settings.diamond_me {
+    crate::me::DiamondSearch::estimate_motion_ss2
+  } else {
+    crate::me::FullSearch::estimate_motion_ss2
+  };
+
   let fc = if fi.primary_ref_frame == PRIMARY_REF_NONE {
     CDFContext::new(fi.base_q_idx)
   } else {
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1112,14 +1112,16 @@ pub fn motion_compensate<T: Pixel>(
 
 pub fn save_block_motion<T: Pixel>(
    fs: &mut FrameState<T>,
-   w_in_b: usize, _h_in_b: usize,
+   w_in_b: usize, h_in_b: usize,
    bsize: BlockSize, bo: &BlockOffset,
    ref_frame: usize, mv: MotionVector,
 ) {
-  let frame_mvs = &mut fs.frame_mvs;
-  for mi_y in (bo.y)..(bo.y + bsize.height_mi()) {
-    for mi_x in (bo.x)..(bo.x + bsize.width_mi()) {
-      frame_mvs[ref_frame][mi_y][mi_x] = mv;
+  let frame_mvs = &mut fs.frame_mvs[ref_frame];
+  let bo_x_end = (bo.x + bsize.width_mi()).min(w_in_b);
+  let bo_y_end = (bo.y + bsize.height_mi()).min(h_in_b);
+  for mi_y in bo.y..bo_y_end {
+    for mi_x in bo.x..bo_x_end {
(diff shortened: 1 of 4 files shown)


magwyz:
This seems to reduce the compression time. Here are the AWCY runs:
https://beta.arewecompressedyet.com/?job=diamond_me_hres_ref%402019-03-01T16%3A36%3A57.994Z&job=diamond_me_hres_test%402019-03-01T16%3A36%3A29.656Z

tdaede:
Looks like an acceptable loss to take all the time.

src/encoder.rs
```diff
@@ -1106,6 +1106,22 @@ pub fn motion_compensate<T: Pixel>(
   }
 }
 
+pub fn save_block_motion<T: Pixel>(
+   fi: &FrameInvariants<T>, fs: &mut FrameState<T>,
```
lu-zero: Maybe take in directly `w_in_b` and `h_in_b`, as you did for `get_subset_predictors`?

src/partition.rs
```diff
@@ -761,6 +761,10 @@ impl MotionVector {
       col: (self.col / 8) * 8
     }
   }
+
+  pub fn is_null(self) -> bool {
```
lu-zero: `is_zero()` would be less confusing :)

src/me.rs
```diff
@@ -746,29 +758,53 @@ pub fn estimate_motion_ss2<T: Pixel>(
...
+            fi, &po,
+            &fs.input_hres, &rec.input_hres,
+            &predictors, fi.sequence.bit_depth,
+            global_mv, lambda,
+            mvx_min >> 1, mvx_max >> 1, mvy_min >> 1, mvy_max >> 1,
+            blk_w >> 1, blk_h >> 1,
+            &mut best_mv, &mut lowest_cost,
+            &mut None, 0
+          );
+        } else {
```
lu-zero: Make them two functions and push the branch as up as possible in the call tree.

magwyz:
Here are the latest AWCY runs:
https://beta.arewecompressedyet.com/?job=diamond_me_hres_ref%402019-03-14T16%3A45%3A14.540Z&job=diamond_me_hres_test

src/me.rs
magwyz: I extracted the if clause from the loop in the last version of this PR.
I have no issue making two functions. I however do not see in practice how to do it properly.

lu-zero:
I'd rework it to use the same approach used in #1118 :)

src/me.rs
```diff
@@ -731,91 +731,141 @@ pub fn estimate_motion_ss4<T: Pixel>(
...
+    frame_mvs: &[MotionVector], frame_ref_opt: &Option<Arc<ReferenceFrame<T>>>,
+    po: PlaneOffset, rec: &Arc<ReferenceFrame<T>>,
+    global_mv: &[MotionVector; 2], lambda: u32,
+    mvx_min: isize, mvx_max: isize, mvy_min: isize, mvy_max: isize,
+    blk_w: usize, blk_h: usize,
+    best_mv: &mut MotionVector, lowest_cost: &mut u64
+  );
+}
+
+pub struct DiamondSearchSS2 {}
```
lu-zero: We could keep all in a single trait and struct placeholder maybe?

magwyz: Sure.
This is in the last version of the PR.

lu-zero:
It needs a rebase now =/

magwyz:
I've just done the rebase.

src/me.rs
```diff
@@ -397,15 +452,15 @@ pub struct FullSearch {}
...
   fn full_pixel_me<T: Pixel>(
     fi: &FrameInvariants<T>, fs: &FrameState<T>, rec: &Arc<ReferenceFrame<T>>,
-    po: &PlaneOffset, bo: &BlockOffset, lambda: u32, ref_slot: usize,
+    po: &PlaneOffset, bo: &BlockOffset, lambda: u32,
     cmv: MotionVector, pmv: [MotionVector; 2], mvx_min: isize, mvx_max: isize,
     mvy_min: isize, mvy_max: isize, blk_w: usize, blk_h: usize,
     best_mv: &mut MotionVector, lowest_cost: &mut u64, ref_frame: usize
   ) {
-    let frame_mvs = &fs.frame_mvs[ref_slot];
+    let frame_mvs = &fs.frame_mvs[ref_frame - LAST_FRAME];
```
tdaede: I know this pattern is from libaom.... but I hate it :)

I'll change the whole codebase in my own patch.

PR #1052, opened 2019-03-01
Move the decode test to aom-sys

lu-zero:


tdaede:
Can this now drop the submodule too?

lu-zero:
We have a benchmark and a test that depends on it, we could remove the submodule and that code.

tdaede:
OK. Regardless this looks fine, as long as the decode test is just as easy to run. Just installing libaom system wide makes it work, correct?

lu-zero:
Correct

PR #1053, opened 2019-03-01
Remove meaningless operations

vladimir-kazakov:
This PR enables the Clippy's [erasing_op](https://rust-lang.github.io/rust-clippy/master/index.html#erasing_op), [identity_op](https://rust-lang.github.io/rust-clippy/master/index.html#identity_op), and [neg_multiply](https://rust-lang.github.io/rust-clippy/master/index.html#neg_multiply) lints, and fixes all warnings and errors caused by them.

I disabled some lints for some functions, because they have code like [this](https://github.com/xiph/rav1e/blob/714765bded83d76d93e804f44cff06f4c067eaa7/src/transform/forward.rs#L1575), [this](https://github.com/xiph/rav1e/blob/714765bded83d76d93e804f44cff06f4c067eaa7/src/cdef.rs#L141), and [this](https://github.com/xiph/rav1e/blob/714765bded83d76d93e804f44cff06f4c067eaa7/src/encoder.rs#L607). I don't fully understand the meaning of these numbers, so it's hard to say how important it is to keep the original style, but I see a seemingly important patterns there, so I decided to not touch them. Well, I found CDEF directions in the AV1 spec, which somewhat confirmed my suspicions.

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -114,12 +114,9 @@ fn constrain(diff: i32, threshold: i32, damping: i32) -> i32 {
   if threshold != 0 {
     let shift = cmp::max(0, damping - msb(threshold));
     let magnitude = cmp::min(diff.abs(), cmp::max(0, threshold - (diff.abs() >> shift)));
-    if diff < 0 {
-      -1 * magnitude
+
+    if diff < 0 { -magnitude } else { magnitude }
     } else {
-      magnitude
-    }
-  } else {
     0
   }
 }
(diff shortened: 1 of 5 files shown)


tdaede:
Those numbers are pairs of (x, y) sample locations, where some of the x and y's are 0. The 0 is not particularily meaningful either way, I don't have a preference on whether to exclude them or not. The SGR section is similar.

src/cdef.rs
```diff
@@ -114,19 +114,17 @@ fn constrain(diff: i32, threshold: i32, damping: i32) -> i32 {
   if threshold != 0 {
     let shift = cmp::max(0, damping - msb(threshold));
     let magnitude = cmp::min(diff.abs(), cmp::max(0, threshold - (diff.abs() >> shift)));
-    if diff < 0 {
-      -1 * magnitude
+
+    if diff < 0 { -magnitude } else { magnitude }
```
tmatth: I think this is fine but will mention: https://docs.rs/num/0.1.42/num/fn.signum.html

vladimir-kazakov:
@xiphmont, any comments?

PR #1055, opened 2019-03-02
Refactor encode-decode tests to reduce duplication

### LINKED ISSUE #956: Reduce duplication in encode-decode tests
Right now everything is duplicated, which is very annoying for making changes.

We have two separate but nearly identical implementations, one using an internal copy of aom and one using [dav1d-sys](http://crates.io/crates/dav1d-sys):
```
src/test_encode_decode_aom.rs 
src/test_encode_decode_dav1d.rs
```
- [ ] Move the decoder-specific code in a trait
- [ ] Move all the shared code in `src/test_encode_decode.rs` and make it use the trait you created
- [ ] Use [interpolate_test](https://crates.io/crates/interpolate_name) and `cfg_attr` to not repeat yourself

shssoichiro:
Closes #956

Some dav1d tests are known to fail due to rounding errors.
This is an existing issue and is not resolved in this commit.

### DIFF
--- a/src/test_encode_decode.rs
+++ b/src/test_encode_decode.rs
@@ -1,0 +1,302 @@
+// Copyright (c) 2018, The rav1e contributors. All rights reserved
+//
+// This source code is subject to the terms of the BSD 2 Clause License and
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+use super::*;
+use rand::{ChaChaRng, Rng};
+use std::sync::Arc;
+use crate::util::Pixel;
+#[cfg(feature="decode_test")]
+use crate::test_encode_decode_aom::AomDecoder;
+#[cfg(feature="decode_test_dav1d")]
+use crate::test_encode_decode_dav1d::Dav1dDecoder;
+
+fn fill_frame<T: Pixel>(ra: &mut ChaChaRng, frame: &mut Frame<T>) {
+  for plane in frame.planes.iter_mut() {
+    let stride = plane.cfg.stride;
+    for row in plane.data.chunks_mut(stride) {
+      for pixel in row {
+        let v: u8 = ra.gen();
+        *pixel = T::cast_from(v);
+      }
+    }
+  }
+}
+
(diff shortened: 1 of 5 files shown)


src/test_encode_decode.rs
```diff
@@ -0,0 +1,305 @@
...
+    assert_eq!(&line.0[..width], &line.1[..width]);
+  }
+}
+
+pub(crate) fn setup_encoder<T: Pixel>(
+  w: usize, h: usize, speed: usize, quantizer: usize, bit_depth: usize,
+  chroma_sampling: ChromaSampling, min_keyint: u64, max_keyint: u64,
+  low_latency: bool, bitrate: i32
+) -> Context<T> {
+  unsafe {
```
lu-zero: This should go away.

lu-zero:
Has the problem already been reported to upstream dav1d btw?

shssoichiro:
Does not appear so. I can file a report upstream.

src/test_encode_decode.rs
shssoichiro: Removed

shssoichiro:
Actually, it's possible that this is a rounding error in rav1e. If I change the seed of the test RNG from `[0; 32]` to `[1; 32]`, one of the AOM tests also fails on the same data as the dav1d test. Although it's strange that that isn't happening with the existing RNG seed.

shssoichiro:
I think this depends on https://github.com/xiph/rav1e/pull/1046 for the tests to go green.

tdaede:
Or all of the above. Can you dump the [0; 32] seed IVF?

shssoichiro:
Yes, here is the IVF from one of the failing tests (`bitrate_dav1d`): [file](condor3280.startdedicated.com/files/out-64x80-s10-q172-Cs420.ivf)

src/test_encode_decode_dav1d.rs
```diff
@@ -8,272 +8,127 @@
...
-  }
-}
+          let packet = pkt.data;
 
-macro_rules! test_speeds {
-  ($($S:expr),+) => {
-    $(
-        paste::item!{
-            #[test]
-            #[ignore]
```
lu-zero: You should get the `#[ignore]` markers in the new code as well.

src/test_encode_decode_dav1d.rs
```diff
@@ -8,272 +8,127 @@
...
-  let q = 100;
-
-  for keyint in &[4, 5, 6] {
-    encode_decode(w, h, speed, q, limit, 8, Default::default(), *keyint, *keyint, false, 0);
-  }
-}
-
-#[test]
-#[ignore]
-fn odd_size_frame_with_full_rdo() {
```
lu-zero: This seems to be the current offender, adding `#[ignore]` to this test should keep travis happy.

PR #1056, opened 2019-03-02
Remove all the remaining use of the custom aom

lu-zero:
If we want to improve the `predict.rs` test coverage we could just lift a plain implementation of the same code and use it for the tests.


build.sh
```diff
@@ -127,11 +75,11 @@ mpv --loop $DEC_FILE
 
 # Repeat for high bit depth clips
 #cargo run --bin rav1e --release -- $SEQ10 -o $ENC_FILE -s 3 -r $REC_FILE
-#${AOM_TEST}/aomdec $ENC_FILE -o $DEC_FILE
+aomdec $ENC_FILE -o $DEC_FILE
```
barrbrain: Should leave this commented out.

build.sh
```diff
@@ -127,11 +75,11 @@ mpv --loop $DEC_FILE
...
 #cmp <(tail -n+2 $DEC_FILE) <(tail -n+2 $REC_FILE)
 #mpv --loop $DEC_FILE
 
 #cargo run --bin rav1e --release -- $SEQ12 -o $ENC_FILE -s 3 -r $REC_FILE
-#${AOM_TEST}/aomdec $ENC_FILE -o $DEC_FILE
+aomdec $ENC_FILE -o $DEC_FILE
```
barrbrain: Likewise, comment.

src/predict.rs
```diff
@@ -900,389 +888,3 @@ where
...
-      let (o1, o2) = do_smooth_v_pred(&mut ra);
-      assert_eq!(o1, o2);
-
-      let (o1, o2) = do_cfl_pred(&mut ra);
-      assert_eq!(o1, o2);
-    }
-  }
-
-  #[test]
-  fn pred_matches_u8() {
```
barrbrain: This test case might not depend on libaom.

src/predict.rs
```diff
@@ -900,389 +888,3 @@ where
...
-    let mut ra = ChaChaRng::from_seed([0; 32]);
-    for _ in 0..MAX_ITER {
-      let (_, o2) = do_dc_pred(&mut ra);
-
-      assert_same(o2.data)
-    }
-  }
-
-  #[test]
-  fn pred_max() {
```
barrbrain: This test case also seems independent of libaom.

ycho:
Thanks and wow, finally this is happening! I am sorry to see this going, it has helped rav1e started walk... :)

build.sh
lu-zero: You are right :)

lu-zero:
Patch update and issues fixed :)

PR #1067, opened 2019-03-06
Simplify plane_offset()

rom1v:
There is no need to compute the superblock offset and the block offset &inside the superblock separately.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1169,14 +1169,9 @@ impl BlockOffset {
 
   /// Offset of the top-left pixel of this block.
   pub fn plane_offset(&self, plane: &PlaneConfig) -> PlaneOffset {
-    let po = self.sb_offset().plane_offset(plane);
-
-    let x_offset = self.x & LOCAL_BLOCK_MASK;
-    let y_offset = self.y & LOCAL_BLOCK_MASK;
-
     PlaneOffset {
-        x: po.x + (x_offset as isize >> plane.xdec << BLOCK_TO_PLANE_SHIFT),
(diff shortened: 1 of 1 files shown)


ycho:
Also, I've also checked with assert for verification, in case you have tested another way.
lassen:rav1e yushin$ git diff
diff --git a/src/context.rs b/src/context.rs
index ef7b267..4112c0a 100644
--- a/src/context.rs
+++ b/src/context.rs
@@ -1174,10 +1174,20 @@ impl BlockOffset {
     let x_offset = self.x & LOCAL_BLOCK_MASK;
     let y_offset = self.y & LOCAL_BLOCK_MASK;
 
-    PlaneOffset {
+    let tmp_poffset = PlaneOffset {
         x: po.x + (x_offset as isize >> plane.xdec << BLOCK_TO_PLANE_SHIFT),
         y: po.y + (y_offset as isize >> plane.ydec << BLOCK_TO_PLANE_SHIFT)
-    }
+    };
+
+    let tmp_poffset_new = PlaneOffset {
+        x: (self.x >> plane.xdec << BLOCK_TO_PLANE_SHIFT) as isize,
+        y: (self.y >> plane.ydec << BLOCK_TO_PLANE_SHIFT) as isize
+    };
+
+    assert!(tmp_poffset.x == tmp_poffset_new.x);
+    assert!(tmp_poffset.y == tmp_poffset_new.y);
+
+    tmp_poffset
   }

ycho:
oh, I just learned about ''' MarkUp, then you already did that. Thx.

PR #1068, opened 2019-03-06
Use u8 pixels for sequences of bitdepth 8

barrbrain:
Fix all of the paths that do not currently support u8 and at least match the performance of u16.

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -16,7 +16,6 @@ use crate::plane::*;
 use crate::util::{clamp, msb, Pixel, CastFromPrimitive};
 
 use std::cmp;
-use std::mem;
 
 pub struct CdefDirections {
   dir: [[u8; 8]; 8],
@@ -128,10 +127,9 @@ fn constrain(diff: i32, threshold: i32, damping: i32) -> i32 {
 // of the 2-pixel padding around the block, not the block itself.
 // The destination is unpadded.
 unsafe fn cdef_filter_block<T: Pixel>(
-  dst: *mut T, dstride: isize, input: *const T, istride: isize, pri_strength: i32,
+  dst: *mut T, dstride: isize, input: *const u16, istride: isize, pri_strength: i32,
   sec_strength: i32, dir: usize, damping: i32, xsize: isize, ysize: isize, coeff_shift: i32
 ) {
-  assert!(mem::size_of::<T>() == 2, "only implemented for u16 for now");
   let cdef_pri_taps = [[4, 2], [3, 3]];
   let cdef_sec_taps = [[2, 1], [2, 1]];
   let pri_taps = cdef_pri_taps[((pri_strength >> coeff_shift) & 1) as usize];
@@ -157,11 +155,10 @@ unsafe fn cdef_filter_block<T: Pixel>(
         let p1 = *ptr_in.offset(-cdef_directions[dir][k]);
         sum += pri_taps[k] * constrain(i32::cast_from(p0) - i32::cast_from(x), pri_strength, damping);
         sum += pri_taps[k] * constrain(i32::cast_from(p1) - i32::cast_from(x), pri_strength, damping);
-        // FIXME this is just to make it compile, but it's incorrect if T == u8
-        if p0 != T::cast_from(CDEF_VERY_LARGE) {
+        if p0 != CDEF_VERY_LARGE {
           max = cmp::max(p0, max);
         }
-        if p1 != T::cast_from(CDEF_VERY_LARGE) {
+        if p1 != CDEF_VERY_LARGE {
           max = cmp::max(p1, max);
         }
         min = cmp::min(p0, min);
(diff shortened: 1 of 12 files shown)


src/me.rs
```diff
@@ -97,18 +127,58 @@ mod nasm {
     sum
   }
 
+  #[target_feature(enable = "sse2")]
+  unsafe fn sad_sse2<T: Pixel>(
+    plane_org: &PlaneSlice<'_, T>, plane_ref: &PlaneSlice<'_, T>, blk_h: usize,
+    blk_w: usize
+  ) -> u32 {
+    assert!(mem::size_of::<T>() == 1, "only implemented for u8 for now");
```
tdaede: Uhhhh this is pretty weird. Is there syntax for writing this for just one type? E.g. sad_sse2<T: u8>? @lu-zero ?

lu-zero: I can try to look it up a shorter way but moving it to a trait seems the simplest way to achieve what you want.

rom1v: > Is there syntax for writing this for just one type? E.g. sad_sse2<T: u8>?

If the function supports only `u8` temporarily (i.e. it will also support `u16` in the future), this assert seems good.

On the contrary, if it will never support `u16`, it needs not be generic over `T`:

```rust
  unsafe fn sad_sse2_u8(
    plane_org: &PlaneSlice<'_, u8>, plane_ref: &PlaneSlice<'_, u8>, blk_h: usize,
    blk_w: usize
  ) -> u32 {
    // …
  }
```

However, it is not possible to select the implementation based on the parameter type, like in C++:

```cpp
#include <iostream>

template <typename T>
void print(T value);

template <>
void print(int value) {
    std::cout << "int: " << value << std::endl;
}

template <>
void print(double value) {
    std::cout << "double: " << value << std::endl;
}

int main() {
    print(42);
    print(42.);
}
```

```
$ ./test
int: 42
double: 42
```

The selection of the correct implementation based on the generic type can be done via an explicit test (e.g. `mem::size_of::<T>()`, or via additional methods on the `Pixel` trait).

src/cdef.rs
```diff
@@ -128,10 +127,9 @@ fn constrain(diff: i32, threshold: i32, damping: i32) -> i32 {
 // of the 2-pixel padding around the block, not the block itself.
 // The destination is unpadded.
 unsafe fn cdef_filter_block<T: Pixel>(
-  dst: *mut T, dstride: isize, input: *const T, istride: isize, pri_strength: i32,
+  dst: *mut T, dstride: isize, input: *const u16, istride: isize, pri_strength: i32,
```
ycho: Are we then roll back to Monty's original code that always used 16bit for input? I remember you have rescued him when all those Pixel Type changes (by rom1v) has perplexed him with haunting rebasing the other day.

src/rdo.rs
```diff
@@ -1271,19 +1271,18 @@ pub fn rdo_loop_decision<T: Pixel>(sbo: &SuperBlockOffset, fi: &FrameInvariants<
...
-    lrf_output = cdef_sb_frame(fi, &fs.rec);
-  }
+  assert!(fi.sequence.enable_cdef);
+  // all stages; reconstruction goes to cdef so it must be additionally padded
+  // TODO: use the new plane padding mechanism rather than this old kludge.  Will require
+  // altering CDEF code a little.
+  cdef_input = cdef_sb_padded_frame_copy(fi, sbo, &fs.rec, 2);
+  lrf_input = cdef_sb_frame(fi, &fs.rec);
+  lrf_output = cdef_sb_frame(fi, &fs.rec);
+  // FIXME: T:T padded frame copy required for CDEF disabled path.
```
ycho: So, CDEF disabled path will be handled in separate patch?
Then, CDEF cannot be disabled with this patch, correct?

src/cdef.rs
barrbrain: The input code should be much as it was before the Pixel type changes. On the output side, we now additionally have 8bit via the Pixel type.

src/rdo.rs
barrbrain: Yes, CDEF cannot be disabled after this patch. As the code was, the CDEF direction search is run even when CDEF is disabled. I'd like to follow up in a patch that enables toggling CDEF at runtime.

tdaede: Yeah,  the cdef disabling wasn't wired up to any speed level anyway.

src/plane.rs
```diff
@@ -52,21 +53,41 @@ impl<T: Pixel> Debug for Plane<T>
...
+  #[cfg(windows)]
   const STRIDE_ALIGNMENT_LOG2: usize = 4;
+  #[cfg(not(windows))]
+  const STRIDE_ALIGNMENT_LOG2: usize = 5;
 
   /// Data alignment in bytes.
+  #[cfg(windows)]
   const DATA_ALIGNMENT_LOG2: usize = 4;
+  #[cfg(not(windows))]
+  const DATA_ALIGNMENT_LOG2: usize = 5;
```
tdaede: Is this because we aren't executing AVX2 on Windows? Not sure I'd bother with this considering we'd want to eventually have it on Windows anyway.

barrbrain: This is because even without any assembly, the win64 binary segfaults with 32-byte alignment. Deferring analysis of that for when we enable nasm there.

PR #1071, opened 2019-03-06
Avoid out-of-slice accesses from PlaneMutSlice

rom1v:
The function `get_intra_edges()` acceded 1 pixel above and 1 pixel on the left of the given `PlaneSlice`. This behavior will be problematic for tiling.

Instead, make the caller pass the extended region and the offset of the origin, so that a subslice never exceeds its parent slice.

Then remove `go_up()` and `go_left()` from `PlaneMutSlice`.

---

I will probably do the same for the remaining `go_up()` and `go_left()` from `PlaneSlice` later.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -955,7 +955,23 @@ pub fn encode_tx_block<T: Pixel>(
 
   if mode.is_intra() {
     let bit_depth = fi.sequence.bit_depth;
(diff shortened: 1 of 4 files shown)


rom1v:
> Using x_offset instead of x is correct, but looks weirder than before.

Do you prefer #1082?

rom1v:
Superseeded by #1082.

Issue #1064, opened 2019-03-06
Cap thread usage


tdaede:
On my Threadripper, running with 64 threads is way slower than 1. The existing rayon code is way too fine grained for this to work. We need to cap it.

PR #1072, opened 2019-03-06
Set default number of threads to 2

gnafuthegreat:
As a temporary measure related to #1064, set default number of threads to 2 to avoid slowing things down on systems with lots of available threads.


lu-zero:
Ok, let make the API threads aware and have everything use a custom threadpool.

gnafuthegreat:
Looks like @lu-zero handled this more thoroughly with #1083, so I'm closing this PR.

PR #1082, opened 2019-03-07
Avoid out-of-slice accesses from PlaneMutSlice (alternative)

rom1v:
This is an alternative to #1071.

This PR introduces a new structure to hold a slice with its edges, so that a subslice never exceeds its parent slice, and uses it to get rid of `go_up()` and `go_left()` from `PlaneMutSlice`.

I think the resulting code is less confusing than #1071.

### DIFF
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -817,7 +817,7 @@ pub enum MvJointType {
 }
 
 pub fn get_intra_edges<T: Pixel>(
-  dst: &PlaneSlice<'_, T>,
+  dst: &EdgedPlaneSlice<'_, T>,
   tx_size: TxSize,
   bit_depth: usize,
   plane_cfg: &PlaneConfig,
@@ -825,6 +825,10 @@ pub fn get_intra_edges<T: Pixel>(
   frame_h_in_b: usize,
   opt_mode: Option<PredictionMode>
 ) -> AlignedArray<[T; 4 * MAX_TX_SIZE + 1]> {
+  let (left_edge, top_edge) = (dst.left_edge, dst.top_edge);
+  debug_assert!(left_edge <= 1);
+  debug_assert!(top_edge <= 1);
+
   let mut edge_buf: AlignedArray<[T; 4 * MAX_TX_SIZE + 1]> =
     UninitializedAlignedArray();
   let base = 128u16 << (bit_depth - 8);
@@ -834,8 +838,9 @@ pub fn get_intra_edges<T: Pixel>(
     let (left, not_left) = edge_buf.array.split_at_mut(2*MAX_TX_SIZE);
     let (top_left, above) = not_left.split_at_mut(1);
 
-    let x = dst.x;
-    let y = dst.y;
+    let dst_content = dst.without_edges();
+    let x = dst_content.x as usize;
+    let y = dst_content.y as usize;
 
     let mut needs_left = true;
     let mut needs_topleft = true;
@@ -845,7 +850,7 @@ pub fn get_intra_edges<T: Pixel>(
 
     if let Some(mut mode) = opt_mode {
       mode = match mode {
-        PredictionMode::PAETH_PRED => match (x, y) {
+        PredictionMode::PAETH_PRED => match (left_edge, top_edge) {
           (0, 0) => PredictionMode::DC_PRED,
           (_, 0) => PredictionMode::H_PRED,
           (0, _) => PredictionMode::V_PRED,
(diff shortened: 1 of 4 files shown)


src/plane.rs
```diff
@@ -168,6 +175,12 @@ impl<T: Pixel> Plane<T> {
     self.mut_slice(&PlaneOffset { x: 0, y: 0 })
   }
 
+  #[inline]
+  pub fn edged_slice(&self, po: &PlaneOffset, left_edge: usize, top_edge: usize) -> EdgedPlaneSlice<'_, T> {
```
lu-zero: `left_edge` and `top_edge` could be anything different than 1.min(po.x) or 1.min(po.y) ?

rom1v: There are `go_left(3)` and `go_up(3)` with `PlaneSlice` elsewhere in the code base.

But I think the `.min(po.x)` could be done there.

lu-zero: it would make the code even simpler :)

rom1v: I updated the PR :)

src/partition.rs
```diff
@@ -857,24 +862,23 @@ pub fn get_intra_edges<T: Pixel>(
...
+      needs_top = mode != PredictionMode::H_PRED && (!dc_or_cfl || top_edge != 0);
       needs_topright = mode == PredictionMode::D45_PRED || mode == PredictionMode::D63_PRED;
       needs_bottomleft = mode == PredictionMode::D207_PRED;
     }
 
     // Needs left
     if needs_left {
-      if x != 0 {
-        let left_slice = dst.go_left(1);
+      if left_edge != 0 {
```
ycho: Shouldn't this be "x != left_edge" ? I think current code will only work for 1st tile.
If we start to use multiple tiles and x and y come from "dst_content = dst.without_edges();",
then I think the border condition check here and all below are wrong.
We also needs asserts to check x >= left_edge and y >= top_edge?

Can you explain why not using 
x = dst.x; and  y = dst.y; 
given that, "dst: &EdgedPlaneSlice"?

src/partition.rs
```diff
@@ -883,20 +887,18 @@ pub fn get_intra_edges<T: Pixel>(
...
-        _ => dst.go_up(1).go_left(1).p(0, 0)
+        _ => dst.ps[0][0],
       };
     }
 
     // Needs top
     if needs_top {
-      if y != 0 {
-        above[..tx_size.width()].copy_from_slice(&dst.go_up(1)[0][..tx_size.width()]);
+      if top_edge != 0 {
```
ycho: Should be y != top_edge

src/partition.rs
```diff
@@ -883,20 +887,18 @@ pub fn get_intra_edges<T: Pixel>(
...
+        above[..tx_size.width()].copy_from_slice(&dst.ps[0][left_edge..left_edge + tx_size.width()]);
       } else {
-        let val = if x != 0 { dst.go_left(1).p(0, 0) } else { T::cast_from(base - 1) };
+        let val = if left_edge != 0 { dst.ps[0][0] } else { T::cast_from(base - 1) };
```
ycho: should be x != left_edge

src/partition.rs
```diff
@@ -908,27 +910,25 @@ pub fn get_intra_edges<T: Pixel>(
...
+        y: y >> (2 - plane_cfg.ydec)
+      };
 
       let bsize = BlockSize::from_width_and_height(
           tx_size.width() << plane_cfg.xdec,
           tx_size.height() << plane_cfg.ydec
         );
 
-      let num_avail = if y != 0 && has_tr(&bo, bsize) {
+      let num_avail = if top_edge != 0 && has_tr(&bo, bsize) {
```
ycho: should be y != top_edge

src/partition.rs
```diff
@@ -943,28 +943,25 @@ pub fn get_intra_edges<T: Pixel>(
...
+        y: y >> (2 - plane_cfg.ydec)
         };
 
       let bsize = BlockSize::from_width_and_height(
         tx_size.width() << plane_cfg.xdec,
         tx_size.height() << plane_cfg.ydec
         );
 
-      let num_avail = if x != 0 && has_bl(&bo, bsize) {
+      let num_avail = if left_edge != 0 && has_bl(&bo, bsize) {
```
ycho: should be x != left_edge

src/partition.rs
rom1v: > Shouldn't this be "x != left_edge" ?

`left_edge` represents the number of pixels in the left edge.

PR #1084, opened 2019-03-07
Expose T::size() for convenience

rom1v:
Replace `std::mem::size_of::<T>()` by `T::size()`.

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -16,7 +16,6 @@ use crate::plane::*;
 use crate::util::{clamp, msb, Pixel, CastFromPrimitive};
 
 use std::cmp;
-use std::mem;
 
 pub struct CdefDirections {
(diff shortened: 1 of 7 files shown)


tmatth:
Is this evaluated at compile time?

rom1v:
> Is this evaluated at compile time?

That's what I was wondering. I guess that in practice it is due to `#[inline(always)]` which inlines a raw literal (`1` or `2`).

Unfortunately, a trait function may not be `const`, so this is not as powerful as `size_of::<T>`. For example, we could not create an array having type `[i32; T::size()]`.

So I think we should just keep `mem::size_of()`.

shssoichiro:
Personally I don't see what this gains s, other than fewer characters when using `T::size()`. `mem::size_of` is a const fn and inlineable, so performance between the two should be equivalent.

Issue #1079, opened 2019-03-07
Document the Color information Structs


lu-zero:
- [x] ChromaSampling
- [x] ChromaSamplePosition
- [x] PixelRange
- [x] MatrixCoefficients
- [x] ColorPrimaries
- [x] TransferCharacteristics
- [x] ColorDescription
- [x] MasteringDisplay
- [x] ContentLight

Issue #1076, opened 2019-03-07
Document Config


lu-zero:
- [ ] Document the public fields
- [ ] Document the impl

Issue #1075, opened 2019-03-07
Document Context


lu-zero:
- [ ] Document the public fields
- [ ] Document the impl

PR #1089, opened 2019-03-10
Add documentation to the API

vibhoothiiaanand:
This commit Adds documentation to multiple places
Fixes the follwing issues
- [ ] #1075 
- [ ] #1076 
- [ ] #1077 
- [ ] #1078 
- [ ] #1079 

I Created a single PR so it will be easy for reviewing and closing rather than 5 different PRs for 5 issues.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -408,7 +408,7 @@ impl Default for ColorPrimaries {
 }
 
 arg_enum!{
-/// Contains differnt types of Transfer Characteristics, they define the optoelectronic tranfer charecteristic of the source picture aka Gamma.
+  /// Contains differnt types of Transfer Characteristics, they define the optoelectronic tranfer charecteristic of the source picture aka Gamma.
   #[derive(Debug, Clone, Copy, PartialEq)]
   #[repr(C)]
   pub enum TransferCharacteristics {
@@ -543,6 +543,7 @@ impl Config {
   }
 }
 
+/// Contain all the context configurations
 pub struct Context<T: Pixel> {
   //    timebase: Rational,
   frame_count: u64,
@@ -604,6 +605,8 @@ impl<T: Pixel> fmt::Display for Packet<T> {
 }
 
 impl<T: Pixel> Context<T> {
+
+  /// Create new frame  with details like width, height and also chroma sampling from the encoder
   pub fn new_frame(&self) -> Arc<Frame<T>> {
     Arc::new(Frame::new(
       self.config.enc.width,
@@ -612,6 +615,7 @@ impl<T: Pixel> Context<T> {
     ))
   }
 
+  /// Send the details of the frame by passing the frame details as input
   pub fn send_frame<F>(&mut self, frame: F) -> Result<(), EncoderStatus>
   where
     F: Into<Option<Arc<Frame<T>>>>,
@@ -623,32 +627,39 @@ impl<T: Pixel> Context<T> {
     Ok(())
   }
 
+  /// Returns the details of a specific frame
   pub fn get_frame(&self, frame_number: u64) -> Arc<Frame<T>> {
     // Clones only the arc, so low cost overhead
     self.frame_q.get(&frame_number).as_ref().unwrap().as_ref().unwrap().clone()
   }
 
+  /// Returns the frame count
   pub fn get_frame_count(&self) -> u64 {
     self.frame_count
   }
 
+  /// Set the limit which is passed as argument
   pub fn set_limit(&mut self, limit: u64) {
     self.limit = limit;
   }
 
+  /// Allow to read frames ahead of current ones
   pub fn needs_more_lookahead(&self) -> bool {
     self.needs_more_frames(self.frame_count) && self.frames_processed + LOOKAHEAD_FRAMES > self.frame_q.keys().last().cloned().unwrap_or(0)
   }
 
+  /// Request for more frames
   pub fn needs_more_frames(&self, frame_count: u64) -> bool {
     self.limit == 0 || frame_count < self.limit
   }
 
+  /// Video sequences are headed by a sequence header which was starting value, various other things like chroma sampling x and y, sequence bit depth
   pub fn container_sequence_header(&mut self) -> Vec<u8> {
     fn sequence_header_inner(seq: &Sequence) -> io::Result<Vec<u8>> {
       let mut buf = Vec::new();
 
       {
+        // Create a bitwriter instance with BigEndian type ie. an order in which the "big end" (most significant value in the sequence) is stored first.
         let mut bw = BitWriter::endian(&mut buf, BigEndian);
         bw.write_bit(true)?; // marker
         bw.write(7, 1)?; // version
@@ -775,6 +786,10 @@ impl<T: Pixel> Context<T> {
     (fi, true)
   }
 
+  /// Receive packet functions provide an Encode API with decoupled packets/frames, On sucess it returns packet with compressed frame.
+  /// If it has needs_more_lookahead and if everything is encoded then it will be returning the encode status as "NeedMoreFrames"
+  /// Frames to analyze lookahead, when the frame queue is nearly full it returns with status as "Enough Data". At the begining the encoder
+  /// accepts multiple frames as input untill the buffer is full. When it is end of stream it requires flushing aka. draining.
   pub fn receive_packet(&mut self) -> Result<Packet<T>, EncoderStatus> {
     if self.needs_more_lookahead() {
       return Err(EncoderStatus::NeedMoreFrames);
@@ -910,6 +925,9 @@ impl<T: Pixel> Context<T> {
     }
   }
 
+  /// Flush the output, the flush makes the current frame to the frame_q and then increments the frame count for the next process.
+  /// This also means that some packets might not produce output immediately and might requrie to flushed at the end of decoding to
+  /// get all the decoded data.
   pub fn flush(&mut self) {
     self.frame_q.insert(self.frame_count, None);
     self.frame_count += 1;
--- a/src/api.rs
+++ b/src/api.rs
@@ -48,7 +48,8 @@ pub struct Point {
   pub y: u16
 }
 
-/// Contain all the encoder configuration
+/// Contain all the encoder configuration which are taken for encoded output.
+/// Encoder configurations which doesn't affect the encoded output are in 'config'.
 #[derive(Clone, Debug)]
 pub struct EncoderConfig {
   // output size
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -89,6 +93,10 @@ impl Default for EncoderConfig {
 }
 
 impl EncoderConfig {
+  /// This is the preset set with width and height to 640 and 480 respectively,
```
lu-zero: Maybe omit the details that could and will change and just mention that provides a default setting according to a speed value in the range `0-10`.

Mention what happens when you pass a number higher than `10`

src/api.rs
```diff
@@ -81,6 +82,9 @@ pub struct EncoderConfig {
   pub train_rdo: bool,
 }
 
+/// Default preset for EncoderConfig, In this DEFAULT_SPEED is set to 5 
+/// and called the with_speed_preset with this value(5). We are using 5 
+/// as it is in the middle so it will have good balance of quality and speed.
```
lu-zero: Omit internal information that might or might not change, just mention that the default is a balance between speed and quality and mention that the user could use `with_speed_preset` to have a default structure with a specific speed level.

src/api.rs
```diff
@@ -154,6 +164,9 @@ impl Default for SpeedSettings {
 }
 
 impl SpeedSettings {
+  /// This preset is set to assign the speed, where the speed is passed as arguement,
+  /// the parametes like min_block_size, multi reference preditions, distortion are assigned,
+  /// for instance  *speed_settings: SpeedSettings::from_preset(10);*
```
lu-zero: "Set the speed setting according to a numeric speed preset"

Mention which is the range and what's slower/higher quality and what's faster.

src/api.rs
```diff
@@ -282,14 +296,16 @@ pub enum ChromaSampling {
   Cs400,
 }
 
+/// Make 4:2:0 as default chroma sampling type as it is almost loseless since it
+/// is removing half of the horizondtal and half of vertical color information
```
lu-zero: It is not lossless. You may omit the default description.

src/api.rs
```diff
@@ -309,13 +326,15 @@ pub enum ChromaSamplePosition {
...
 
+/// Assign default ChromaSamplePosition as Unknown
 impl Default for ChromaSamplePosition {
   fn default() -> Self {
     ChromaSamplePosition::Unknown
   }
 }
 
 arg_enum!{
+  /// Contains details of Pixel color range
```
lu-zero: It does not add more information, you may add a reference to what is a color range is and the ITU H.273 section that mentions it.

src/api.rs
```diff
@@ -325,13 +344,15 @@ arg_enum!{
   }
 }
 
+/// Assign default of PixelRange as Unspecified
```
lu-zero: Omit it.

src/api.rs
```diff
@@ -325,13 +344,15 @@ arg_enum!{
...
 impl Default for PixelRange {
     fn default() -> Self {
         PixelRange::Unspecified
     }
 }
 
 arg_enum!{
+  /// Contains different types of Matrix Coefficients which is used in deriving the luma and chroma (YCbCr) from the R'G'B' primaries
```
lu-zero: A reference to ITU H.273 would be good.

src/api.rs
```diff
@@ -352,13 +373,15 @@ arg_enum!{
   }
 }
 
+/// Assign default of MatrixCoefficients as Unspecified
```
lu-zero: Omit it.

src/api.rs
```diff
@@ -377,13 +400,15 @@ arg_enum!{
   }
 }
 
+/// Assign default of ColorPrimaries as Unspecified
```
lu-zero: Omit it.

src/api.rs
```diff
@@ -377,13 +400,15 @@ arg_enum!{
...
 impl Default for ColorPrimaries {
     fn default() -> Self {
         ColorPrimaries::Unspecified
     }
 }
 
 arg_enum!{
+  /// Contains differnt types of Transfer Characteristics, they define the optoelectronic tranfer charecteristic of the source picture aka Gamma.
```
lu-zero: Typo here -> `different`.

(mention to the H.273 section would be great)

src/api.rs
```diff
@@ -407,19 +432,23 @@ arg_enum!{
   }
 }
 
+/// Assign default of TransferCharacteristics as Unspecified
```
lu-zero: Feel free to omit this.

src/api.rs
```diff
@@ -445,6 +476,7 @@ pub struct Config {
 const MAX_USABLE_THREADS: usize = 4;
 
 impl Config {
+  /// Contain the values of the encoder configuration which are to be parsed
```
lu-zero: "Sets a specific configuration `key` parameter by parsing the `value` string"

src/api.rs
```diff
@@ -459,11 +491,13 @@ impl Config {
 
     Ok(())
   }
-
+  /// Create a new context for the encoder. Context contatins
```
lu-zero: The context is the full encoder state.

Please omit implementation details that may and will change and mention what the context is useful for :)

src/api.rs
```diff
@@ -477,7 +511,7 @@ impl Config {
     };
 
     let pool = rayon::ThreadPoolBuilder::new().num_threads(threads).build().unwrap();
-
+    // Initialize default values to context objects
```
lu-zero: May be omitted.

src/api.rs
```diff
@@ -509,6 +543,7 @@ impl Config {
   }
 }
 
+/// Contain all the context configurations
```
lu-zero: Omit it or expand it.

src/api.rs
```diff
@@ -570,6 +605,8 @@ impl<T: Pixel> fmt::Display for Packet<T> {
 }
 
 impl<T: Pixel> Context<T> {
+
+  /// Create new frame  with details like width, height and also chroma sampling from the encoder
```
lu-zero: "Create a new empty frame according to the current encoder settings"

src/api.rs
```diff
@@ -578,6 +615,7 @@ impl<T: Pixel> Context<T> {
     ))
   }
 
+  /// Send the details of the frame by passing the frame details as input
```
lu-zero: Send a frame to be encoded.

Mention that the function may return an error if the input queue is full.

src/api.rs
```diff
@@ -589,32 +627,39 @@ impl<T: Pixel> Context<T> {
     Ok(())
   }
 
+  /// Returns the details of a specific frame
```
lu-zero: You may omit this, (and this function should not be public btw).

src/api.rs
```diff
@@ -589,32 +627,39 @@ impl<T: Pixel> Context<T> {
...
   pub fn get_frame(&self, frame_number: u64) -> Arc<Frame<T>> {
     // Clones only the arc, so low cost overhead
     self.frame_q.get(&frame_number).as_ref().unwrap().as_ref().unwrap().clone()
   }
 
+  /// Returns the frame count
```
lu-zero: You should explain what the frame count is.

src/api.rs
```diff
@@ -589,32 +627,39 @@ impl<T: Pixel> Context<T> {
...
   pub fn get_frame_count(&self) -> u64 {
     self.frame_count
   }
 
+  /// Set the limit which is passed as argument
```
lu-zero: You should explain what the limit is.

src/api.rs
```diff
@@ -589,32 +627,39 @@ impl<T: Pixel> Context<T> {
...
   pub fn set_limit(&mut self, limit: u64) {
     self.limit = limit;
   }
 
+  /// Allow to read frames ahead of current ones
```
lu-zero: This function should not be public. (and does not do that)

src/api.rs
```diff
@@ -589,32 +627,39 @@ impl<T: Pixel> Context<T> {
...
   pub fn needs_more_lookahead(&self) -> bool {
     self.needs_more_frames(self.frame_count) && self.frames_processed + LOOKAHEAD_FRAMES > self.frame_q.keys().last().cloned().unwrap_or(0)
   }
 
+  /// Request for more frames
```
lu-zero: Returns true if more frames should be sent to the encoder in order.

Mention `send_frame()`.

src/api.rs
```diff
@@ -589,32 +627,39 @@ impl<T: Pixel> Context<T> {
...
   pub fn needs_more_frames(&self, frame_count: u64) -> bool {
     self.limit == 0 || frame_count < self.limit
   }
 
+  /// Video sequences are headed by a sequence header which was starting value, various other things like chroma sampling x and y, sequence bit depth
```
lu-zero: You should document what the function returns. Bonus point if you add a reference to the AV1 specification.

src/api.rs
```diff
@@ -741,6 +786,10 @@ impl<T: Pixel> Context<T> {
     (fi, true)
   }
 
+  /// Receive packet functions provide an Encode API with decoupled packets/frames, On sucess it returns packet with compressed frame.
```
lu-zero: Do not mention implementation details that may and should be changed soon :)

"Returns an encoded packet or signals that additional frame data must be provided"

Mention `flush()` and `send_frame()` .

src/api.rs
```diff
@@ -876,6 +925,9 @@ impl<T: Pixel> Context<T> {
     }
   }
 
+  /// Flush the output, the flush makes the current frame to the frame_q and then increments the frame count for the next process.
```
lu-zero: Please omit internal details.

Mention `receive_packet()` and `send_frame()`.

src/api.rs
```diff
@@ -48,6 +48,7 @@ pub struct Point {
   pub y: u16
 }
 
+/// Contain all the encoder configuration
```
shssoichiro: Can you add that this should contain settings which affect the encoded output? Any options which do not affect the output should be in `Config` instead.

vibhoothiiaanand: Sure :) 
Updated

vibhoothiiaanand: Done

vibhoothiiaanand:
```


/// Contain all the encoder configuration which are taken for encoded output.
/// Encoder configurations which doesn't affect the encoded output are in 'config'.
#[derive(Clone, Debug)]
```

src/api.rs
vibhoothiiaanand: Omitting it as the function has explanation below

rzumer:
Did you forget to push changes?

vibhoothiiaanand:
> Did you forget to push changes? I don't see the fixes that should be there.

I didn't push, was bit busy with Midterms and project proposal and also was waiting for Luca to fix up the issues of cargo doc. Now all good, Will update few more then rebase and will push :)

lu-zero:
Please open a new pr with a single patch addressing a single issue. (Possibly not touching the Context since is being actively reworked)

PR #1090, opened 2019-03-10
Use cargo fetch to generate the Cargo.lock needed by kcov

lu-zero:


tdaede:
Small speed improvement?

lu-zero:
Really tiny if ever, I'm thinking about restricting the tests run for the coverage instead

PR #1092, opened 2019-03-11
Remove fake-genericity from sad functions

rom1v:
The functions `sad_ssse3()` and `sad_sse2()` only support `u16` and `u8` respectively, so they are not generic.

Make the caller pass the expected type.

sad

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -92,7 +92,7 @@ mod nasm {
   }
 
   #[target_feature(enable = "ssse3")]
-  unsafe fn sad_ssse3(
+  unsafe fn sad_hbd_ssse3(
     plane_org: &PlaneSlice<'_, u16>, plane_ref: &PlaneSlice<'_, u16>, blk_h: usize,
     blk_w: usize, bit_depth: usize
   ) -> u32 {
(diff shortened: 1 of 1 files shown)


src/me.rs
```diff
@@ -92,14 +92,13 @@ mod nasm {
   }
 
   #[target_feature(enable = "ssse3")]
-  unsafe fn sad_ssse3<T: Pixel>(
-    plane_org: &PlaneSlice<'_, T>, plane_ref: &PlaneSlice<'_, T>, blk_h: usize,
+  unsafe fn sad_ssse3(
```
barrbrain: Perhaps we can rename `sad_ssse3` to `sad_hbd_ssse3` for clarity.

src/me.rs
```diff
@@ -175,11 +173,15 @@ mod nasm {
     {
       if mem::size_of::<T>() == 2 && is_x86_feature_detected!("ssse3") && blk_h >= 4 && blk_w >= 4 {
         return unsafe {
+          let plane_org = &*(plane_org as *const _ as *const PlaneSlice<'_, u16>);
+          let plane_ref = &*(plane_ref as *const _ as *const PlaneSlice<'_, u16>);
```
barrbrain: Wow, pointer casting is ugly. I guess that is by design so that we feel bad doing it. 😅

src/me.rs
rom1v: hbd?

barrbrain: This function delegates to `rav1e_sad_4x4_hbd_ssse3` and friends. The high bit depth (hbd) label follows dav1d convention. In libaom, highbd is more commonly used.

rom1v: :+1: done

PR #1093, opened 2019-03-11
Use a procedural macro instead a local copy of arg_enum

lu-zero:
Make cargo doc happier and make crav1e not depend on compiler
unstable feature.


tdaede:
Looks prettier, OK with me.

PR #1096, opened 2019-03-12
cdef: make sec_tap depend on sec_strength

tmatth:
Currently this will have no impact because pri_strength == sec_strength
and cdef_sec_taps[0] == cdef_sec_taps[1].


tmatth:
@barrbrain ah good catch, the spec makes this much clearer.

PR #1102, opened 2019-03-13
Plane custom aligned storage

### LINKED ISSUE #1101: Replace Plane's data vec with a custom aligned type
We currently create the vec with from_raw_parts, but this is illegal because we need to have a custom free function.

lu-zero:
To fix #1101 I'll need to rework `Plane::data` a bit.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,7 +17,6 @@ extern crate bincode;
 extern crate dav1d_sys;
 
 #[cfg(test)]
-#[macro_use]
 extern crate interpolate_name;
 
 #[cfg(test)]
--- a/src/plane.rs
+++ b/src/plane.rs
@@ -42,7 +42,6 @@ pub struct PlaneOffset {
 pub struct PlaneData<T: Pixel> {
   ptr: std::ptr::NonNull<T>,
   _marker: PhantomData<T>,
-  layout: Layout,
   len: usize,
 }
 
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -17,7 +17,6 @@ extern crate bincode;
 extern crate dav1d_sys;
 
 #[cfg(test)]
-#[macro_use]
```
barrbrain: This is used by features decode_test and decode_test_dav1d.

lu-zero: I forgot to add the `use interpolate_name::interpolate_test` indeed, now fixed :)

lu-zero:
@tdaede could you please give it a go on Windows?

src/plane.rs
```diff
@@ -38,9 +38,105 @@ pub struct PlaneOffset {
...
+    }
+  }
+}
+
+impl<T: Pixel> PlaneData<T> {
+  /// Data alignment in bytes.
+  #[cfg(windows)]
+  const DATA_ALIGNMENT_LOG2: usize = 4;
+  #[cfg(not(windows))]
+  const DATA_ALIGNMENT_LOG2: usize = 5;
```
rom1v: Why is it different for Windows?

barrbrain: This is rolled over from the existing aligned allocation. That gave SEGFAULT on win64 for `DATA_ALIGNMENT_LOG2=5`, presumably because of an illegal `dealloc` with unmatching alignment. We should be able to drop it safely now that we have our own container.

PR #1106, opened 2019-03-13
Call assembly functions on Windows.

tdaede:


src/mc.rs
```diff
@@ -7,7 +7,7 @@
 // Media Patent License 1.0 was not distributed with this source code in the
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
-#[cfg(all(target_arch = "x86_64", not(windows), feature = "nasm"))]
+#[cfg(all(target_arch = "x86_64", feature = "nasm"))]
 pub use self::nasm::*;
 #[cfg(any(not(target_arch = "x86_64"), windows, not(feature = "nasm")))]
```
barrbrain: Need to update the counter-condition also.
```suggestion
#[cfg(any(not(target_arch = "x86_64"), not(feature = "nasm")))]
```

EwoutH:
In a very quick run on 20 frames rav1e was 44% faster with its assembly functions! The output it not bit exact however, is that expected?
```
rav1e0314 "E:\Morocco8K-8b-1920.y4m" -o NULL.ivf -l 20 --psnr
1920x1080 @ 19001/317 fps
encoded 20/20 frames, 0.099 fps, 1013.03 Kb/s, est. size: 0.04 MB, est. time: 0 s
Key Frames:      1    avg size:   18877 B
Inter:          19    avg size:    1283 B
Mean PSNR: Y: 47.7122  Cb: 51.6057  Cr: 50.7179  Avg: 50.0119

rav1e0314-asm "E:\Morocco8K-8b-1920.y4m" -o NULL.ivf -l 20 --psnr
1920x1080 @ 19001/317 fps
encoded 20/20 frames, 0.143 fps, 1008.80 Kb/s, est. size: 0.04 MB, est. time: 0 s
Key Frames:      1    avg size:   18877 B
Inter:          19    avg size:    1274 B
Mean PSNR: Y: 47.7231  Cb: 51.6010  Cr: 50.7246  Avg: 50.0163
```

tdaede:
uhhhh... no that's not expected, as we are supposed to have unit tests that catch that. Could you maybe try checking out previous version of individual files and narrow down the difference? (I'll also check Linux, as it's probably happening there too)

tdaede:
I can reproduce this on Linux. Dunno if that should block the patch (Windows now because the same as Linux/Mac) but we should still fix it.

barrbrain:
The difference between RUst-native and assembly only seems to affect RDO, no decode mismatches have been witnessed. We can chase it down after this. The only thing blocking my approval is the comment I left on `src/mc.rs`.

PR #1107, opened 2019-03-13
Retrieve size from plane_cfg in get_intra_edges()

rom1v:
To compute the number of available pixels in an edge, `get_intra_edges()` used `frame_w_in_b` (`MiCols`) and `frame_h_in_b` (`MiRows`), initialized [as follow](https://aomediacodec.github.io/av1-spec/#compute-image-size-function):

    MiCols = 2 * ( ( FrameWidth + 7 ) >> 3 )
    MiRows = 2 * ( ( FrameHeight + 7 ) >> 3 )

As a consequence, the sizes computed by `get_intra_edges()` were basically the frame size rounded up to the next multiple of 8 (i.e. including some
padding), decimated:

    (MI_SIZE >> plane_cfg.xdec) * frame_w_in_b
    (MI_SIZE >> plane_cfg.ydec) * frame_h_in_b

Using edges pixels in the frame padding here will make no sense with tiling, so it should be equivalent to just using the plane size.

---
I don't really know why it was implemented that way. Maybe there was a good reason.

ihniwid

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -963,8 +963,6 @@ pub fn encode_tx_block<T: Pixel>(
       tx_size,
       bit_depth,
       &fs.input.planes[p].cfg,
-      fi.w_in_b,
-      fi.h_in_b,
       Some(mode),
     );
     mode.predict_intra(&mut rec.mut_slice(po), tx_size, bit_depth, &ac, alpha, &edge_buf);
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -821,8 +821,6 @@ pub fn get_intra_edges<T: Pixel>(
   tx_size: TxSize,
   bit_depth: usize,
   plane_cfg: &PlaneConfig,
-  frame_w_in_b: usize,
-  frame_h_in_b: usize,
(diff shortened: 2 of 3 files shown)


tdaede:
Because this rounded up to 8, but tx_size goes down to 4, this won't always produce the same result, will it?

rom1v:
> Because this rounded up to 8, but tx_size goes down to 4, this won't always produce the same result, will it?

Yes, I agree, the result is not necessarily the same, but I hoped this was harmless. :smile: 

In fact, I know understand why it can work with tiling: in `tile_info()`, the last tile size is initialized from `MiCols` and `MiRows`, not `FrameWidth` and `FrameHeight`, so the last tile also has this padding:

```
MiColStarts[i] = MiCols
// …
MiRowStarts[i] = MiRows
```

(and now, I have to adapt my tiling iterator to match this behavior)

rom1v:
Oh, but the plane dimensions are also rounded up to the next multiple of 8.

I opened a new PR: https://github.com/xiph/rav1e/pull/1109.

PR #1110, opened 2019-03-14
Use sccache in CI scripts

barrbrain:
For local builds this can halve build times with a hot cache. In CI there will be some overhead managing the cache.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,10 @@ before_install:
     - ninja && sudo ninja install
     - cd ../..
 
+cache:
+  directories:
+  - "$HOME/.cache/sccache"
+
 after_script:
 - sccache -s
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,17 +13,15 @@ before_install:
     - hash -r
     - which cmake
     - cmake --version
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -13,27 +13,37 @@ before_install:
...
     - mkdir -p .build
     - cd .build
-    - cmake -GNinja .. -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=0 -DENABLE_DOCS=0 -DCONFIG_LOWBITDEPTH=1 -DCMAKE_INSTALL_PREFIX=/usr -DCONFIG_PIC=1
+    - cmake -GNinja .. -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=0 -DENABLE_DOCS=0 -DCONFIG_LOWBITDEPTH=1 -DCMAKE_INSTALL_PREFIX=/usr -DCONFIG_PIC=1
     - ninja && sudo ninja install
     - cd ../..
 
+cache:
+  directories:
+  - "$HOME/.cache/sccache"
```
lu-zero: I wonder if adding .cargo in the cache would give us even more gains.

barrbrain: I'll follow that up separately. I think this PR is a complete integration of sccache.

PR #1111, opened 2019-03-14
Fix get_sad() tests

rom1v:
In tests, the function `get_sad()` was called with block width and block height parameters swapped.

As a consequence, associate the precomputed SAD values to the transposed block size.


shssoichiro:
It might be better to fix the function itself so it takes the parameters in the "normal" order. We do WxH everywhere else, so this is likely to throw people off in the future if it's different in one method.

rom1v:
@shssoichiro I agree, but I'd prefer to swap the parameters everywhere later, because every line will conflict with my local branch about tiling.

PR #1112, opened 2019-03-14
Cache the cargo directory in CI

barrbrain:
This is potentially a substantial saving on the time spent in the prelude of CI tasks. The downsides are less clear than with sccache. Proceeding with cautious optimism.


barrbrain:
I compared a build with hot cache before and after this change. There was no clear improvement in build times.

Issue #689, opened 2018-10-30
Low latency API


tguillem:
Hello,

I just read the C API and have a remark.

The current API doesn't allow to use the encoder asynchronously.
What most encoders and rav1e are doing is: queuing input frames and dequeing output packets from the same thread or same loop.

Queuing input frames is generally paced by I/O. If your encoder has an output packet ready but the application using your API is stuck in I/O, the output packet will be sent with a high delay.

To solve this issue, the encoder should either provide a callback, that is called when an output packet is encoded. In that case, one application could process the output packet directly from this callback.

An other way to solve this is issue is to make your queuing and dequeing functions thread-safe (rav1e_send_frame() and rav1e_receive_packet()). That way, one application could wait for output packets from their own thread via rav1e_receive_packet() (you should then use a condition variable somewhere). You should also provide a way to cancel the previous call, in order to interrupt the output thread without delays.

ycho:
Yes, I agree that rav1e needs this better way of handling input frames to it and output bitstream from it asynchronously. Thank you for pointing this out!

vibhoothiiaanand:
With help of #2118 Channel API, we can achieve something similar :)

PR #1117, opened 2019-03-15
Rework the internal encoding logic

lu-zero:
- [ ] move the encoding loop in a stand alone thread (needed by #689)
- [x] feed the frames to the scenechange detector, queue them up until:
  - the scenechange is detected
  - the limit has been hit
  - a flush has been hit
- [ ] feed the group to the reordering logic and prepare the frame-invariants
- [ ] encode the groups using a threadpool

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -9,6 +9,8 @@
 
 use arg_enum_proc_macro::ArgEnum;
 use bitstream_io::*;
+use crossbeam::channel::*;
+
 use crate::encoder::*;
 use crate::metrics::calculate_frame_psnr;
 use crate::partition::*;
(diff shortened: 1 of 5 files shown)


src/api.rs
```diff
@@ -25,7 +25,7 @@ use std::sync::Arc;
 use std::collections::BTreeSet;
 use std::path::PathBuf;
 
-const LOOKAHEAD_FRAMES: u64 = 10;
+const LOOKAHEAD_FRAMES: u64 = 7;
```
shssoichiro: Out of curiosity, what is the significance of changing this to 7?

lu-zero: Letting me test on a sample of 20 frames w/out having to edit it :) (I set it back to 10 now)

lu-zero:
Now I have something that builds but it is almost surely wrong (off by ones and other logic faults expected). Once it proves to work the code can be simplified a lot.

PR #1119, opened 2019-03-15
Avoid superfluous memset in transforms and coeff coding

barrbrain:
Benchmarks `write_b_bench(TX_4X4, _)` improved by ~20%; `write_b_bench(TX_8X8, _)` by ~8%.

### DIFF
--- a/src/transform/forward.rs
+++ b/src/transform/forward.rs
@@ -1740,7 +1740,8 @@ trait FwdTxfm2D: Dim {
     input: &[i16], output: &mut [i32], stride: usize, tx_type: TxType,
     bd: usize
   ) {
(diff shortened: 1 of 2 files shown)


src/transform/forward.rs
```diff
@@ -1855,13 +1856,13 @@ pub fn fht64x64(
   bit_depth: usize
 ) {
   assert!(tx_type == TxType::DCT_DCT);
-  let mut tmp = [0 as i32; 4096];
+  let mut tmp: AlignedArray<[i32; 4096]> = UninitializedAlignedArray();
```
lu-zero: maybe keep the array in a second variable and make tmp a reference to `.array`.

Issue #1081, opened 2019-03-07
Frame flushing regression - last sub GOP constructed incorrectly


SmilingWolf:
I perform semi-regular encoder benchmarks on a set of anime-centered clips.
While preparing this month's report I noticed a massive drop in BD rate (up to 12% less efficiency).

I bisected the commits between the last "good" tests, performed on commit 64b9f501, and the one where I first noticed the drop, commit 57e4f926. Turned out the bad commit is 3202bf70 (ping @shssoichiro):
```
# sh bd_rate.sh /c/Users/utente/Desktop/regressionTesting/rav1e-0.1.0-1219-be245fc0.out /c/Users/utente/Desktop/regressionTesting/rav1e-0.1.0-1222-3202bf70.out
        RATE (%) DSNR (dB)
 MSSSIM 12.5511 -0.502696
PSNRHVS 12.6137 -0.699069
  HVMAF 12.4122 -1.06813
```
BD rate report up to the commit before the bad one:
```
# sh bd_rate.sh /c/Users/utente/Desktop/regressionTesting/rav1e-0.1.0-977-64b9f501.out /c/Users/utente/Desktop/regressionTesting/rav1e-0.1.0-1219-be245fc0.out
        RATE (%) DSNR (dB)
 MSSSIM -3.70748 0.161186
PSNRHVS -3.5851 0.215709
  HVMAF -5.32006 0.524379
```
BD rate between 64b9f501 and a more recent one (d3b81ca1):
```
# sh bd_rate.sh /c/Users/utente/Desktop/regressionTesting/rav1e-0.1.0-977-64b9f501.out /c/Users/utente/Desktop/regressionTesting/rav1e-0.1.0-1768-d3b81ca1.out
        RATE (%) DSNR (dB)
 MSSSIM 9.60217 -0.394358
PSNRHVS 9.65825 -0.547198
  HVMAF 6.88864 -0.659584
```
The regression can be reliably replicated on a variety anime clips with different resolutions (1280x720, 1920x1080), so I'm sure it's not just a pathological case with this particular clip.

Note that the regression is not nearly as bad on more "real life" content, for example ToS:
```
# sh bd_rate.sh /c/Users/utente/Desktop/codecComparison/TearsOfSteel/statistics/rav1e-0.1.0-977-64b9f501.out /c/Users/utente/Desktop/codecComparison/TearsOfSteel/statistics/rav1e-0.1.0-1726-57e4f926.out
        RATE (%) DSNR (dB)
[...]

PR #1121, opened 2019-03-16
Workaround the flushing logic problem

lu-zero:
Should address #1081.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -568,7 +568,11 @@ impl<T: Pixel> Context<T> {
     T: Pixel,
   {
     let idx = self.frame_count;
-    self.frame_q.insert(idx, frame.into());
+    let frame = frame.into();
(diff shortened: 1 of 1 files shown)


lu-zero:
I'd land it before the other changes since it has a little more priority

shssoichiro:
This does not seem to address the issue in #1081. If I run it through @SmilingWolf's scripts, the ffmpeg output matches 3202bf7 rather than the good commit 64b9f50.

[bad.log](https://github.com/xiph/rav1e/files/2974091/bad.log)
[good.log](https://github.com/xiph/rav1e/files/2974092/good.log)
[test.log](https://github.com/xiph/rav1e/files/2974093/test.log)

lu-zero:
I guess it is time to go with the plan I laid out and see if it works better =/

PR #1123, opened 2019-03-17
Enable the Clippy's if_same_then_else lint

vladimir-kazakov:
This PR enables the Clippy's [if_same_then_else](https://rust-lang.github.io/rust-clippy/master/index.html#if_same_then_else) lint, and fixes all errors caused by it.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1854,10 +1854,7 @@ fn encode_partition_topdown<T: Pixel>(
     PARTITION_SPLIT |
     PARTITION_HORZ |
     PARTITION_VERT => {
-      let num_modes = if partition == PARTITION_SPLIT { 1 }
-      else { 1 };
-
-      if rdo_output.part_modes.len() >= num_modes {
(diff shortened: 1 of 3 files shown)


src/encoder.rs
```diff
@@ -1854,8 +1854,7 @@ fn encode_partition_topdown<T: Pixel>(
     PARTITION_SPLIT |
     PARTITION_HORZ |
     PARTITION_VERT => {
-      let num_modes = if partition == PARTITION_SPLIT { 1 }
-      else { 1 };
+      let num_modes = 1;
 
       if rdo_output.part_modes.len() >= num_modes {
```
barrbrain: If we further merge this constant into the only statement where it is used, we can simply to the following:
```Rust
if !rdo_output.part_modes.is_empty() {
```

vladimir-kazakov: Initially I thought that leaving this variable may clarify what actually has to be done, but you're right - no need to leave references to incomplete features. A person that will be able to complete the implementation won't be stopped by the absence of this variable.

PR #1126, opened 2019-03-18
Tile encoding

rom1v:
_(description updated on 16 april 2019)_

This PR implements tile encoding (#631).

 1. The (many) first commits introduce tiling structures, which allow to expose simultaneous tiled regions of the whole frame data.
 2. Following commits use the tiling structures where necessary in the whole codebase.
 3. Then, command line arguments are added, the encoder encodes tiles separately, and separate tiles are written to the bitstream.
 4. Finally, parallelization is enabled (spoiler: 1 line of code).

## Context

Encoding a frame first involves frame-wise accesses (initialization, etc.), then tile-wise accesses (to encode tiles in parallel), then frame-wise accesses using the results of tile-encoding (deblocking, cdef, …):

```
                                \
      +----------------+         |
      |                |         |
      |                |         |  Frame-wise accesses
      |                |          >
      |                |         |   - FrameState<T>
      |                |         |   - Frame<T>
      +----------------+         |   - Plane<T>
                                /    - ...

              ||   tiling views
              \/
                                \
  +---+  +---+  +---+  +---+     |
  |   |  |   |  |   |  |   |     |  Tile encoding (possibly in parallel)
  +---+  +---+  +---+  +---+     |
                                 |
  +---+  +---+  +---+  +---+     |  Tile-wise accesses
  |   |  |   |  |   |  |   |      >
  +---+  +---+  +---+  +---+     |   - TileStateMut<'_, T>
                                 |   - TileMut<'_, T>
  +---+  +---+  +---+  +---+     |   - PlaneRegionMut<'_, T>
  |   |  |   |  |   |  |   |     |
  +---+  +---+  +---+  +---+     |
                                /

              ||   vanishing of tiling views
              \/
                                \
      +----------------+         |
      |                |         |
      |                |         |  Frame-wise accesses
      |                |          >
      |                |         |  (deblocking, cdef, ...)
      |                |         |
      +----------------+         |
                                /
```

## Tiling

As you know, in Rust, it is not sufficient not to read/write the same memory from several threads, it must be impossible to write (safe) code that could do it. More precisely, a mutable reference may not alias any other reference to the same memory.

That's the reason why, as a preliminary step, I replaced accesses using the whole plane as a raw slice in addition to the stride information by `PlaneSlice` (#1035) and `PlaneMutSlice` (#1043).

But `Plane(Mut)Slice` still borrows the whole plane slice, so it does not, in itself, solves the problem.

There are several structures to be tiled, which form a tree:

```
 +- FrameState → TileState
 |  +- Frame → Tile
 |  |  +- Plane → PlaneRegion 
 |  +  RestorationState → TileRestorationState
 |  |  +- RestorationPlane → TileRestorationPlane
 |  |     +- FrameRestorationUnits → TileRestorationUnits
 |  +  FrameMotionVectors → TileMotionVectors
 +- FrameBlocks → TileBlocks
```

Most of them exist both in const and mutable version (e.g. `PlaneRegion` and `PlaneRegionMut`).

#### `PlaneRegion`

This is a view of bounded region of a `Plane`. It is similar to `PlaneSlice`, except that it does not borrow the whole underlying raw slice. That way, it is possible to get several non-overlapping regions simultaneously.

_In the end, we should probably merge it with `PlaneSlice`, but it requires more work because some frame-wise code still uses `PlaneSlice` in the code base._

It is possible to retrieve a subregion of a region (which may not exceed its parent). In theory, a subregion is defined by a rectangle (for example: x, y, width, height), but in practice, we need more flexibility. For example, we often need to retrieve a region from an offset, using the same bottom-right corner as its parent without providing width and height.

For that purpose, I propose a specific `Area` structure (actually, a Rust `enum`) to describe subregion bounds. Here are some usage examples:

```rust
let region = plane.region(Area::Rect { x: 32, y: 32, width: 512, height: 512 });

// the area is relative to the parent region
let subregion = region.subregion(Area::StartingAt { x: 128, y: 128 });
// it is equivalent to
let subregion = region.subregion(Area::Rect { x: 128, y: 128, width: 384, height: 384 });
// or
let subregion = plane.region(Area:: Rect { x: 160, y: 160, width: 384, height: 384 });
```

Retrieving a subregion from a `BlockOffset` is so common accross the code base that I decided to expose it directly:

```
let bo = BlockOffset { x: 2, y: 3 };
let subregion = region.subregion(Area::BlockStartingAt { bo });
```

Like `Plane(Mut)Slice`, it provides _operator[]_ and iterators over its rows:

```rust
let row5 = &region[5];
let value = region[3][4];
for row in region.rows_iter() {
    let _first_four_values = &row[..4];
}
```

The mutable versions of the structure (`PlaneRegionMut`) and methods are also provided.

#### `Tile`

A `Tile` is a view of 3 colocated plane regions (`Tile` is to a `PlaneRegion` as a `Frame` is to a `Plane`).

The mutable version (`TileMut`) is also provided.

#### `TileState`

The way the `FrameState` fields are mapped in `TileState` depends on how they are accessed tile-wise and frame-wise.

Some fields (like `qc`) are only used during tile-encoding, so they are only stored in `TileState`.

Some other fields (like `input` or `segmentation`) are not written tile-wise, so they just reference the matching field in `FrameState`.

Some others (like `rec`) are written tile-wise, but must be accessible frame-wise once the tile views vanish (e.g. for deblocking).

It contains 2 tiled views: `TileRestorationState` and a vector of `TileMotionVectorsMut` (a tiled view of `FrameMotionVectors`).

This structure is only provided as mutable (`TileStateMut`). A const version is not necessary, and would require to instantiate a const version of all its embedded tiled views.

#### `TileBlocks`

`TileBlocks` is a tiled view of `FrameBlocks`. It exposes the blocks associated to the tile.

The mutable version (`TileBlocksMut`) is also provided.

### Splitting into tiles

A `TilingInfo` structure computes all the details about tiling from the frame width and height and the (log2 of the) number of tile columns and rows. The details are accessible for initializing data or writing into the bitstream.

It provides an iterator over tiles (yielding one `TileStateMut` and one `TileBlocksMut` for each tile).

## Frame offsets vs tile offsets

In `encode_tile()`, super-block, block and plane offsets are expressed relative to the tile. The tiling views expose its data relative to the tile:
 - `plane_region[y][x]` is pixel (x, y) relative to the plane region,
 - `tile_blocks[boy][box]` contains the `Block` at (box, boy) relative to the tile,
 - …

`TileStateMut` exposes some references to frame-level data stored in `FrameState`:
 - `input` is a reference to the whole frame,
 - `input_hres` and `input_qres` are references to the whole planes.

When accessing these frame-level data, tile offsets are converted to frame offsets, for example by:

```rust
let frame_bo = ts.to_frame_block_offset(bo);
```

## Current state

It works.

Need more tests and reviews.


rom1v:
I added `#[derive(Copy)]` to `BlockOffset`, and used `BlockOffset` by value in the `Area` enum. Using a reference would add a lifetime parameter (`Area<'_>`), which would be overkill here.

But since I added `Copy`, _clippy_ warns about `BlockOffset` everywhere:

```
error: this argument is passed by reference, but would be more efficient if passed by value
    --> src/partition.rs:1234:19
     |
1234 | pub fn has_tr(bo: &BlockOffset, bsize: BlockSize) -> bool {
     |                   ^^^^^^^^^^^^ help: consider passing by value instead: `BlockOffset`
     |
     = note: `-D clippy::trivially-copy-pass-by-ref` implied by `-D warnings`
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref
```

I think we should replace all `&BlockOffset` to `BlockOffset` in function parameters. Do you agree I do this on `master`?

rom1v:
tiling

At least, we can see the tiles :trollface:

rom1v:
Here is my first working tiled (2×2) video encoded with rav1e: [bbb_tiled3.mp4](http://h.rom1v.com/bbb_tiled3.mp4) | [bbb_tiled3.ivf](http://h.rom1v.com/bbb_tiled3.ivf)

From this version: https://github.com/rom1v/rav1e/commits/tiling.firstworking

```
ffmpeg -i big_buck_bunny_720p.mp4 -ss 10 -t 10 bbb.y4m
target/release/rav1e samples/bbb.y4m --tile-rows-log2 1 --tile-cols-log2 1 -o bbb_tiled3.ivf
ffmpeg -i bbb_tiled3.ivf -c:v copy bbb_tiled3.mp4
```

:tada:

rom1v:
> Would par_iter() keeps the run order of encode_tile()?

No, the point is to execute them in parallel.

> I.e. Preserving the order of generated bitstream from each tile.

Each encode_tile() returns a byte array (`Vec<u8>`), which are written in order (following the AV1 spec) at the end.

See 806005fb51f906ed92ed407ea55f213a91e615e2.

rom1v:
> Would 'rec' covers a tile only, it comes from : "let rec = &mut ts.rec.planes[p];"
> I read that it still covers whole frame (plane p) ?

`ts` is a `TileStateMut<'_, T>`, `ts.rec` is a `TileMut<'_, T>`, `ts.rec.planes[p]` is a `PlaneRegion<'_, T>`. It only gives access to the plane region covered by the tile.

rom1v:
1 tile vs 4 tlies: https://beta.arewecompressedyet.com/?job=%402019-04-05T21%3A29%3A59.430Z_ref_1_tile&job=%402019-04-05T21%3A28%3A52.445Z_4_tiles

rom1v:
1 tile vs 4 tiles: https://beta.arewecompressedyet.com/?job=%402019-04-05T21%3A29%3A59.430Z_ref_1_tile&job=%402019-04-05T21%3A28%3A52.445Z_4_tiles

Of course, there is a cost in quality (SSIM, PSNR…) because it loose the possibility to exploit redundancy across tiles. For now, the current version only saves CDF from tile 0 (it should choose the bigger tile in bytes instead), and always store tile sizes on 4 bytes. It can (and will) be improved.

The encoding time is worse with tiling on AWCY because I think that it uses only 1 core per instance.

rom1v:
Unfortunately (but as expected), the tiling structures are not a zero cost abstraction. They add an overhead in encoding time between 1~3%.

Concretely, if we compare the version compiled from `master` branch with the version compiled from this pull request, and encode a video without tiling (1 tile), the latter takes a bit more time to encode. Said otherwise, we pay a (small) cost for tile encoding even if we don't use it.

As an example, on my laptop, an encoding takes 3mn33,245 on `master` and 3mn35,722 on `tiling` (it's just an example, but the difference is quite stable).

You can compare encoding times on AWCY for 1 tile: https://beta.arewecompressedyet.com/?job=master-70005e353aa8ce21e3ecd257c927f71d4012a117&job=%402019-04-05T21%3A29%3A59.430Z_ref_1_tile

Maybe some work could be done to minimize this overhead (for example using more `#[inline(always)]`), but I think the overhead cannot be removed (within a memory-safe language).

EwoutH:
I did a [scaling test](https://docs.google.com/spreadsheets/d/1JpPz-Tvj8OO9VGyv-0_VQQRlMVpChuwYeRwr3L3Ss1E/edit#gid=1605614229) with 1 to 32 tiles measuring speed, bitrate and psnr on a Ryzen 5 1600 (6c/12t, AVX2):

| Tiles |   FPS | Bitrate |   PSNR |
|-------|------:|--------:|-------:|
| 0x0   | 0,042 |    2443 | 50,757 |
| 0x1   | 0,078 |    2566 | 50,750 |
| 1x1   | 0,111 |    2586 | 50,734 |
| 1x2   | 0,103 |    2655 | 50,730 |
| 2x2   | 0,072 |    2686 | 50,727 |
| 2x3   | 0,046 |    2827 | 50,727 |

And normalized:

| Tiles |  FPS | Bitrate |    PSNR |
|-------|-----:|--------:|--------:|
| 0x0   | 100% |  100,0% | 100,00% |
| 0x1   | 186% |  105,0% |  99,99% |
| 1x1   | 264% |  105,8% |  99,95% |
| 1x2   | 245% |  108,7% |  99,95% |
| 2x2   | 171% |  109,9% |  99,94% |
| 2x3   | 110% |  115,7% |  99,94% |

rom1v:
@EwoutH Thank you for your tests!

The number of threads is limited to 8 (on this branch, 4 on master). For 2×2 and 2×3 log2-tiles (4×4 tand 4×8 tiles), did you increase the number of threads? (option `--threads`)

EwoutH:
@rom1v No I didn't, now rerunning with `--threads 12` for 1x1 to 2x3.

I'm now also testing on a 2x Intel Xeon E5-2680v4  (2x 14c/28t, AVX2) system, which should give better scaling results for 8 to 32 tiles. Results are coming in [here](https://docs.google.com/spreadsheets/d/1JpPz-Tvj8OO9VGyv-0_VQQRlMVpChuwYeRwr3L3Ss1E/edit#gid=1605614229).

EwoutH:
Updated results.

2160p 8-bit 4:2:0 on a Ryzen 5 1600 (6c/12t, AVX2) system:

Tiles | FPS | Bitrate | PSNR | FPS | Bitrate | PSNR
-- | -- | -- | -- | -- | -- | --
0x0 | 0,042 | 2443 | 50,757 | 100% | 100,0% | 100,00%
0x1 | 0,078 | 2566 | 50,750 | 186% | 105,0% | 99,99%
1x1 | 0,113 | 2586 | 50,734 | 269% | 105,8% | 99,95%
1x2 | 0,114 | 2655 | 50,730 | 271% | 108,7% | 99,95%
2x2 | 0,077 | 2686 | 50,727 | 183% | 109,9% | 99,94%
2x3 | 0,050 | 2827 | 50,727 | 119% | 115,7% | 99,94%

1080p 8-bit 4:2:0 on a 2x Intel Xeon E5-2680v4  (2x 14c/28t, AVX2) system:

| Tiles |   FPS | Bitrate |    PSNR | FPS  | Bitrate | PSNR    |
|-------|------:|--------:|--------:|------|---------|---------|
| 0x0   | 0,103 |  722,02 |  49,680 | 100% | 100,0%  | 100,00% |
| 0x1   | 0,184 |  825,25 |  49,694 | 179% | 114,3%  | 100,03% |
| 1x1   | 0,311 |  835,29 | 49,6932 | 302% | 115,7%  | 100,03% |
| 1x2   | 0,465 |  875,76 | 49,6631 | 451% | 121,3%  | 99,97%  |
| 2x2   | 0,431 |  917,09 | 49,6699 | 418% | 127,0%  | 99,98%  |
| 2x3   | 0,418 | 1017,15 | 49,6727 | 406% | 140,9%  | 99,98%  |

rom1v:
Thanks!

Did you use 16 threads (or more) for 2x2 (4x4) and 32 threads (or more) for 2x3 (4x8)?

EwoutH:
@rom1v On Ryzen 5 12 threads for 1x1 to 2x3, on Intel Xeon 16 threads for 1x1, 32 for 1x2 and 56 for 2x2 and 2x3. CPU was maxed out on Ryzen.

ycho:
With intra frame only, i.e. 1 frame only, the loss of coding efficiency is much less.

[AWCY for 1 frame, 1 tile vs 4 tiles](https://beta.arewecompressedyet.com/?job=%402019-04-05T21%3A29%3A59.430Z_ref_1_tile_1f%402019-04-08T18%3A16%3A54.523Z&job=%402019-04-05T21%3A28%3A52.445Z_4_tiles_1f%402019-04-08T18%3A17%3A34.602Z)

With largest with shield, >30%.

rom1v:
@ycho Note that the rate distortion curves (at least when there are several frames) are meaningless until the corrupted frames bugs are solves (this [result](https://user-images.githubusercontent.com/543275/55742123-3aebb780-5a2f-11e9-95dd-29d0a2beefb5.jpg) has necessarily a poor psnr).

tdaede:
That speed loss isn't so bad and worth it (the same infra will thread lrf etc too), but I think we could probably do a bit better. Do you have an idea where it's coming from? Could it be measured on a per-commit basis to narrow it down? Wouldn't surprise me if it could be eliminated by some strategic #[inline]s.

ycho:
I think I've fixed "Issue 3: assertion failure at speed 0". 
Apply:
https://github.com/xiph/rav1e/pull/1196/commits/df716da15e43a8f60017f7704481ae82046fe4bc
(Or wait for PR #1196, which has this commit,  lands)

Turns out we've still have problems with directional intra prediction! And this was also the cause for my intar tx partitioning at depth 2.

rom1v:
@ycho I confirm that https://github.com/xiph/rav1e/commit/df716da15e43a8f60017f7704481ae82046fe4bc prevents the assertion to fali.

However, can you make it fail on current `master` in some way? If not, chances are that tiling code break something, what do you think?

rom1v:
> Could it be measured on a per-commit basis to narrow it down?

Unfortunately, the end-to-end encoding only works from when the tiling structures are actually used, and all are used at once.

> Wouldn't surprise me if it could be eliminated by some strategic #[inline]s.

You're right. I tried to put `#[inline(always)]` almost everywhere in tiling structures, the overhead almost vanishes.

rom1v:
@ycho, oh ok, I managed to make the assertion fail on `master` + https://github.com/xiph/rav1e/pull/1196/commits/78788193d0bb57238e5dccdc4ddcdbb8363e26bc (without disabling directional intra prediction).

EwoutH:
I ran an additional 3 runs and averaged the results out, reducing the noise a little.

| FPS | master | tiling | tiling.108 |
|-----|-------:|-------:|-----------:|
| 0x0 |  0,103 |  0,103 |      0,103 |
| 1x1 |      - |  0,304 |      0,308 |

| User time (s) | master | tiling | tiling.108 |
|---------------|-------:|-------:|-----------:|
| 0x0           |  705,4 |  719,0 |      719,2 |
| 1x1           |      - |  856,4 |      861,4 |

- There is no noticeable difference in single-tile performance (fps), but the used CPU time increase a little (about 2%), which would cause a small slow-down on a single-core system.
- It looks like the tiling.108 branch uses a little more cpu time (about 0.5%) with 4 tiles than the tiling branch, but the performance impact is inconclusive.
- [Full data here](https://docs.google.com/spreadsheets/d/1JpPz-Tvj8OO9VGyv-0_VQQRlMVpChuwYeRwr3L3Ss1E/edit#gid=350324363).

ycho:
Not a fix for issue 1 or 2, but fix of 1/4 ME/
https://github.com/ycho/rav1e/commit/4b3693e6c26a3b05ee21cf2d3cfff4d1e56b4531

Also, if change 192 to 64 in
"let range_x = 192 * fi.me_range_scale as isize;" in estimate_motion_ss4(), then it seems working with nyan.y4m

ycho:
For issues 2 and 3, when 2x2 tiles are used, It seems SBs in right side tiles has wrong row mv values, where it should be zero MVs but somehow very large negative value for mv_x.

2nd frame decoded by dav1d shown below:

Screen Shot 2019-04-10 at 10 57 31 AM

ycho:
Issue 1 seems not occurring if column tiles are not used, for ex: --tile-rows-log2 1 --tile-cols-log2 0
But still there if column tiles are used, for ex --tile-rows-log2 0 --tile-cols-log2 1

ycho:
This probably might be out of this PR, but enforcing to disable diamond ME gives assert fail.
Note that diamond ME is default, but current 1/4 ME uses full search ME. So, I thought this finding might be related.

To enforce disabling diamond ME,
```
  fn diamond_me_preset(_speed: usize) -> bool {
    false
  }
```

```
480x304 @ 30/1 fps
encoded 1/30 frames, 3.922 fps, 1841.48 Kb/s, est. size: 0.22 MB, est. time: 7 s                    thread 'thread '<unnamed><unnamed>' panicked at '' panicked at 'assertion failed: rect.y >= 0 && rect.y as usize <= self.rect.heightassertion failed: rect.x >= 0 && rect.x as usize <= self.rect.width', ', src/tiling/plane_region.rssrc/tiling/plane_region.rs::278278::11

thread '<unnamed>' panicked at 'assertion failed: rect.x >= 0 && rect.x as usize <= self.rect.width', src/tiling/plane_region.rs:278:1
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::begin_panic
   4: rav1e::me::MotionEstimation::motion_estimation
   5: rav1e::rdo::rdo_mode_decision
   6: rav1e::rdo::rdo_partition_decision
   7: rav1e::encoder::encode_partition_topdown
   8: rav1e::encoder::encode_tile
   9: rayon::iter::plumbing::bridge_producer_consumer::helper
  10: rayon_core::join::join_context::{{closure}}
  11: rayon::iter::plumbing::bridge_producer_consumer::helper
  12: <rayon_core::job::StackJob<L, F, R> as rayon_core::job::Job>::execute
  13: rayon_core::registry::WorkerThread::wait_until_cold
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::begin_panic
   4: rav1e::me::MotionEstimation::motion_estimation
   5: rav1e::rdo::rdo_mode_decision
   6: rav1e::rdo::rdo_partition_decision
   7: rav1e::encoder::encode_partition_topdown
   8: rav1e::encoder::encode_tile
   9: rayon::iter::plumbing::bridge_producer_consumer::helper
  10: <rayon_core::job::StackJob<L, F, R> as rayon_core::job::Job>::execute
  11: rayon_core::registry::WorkerThread::wait_until_cold
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::begin_panic
   4: rav1e::me::MotionEstimation::motion_estimation
   5: rav1e::rdo::rdo_mode_decision
   6: rav1e::rdo::rdo_partition_decision
   7: rav1e::encoder::encode_partition_topdown
   8: rav1e::encoder::encode_tile
   9: rayon::iter::plumbing::bridge_producer_consumer::helper
  10: <rayon_core::job::StackJob<L, F, R> as rayon_core::job::Job>::execute
  11: rayon_core::registry::WorkerThread::wait_until_cold
```

ycho:
I think I found a reason (at least part of it), I haven't tried this yet.
In short, when doing 1/4 ME, the original input is tile and po is plane offset, 
but full_search() assumed original input is a plane.
i.e. in estimate_motion_ss4(),
```
    full_search(
      x_lo,
      x_hi,
      y_lo,
      y_hi,
      blk_h >> 2,
      blk_w >> 2,
      &ts.input_qres,   <==== Tile input source passed
      &rec.input_qres,
      &mut best_mv,
      &mut lowest_cost,
      po,  <==== plane offset passed
      1,
      fi.sequence.bit_depth,
      lambda,
      [MotionVector::default(); 2],
      fi.allow_high_precision_mv
    );
```

Then in full_search(),
```
    let (cost, mv) = search_area.map(|(y, x)| {
    let plane_org = p_org.region(Area::StartingAt { x: po.x, y: po.y });   <====  plane offset used for tile 
               input source 
```
I can try tonight if I can find time! :)
Again,  remind that 1/4 initial coarse ME is the only place full ME search with current master.

ycho:
Partial fix for issue 1 (and maybe 2?). Now you can see 2nd frame!
https://github.com/ycho/rav1e/commit/a2bb9e9be4e08713a42a31fa9e16502502b2f1ff

rom1v:
@ycho Thank you for your investigations.

> In short, when doing 1/4 ME, the original input is tile and po is plane offset, but full_search() assumed original input is a plane.

In calls to `full_search()`, the input (`&ts.input.planes[0]`) is a full plane (`ts.input` is a `Frame<T>`), and the plane offset passed as argument is `frame_bo` (relative to the frame).

`full_search()` should never handle offset relative to a tile (it does not receive any tiling view anyway).

> ```rust
> let plane_org = p_org.region(Area::StartingAt { x: bo.x as isize, y: bo.y as isize });
> ```

This may not be correct, `Area::StartingAt` expects plane offsets, not block offsets (contrary to `Area::BlockStartingAt`).

rom1v:
> This probably might be out of this PR, but enforcing to disable diamond ME gives assert fail.

Oh, thank you :+1: I just fixed it:

```diff --git a/src/me.rs b/src/me.rs
index 436f6062..abd5a398 100644
--- a/src/me.rs
+++ b/src/me.rs
@@ -950,7 +950,7 @@ fn telescopic_subpel_search<T: Pixel>(
           );
         }
 
-        let plane_org = ts.input_tile.planes[0].subregion(Area::StartingAt { x: po.x, y: po.y });
+        let plane_org = ts.input.planes[0].region(Area::StartingAt { x: po.x, y: po.y });
         let plane_ref = tmp_plane.as_region();
 
         let sad = get_sad(&plane_org, &plane_ref, blk_h, blk_w, fi.sequence.bit_depth);
```
(the given plane offset is relative to the frame)

ycho:
> > ```
> > &ts.input_qres,   <==== Tile input source passed
> > ```
> 
> No, `input_qres` is a `&'a Plane<T>` (a reference to the whole plane stored in `FrameState<T>`), so it's a frame plane (not a region).

Oh, you are right. It is NOT tile data.

```
      input_hres: &fs.input_hres,
      input_qres: &fs.input_qres,
```

src/me.rs
```diff
@@ -942,9 +980,8 @@ fn full_search<T: Pixel>(
     let search_area = search_range_y.flat_map(|y| { search_range_x.clone().map(move |x| (y, x)) });
 
     let (cost, mv) = search_area.map(|(y, x)| {
-      let plane_org = p_org.slice(po);
-      let plane_ref = p_ref.slice(PlaneOffset { x, y });
-
+      let plane_org = p_org.region(Area::StartingAt { x: po.x, y: po.y });
+      let plane_ref = p_ref.region(Area::StartingAt { x, y });
```
ycho: I think .region(Area::StartingAt { x, y } ) does not get panic when x or y is invalid value. For ex, if x is out of plane border (and crossing padded area, which is 22 pixels for 16x16 blk size at 1/4 ME). 
I tested this in estimate_motion_ss4( )
```
   full_search(
      x_lo - 200,
```
then encoder runs w/o any panic.

rom1v: Oh, good catch!

Initially, `PlaneRegion` used unsigned coordinates, so it only asserted max values, but then I needed to convert them to signed, but I forgot to assert min values.

I just fixed it.

rom1v:
```
rav1e -o video.ivf samples/objective-1-fast/Netflix_Crosswalk_1920x1080_60fps_8bit_420_60f.y4m -s10 --tile-rows-log2 1 --tile-cols-log2 3
```

The bitstream is corrupted, but dav1d can more or less decode it, and AOMAnalyzer can get some frames.

Here is frame 2:

rav1e-tiles

There is clearly a problem at tiles boundaries. The corrupted squares we see on the image do not even match the frame blocks.

rom1v:
It works \o/ https://github.com/rom1v/rav1e/commits/tiling.115

But I don't really understand why I need to avoid mv outside the current tile here, it can probably be fixed "earlier".

ycho:
I think we are missing "if ( !is_inside(mvRow,mvCol) ) break" both in scan_row_mbmi() and scan_cole_mbmi(). In av1 spec, they are https://aomediacodec.github.io/av1-spec/#scan-row-process and https://aomediacodec.github.io/av1-spec/#scan-col-process
The is_inside() check whether mv is inside the tile or not. https://aomediacodec.github.io/av1-spec/#scan-col-process

rom1v:
@ycho good point, but it seems insufficient: https://github.com/rom1v/rav1e/commit/dc586ac371ee0bb0b717f10414b6585d648ed58b

rom1v:
I finally found the culprit of the bitstream corruption, and [fixed it](https://github.com/xiph/rav1e/pull/1126/commits/215375572b5b0c3a963957dedcf9134ba379e745)!

---

It remained a problem with restoration with stretched restoration units (`decode_test` failed). Therefore, I refactored the restoration units (https://github.com/xiph/rav1e/pull/1213), then I rewrote the way `RestorationState` was tiled on my tiling branch (tiles never need to share restoration units at tile level anymore). And now, `decode_test` pass. @xiphmont, I would like your confirmation that the principle is ok.

If for any reason, we need to share RU at tile-level in the future, the new tiled view of `RestorationState` is still better IMO (we would just need to pass the same unit to different tile in a `Mutex`).

---

So, now, AFAIK, everything works. If you could test on more samples/conditions, that would be great :wink:

Here are the AWCY results:
https://beta.arewecompressedyet.com/?job=master-cd3ef475ba6a2a7d9832e78dbc296f64e6e7a859&job=%402019-04-16T10%3A39%3A22.178Z

rom1v:
I encoded all [`objective-1-fast` samples](https://media.xiph.org/video/derf/objective-1-fast.tar.gz) (from [here](https://media.xiph.org/video/derf/)) with `--tile-cols-log2 2 --tile-rows-log2 2` (up to 4×4 tiles), then decoded them.

```bash
# encoding
cd samples
mkdir -p out
for i in *.y4m; do ../target/release/rav1e --tile-cols-log2 2 --tile-rows-log2 2 "$i" -o "out/${i%.*}.ivf"; done

# decoding
cd out
mkdir -p dec
for i in *.ivf; do aomdec "$i" -o "dec/${i%.*}.y4m" || { echo '!!!!! FAIL !!!!!' >&2; }; done
```

Everything worked, and the decoded videos were ok in VLC.

I can now remove the `[WiP]` tag :smile: 

Thank you for your reviews/tests :wink:

src/api.rs
```diff
@@ -458,7 +462,7 @@ pub struct Config {
   pub threads: usize
 }
 
-const MAX_USABLE_THREADS: usize = 4;
+const MAX_USABLE_THREADS: usize = 8;
```
tdaede: The reason for this is because the intra mode rayon usage is really too fine grained to scale well, while for tile threading we don't need this limit. Instead of bumping this number we should probably remove or limit the intra mode parallelism.

src/encoder.rs
```diff
@@ -2066,17 +2150,50 @@ fn encode_tile_group<T: Pixel>(fi: &FrameInvariants<T>, fs: &mut FrameState<T>)
...
+
+  // use the biggest tile (in bytes) for CDF update
+  fs.context_update_tile_id = idx_max;
+  fs.cdfs = cdfs[idx_max];
   fs.cdfs.reset_counts();
 
-  data
+  let max_tile_size_bytes = ((max_len as u32).ilog() + 7) / 8;
+  debug_assert!(max_tile_size_bytes > 0 && max_tile_size_bytes <= 4);
+  fs.max_tile_size_bytes = max_tile_size_bytes;
```
tdaede: any reason this is stored in framestate?

src/encoder.rs
```diff
@@ -383,16 +396,28 @@ impl Sequence {
       second_forward_idx >= 0
     }
   }
+
+  #[inline(always)]
+  pub fn sb_size_log2(&self) -> usize {
```
tdaede: really wish this could be a const fn augh

src/api.rs
rom1v: OK, so I remove this commit for now?

Indeed, the solution is probably to not submit too fine grained tasks to rayon and totally remove this limit.

src/encoder.rs
rom1v: This is a property at frame level (there is 1 `FrameState` per frame, and 1 `TileState` per tile, i.e. _n_ `TileState`s per frame).

It is then retrieved from `FrameState` to write the header (see commit `Optimizer tile size bytes in bitstream`, currently 1558b052119eaf58fdac48eec661d2f28be5918b).

src/encoder.rs
rom1v: It cannot be `const fn`, since it should depend on a (non-existing yet) command-line parameter (64×64 vs 128×128).

We could store them as `FrameInvariants` fields though:

```
    pub sb_size: usize,
    pub sb_size_log2: usize,
```

src/encoder.rs
tdaede: OK. Makes me somewhat uncomfortable how it sits invalid for so long, but I guess the assert for bytes > 0 should cover that case.

src/api.rs
tdaede: Given discussion on irc, just drop the pmv par_iter and remove the MAX_USABLE_THREADS threshold entirely.

src/encoder.rs
rom1v: > Makes me somewhat uncomfortable how it sits invalid for so long

This is exactly the same for the `cdfs` variable in `FrameState`, it is initialized only there (4 lines above, `fs.cdfs = cdfs;`).

rom1v:
Here are the new results after the recent changes on the `tiling` branch.

I tested on my laptop (8 CPUs) several encodings to compare encoding performance (this is not a good benchmark, but it gives an idea, you are encouraged to run your own tests), and launched AWCY runs.

## 1 tile

Since commits `Reduce overhead of tiling abstraction` (currently https://github.com/xiph/rav1e/pull/1126/commits/0d9e3dd7b39b93b46491b1a65d744950d5579d33) and `Inline all TileBlocks methods ` (https://github.com/xiph/rav1e/pull/1126/commits/43e2a566b576cac38707f3fd845bd1907286f6c7), with 1 tile (i.e. no tile encoding), `tiling` is similar / a bit faster than `master` now.

On this specific test:

```
time target/release/rav1e \
    Netflix_Crosswalk_1920x1080_60fps_8bit_420_60f.y4m \
    -o video.ivf
```

 - `master`: 7m5,630s
 - `tiling`: 7m2,336s (speedup: **1.008x**)

The difference happens to be visible on [AWCY](https://beta.arewecompressedyet.com/?job=tiling.137_reference_master%402019-04-18T18%3A42%3A18.213Z&job=tiling.137_1_tile%402019-04-18T17%3A42%3A49.905Z) (Metric: Encoding Time). This proves nothing since it probably depends on the server load, but the difference is quite consistant across different runs.

They produce the same results:

```
0.0000 |  0.0000 |  0.0000 |   0.0000 | 0.0000 |  0.0000 |     0.0000
```

## 2 tiles

```
time target/release/rav1e \
    Netflix_Crosswalk_1920x1080_60fps_8bit_420_60f.y4m \
    -o video.ivf
    --tile-cols-log2 1
```

 - 1 tile: 7m2,336s
 - 2 tiles: 3m53,578s (speedup: **1.81x**)

[AWCY](https://beta.arewecompressedyet.com/?job=tiling.137_1_tile%402019-04-18T17%3A42%3A49.905Z&job=tiling.137_2_tiles%402019-04-18T17%3A43%3A20.621Z) without low latency:

```
2.6419 |  2.4644 |  2.4862 |   2.6375 | 2.7437 |  2.7179 |     2.8091
```

[AWCY](https://beta.arewecompressedyet.com/?job=tiling.137_1_tile_low_latency%402019-04-18T18%3A41%3A43.143Z&job=tiling.137_2_tiles_low_latency%402019-04-18T18%3A31%3A33.779Z) with low latency (`--low_latency=true`):

```
1.3688 |  0.7373 |  1.1210 |   1.3709 | 1.4007 |  1.4120 |     1.3294
```


## 4 tiles

```
time target/release/rav1e \
    Netflix_Crosswalk_1920x1080_60fps_8bit_420_60f.y4m \
    -o video.ivf
    --tile-cols-log2 1 --tile-rows-log2 1
```

 - 1 tile: 7m2,336s
 - 4 tiles: 2m12,995s (speedup: **3.18x**)

[AWCY](https://beta.arewecompressedyet.com/?job=tiling.137_1_tile%402019-04-18T17%3A42%3A49.905Z&job=tiling.137_4_tiles%402019-04-18T17%3A43%3A55.951Z) without low latency:

```
5.9145 |  5.3288 |  5.8706 |   5.8448 | 6.1259 |  6.0573 |     6.0141
```

[AWCY](https://beta.arewecompressedyet.com/?job=tiling.137_1_tile_low_latency%402019-04-18T18%3A41%3A43.143Z&job=tiling.137_4_tiles_low_latency%402019-04-18T18%3A31%3A54.921Z) with low latency (`--low_latency=true`):

```
2.9473 |  1.8566 |  2.4672 |   2.9412 | 3.0758 |  3.0709 |     2.9937
```


## 8 tiles

```
time target/release/rav1e \
    Netflix_Crosswalk_1920x1080_60fps_8bit_420_60f.y4m \
    -o video.ivf
    --tile-cols-log2 2 --tile-rows-log2 1
```

 - 1 tile: 7m2,336s
 - 8 tiles: 1m57,533s (speedup: **3.59x**)

[AWCY](https://beta.arewecompressedyet.com/?job=tiling.137_1_tile%402019-04-18T17%3A42%3A49.905Z&job=tiling.137_8_tiles%402019-04-18T17%3A44%3A15.874Z) without low latency:

```
10.4391 |  9.9325 | 10.3241 |  10.2902 | 10.8110 | 10.7126 |    10.7646
```

[AWCY](https://beta.arewecompressedyet.com/?job=tiling.137_1_tile_low_latency%402019-04-18T18%3A41%3A43.143Z&job=tiling.137_8_tiles_low_latency%402019-04-18T18%3A32%3A21.077Z) with low latency (`--low_latency=true`):

```
5.1544 |  4.1440 |  4.6347 |   5.1364 | 5.3689 |  5.3737 |     5.2127
```

rom1v:
> (Since low_latency=false seems not working correctly)

Actually, in absolute, `--low_latency=false` gives better objective quality results. But relatively, tiling cause less loss with `--low_latency=true`: [AWCY](https://beta.arewecompressedyet.com/?job=tiling.137_1_tile%402019-04-18T17%3A42%3A49.905Z&job=tiling.137_4_tiles%402019-04-18T17%3A43%3A55.951Z&job=tiling.137_1_tile_low_latency%402019-04-18T18%3A41%3A43.143Z&job=tiling.137_4_tiles_low_latency%402019-04-18T18%3A31%3A54.921Z).

I don't know why.

ycho:
> Actually, in absolute, `--low_latency=false` gives better objective quality results.

Sure, far better (> 10%), that is why we wanted to have pyramid and/or frame-reordering.

rom1v:
I published a blog post about this feature: https://blog.rom1v.com/2019/04/implementing-tile-encoding-in-rav1e/

PR #1127, opened 2019-03-19
Add struct FrameMotionVectors

rom1v:
The motion vectors were stored in a `Vec<Vec<MotionVector>>`.

The innermost `Vec` contains a flatten matrix (`fi.w_in_b` x `fi.h_in_b`) of `MotionVector`s, and there are `REF_FRAMES` instances of them (the outermost `Vec`).

Introduce a typed structure to replace the innermost `Vec`:
 - this improves readability;
 - this allows to expose it as a 2D array, thanks to `Index` and `IndexMut` traits;
 - this will allow to split it into (non-overlapping) tiled views, containing only the motion vectors for a bounded region of the plane (see <https://github.com/xiph/rav1e/pull/1126>).

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -19,6 +19,7 @@ use crate::partition::*;
 use crate::plane::*;
 use crate::util::Pixel;
 
+use std::ops::{Index, IndexMut};
 use std::sync::Arc;
 
 #[cfg(all(target_arch = "x86_64", feature = "nasm"))]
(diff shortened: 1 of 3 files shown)


src/me.rs
```diff
@@ -216,6 +217,38 @@ mod native {
   }
 }
 
+#[derive(Debug, Clone)]
+pub struct FrameMotionVectors {
+  mvs: Box<[MotionVector]>,
```
shssoichiro: I think this is equivalent structurally to a `Vec`, but has less flexibility. Is there a specific reason we would need to store it this way?

rom1v: The array size is constant (cols × rows), this avoids to expose methods which could resize the array (like `.push()`).

lu-zero: It also has the additional bonus of being slightly smaller and if we have plenty of motion vectors it could have some impact :)

PR #1129, opened 2019-03-19
Create RefFrame enum.

tdaede:
Moves all functions that previously used usize to this type.
Instead of direct conversions to a slot number, use a to_index fn.

This also changes the size of the global mv state and context
ref counting arrays as they don't need LAST_FRAME.

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -16,6 +16,7 @@ use crate::encoder::ReferenceFrame;
 use crate::FrameInvariants;
 use crate::FrameState;
 use crate::partition::*;
+use crate::partition::RefType::*;
 use crate::plane::*;
 use crate::util::Pixel;
 
@@ -341,7 +342,7 @@ pub trait MotionEstimation {
     cmv: MotionVector, pmv: [MotionVector; 2],
     mvx_min: isize, mvx_max: isize, mvy_min: isize, mvy_max: isize,
     blk_w: usize, blk_h: usize, best_mv: &mut MotionVector,
-    lowest_cost: &mut u64, ref_frame: usize
+    lowest_cost: &mut u64, ref_frame: RefType
   );
 
   fn sub_pixel_me<T: Pixel>(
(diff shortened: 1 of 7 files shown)


src/partition.rs
```diff
@@ -21,15 +21,60 @@ use crate::plane::*;
...
-pub const LAST2_FRAME: usize = 2;
-pub const LAST3_FRAME: usize = 3;
-pub const GOLDEN_FRAME: usize = 4;
-pub const BWDREF_FRAME: usize = 5;
-pub const ALTREF2_FRAME: usize = 6;
-pub const ALTREF_FRAME: usize = 7;
+// LAST_FRAME through ALTREF_FRAME correspond to slots 0-6.
+#[derive(PartialEq, Copy, Clone)]
+pub enum RefType {
+  NONE_FRAME = 8,
```
smarter: Is there some convention to put the none value at the beginning ? Otherwise I'd put it at the end to keep things in order. Also, are there situations where we know that NONE_FRAME cannot occur? If so I'd remove it from the enum and use `Option<RefType>` where we need it.

src/partition.rs
```diff
@@ -21,15 +21,60 @@ use crate::plane::*;
...
+  LAST_FRAME,
+  LAST2_FRAME,
+  LAST3_FRAME,
+  GOLDEN_FRAME,
+  BWDREF_FRAME,
+  ALTREF2_FRAME,
+  ALTREF_FRAME
+];
+
+pub const ALL_REFS: [RefType; 9] = [
```
smarter: Seems like there's some risk we could confuse the index used for ALL_INTER_REFS and the one for ALL_REFS, maybe they should be enums too ?

src/partition.rs
tdaede: It was that way in libaom. I will switch it. Yes, there are a lot of situations where both NONE_FRAME and INTRA_FRAME cannot occur (basically everywhere to_index() is used). I'm going to actually make a separate enum type for those cases. Then maybe we can drop RefType entirely.

src/partition.rs
tdaede: How about I move ALL_REFS to inside get_mvref_ref_frames, the only place it's used? (I actually want to delete that function entirely if possible...)

smarter: sure, especially If it's going to be removed eventually then no need to make an enum for it.

src/me.rs
```diff
@@ -548,7 +549,7 @@ impl MotionEstimation for DiamondSearch {
           mvx_min >> 1, mvx_max >> 1, mvy_min >> 1, mvy_max >> 1,
           blk_w >> 1, blk_h >> 1,
           best_mv, lowest_cost,
-          &mut None, 0
+          &mut None, LAST_FRAME
```
tdaede: @magwyz

PR #1130, opened 2019-03-20
Make BlockOffset derive Copy

rom1v:
`BlockOffset` has a size of 128 bits (the same as a slice), and is trivially copyable, so make it derive `Copy`.

Once it derives `Copy`, [clippy suggests to never pass it by reference](https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref).

So pass it by value everywhere to simplify usage.

In particular, this avoids lifetimes bounds where not necessary:

```diff

@@ -1044,8 +1044,8 @@ pub fn rdo_tx_type_decision<T: Pixel>(
   best_type
 }
 
-pub fn get_sub_partitions<'a>(four_partitions: &[&'a BlockOffset; 4],
-                              partition: PartitionType) -> Vec<&'a BlockOffset> {
+pub fn get_sub_partitions(four_partitions: &[BlockOffset; 4],
+                          partition: PartitionType) -> Vec<BlockOffset> {
   let mut partitions = vec![ four_partitions[0] ];
 
   if partition == PARTITION_NONE {
```

See <https://github.com/xiph/rav1e/pull/1126#issuecomment-474532123>.

---

This PR is basically a quick&dirty replacement:
```bash
find -name '*.rs' -exec sed -i 's/&BlockOffset/BlockOffset/g;s/&bo/bo/g' {} ';'
```
followed by around twenty manual fixes.

If you agree with this one, we should do it for `PlaneOffset` and `SuperBlockOffset`too.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -77,7 +77,7 @@ fn write_b_bench(b: &mut Bencher, tx_size: TxSize, qindex: usize) {
               &mut cw,
               &mut w,
               p,
-              &bo,
+              bo,
               mode,
               tx_size,
               tx_type,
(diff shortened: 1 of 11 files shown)


lu-zero:
I like your plan :)

rom1v:
OK :+1:

Then I did the same for `SuperBlockOffset` and `PlaneOffset`.

PR #1141, opened 2019-03-22
Bench the sad function for 8 bit and 10 bit

magwyz:
They have different paths and ASM functions.

### DIFF
--- a/benches/me.rs
+++ b/benches/me.rs
@@ -33,13 +33,12 @@ fn new_plane<T: Pixel>(ra: &mut ChaChaRng, width: usize, height: usize) -> Plane
   p
 }
 
-fn bench_get_sad(b: &mut Bencher, bs: &BlockSize) {
+fn bench_get_sad(b: &mut Bencher, &(bs, bit_depth): &(BlockSize, usize)) {
   let mut ra = ChaChaRng::from_seed([0; 32]);
   let bsw = bs.width();
   let bsh = bs.height();
   let w = 640;
   let h = 480;
-  let bit_depth = 10;
   let input_plane = new_plane::<u16>(&mut ra, w, h);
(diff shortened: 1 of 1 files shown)


magwyz:
Maybe we should have two tests instead?

benches/me.rs
```diff
@@ -33,13 +33,13 @@ fn new_plane<T: Pixel>(ra: &mut ChaChaRng, width: usize, height: usize) -> Plane
   p
 }
 
-fn bench_get_sad(b: &mut Bencher, bs: &BlockSize) {
+fn bench_get_sad(b: &mut Bencher, input_type: &(BlockSize, usize)) {
+  let (bs, bit_depth) = input_type;
```
shssoichiro: I'd prefer these as two parameters to the function, rather than as one `input_type` tuple.

benches/me.rs
```diff
@@ -33,13 +33,13 @@ fn new_plane<T: Pixel>(ra: &mut ChaChaRng, width: usize, height: usize) -> Plane
   p
 }
 
-fn bench_get_sad(b: &mut Bencher, bs: &BlockSize) {
+fn bench_get_sad(b: &mut Bencher, input_type: &(BlockSize, usize)) {
```
rom1v: The `(BlockSize, usize)` could be passed by value, there is no need for reference here.

benches/me.rs
magwyz: I think we are indeed limited by the `bench_function_over_inputs` function that takes a single `inputs` array as parameter.

rzumer:
> 
> 
> Maybe we should have two tests instead?

I would lean towards yes, since it would make it easier to identify whether a regression affects one path or the other (or both) without modifying the test.

PR #1144, opened 2019-03-22
Transorm block partition for intra

ycho:
Currently there is no RDO and split one leve partition of transforms for intra block.
Also, for intra frame only because enabling tx_mode_select (in frame header) in any inter frame requires necessary signaling of var-tx for inter mode blocks, which we don't have yet.

Deblocking is disabled since it causes mismatch between encoder and decoder's reconstructed images. Communicated with Monty and he figured out why and stared to modify deblocking filter code.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1318,6 +1318,8 @@ pub struct BlockContextCheckpoint {
   cdef_coded: bool,
   above_partition_context: Vec<u8>,
   left_partition_context: [u8; MAX_MIB_SIZE],
+  above_tx_context: Vec<u8>,
+  left_tx_context: [u8; MAX_MIB_SIZE],
   above_coeff_context: [Vec<u8>; PLANES],
   left_coeff_context: [[u8; MAX_MIB_SIZE]; PLANES],
 }
@@ -1332,6 +1334,8 @@ pub struct BlockContext {
   pub preskip_segid: bool,
   above_partition_context: Vec<u8>,
   left_partition_context: [u8; MAX_MIB_SIZE],
+  above_tx_context: Vec<u8>,
+  left_tx_context: [u8; MAX_MIB_SIZE],
   above_coeff_context: [Vec<u8>; PLANES],
   left_coeff_context: [[u8; MAX_MIB_SIZE]; PLANES],
   blocks: FrameBlocks,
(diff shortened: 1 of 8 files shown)


ycho:
Screen Shot 2019-03-22 at 3 44 39 PM

ycho:
Monty has fixed deblocking with 2e2f0547 and I confirm it really fixed! Thanks!

src/context.rs
```diff
@@ -1260,6 +1312,8 @@ pub struct BlockContext {
   pub preskip_segid: bool,
   above_partition_context: Vec<u8>,
```
lu-zero: the following fields look all similar, would make sense having something like the following?
``` rust
struct EdgeContext {
  above: Vec<u8>,
  left: [u8; MAX_MIB_SIZE],
}
```

src/context.rs
```diff
@@ -1395,7 +1455,39 @@ impl BlockContext {
...
-  //TODO(anyone): Add reset_left_tx_context() here then call it in reset_left_contexts()
+
+  pub fn update_tx_size_context(
+    &mut self, bo: BlockOffset, bsize: BlockSize, tx_size: TxSize, skip: bool
+  ) {
+    let n4_w = bsize.width_mi();
+    let n4_h = bsize.height_mi();
+    let mut tx_w = tx_size.width() as u8;
+    let mut tx_h = tx_size.height() as u8;
+
```
lu-zero: What about?
``` rust
let (tx_w, tx_h) = if skip {
  (n4_w as u8, n4_h as u8)
} else {
  (tx_size.width() as u8, tx_size.height() as u8)
};
```

src/context.rs
```diff
@@ -1395,7 +1455,39 @@ impl BlockContext {
...
+      &mut self.above_tx_context[bo.x..bo.x + n4_w as usize];
+    let left_ctx = &mut self.left_tx_context
+      [bo.y_in_sb()..bo.y_in_sb() + n4_h as usize];
+
+    if skip {
+      tx_w = n4_w as u8;
+      tx_h = n4_h as u8;
+    }
+
+    for i in 0..n4_w {
```
lu-zero:
``` rust 
for v in above_ctx[0..n4_w] {
  *v = tx_w;
}
```
maybe?

src/context.rs
ycho: You mean, factor out of BlockContext struct as separate stuct?
I am no sure. These new contexts are part of av1 spec like the existing *_coeff_context and partition_context, that are used to choose the CDFs, thus should be saved and restored to encode/decode symbols correctly.

src/context.rs
ycho: Thanks for suggestion. Done.

src/context.rs
ycho: Thanks again. (You meant " for v in above_ctx[0..n4_w].iter_mut()", btw )

PR #1145, opened 2019-03-22
Use optimization level 3 for dev builds.

tdaede:
This is required for a lot of inlining to happen, most critically
for the RefType eq method, which otherwise obliterates perf.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,7 @@ name = "bench"
 harness = false
 
 [profile.dev]
-opt-level = 2
+opt-level = 3
 
 [profile.release]
 codegen-units = 1  # if > 1 enables parallel code generation which improves


tdaede:
BTW I filed https://github.com/rust-lang/rust/issues/59375 about this.

shssoichiro:
I wouldn't recommend this. Because opt level 3 inlines so aggressively, it breaks a lot of backtraces by removing contextual info.

tdaede:
I can try bumping the runtime even higher for #1129, I suppose. I could open a PR to libstd, though I haven't actually looked at the derive(eq) implementation yet.

shssoichiro:
I think this is okay temporarily if we need it to unblock a PR. But I also think we need a better long-term solution to the tests timing out, since we keep running into that issue.

lu-zero:
As discussed on irc, ideally I can come up with a `AlwaysInlinedPartialEq` proc_macro while upstream hopefully fixes it.

tdaede:
I'm going to leave this as is for now, I found a workaround for timeouts.

PR #1151, opened 2019-03-26
Add checkpoint structure for BlockContext

rom1v:
`BlockContext::checkpoint()` returned a whole copy of `BlockContext`, while only a subset of its fields were actually part of the checkpoint. The other fields, unused, were filled by default values.

Instead, use a structure dedicated to `BlockContext` checkpoints, containing only the relevant fields.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1284,6 +1284,15 @@ impl IndexMut<usize> for FrameBlocks {
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


rom1v:
Could you confirm that `rows` and `cols` fields of `BlockContext` never change (that were copied to and restored from the checkpoint, I did not included them here)?

PR #1152, opened 2019-03-26
Refactor FrameBlocks

rom1v:
I could have open several PR, but since these changes conflict with each other, I decided to open one for all. If necessary (e.g. if not all changes are accepted), I can split as needed.

_(To minimize confusion, I suggest to review per commit.)_

---

The first two commits remove `BlockContext` methods which return `Block`s by value.

> The methods above_of(), left_of() and above_left_of() returned the
> matching block by value, or a default block if the offset resulted in a
> block outside boundaries.
>
> The Block structure is quite big (std::mem::size_of::<Block>() == 136).
> For reading a field, it is probably not optimal to return a whole Block
> copy or a new default block (although the compiler might optimize such
> accesses).
>
> Moreover, the boundaries checks were often redundant, because already
> done by the callers.
>
> Instead, let the callers access the blocks by their index. This
> simplifies the BlockContext API (which will be adapted for tiling).

The next two commits move methods from `BlockContext` to `FrameBlocks` (they are very specific to blocks, and don't need the context).

For convenience, the last two commits implement `Index` and `IndexMut` traits with `BlockOffset` index for `FrameBlocks`. That way, we can index a block directly from a `BlockOffset`:

```rust
let bo = BlockOffset { x, y };
// before
bc.blocks.at_mut(bo).neighbors_ref_counts = ref_counts;
// after
bc.blocks[bo].neighbors_ref_counts = ref_counts;
```

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -221,10 +221,10 @@ pub fn cdef_analyze_superblock<T: Pixel>(
       // in the main frame.
       let global_block_offset = sbo_global.block_offset(bx<<1, by<<1);
       if global_block_offset.x < bc_global.cols && global_block_offset.y < bc_global.rows {
-        let skip = bc_global.at(global_block_offset).skip
-          & bc_global.at(sbo_global.block_offset(2*bx+1, 2*by)).skip
-          & bc_global.at(sbo_global.block_offset(2*bx, 2*by+1)).skip
-          & bc_global.at(sbo_global.block_offset(2*bx+1, 2*by+1)).skip;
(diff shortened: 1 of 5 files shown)


rom1v:
I don't know why GitHub still does not display commit in topological order instead of chronological order in pull requests, that's so confusing. I'll rewrite the dates.

rom1v:
Sorry, this PR is not review-friendly. I will keep them on my local branch and submit each thing separately one after the other.

Here is the first: https://github.com/xiph/rav1e/pull/1153

PR #1153, opened 2019-03-26
Return blocks by reference, not by value

rom1v:
The methods `above_of()`, `left_of()` and `above_left_of()` returned the matching block by value, or a default block if the offset resulted in a block outside boundaries.

The Block structure is quite big (`std::mem::size_of::<Block>() == 120`). For reading a field, it is probably not optimal to return a whole `Block` copy or a new default block (although the compiler might optimize such accesses).

Moreover, the boundaries checks were often redundant, because already done by the callers.

Instead, let the callers check boundaries and return a reference to the matching block.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1253,7 +1253,7 @@ impl BlockOffset {
   }
 }
 
-#[derive(Copy, Clone)]
+#[derive(Clone)]
 pub struct Block {
   pub mode: PredictionMode,
   pub partition: PartitionType,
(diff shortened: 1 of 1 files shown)


src/context.rs
```diff
@@ -1360,30 +1360,6 @@ impl BlockContext {
     &self.blocks[bo.y][bo.x]
   }
 
-  pub fn above_of(&self, bo: BlockOffset) -> Block {
```
lu-zero: You may change it to return an Option<&Block> and have the code use `if let` or `map().unwrap_or()` instead.

rom1v: Yes, but this would only avoid to return `Block` by value.

The boundaries checks are (often) redundant (already done by the callers), so I would like to also remove them. When they are already checked by the callers, it would be annoying to (also) unwrap the `Option`.

IMO, these helpers functions do not really help compared to direct indexing.

src/context.rs
```diff
@@ -1709,21 +1677,17 @@ impl BlockContext {
 
     match (has_above, has_left) {
       (true, true) => {
-        let above_intra = !self.above_of(bo).is_inter();
-        let left_intra = !self.left_of(bo).is_inter();
+        let above_intra = !self.blocks[bo.y - 1][bo.x].is_inter();
```
ycho: Well, to me this comes like 'rephrasing' and make my debugging much more quicker.

rom1v: Out of curiosity, why `self.blocks[bo.y - 1][bo.x]` makes your debugging easier than `self.above_of(bo)`?

ycho: Well, quite exaggerated, but because I don't need to the function and check the self.blocks[bo.y][bo.x] and its neighbors including the one being read [bo.y-1][bo.x], that is one step deeper. Maybe would not make good sense to others :)

lu-zero:
Looks good :) (and needs a rebase now =/)

rom1v:
Thank you for the review :)

PR #1154, opened 2019-03-26
RDO of transform size for intra block

ycho:
Enabled for speed <= 2, i.e. when fi.config.speed_settings.rdo_tx_decision == true
&& INTRA frame only.
TODO: Currently, partition depths are 0 or 1. Plan to fix for max depth 2 in a separate patch.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -364,39 +364,62 @@ pub fn rdo_tx_size_type<T: Pixel>(
   luma_mode: PredictionMode, ref_frames: [usize; 2], mvs: [MotionVector; 2], skip: bool
 ) -> (TxSize, TxType) {
   use crate::context::max_txsize_rect_lookup;
-  let tx_size = if !fi.tx_mode_select || !luma_mode.is_intra() {
-    max_txsize_rect_lookup[bsize as usize]
-  } else {
-    sub_tx_size_map[max_txsize_rect_lookup[bsize as usize] as usize]  // 1 level partition
-  };
-  debug_assert!(tx_size.width_log2() <= bsize.width_log2());
-  debug_assert!(tx_size.height_log2() <= bsize.height_log2());
 
-  // Luma plane transform type decision
+  let mut tx_size = max_txsize_rect_lookup[bsize as usize];
+  let mut best_tx_type = TxType::DCT_DCT;
+  let mut best_tx_size = tx_size;
+  let mut best_rd = std::f64::MAX;
   let is_inter = !luma_mode.is_intra();
-  let tx_set = get_tx_set(tx_size, is_inter, fi.use_reduced_tx_set);
 
-  let tx_type =
-    if tx_set > TxSet::TX_SET_DCTONLY && fi.config.speed_settings.rdo_tx_decision && !skip {
-      rdo_tx_type_decision(
-        fi,
-        fs,
-        cw,
-        luma_mode,
-        ref_frames,
-        mvs,
-        bsize,
-        bo,
-        tx_size,
-        tx_set,
-      )
-    } else {
-      TxType::DCT_DCT
-    };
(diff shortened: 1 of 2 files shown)


ycho:
[AWCY result for speed 2](https://beta.arewecompressedyet.com/?job=master-0620491_s2%402019-03-26T21%3A19%3A09.911Z&job=test0_rdo_tx_size_intra_s2%402019-03-26T23%3A41%3A18.757Z)

| -1.3858 | -1.1944 | -0.8549 |  -0.9902 | -1.3141 | -0.9488 |    -1.1913 |

ycho:
For speed levels (3~10) that does not use tx partition, encode tx_mode_select = false in frame header. Because if tx_mode_select is true in frame header then each block signals 'no partition' which causes loss in coding efficiency.

src/rdo.rs
```diff
@@ -364,39 +364,67 @@ pub fn rdo_tx_size_type<T: Pixel>(
...
-  debug_assert!(tx_size.height_log2() <= bsize.height_log2());
 
-  // Luma plane transform type decision
+  let mut tx_size = max_txsize_rect_lookup[bsize as usize];
+  let mut best_tx_type = TxType::DCT_DCT;
+  let mut best_tx_size = tx_size;
+  let mut best_rd = std::f64::MAX;
   let is_inter = !luma_mode.is_intra();
-  let tx_set = get_tx_set(tx_size, is_inter, fi.use_reduced_tx_set);
+  let mut tx_set;
```
rom1v: Couldn't it be declared inside the loop (where it is assigned)?

src/rdo.rs
```diff
@@ -364,39 +364,67 @@ pub fn rdo_tx_size_type<T: Pixel>(
...
+
+  for _ in 0..=rdo_tx_depth {
+    tx_set = get_tx_set(tx_size, is_inter, fi.use_reduced_tx_set);
+
+    let do_rdo_tx_type = tx_set > TxSet::TX_SET_DCTONLY &&
+        fi.config.speed_settings.rdo_tx_decision && !skip;
+
+    if rdo_tx_depth == 0 && !do_rdo_tx_type { return (best_tx_size, best_tx_type) };
+
+    if do_rdo_tx_type {
```
rom1v: Same here, I think that `tx_types` could be declared here:
```
let tx_types = if do_rdo_tx_type {…} else {…};
```

ycho: Done, and thanks again!

src/rdo.rs
```diff
@@ -364,39 +364,62 @@ pub fn rdo_tx_size_type<T: Pixel>(
...
+  let cw_checkpoint = cw.checkpoint();
+
+  for _ in 0..=rdo_tx_depth {
+    let tx_set = get_tx_set(tx_size, is_inter, fi.use_reduced_tx_set);
+
+    let do_rdo_tx_type = tx_set > TxSet::TX_SET_DCTONLY &&
+        fi.config.speed_settings.rdo_tx_decision && !skip;
 
-  assert!(tx_size.sqr() <= TxSize::TX_32X32 || tx_type == TxType::DCT_DCT);
+    if rdo_tx_depth == 0 && !do_rdo_tx_type { return (best_tx_size, best_tx_type) };
```
ycho: @lu_zero, by 'early early', I guess you meant here?
It should happen at speed 3~10 (i.e. fi.config.speed_settings.rdo_tx_decision is false).
I see that the condition "rdo_tx_depth == 0 && !do_rdo_tx_type" seems a bit confusing.
It is equivalent to "!do_rdo_tx_size && !do_rdo_tx_type".

src/rdo.rs
ycho: Done. Thanks for catching these! It also saved SIX lines!
(Somehow, the message I entered did not go through bcz I didn't hit "Comment".

PR #1155, opened 2019-03-28
Remove unused get_mvref_ref_frames function.

tdaede:


tdaede:
The RefType commit. It was only ever used to initialize one variable. (in the libaom decoder it's used for symbol decoding I think...)

PR #1158, opened 2019-03-28
Replace at() and at_mut() by operator[]

rom1v:
For convenience, implement `Index` and `IndexMut` traits with `BlockOffset` index for `FrameBlocks`. That way, we can index a block directly from a `BlockOffset`:

```rust
let bo = BlockOffset { x, y };
// before
bc.blocks.at_mut(bo).neighbors_ref_counts = ref_counts;
// after
bc.blocks[bo].neighbors_ref_counts = ref_counts;
```
---

This is a superficial change, but I propose it now because I will do (did on my local branch) it for the tiled version on my local branch.


rom1v:
Thank you @shssoichiro and @lu-zero for the review :+1:

PR #1160, opened 2019-03-29
Refactor encode_block functions into a struct

shssoichiro:
This is the first step in a larger refactor which aims to make the
encoding process follow along with the AV1 spec, making it easier to add
new functionality. The next step will be to break out portions of the
encode methods into their own methods, labeled according to the AV1
spec.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1137,187 +1137,201 @@ pub fn save_block_motion<T: Pixel>(
   }
 }
 
-pub fn encode_block_a<T: Pixel>(
-  seq: &Sequence, fs: &FrameState<T>,
-  cw: &mut ContextWriter, w: &mut dyn Writer,
-  bsize: BlockSize, bo: BlockOffset, skip: bool
-) -> bool {
-  cw.bc.blocks.set_skip(bo, bsize, skip);
-  if fs.segmentation.enabled && fs.segmentation.update_map && fs.segmentation.preskip {
-    cw.write_segmentation(w, bo, bsize, false, fs.segmentation.last_active_segid);
-  }
-  cw.write_skip(w, bo, skip);
-  if fs.segmentation.enabled && fs.segmentation.update_map && !fs.segmentation.preskip {
-    cw.write_segmentation(w, bo, bsize, skip, fs.segmentation.last_active_segid);
-  }
-  if !skip && seq.enable_cdef {
-    cw.bc.cdef_coded = true;
-  }
-  cw.bc.cdef_coded
+pub struct BlockEncoder<'a, T: Pixel> {
+  pub seq: &'a Sequence,
+  pub fs: &'a mut FrameState<T>,
+  pub cw: &'a mut ContextWriter,
+  pub bsize: BlockSize,
+  pub bo: BlockOffset,
+  pub fi: &'a FrameInvariants<T>,
+  pub luma_mode: PredictionMode,
+  pub chroma_mode: PredictionMode,
+  pub ref_frames: [RefType; 2],
+  pub mvs: [MotionVector; 2],
+  pub cfl: CFLParams,
+  pub tx_size: TxSize,
+  pub tx_type: TxType,
+  pub mode_context: usize,
+  pub mv_stack: &'a [CandidateMV],
+  pub rdo_type: RDOType,
+  pub skip: bool,
+  pub for_rdo_use: bool,
 }
 
-pub fn encode_block_b<T: Pixel>(
-  fi: &FrameInvariants<T>, fs: &mut FrameState<T>,
-  cw: &mut ContextWriter, w: &mut dyn Writer,
-  luma_mode: PredictionMode, chroma_mode: PredictionMode,
-  ref_frames: [RefType; 2], mvs: [MotionVector; 2],
-  bsize: BlockSize, bo: BlockOffset, skip: bool,
-  cfl: CFLParams, tx_size: TxSize, tx_type: TxType,
-  mode_context: usize, mv_stack: &[CandidateMV],
-  rdo_type: RDOType, for_rdo_use: bool
-) -> i64 {
-  let is_inter = !luma_mode.is_intra();
-  if is_inter { assert!(luma_mode == chroma_mode); };
-  let sb_size = if fi.sequence.use_128x128_superblock {
-    BlockSize::BLOCK_128X128
-  } else {
-    BlockSize::BLOCK_64X64
-  };
-  let PlaneConfig { xdec, ydec, .. } = fs.input.planes[1].cfg;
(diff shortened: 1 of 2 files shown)


src/encoder.rs
```diff
@@ -1137,187 +1137,201 @@ pub fn save_block_motion<T: Pixel>(
...
+    self.cw.bc.cdef_coded
   }
-  cw.bc.code_deltas = false;
 
-  if fi.frame_type == FrameType::INTER {
-    cw.write_is_inter(w, bo, is_inter);
-    if is_inter {
-      cw.fill_neighbours_ref_counts(bo);
-      cw.write_ref_frames(w, fi, bo);
+  pub fn encode_block_b<'w>(&mut self, w: &'w mut dyn Writer) -> i64 {
```
rom1v: An implicit lifetime for `'w` is equivalent:

```rust
pub fn encode_block_b(&mut self, w: &mut dyn Writer) -> i64 {
```

shssoichiro: Fixed

rzumer:
SGTM from a design perspective, since it avoids passing mostly the same set of parameters twice, but I don't think that following the AV1 spec closely is a design goal for rav1e. Maybe @tdaede can chime in on that.

tdaede:
Correct, if there are places it makes sense to follow the spec or easier to read then do so (especially in context.rs) But the encode process is not normative so we will almost certainly diverge. There is an obvious case for separating encode_block_a/b farther though - that's to replace portions of them with faster approximations when doing the RDO steps.

Moving everything into a struct is nice for dropping the huge argument list, but I'm a little bit concerned about hiding what state functions can mutate. It's not really an issue here because it's only done for encode_block_a/encode_block_b which take almost the same parameters, but it's worth keeping an eye on.

I'll take a more detailed pass through the patch before r+ing

src/encoder.rs
```diff
@@ -1137,187 +1137,201 @@ pub fn save_block_motion<T: Pixel>(
...
-  }
-  cw.write_skip(w, bo, skip);
-  if fs.segmentation.enabled && fs.segmentation.update_map && !fs.segmentation.preskip {
-    cw.write_segmentation(w, bo, bsize, skip, fs.segmentation.last_active_segid);
-  }
-  if !skip && seq.enable_cdef {
-    cw.bc.cdef_coded = true;
-  }
-  cw.bc.cdef_coded
+pub struct BlockEncoder<'a, T: Pixel> {
```
rom1v: I attempted to rebase my local branch about tiling on this PR, but this structure adds a lot of constraints on lifetimes.

In particular, I use structs with lifetimes, so I replace:
 - `&'a mut ContextWriter` by `&'a mut ContextWriter<'a>`
 - `&'a mut FrameState<T>` by `&'a mut TileStateMut<'a, T>`

The lifetimes constraints cannot be expressed on closure parameters:

```rust
struct X<'a>(&'a mut i32, &'a mut i32);

fn main() {
    fn function<'a>(a: &'a mut i32, b: &'a mut i32) -> X<'a> { X(a, b) }
    let lambda = |a: &mut i32, b: &mut i32| { X(a, b) };
    let mut a = 0;
    let mut b = 1;
    let x = function(&mut a, &mut b);
    // let x = lambda(&mut a, &mut b);
}

It happens to work on this example if we remove the explicit labels:
```
let lambda = |a, b| { X(a,b) }
```
but this works only if the compiler is able to infer them properly.

Therefore, with this change, I need to convert this [closure](https://github.com/xiph/rav1e/blob/master/src/rdo.rs#L582) (which itself contains another closure which should probably be converted too) to a function (which, in itself, is not a bad idea). But then, it does not capture the variables from the environment anymore, so I also need to add `&'a FrameInvariants<T>` and other fields.

But then, all the lifetimes are too constrained.

```rust
error[E0495]: cannot infer an appropriate lifetime for lifetime parameter 'a in function call due to conflicting requirements
   --> src/rdo.rs:716:5
    |
716 | /     luma_chroma_mode_rdo(fi, luma_mode, ts, cw, bsize, bo, &mut best, mvs, ref_frames_set[i], rdo_type,
717 | |              &mode_set_chroma, false, mode_contexts[i], &mv_stacks[i]);
    | |______________________________________________________________________^
    |
note: first, the lifetime cannot outlive the lifetime  as defined on the body at 694:28...
   --> src/rdo.rs:694:28
    |
694 |   mode_set.iter().for_each(|&(luma_mode, i)| {
    |                            ^^^^^^^^^^^^^^^^^
note: ...so that reference does not outlive borrowed content
   --> src/rdo.rs:716:41
    |
716 |     luma_chroma_mode_rdo(fi, luma_mode, ts, cw, bsize, bo, &mut best, mvs, ref_frames_set[i], rdo_type,
    |                                         ^^
note: but, the lifetime must be valid for the lifetime 'a as defined on the function body at 448:26...
   --> src/rdo.rs:448:26
    |
448 | pub fn rdo_mode_decision<'a, T: Pixel>(
    |                          ^^
    = note: ...so that the expression is assignable:
            expected &mut tiling::tile_state::TileStateMut<'_, T>
               found &mut tiling::tile_state::TileStateMut<'a, T>
```
(and many more)

I cannot do what the compiler suggests, since the structure requires all the lifetimes to be the same (`'a`).

One solution to relax these constraints would be to use different lifetimes for all fields, but then that's very verbose:
```rust
pub struct BlockEncoder<'a, 'b, 'c, 'd', 'e, 'f, … T: Pixel> { … }
```

So in practice, it adds a lot of complexity.

Note that I also choose to use the same lifetime on my PR #1162 (https://github.com/xiph/rav1e/pull/1162/files#diff-e0934bb6f70a4f8c3baf22a773c14ec7R1908), but it does not cause much problems in practice (just to constraint the lifetimes of one fonction).

tdaede: Could this complexity be avoided by removing everything that's framewide from the struct? (e.g. everything with 'a)

rom1v: So all references? It would defeat the purpose of this structure I think.

lu-zero:
I wouldn't do that now, @rom1v takes priority and I'd rather avoid yet another large R+W structure if we could avoid it.

shssoichiro:
It looks like there might be some issues with this. I'll at least hold on this until @rom1v is finished with his tiling changes, and then see if there's a way to make this work well.

rom1v:
@shssoichiro Thank you :)

tdaede:
Do you still want to work on this, or drop it for now? (I had started working on something similar to struct-ify some of the many args patched, and then remembered this pr)

shssoichiro:
@tdaede This is no doubt quite out of date, it's probably worth it to scrap this PR and continue with what you were working on.

PR #1162, opened 2019-03-29
Use FrameBlocks and CDFContext by reference

rom1v:
There will be one instance of `BlockContext` and `ContextWriter`  per-tile.

However, they owned respectively `FrameBlocks` and `CDFContext`, which will survive tile-encoding to be used frame-wise, so they will have a longer lifetime.

Therefore, make `BlockContext` and `ContextWriter` reference `FrameBlocks` and `CDFContext` (instead of taking ownership).

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -43,12 +43,12 @@ fn write_b_bench(b: &mut Bencher, tx_size: TxSize, qindex: usize) {
   let sequence = Sequence::new(&Default::default());
   let mut fi = FrameInvariants::<u16>::new(config, sequence);
   let mut w = ec::WriterEncoder::new();
-  let fc = CDFContext::new(fi.base_q_idx);
+  let mut fc = CDFContext::new(fi.base_q_idx);
   let mut fb = FrameBlocks::new(fi.sb_width * 16, fi.sb_height * 16);
   let bc = BlockContext::new(&mut fb);
   let mut fs = FrameState::new(&fi);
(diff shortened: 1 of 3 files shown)


tdaede:
I would like to land this before #1160 as they both touch the same areas of the code, and this one is more constraining.

tdaede:
This looks OK, but since CDFContext is small, it could also simply be copied out of the appropriate tile at the end of the frame.

rom1v:
> This looks OK, but since CDFContext is small, it could also simply be copied out of the appropriate tile at the end of the frame.

Yes, I first did this, but with a reference it leaves more choices about the way we implement it. For example, currently, at frame level, I create the initial `CDFContext` in n copies, then pass them a mutable reference to `encode_tile()`, which just returns the `Vec<u8>` (and we can then read them frame wide, without any copy).

But it could be done with copies, `encode_tile()` could return a `(CDFContext, Vec<u8>)`, this would avoid the reference here. Currently, I prefer with the reference (and it avoids a copy, `std::mem::size_of::<CDFContext>() == 17510`, it's still "big").

lu-zero:
We can go this way initially, but it makes sound that the order in which we encode the tiles is important and we have a dependency between tiles.

rom1v:
> it makes sound that the order in which we encode the tiles is important and we have a dependency between tiles.

No, tiles are encoded independently (in parallel), each one writes to its own `CDFContext`, but at the end of tile encoding, the framewide code can access all the `CDFContext` instances.

PR #1163, opened 2019-03-30
Preliminaries api rework

lu-zero:
Some patches that could land as they are while I keep working on addressing the scenechange+reorder interaction.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -2251,51 +2251,52 @@ pub fn encode_show_existing_frame<T: Pixel>(
   packet
 }
 
-pub fn encode_frame<T: Pixel>(fi: &mut FrameInvariants<T>, fs: &mut FrameState<T>) -> Vec<u8> {
+pub fn encode_frame<T: Pixel>(
+  fi: &mut FrameInvariants<T>, fs: &mut FrameState<T>
+) -> Vec<u8> {
   debug_assert!(!fi.show_existing_frame);
   let mut packet = Vec::new();
-  {
-    if !fi.intra_only {
-      for i in 0..INTER_REFS_PER_FRAME {
-        fi.ref_frame_sign_bias[i] =
-          if !fi.sequence.enable_order_hint {
-            false
-          } else if let Some(ref rec) = fi.rec_buffer.frames[fi.ref_frames[i] as usize] {
-            let hint = rec.order_hint;
-            fi.sequence.get_relative_dist(hint, fi.order_hint) > 0
-          } else {
-            false
-          };
-      }
+  if !fi.intra_only {
+    for i in 0..INTER_REFS_PER_FRAME {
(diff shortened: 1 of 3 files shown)


src/encoder.rs
```diff
@@ -2231,57 +2231,68 @@ fn write_tile_group_header(tile_start_and_end_present_flag: bool) ->
     buf.clone()
   }
 
-pub fn encode_frame<T: Pixel>(fi: &mut FrameInvariants<T>, fs: &mut FrameState<T>) -> Vec<u8> {
+pub fn encode_show_existing_frame<T: Pixel>(
```
rzumer: I would like a second opinion, but maybe `encode_existing_frame()` or something similar would be a better name. `encode_show_existing_frame()` makes sense, but is a bit unintuitive in my opinion. Maybe a short function comment would help.

src/scenechange/mod.rs
```diff
@@ -40,34 +40,46 @@ impl<T: Pixel> Default for SceneChangeDetector<T> {
...
-        let delta_yuv = (
-          (delta_yuv.0 / len as u64) as u16,
-          (delta_yuv.1 / len as u64) as u16,
-          (delta_yuv.2 / len as u64) as u16
-        );
-        let delta_avg = ((delta_yuv.0 + delta_yuv.1 + delta_yuv.2) / 3) as u8;
-        is_change = delta_avg >= self.threshold;
+    if let Some((last_num, ref last_frame)) = self.last_frame {
+      if last_num == frame_num - 1 {
+        is_change = is_changed(last_frame, &curr_frame, self.threshold);
```
rzumer: Extracting this part into a function in my opinion does not do much, as it reduces this function to only a half dozen lines that defer the processing elsewhere. Is there a constraint that requires `is_changed()` to be in its own function?

If kept, I would suggest to change the name to a clearer one with context of what is "changed", as without reading it in context, it is hard to tell what it means at a glance.

src/encoder.rs
lu-zero: Naming things is hard :)

I'll try to put a reference to the specification since the concept itself is indeed not so intuitive.

src/scenechange/mod.rs
lu-zero: It is a step in between some larger rework. I can rename it to something different or drop this patch for the time being. Ideally I could do in a different way.

rzumer: If it is needed for later work then it is fine to merge, I would just like to know the purpose of factoring it out more specifically.

src/encoder.rs
rzumer: Good idea.

src/scenechange/mod.rs
lu-zero: Patch dropped from this set for now.

Issue #1149, opened 2019-03-25
Chroma 422 mode causes enc/dec mismatch and its test cannot detect this


ycho:
It passes "cargo test --release --features=decode_test chroma_sampling_422_aom",
but encode and decoder mismatches if tried for larger size input like:
./target/release/rav1e https://media.xiph.org/video/derf/y4m/flower_garden_422_ntsc.y4m -o test.ivf -r test_rec.y4m --quantizer 100 --speed=5 --limit=1

FYI, that current chroma 422 and 444 are all encoded as INTRA frames, i.e. inter modes are NOT used.

lu-zero:
**aom**:
```
Diff < left / right > :
 [
     43,
     173,
     102,
     133,
     233,
     188,
     76,
     216,
     62,
     170,
     109,
     60,
     86,
     124,
     202,
     36,
     42,
     134,
     240,
     183,
     105,
     215,
     144,
     21,
     16,
     237,
     109,
     130,
     143,
     162,
     6,
     14,
     39,
     136,
     127,
     255,
     118,
     37,
     22,
     41,
     106,
     139,
     167,
     226,
     251,
     185,
     58,
     1,
     94,
     192,
     117,
     14,
     148,
     62,
     7,
     181,
     164,
     15,
     16,
     212,
     202,
     4,
     129,
     1,
     12,
     73,
     201,
     35,
     164,
     97,
     109,
     87,
     248,
     36,
     209,
     126,
     54,
     155,
     35,
     141,
     248,
     58,
     13,
     73,
     39,
     75,
     249,
     135,
     30,
     78,
     53,
     246,
     122,
     55,
     39,
     230,
     174,
     12,
     79,
     178,
     81,
     93,
     41,
     20,
     125,
     197,
     189,
     134,
     248,
     100,
     22,
     138,
     125,
     154,
     126,
     232,
     69,
     179,
     210,
     118,
     64,
     163,
     16,
     120,
     163,
     97,
     230,
     102,
     103,
     43,
     198,
     139,
     219,
     185,
     80,
     137,
     42,
     160,
     67,
     22,
     135,
     128,
     27,
     199,
     159,
     71,
     199,
     210,
     89,
     173,
     231,
     54,
     215,
     104,
     229,
     123,
     138,
     191,
     180,
     228,
<    40,
>    39,
     254,
     48,
     17,
     67,
     229,
<    69,
>    70,
     109,
     105,
     218,
     218,
     80,
     136,
     123,
     61,
[...]

PR #1165, opened 2019-03-30
Update the chroma_sampling test to encode a larger frame

lu-zero:
Makes #1149 apparent in the test.

### DIFF
--- a/src/test_encode_decode.rs
+++ b/src/test_encode_decode.rs
@@ -370,7 +370,7 @@ fn chroma_sampling(decoder: &str, cs: ChromaSampling) {
   let quantizer = 100;
   let limit = 3; // Include inter frames
   let speed = 0; // Test as many tools as possible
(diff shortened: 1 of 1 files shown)


lu-zero:
Funny enough I manage to get an error locally with both my mac and my power9 machine, but it survives travis ...

ycho:
Thanks for looking at this that I reported.
Yes, it passes Travis test, but if I use real sequences it fails. 
So, we have issue(s) to fix in encoder path for 422.

lu-zero:
Let me repeat, this update **fails** on both ppc and a system with the full simd set we support :)

So there is something we do in generic code but we do not in only one of the many impl we are using.

ycho:
OK, I could not find you meant that, if you did, it passes with C code but only fails if simd is used.
So, you are confirming that only simd code needs attention?

PR #1170, opened 2019-04-02
More API rework pending flushing logic fix

lu-zero:
Here some more refactors, some tests and a fix for a corner case condition.

Still in the way of fixing #1081 once for all.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -776,6 +776,10 @@ impl<T: Pixel> ContextInner<T> {
   }
 
   pub fn receive_packet(&mut self) -> Result<Packet<T>, EncoderStatus> {
+    if self.frames_processed == self.limit {
+      return Err(EncoderStatus::EnoughData);
+    }
+
     if self.needs_more_lookahead() {
       return Err(EncoderStatus::NeedMoreFrames);
     }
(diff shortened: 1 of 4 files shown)


src/api.rs
```diff
@@ -489,35 +491,42 @@ impl Config {
       first_pass_data: FirstPassData {
         frames: Vec::new(),
       },
+      pool,
+      },
+      config: self.enc.clone(),
```
shssoichiro: I had trouble reading this because of the indentation, could you please indent the inner struct contents?

src/api.rs
```diff
@@ -489,35 +491,42 @@ impl Config {
...
+  pub(crate) config: EncoderConfig,
   rc_state: RCState,
   maybe_prev_log_base_q: Option<i64>,
   pub first_pass_data: FirstPassData,
   pool: rayon::ThreadPool,
 }
 
+pub struct Context<T: Pixel> {
+  inner: ContextInner<T>,
+  config: EncoderConfig,
```
shssoichiro: What is the use case for having EncoderConfig in both Context and ContextInner? Is it avoidable somehow?

src/api.rs
lu-zero: Sure

src/api.rs
lu-zero: It is avoidable, but in the next set. Keeping it around reduces the amount of code changes.

PR #1174, opened 2019-04-03
Speedup the flush tests

lu-zero:
- Drop the spurious `#[test]` attribute (caused double runs)
- Use speed level 10


rom1v:
Before:
```
$ time cargo test api::test
running 16 tests
real	0m56,919s

$ time cargo test
running 40 tests
real	0m57,700s
```

After:
```
$ time cargo test api::test
running 8 tests
real	0m27,262s

$ time cargo test
running 32 tests
real	0m27,468s
```

That's clearly better, but these tests still represents ~100% of the whole tests. I think they should be enabled only when a specific feature is set.

lu-zero:
Sadly to test that part I need a large number of frames encoded. I can make them ignored by default.

PR #1176, opened 2019-04-03
Remove EdgedPlaneSlice

rom1v:
`EdgedPlaneSlice` was added by #1082 so that a slice never needed to exceed its parent.

However, it added complexity, and happens to be insufficient to adapt `get_intra_edges()` for tiling (we can't retrieve the tile dimensions).

This "reverts" #1082 and solves the problem differently: pass the whole `Plane` (in the future, plane region associated with the tile) and its offset.

It's simpler, and it still does not require a slice to exceed its parent.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -975,13 +975,7 @@ pub fn encode_tx_block<T: Pixel>(
 
   if mode.is_intra() {
     let bit_depth = fi.sequence.bit_depth;
-    let edge_buf = get_intra_edges(
-      &rec.edged_slice(po, 1, 1),
-      tx_size,
-      bit_depth,
-      &fs.input.planes[p].cfg,
-      Some(mode),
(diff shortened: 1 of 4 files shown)


tdaede:
Will this plane be passed read only? Or does the plane region enforce that you only read from the correct region?

ycho:
> Will this plane be passed read only? Or does the plane region enforce that you only read from the correct region?
I think those reference pixels (i.e. left or top or top-right or bottom-left) are read only, but later intra prediction will write the predicted pixels in the block area, as like:
"mode.predict_intra(&**mut** rec.mut_slice(po), "

rom1v:
> Will this plane be passed read only? Or does the plane region enforce that you only read from the correct region?

The plane region enforces that you only read from (or write to, for the mutable version) the correct region.

Issue #1074, opened 2019-03-06
Update initial coarse predicted MVs for 64x64 and 32x32 blocks.


ycho:
This can possibly improve the performance of MVs thus coding efficiency gain.

ycho:
Same issues as https://github.com/xiph/rav1e/pull/1177 and #840.

PR #1177, opened 2019-04-03
Update frame_pmvs[] as soon as better MVs are available for topdown p…

ycho:
#1074 

For topdown partition search, as soon as better (compared to those from initial coarse motion estimation) 64x64 or 32x32 MV is available, update them in frame_pmvs[ ]

| -0.1691 | -0.0341 | -0.1063 |  -0.1481 | -0.1275 | -0.1056 |    -0.1293 |

[AWCY result for default speed 5](https://beta.arewecompressedyet.com/?job=master-bc1ffd6b3f527e4708abbba55ab466311a223a58&job=test0_update_frame_pmvs%402019-04-03T22%3A01%3A10.541Z)

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -456,7 +456,7 @@ impl Default for EncodingSettings {
 pub fn rdo_mode_decision<T: Pixel>(
   fi: &FrameInvariants<T>, fs: &mut FrameState<T>,
   cw: &mut ContextWriter, bsize: BlockSize, bo: BlockOffset,
-  pmvs: &[Option<MotionVector>]
+  pmvs: &mut [Option<MotionVector>]
 ) -> RDOPartitionOutput {
   let mut best = EncodingSettings::default();
 
@@ -525,8 +525,12 @@ pub fn rdo_mode_decision<T: Pixel>(
       let ref_slot = ref_slot_set[i] as usize;
       let cmv = pmvs[ref_slot].unwrap();
 
-
       let b_me = motion_estimation(fi, fs, bsize, bo, ref_frames[0], cmv, pmv);
+
+      if !fi.config.speed_settings.encode_bottomup &&
+        (bsize == BlockSize::BLOCK_32X32 || bsize == BlockSize::BLOCK_64X64) {
+          pmvs[ref_slot] = Some(b_me);
+      };
 
       mvs_from_me.push([
         b_me,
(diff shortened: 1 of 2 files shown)


ycho:
@magwyz , this is what I proposed the other day.

tdaede:
This looks okay, though somehow I was under the impression that pmvs was only for MVs already found in previous frames. We're reading it for current frame mvs as well?

ycho:
> This looks okay, though somehow I was under the impression that pmvs was only for MVs already found in previous frames. We're reading it for current frame mvs as well?

For each frame, before going into SB loop, coarse ME for each SB is done at 1/4 resolution.
Inside SB loop, another coarse ME is done for 32x32 (so four 32x32s in a SB) at 1/2 resolution.
These are stored in frame_pmvs[ ], and they don't change once set.
As I understand, they are used as search center, or 'seed', when ME is done for each block.
Meanwhile, ref MVs (as defined by AV1 spec, which are found by find_mvrefs() then stroed in mv_stack[ ]) are used  as reference MV when encoding the MV.
I am not sure whether Diamond ME is using mvs from previous frame, though. Need to learn this.... :)

src/rdo.rs
```diff
@@ -525,9 +525,13 @@ pub fn rdo_mode_decision<T: Pixel>(
       let ref_slot = ref_slot_set[i] as usize;
       let cmv = pmvs[ref_slot].unwrap();
 
-
       let b_me = motion_estimation(fi, fs, bsize, bo, ref_frames[0], cmv, pmv);
 
+      if !fi.config.speed_settings.encode_bottomup &&
+        (bsize == BlockSize::BLOCK_32X32 || bsize == BlockSize::BLOCK_64X64) {
```
rom1v: Couldn't you just update them unconditionally?

lu-zero:
I'd rather return a new set from `rdo_mode_decision`, but it shouldn't matter much all in all.

src/rdo.rs
magwyz: I guess not because, if I understand well, pmvs are only computed and stored for 64x64 and 32x32 blocks.

ycho: Right, we can but better not doing that. Because, for ex 32x32 has its childrenpartitions (ex: four of 16x16 blocks) split from it during topdown partition search, then if one of 16x16 MV update parent 32x32's pmv then other children may not have better pmv. This case would happen when 32x32 block straddle on object boundary, where 16x16 blocks in it have different MVs.

ycho:
> I am not sure whether Diamond ME is using mvs from previous frame, though. Need to learn this.... :)

I checked that MVs (left/right/top/bottom and co-location) from previous frame is added to predictors for diamond ME for block and for 1/2 coarse ME for 32x32.
The block with // EPZS subset C predictors.
 in get_subset_predictors( )

PR #1182, opened 2019-04-04
Simplify further the API

lu-zero:
This should address the problem found by @dwbuiten (and incidentally make rav1e emit a last keyframe)

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -25,8 +25,6 @@ use std::sync::Arc;
 use std::collections::BTreeSet;
 use std::path::PathBuf;
 
-const LOOKAHEAD_FRAMES: u64 = 10;
-
 // TODO: use the num crate?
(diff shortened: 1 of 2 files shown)


tdaede:
What do you mean by "last keyframe?"

kodabb:
@tdaede the very last frame of the stream is encoded as key

shssoichiro:
Does the last keyframe output resolve some issue, or is it just incidental? I can see the side effect that it would (slightly) increase output filesize, but that's acceptable if it's needed to fix some issue.

dwbuiten:
This PR causes some frames to go missing for me (using crav1e, at least) when encoding foreman_cif.y4m. Specifically frames 241, 242, and 243 vanish.

lu-zero:
Let's drop it for now.

PR #1198, opened 2019-04-09
Do not crash on very tiny images

rom1v:
Commit 8412ff0b70022180c63dffab38140770c410e75a avoided some crashes on tiny images, but it was incomplete: since it did not execute subsampled motion estimation with 32x32 blocks when the image was smaller, it left the array of `Option<MotionVector>` full of `None`, while `rdo_mode_decision()` unwrapped the values.

If the option is `None`, use a default `MotionVector` instead to fix the problem.


shssoichiro:
This is failing Clippy in CI.

Also, would it be easy enough to add a test around this issue?

rom1v:
> Also, would it be easy enough to add a test around this issue?

That would be great. Maybe when I will add tiling tests in `decode_test` feature.

Issue #1197, opened 2019-04-09
Please publish ivf to crates.io


eclipseo:
I'm trying to generate a .crate as part of my build process but I encounter issue with ivf:

```
cargo publish --dry-run
    Updating crates.io index
error: all path dependencies must have a version specified when publishing.
dependency `ivf` does not specify a version
```

```
cargo publish --dry-run --allow-dirty
    Updating crates.io index
warning: manifest has no description, license, license-file, documentation, homepage or repository.
See http://doc.crates.io/manifest.html#package-metadata for more info.
   Packaging rav1e v0.1.0 (/home/bob/packaging/rust-rav1e/rav1e)
   Verifying rav1e v0.1.0 (/home/bob/packaging/rust-rav1e/rav1e)
    Updating crates.io index
error: failed to verify package tarball

Caused by:
  no matching package named `ivf` found
location searched: registry `https://github.com/rust-lang/crates.io-index`
required by package `rav1e v0.1.0 (/home/bob/packaging/rust-rav1e/rav1e/target/package/rav1e-0.1.0)`
```
Would it be possible to publish ivf on crates.io to solve this?

lu-zero:
Should be enough to add a version, but we might start pushing everything to crates.io soon enough.

lu-zero:
Currently `nasm-rs` is a bigger problem.

```
error: crates cannot be published with dependencies sourced from a repository
either publish `nasm-rs` as its own crate and specify a version as a dependency or pull it into this repository and specify it with a path and version
```

lu-zero:
Let me open a better issue regarding publishing.

PR #1203, opened 2019-04-10
Prepare ivf for publishing

lu-zero:
Addresses #1197


eclipseo:
Could you add a License file in the crate too please?

lu-zero:
Doesn't seem needed from what I could [read](https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata)

eclipseo:
> Doesn't seem needed from what I could [read](https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata)

Needed by downstream Linux distro.

lu-zero:
If you have time to send a PR I'd merge it immediately :)

PR #1205, opened 2019-04-12
Use unreachable!() when needed

lu-zero:


lu-zero:
https://github.com/xiph/rav1e/pull/1206 has all the fixes needed.

PR #1208, opened 2019-04-12
Unbreak non-x86

lu-zero:


src/ec.rs
```diff
@@ -558,11 +558,18 @@ where
...
     if cdf.len() == 5 && cfg!(target_feature = "sse2") {
       Self::update_cdf_4_sse2(cdf, s);
     } else {
       Self::update_cdf(cdf, s);
     }
+    }
+    #[cfg(not(target_arch = "x86_64"))]
+    {
+      Self::update_cdf(cdf, s);
+    }
```
barrbrain: This can be simplified by early return.

src/ec.rs
```diff
@@ -558,11 +558,18 @@ where
       }
     }
     self.symbol(s, &cdf[..nsymbs]);
+    #[cfg(target_arch = "x86_64")]
+    {
     if cdf.len() == 5 && cfg!(target_feature = "sse2") {
       Self::update_cdf_4_sse2(cdf, s);
```
barrbrain: Return here to remove the need for an else block.

lu-zero: Sure

PR #1214, opened 2019-04-14
Enable the Clippy's "collapsible_if" lint

vladimir-kazakov:
This PR enables the Clippy's [collapsible_if](https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_if) lint, and fixes all warnings caused by it.


lu-zero:
The code coverage variation is fishy though. Probably another false positive.

PR #1215, opened 2019-04-14
Segment threading support

lu-zero:
Yet another stab at it. Each step needs some serious testing.

- [x] Move the segment decision logic out of the inner encoding loop
- [ ] Make each segment use a separate inner encoding context
- [x] Move the encoding process to a separate thread
- [ ] Use rayon parallel iterators to process the segments in parallel


src/api.rs
```diff
@@ -466,19 +466,18 @@ impl Config {
...
-      inner: ContextInner {
+    let inner = ContextInner {
         frame_count: 0,
         limit: 0,
         idx: 0,
         frames_processed: 0,
         frame_q: BTreeMap::new(),
         frame_invariants: BTreeMap::new(),
         keyframes: BTreeSet::new(),
+        keyframes_tmp: BTreeSet::new(),
```
tdaede: Please come up with a better name for this...

lu-zero: The plan is to have only one of this in the end but it would be `detected_keyframes`. if it has to stay.

EwoutH:
So far a slight reduction in encoding speed (2-3%), and a small drop in bitrate and (both default quantizer). [Settings here](https://docs.google.com/spreadsheets/d/1JpPz-Tvj8OO9VGyv-0_VQQRlMVpChuwYeRwr3L3Ss1E/edit#gid=515838162).

| -s 10   | cd3ef47 | 1a4a449 |        |
|---------|--------:|--------:|-------:|
| FPS     |   0,132 |   0,129 |  97,7% |
| Bitrate |   923,8 |   862,0 |  93,3% |
| PSNR    |  49,675 | 49,1965 | 99,04% |

| -s 5    | cd3ef47 | 1a4a449 |        |
|---------|--------:|--------:|-------:|
| FPS     |   0,108 |   0,105 |  97,2% |
| Bitrate |   791,4 |   770,9 |  97,4% |
| PSNR    |  49,987 |  49,695 | 99,42% |

lu-zero:
The speed difference probably is due the instrumentation, the bitrate variation on the other hand should be discussed ^^;

I'll move the instrumentation out of the way now.

lu-zero:
Thanks to @SmilingWolf for testing and finding a strange interaction with `-l`.

tdaede:
Can you make this landable just with the first commit? We really need to fix the GOP ending issue but I'd rather make the changes on top of that.

tdaede:
(maybe with the second checkbox as well, I dunno)

tdaede:
I did some trial encodes with this patch and the md5s all matched.

src/api.rs
tdaede: What this really is is "first pass data". detected_keyframes works for now.

lu-zero:
I tried to test the `-l 22` and `-l 24` with @SmilingWolf 's sample and I can reproduce the problem

lu-zero:
> In the meantime, as the case like coarse motion estimations, I would not exclude the possibility that introducing another multithreading on the top of tmulti-thread running tile encoding can cause overheads, eating the benefits of multi-threading?

The two have different uses:
- with tile-mt you pay quality for less latency and easier decoding.
- with segment-mt you pay latency for throughput but you would not impact the quality.

Ideally one could mix and match them to find its optimal trade-off. (frame-mt would sit somehow in-between the two).

ycho:
> > In the meantime, as the case like coarse motion estimations, I would not exclude the possibility that introducing another multithreading on the top of tmulti-thread running tile encoding can cause overheads, eating the benefits of multi-threading?
> 
> The two have different uses:
> 
>     * with tile-mt you pay quality for less latency and easier decoding.
> 
>     * with segment-mt you pay latency for throughput but you would not impact the quality.
> 
> 
> Ideally one could mix and match them to find its optimal trade-off. (frame-mt would sit somehow in-between the two).

Right, what you replied are excellence points and addressed important points of variety of trade offs rav1e can possibly achieve over existing tile only (and frame) based multi-threading.
If the computer where encoder runs have enough cores, doing hybrid parallelization, i.e. allowing some level of tile encoding solely for decoder's sake of performance, plus doing parallel encoding elsewhere than tile (that does not sacrifice coding performance as you mentioned) would be best scenario, I think.

ycho:
@lu_zero, By 'segment', do you mean the segment indicated by segment_ID in a frame, i.e. max 8 segments by av1 spec. Otherwise, GOP (Group of Pictures), given that the term 'group' in rav1e codebase indicates a group of frames involved forming one pyramid structure?

ycho:
> @lu_zero, By 'segment', do you mean the segment indicated by segment_ID in a frame, i.e. max 8 segments by av1 spec. Otherwise, GOP (Group of Pictures), given that the term 'group' in rav1e codebase indicates a group of frames involved forming one pyramid structure?

Okay- reading the comments it was indeed GOP.

vibhoothiiaanand:
@lu-zero is this going to be replaced with #2118 Channel API ? 
Or is it different one?

lu-zero:
The channel API is a prerequisite this can be closed for now.

PR #1216, opened 2019-04-15
Document EncoderConfig

vibhoothiiaanand:
Add documentation of the encoderConfig for both impl and  public fields
Ref #1078

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -85,6 +85,8 @@ pub struct EncoderConfig {
   pub train_rdo: bool,
 }
 
+/// Default preset for EncoderConfig,it is a balance between quality and speed.
+/// User can use with_speed_preset to have a default structure with a specific speed level.
 impl Default for EncoderConfig {
   fn default() -> Self {
     const DEFAULT_SPEED: usize = 5;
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -83,6 +83,8 @@ pub struct EncoderConfig {
   pub train_rdo: bool,
 }
 
+/// Default preset for EncoderConfig,it is a balance between quality and speed.
```
lu-zero:
```suggestion
/// Default preset for EncoderConfig: it is a balance between quality and speed.
```

src/api.rs
```diff
@@ -83,6 +83,8 @@ pub struct EncoderConfig {
...
+/// User can use with_speed_preset to have a default structure with a specific speed level.
```
lu-zero:
```suggestion
/// See [EncoderConfig::with_speed_preset].
```

src/api.rs
```diff
@@ -91,6 +93,9 @@ impl Default for EncoderConfig {
 }
 
 impl EncoderConfig {
+
+  /// This is a preset which provides default settings according to a speed value in the specific range 0-10,
```
lu-zero: You should reword it so it points to the SpeedSetting documentation maybe.

tdaede:
@lu-zero are your changes addressed now?

PR #1218, opened 2019-04-15
Document the Color Information structs

vibhoothiiaanand:
This commit adds documentation for

- [ ]   ChromaSampling
- [x] ChromaSamplePosition
- [x] PixelRange
- [x] MatrixCoefficients
- [x] ColorPrimaries
- [ ] TransferCharacteristics
- [x] ColorDescription
- [x] MasteringDisplay
- [x] ContentLight

Ref #1079

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -124,7 +124,7 @@ impl EncoderConfig {
   }
 }
 
-/// Contains all the speed settings
+/// Contain all the speed settings
 #[derive(Clone, Copy, Debug)]
 pub struct SpeedSettings {
   pub min_block_size: BlockSize,
@@ -301,6 +301,7 @@ pub enum PredictionModesSetting {
   ComplexAll,
 }
 
+/// Contain all details of chroma sampling for different types like 4:2:0, 4:2:2, 4:4:4, 4:0:0
 #[derive(Copy, Clone, Debug, PartialEq)]
 #[repr(C)]
 pub enum ChromaSampling {
@@ -317,7 +318,8 @@ impl Default for ChromaSampling {
 }
 
 impl ChromaSampling {
-  // Provides the sampling period in the horizontal and vertical axes.
+
+  /// Provides the sampling period in the horizontal and vertical axes.
   pub fn sampling_period(self) -> (usize, usize) {
     use self::ChromaSampling::*;
     match self {
@@ -329,6 +331,7 @@ impl ChromaSampling {
   }
 }
 
+/// Contain the detail of the chroma sampling position ie. the location of chroma samples wrt luma samples.
 #[derive(Copy, Clone, Debug, PartialEq)]
 #[repr(C)]
 pub enum ChromaSamplePosition {
@@ -343,6 +346,11 @@ impl Default for ChromaSamplePosition {
   }
 }
 
+/// Pixel data can be encoded over 8 or 10 (or 12 but unsupported) bits. When data is expressed in:
+///  - Limited range: Take into account different lower and higher bounds that can be used to express color information, for
+///    Luma 8 bit limited range is 16-235, anything below 16 will be treated as pure black, anything above 235 will be treated
+///    as pure white, whereas for 10 bit the same limits apply between 64 - 940. While Chroma has a different range(16-240).
+///  - Full range: Use the whole width of the available levels (0-255 for 8 bit, 0-1023 for 10 bit) to represent data.
 #[derive(ArgEnum, Debug, Clone, Copy, PartialEq)]
 #[repr(C)]
 pub enum PixelRange {
@@ -357,6 +365,12 @@ impl Default for PixelRange {
     }
 }
 
+/// Contain different types of Matrix Coefficients which are used for deriving the luma and chroma (YCbCr) from the R, G, B primaries.
+/// - This is used to describe the color parameters for the video to be encoded, the default value is `Unspecified`.
+/// - Matrix Coefficients can also covert across a multitude of systems.
+/// - For more information please refer to `Table 4` of [`ISO/IEC 23008-9's ITU H.273 Standard`].
+///
+/// [`ISO/IEC 23008-9's ITU H.273 Standard`]:  https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-H.273-201612-I!!PDF-E&type=items
 #[derive(ArgEnum, Debug, Clone, Copy, PartialEq)]
 #[repr(C)]
 pub enum MatrixCoefficients {
@@ -382,6 +396,7 @@ impl Default for MatrixCoefficients {
     }
 }
 
+/// Contain different types of color primaries according to av1 spec, this also defines the chromaticity co-ordinate of the source picture.
 #[derive(ArgEnum, Debug, Clone, Copy, PartialEq)]
 #[repr(C)]
 pub enum ColorPrimaries {
@@ -405,6 +420,10 @@ impl Default for ColorPrimaries {
     }
 }
 
+/// Contain different types of Transfer Characteristics, they define the optoelectronic transfer charecteristic of the source picture aka Gamma.
+/// - For more information please refer to `Table 3` of [`ISO/IEC 23008-9's ITU H.273 Standard`].
+///
+/// [`ISO/IEC 23008-9's ITU H.273 Standard`]:  https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-H.273-201612-I!!PDF-E&type=items
 #[derive(ArgEnum, Debug, Clone, Copy, PartialEq)]
 #[repr(C)]
 pub enum TransferCharacteristics {
@@ -433,6 +452,7 @@ impl Default for TransferCharacteristics {
     }
 }
 
+/// Contain the description of color which includes Color Primaries, Transfer Characteristics and also Matrix Coefficients.
 #[derive(Copy, Clone, Debug)]
 pub struct ColorDescription {
     pub color_primaries: ColorPrimaries,
@@ -440,6 +460,8 @@ pub struct ColorDescription {
     pub matrix_coefficients: MatrixCoefficients
 }
 
+/// Contain all the detail of picture like chromaticity coordinate of primaries, the chromaticity coordinates of the white point.
+/// The maximum and the minimum number of nits of the display used to master the content.
 #[derive(Copy, Clone, Debug)]
 pub struct MasteringDisplay {
     pub primaries: [Point; 3],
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -298,7 +299,8 @@ impl Default for ChromaSampling {
 }
 
 impl ChromaSampling {
-  // Provides the sampling period in the horizontal and vertical axes.
+
+  /// Provides the sampling period in the horizontal and vertical axes.
```
kodabb: sampling period? i never heard of this way of calling the subsampling amount

src/api.rs
```diff
@@ -310,6 +312,7 @@ impl ChromaSampling {
   }
 }
 
+/// Contains details of chroma sampling position
```
kodabb: "details" is a bit loose, it provide 1 detail, that is the location of chroma wrt luma samples, a brief explanation like below would be nice

src/api.rs
```diff
@@ -324,6 +327,12 @@ impl Default for ChromaSamplePosition {
   }
 }
 
+/// Pixel data can be encoded over 8 or 10 (or 12 but unsupported) bits. When data is expressed in limited range
+/// you need to take into account different lower and higher bounds that can be used to express color information:
+///  - 8 bit limited ranges are within 16-235, anything below 16 will be treated as pure black, anything above 235
+///    will be treated as pure white, whereas for 10 bit the same limits apply between 64 - 940.
+///  - Of course when using full range you may use the whole width of the available levels (0-255 for 8 bit,
+///    and 0-1023 for 10 bit) to represent data.
```
kodabb: these amounts are true for luma only, chroma has a different range (16-240)

also it is a bit verbose, this could easily fit in a single sentence: when writing documentation you can usually avoid adverbs and connecting sentences with "of course"

src/api.rs
```diff
@@ -338,6 +347,9 @@ impl Default for PixelRange {
     }
 }
 
+/// Contains different types of Matrix Coefficients which is used in deriving the luma and chroma (YCbCr) from the R'G'B' primaries or X, Y, Z primaries.
```
kodabb: - R'G'B' is not a thing (it exists but kiwf)
- the matrix coeffs do not covert to ycbcr exclusively, they can covert across a multitude of systems (like ycgco)

src/api.rs
```diff
@@ -338,6 +347,9 @@ impl Default for PixelRange {
...
+/// - This is used to describe the color parameters for the video to be encoded, the default value is Unspecfied.
+/// - This is defined as per ITU H.273 Standrad, for more information please refer to "Table 4 - Interpretation of matrix coefficients (MatrixCoefficients) value".
```
kodabb: maybe mention 23008:9 too, also no need to include the whole name of the table

src/api.rs
```diff
@@ -386,6 +399,9 @@ impl Default for ColorPrimaries {
     }
 }
 
+/// Contains different types of Transfer Characteristics, they define the optoelectronic tranfer charecteristic of the source picture aka Gamma.
+/// - 8 Bit field is sufficient for representation of the TransferCharacteristics.
```
kodabb: how is this information useful, considering that it's true for every other color field?

src/api.rs
```diff
@@ -414,13 +430,16 @@ impl Default for TransferCharacteristics {
...
 
+/// Contain description of color which inclues Color Primaries, Transfer Characteristics and also Matrix Coefficients.
 #[derive(Copy, Clone, Debug)]
 pub struct ColorDescription {
     pub color_primaries: ColorPrimaries,
     pub transfer_characteristics: TransferCharacteristics,
     pub matrix_coefficients: MatrixCoefficients
 }
 
+/// Contain all the detail of picture like chromacity coordinate of primaries(RGB), the chromaticity coordinates of the white point in the color space.\
```
kodabb: redundant `(RGB)`

kodabb: extra `\` at the end

kodabb: just "white point" is enough

src/api.rs
```diff
@@ -429,6 +448,8 @@ pub struct MasteringDisplay {
     pub min_luminance: u32,
 }
 
+/// Contain details like the maximum content light level which is the nit value corresponding to the brightest pixel used anywhere in the frame.\
```
kodabb: either use `contains` here, or use `contain` everywhere, consistency is important for documentation
also extra `\`

src/api.rs
rzumer: That's not new in this PR though.

src/api.rs
rzumer:
```suggestion
/// - This is defined as per ITU H.273 Standard, for more information please refer to "Table 4 - Interpretation of matrix coefficients (MatrixCoefficients) value".
```

src/api.rs
```diff
@@ -363,6 +375,7 @@ impl Default for MatrixCoefficients {
     }
 }
 
+/// Contains different types of the color primaries, they defines how the chromaticy coordinates of the source picture
```
rzumer:
```suggestion
/// Contains different types of the color primaries, they defines how the chromaticity coordinates of the source picture
```

The grammar is a bit awkward overall.

src/api.rs
kodabb: the fact that this line's not new to this PR is not an argument to not improve it, but yeah if the naming is used elsewhere it's better to leave it

src/api.rs
```diff
@@ -317,7 +318,8 @@ impl Default for ChromaSampling {
 }
 
 impl ChromaSampling {
-  // Provides the sampling period in the horizontal and vertical axes.
+
```
kodabb: nit: extra empty line

src/api.rs
```diff
@@ -124,7 +124,7 @@ impl EncoderConfig {
   }
 }
 
-/// Contains all the speed settings
+/// Contain all the speed settings
```
kodabb: sometimes you use `contain` and sometimes you (or someone before you) use `contains` or `provides`: i think it's a redundant verb that could be dropped altogether

src/api.rs
```diff
@@ -301,6 +301,7 @@ pub enum PredictionModesSetting {
   ComplexAll,
 }
 
+/// Contain all details of chroma sampling for different types like 4:2:0, 4:2:2, 4:4:4, 4:0:0
```
kodabb: how about just `different types of chroma sampling`?

just listing the types themselves in the documentation itself will only increase maintenance cost, so it's better to either exclude them or document them inline

src/api.rs
```diff
@@ -329,6 +331,7 @@ impl ChromaSampling {
   }
 }
 
+/// Contain the detail of the chroma sampling position ie. the location of chroma samples wrt luma samples.
```
kodabb: don't use `ie` or `wrt` when writing documentation

documentation needs to be readable, not short to write

src/api.rs
```diff
@@ -343,6 +346,11 @@ impl Default for ChromaSamplePosition {
   }
 }
 
+/// Pixel data can be encoded over 8 or 10 (or 12 but unsupported) bits. When data is expressed in:
```
kodabb: why is there a whole sentence about 8, 10, 12 bits? and a whole explanation for each type? also the documentation only needs to convey what this is used for, this chunk of text would work for a help manual

src/api.rs
```diff
@@ -357,6 +365,12 @@ impl Default for PixelRange {
     }
 }
 
+/// Contain different types of Matrix Coefficients which are used for deriving the luma and chroma (YCbCr) from the R, G, B primaries.
+/// - This is used to describe the color parameters for the video to be encoded, the default value is `Unspecified`.
+/// - Matrix Coefficients can also covert across a multitude of systems.
```
kodabb: same problem here, also what is this line for here? it should be coalesced in the first one

src/api.rs
```diff
@@ -382,6 +396,7 @@ impl Default for MatrixCoefficients {
     }
 }
 
+/// Contain different types of color primaries according to av1 spec, this also defines the chromaticity co-ordinate of the source picture.
```
kodabb: everything is according to the av1 spec, no need to mention that

src/api.rs
```diff
@@ -357,6 +365,12 @@ impl Default for PixelRange {
     }
 }
 
+/// Contain different types of Matrix Coefficients which are used for deriving the luma and chroma (YCbCr) from the R, G, B primaries.
+/// - This is used to describe the color parameters for the video to be encoded, the default value is `Unspecified`.
+/// - Matrix Coefficients can also covert across a multitude of systems.
+/// - For more information please refer to `Table 4` of [`ISO/IEC 23008-9's ITU H.273 Standard`].
+///
+/// [`ISO/IEC 23008-9's ITU H.273 Standard`]:  https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-H.273-201612-I!!PDF-E&type=items
```
kodabb: does this really need a link every time?

src/api.rs
kodabb:
```
[14:54:30] <koda>	the only major problem is that information is too sparse
[14:54:37]  <koda>	first there should be the most important thing
[14:54:41]    <koda>	then any ancillary information
[14:54:46]  <koda>	then tiny details if at all
[14:55:05]  <koda>	Matrix Coefficients is a ripe example
[14:55:11]  <koda>	 /// Contain different types of Matrix Coefficients which are used for deriving the luma and chroma (YCbCr) from the R, G, B primaries. 
[14:55:46]  <koda>	besides that it is not correct (or at least incomplete)
[14:55:59]  <koda>	the important part is actually the second line
[14:56:05]  <koda>	This is used to describe the color parameters for the video to be encoded, the default value is `Unspecified`.
[14:56:29]  <koda>	and the third sentence could be merged in the second one
[14:57:51]  <koda>	 // This is used to describe matrix coeffiencnts in the color parameters, the default value is `Unspecified`.   These are used to convert across a multitude of color systems, such as YCbCr to RGB and vice versa
```

src/api.rs
vibhoothiiaanand: Documenting Inline will be better imo.

src/api.rs
vibhoothiiaanand: Okay, WIll remove it

src/api.rs
rzumer: It is also incorrect that 12-bit color is unsupported (unless it was broken recently).

kodabb:
no more updates to this? :-(

vibhoothiiaanand:
@kodabb Luca was working on it recently during refactoring of API into different in https://github.com/xiph/rav1e/pull/1495/commits/850c540670c4dbd35da943901931c15198a9af30 . That was the latest update about this.

PR #1222, opened 2019-04-16
Bring back desync_finder from the dead.

tdaede:
Activate it with feature=desync_finder rather than debug.

Print to stdout rather than stderr because the latter deadlocks.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1866,20 +1866,20 @@ impl FieldMap {
     for (name, start, end) in &self.map {
       // eprintln!("{} {} {} val {}", name, start, end, addr);
       if addr >= *start && addr < *end {
-        eprintln!(" CDF {}", name);
-        eprintln!("");
+        println!(" CDF {}", name);
+        println!();
         return;
       }
     }
 
-    eprintln!("  CDF address not found {}", addr);
+    println!("  CDF address not found {}", addr);
   }
 }
 
 macro_rules! symbol_with_update {
   ($self:ident, $w:ident, $s:expr, $cdf:expr) => {
     $w.symbol_with_update($s, $cdf);
-    #[cfg(debug)] {
+    #[cfg(feature = "desync_finder")] {
       if let Some(map) = $self.fc_map.as_ref() {
         map.lookup($cdf.as_ptr() as usize);
       }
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -1866,22 +1866,23 @@ impl FieldMap {
     for (name, start, end) in &self.map {
       // eprintln!("{} {} {} val {}", name, start, end, addr);
       if addr >= *start && addr < *end {
-        eprintln!(" CDF {}", name);
-        eprintln!("");
+        println!(" CDF {}", name);
+        println!("");
```
lu-zero: clippy is unhappy.

src/context.rs
```diff
@@ -1866,22 +1866,23 @@ impl FieldMap {
...
 
 macro_rules! symbol_with_update {
   ($self:ident, $w:ident, $s:expr, $cdf:expr) => {
     $w.symbol_with_update($s, $cdf);
-    #[cfg(debug)] {
+    #[cfg(feature = "desync_finder")] {
+      //println!("{:?}", $cdf);
       if let Some(map) = $self.fc_map.as_ref() {
-        map.lookup($cdf.as_ptr() as usize);
+                map.lookup($cdf.as_ptr() as usize);
```
lu-zero: while at it, this look strange.

src/context.rs
```diff
@@ -3643,20 +3644,15 @@ impl<'a> ContextWriter<'a> {
...
-        0 => &mut self.fc.eob_flag_cdf16[plane_type][eob_multi_ctx],
-        1 => &mut self.fc.eob_flag_cdf32[plane_type][eob_multi_ctx],
-        2 => &mut self.fc.eob_flag_cdf64[plane_type][eob_multi_ctx],
-        3 => &mut self.fc.eob_flag_cdf128[plane_type][eob_multi_ctx],
-        4 => &mut self.fc.eob_flag_cdf256[plane_type][eob_multi_ctx],
-        5 => &mut self.fc.eob_flag_cdf512[plane_type][eob_multi_ctx],
-        _ => &mut self.fc.eob_flag_cdf1024[plane_type][eob_multi_ctx],
-      }
-    );
+    match eob_multi_size {
```
lu-zero: maybe 
``` rust
let eob_flag_cdf = match eob_mult_size { ... };
symbol_with_update!(..., eob_flag_cdf);
```
works the same way?

tdaede: Nope, as the cdf sizes are variable.

PR #1223, opened 2019-04-16
Add a ivf reader

lu-zero:


ivf/src/lib.rs
```diff
@@ -30,3 +30,54 @@ pub fn write_ivf_frame(
...
+
+  let mut _signature = [0u8; 4];
+  let mut tag = [0u8; 4];
+
+  br.read_bytes(&mut _signature)?;
+  let _v0: u16 = br.read(16)?;
+  let _v1: u16 = br.read(16)?;
+  br.read_bytes(&mut tag)?;
+
+  // log::debug!("sign {:?} version {} {} tag {:?}", &signature, v0, v1, &tag);
```
tdaede: drop?

ivf/src/lib.rs
```diff
@@ -30,3 +30,54 @@ pub fn write_ivf_frame(
...
+  pub timebase_den: u32
+}
+
+pub fn read_header(r: &mut dyn io::Read) -> io::Result<Header> {
+  let mut br = BitReader::endian(r, LittleEndian);
+
+  let mut _signature = [0u8; 4];
+  let mut tag = [0u8; 4];
+
+  br.read_bytes(&mut _signature)?;
```
tdaede: Do you want to actually verify the signature?

ivf/src/lib.rs
lu-zero: Sure

ivf/src/lib.rs
lu-zero: Probably it is a good idea.

PR #1225, opened 2019-04-17
wip: Try to add the aarch build target

lu-zero:


tdaede:
That package is in a separate repo, you'll need to add it (if travis even allows that)

barrbrain:
I think we can shelve this now that Travis has AArch64 hosts.

lu-zero:
Agreed.

Issue #1224, opened 2019-04-16
Improve --low_latency command line option


rom1v:
`--low-latency` is a boolean option, which takes a value (`false` by default).

This is both unnecessary (in general, either we specify the option it or we don't) and very confusing: `rav1e ... --low_latency` means `low_latency` **disabled** (the default is `false`!). `rav1e ... --low_latency true` enabled it.

The behavior should be changed so that `low_latency` is disabled if the option is not present, and enable if it is present.

https://github.com/xiph/rav1e/blob/cd3ef475ba6a2a7d9832e78dbc296f64e6e7a859/src/bin/common.rs#L162-L167

ycho:
Just giving " --low_latency" w/o true or false makes it low_latency disabled?
I agree on what you said, if default is false, then I prefer that it become true when giving "--low-latency".

vladimir-kazakov:
> Just giving " --low_latency" w/o true or false makes it low_latency disabled?

This is the old behavior, when low_latency was an option. In this case, you had to specify `true` to enable low latency.

In #1228 I changed this option to a flag, which means that there is no need to provide true or false anymore. If the flag is present, it enables the low latency mode; otherwise, it's disabled.

PR #1228, opened 2019-04-21
Change the low_latency option to a flag

vladimir-kazakov:
This PR fixes the issue #1224. It changes the low_latency option that takes a boolean value to a flag, which can be used without arguments. If the flag is present, the low latency mode is enabled; otherwise, it's disabled.

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -163,8 +163,6 @@ pub fn parse_cli() -> CliOptions {
         .help("Low latency mode; disables frame reordering\n\
             Has a significant speed-to-quality trade-off")
         .long("low_latency")
-        .takes_value(true)
-        .default_value("false")
     )
     .arg(
       Arg::with_name("TUNE")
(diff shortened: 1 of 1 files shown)


tdaede:
Does it now reject --low_latency true/false? If so I'll need to adjust the AWCY arguments to match.

tdaede:
We don't have a versioning policy yet, so I just made one: https://github.com/xiph/rav1e/pull/1231

vladimir-kazakov:
> Does it now reject --low_latency true/false?

@tdaede, yes, using any value with a flag leads to an error, because flags don't expect values.

> This is a breaking change from the user point of view, maybe we should increase rav1e version to 0.2.0?

I guess the version should change for a specific release, not because of every PR with a breaking change. In other words, it should be okay to have breaking changes in the master (development) branch, but the version number should be increased before making a release.

rom1v:
> I guess the version should change for a specific release, not because of every PR with a breaking change. In other words, it should be okay to have breaking changes in the master (development) branch, but the version number should be increased before making a release.

Yes, absolutely.

lu-zero:
I guess we could put a label so when we cut releases we won't forget.

PR #1229, opened 2019-04-22
Static lookup of pointers for mc functions

KyleSiefring:
Just want to know what people think so far.

Despite what the title says, I haven't put in the tables with indexes correlating to feature level.

### DIFF
--- a/src/cpu_features.rs
+++ b/src/cpu_features.rs
@@ -1,0 +1,69 @@
+// Copyright (c) 2019, The rav1e contributors. All rights reserved
+//
+// This source code is subject to the terms of the BSD 2 Clause License and
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
+pub use native::*;
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+pub use x86::*;
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+mod x86 {
+  use std::env;
+
+  #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd)]
+  pub enum CpuFeatureLevel {
+    NATIVE,
+    AVX2,
+  }
+
+  impl CpuFeatureLevel {
+    pub const fn len() -> usize {
+      CpuFeatureLevel::AVX2 as usize + 1
+    }
+
+    #[inline(always)]
+    pub fn as_index(self) -> usize {
+      const LEN: usize = CpuFeatureLevel::len();
+      assert_eq!(LEN & (LEN - 1), 0);
+      self as usize & (LEN - 1)
+    }
+  }
+
+  pub fn detect_cpu_feature_level() -> CpuFeatureLevel {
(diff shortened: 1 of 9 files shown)


src/mc.rs
```diff
@@ -136,72 +140,285 @@ const SUBPEL_FILTERS: [[[i32; SUBPEL_FILTER_SIZE]; 16]; 6] = [
...
+  *const u16,
+  isize,
+  i32,
+  i32,
+  i32,
+  i32,
+  i32
+);
+
+union PutFnUnion {
```
lu-zero: this could be an enum

KyleSiefring: Wouldn't that use another byte if the structs are passed around?

KyleSiefring:
So, I'm not sure what you mean by using ``T:Pixel``. I would have to define the same number of types, because of the bit_depth parameter only being on hbd.

src/mc.rs
lu-zero: We can check the compiler output, ideally it could be not needed though ^^; (But to try I'll need to have time and that will happen only in few days)

src/mc.rs
```diff
@@ -136,72 +140,288 @@ const SUBPEL_FILTERS: [[[i32; SUBPEL_FILTER_SIZE]; 16]; 6] = [
   ]
 ];
 
+// gets an index that can be mapped to a function for a pair of filter modes
+const fn get_2d_mode_idx(mode_x: FilterMode, mode_y: FilterMode) -> usize {
+  mode_x as usize + 4 * (mode_y as usize)
+}
+
+type PutFn =
+  unsafe extern fn(*mut u8, isize, *const u8, isize, i32, i32, i32, i32);
```
sdroege: It would be nice here and for the other `fn`s to give names to the parameters :)

src/mc.rs
lu-zero: You cannot have both None, I assume.

KyleSiefring: I don't think I understand what you are asking.

lu-zero: I can play with that myself later. this set is taking already enough time :)

tdaede:
To summarize my thoughts on this:

* I don't like using ndarray here as it largely duplicates PlaneSlice / PlaneRegion. You could switch the underlying implementation of PlaneSlice instead if you want.

* Not a huge fan of trying to give the rust versions a C ABI in general. I know this is slightly faster but I don't think it's worth it. TBH if you really want to do that I would almost just make the whole Rust function unsafe and take C pointers.

tdaede:
@lu-zero can you re-review this?

src/encoder.rs
```diff
@@ -526,6 +527,7 @@ pub struct FrameInvariants<T: Pixel> {
   /// how much future frames depend on the block (for example, via
   /// inter-prediction).
   pub block_importances: Box<[f32]>,
+  pub cpu_feature_level: CpuFeatureLevel,
```
YaLTeR: This doesn't look like it belongs in `FrameInvariants`

src/cpu_features.rs
```diff
@@ -0,0 +1,69 @@
...
+#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
+pub use native::*;
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+pub use x86::*;
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+mod x86 {
+  use std::env;
+
+  #[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
```
YaLTeR:
```suggestion
  #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
```

src/encoder.rs
KyleSiefring: I'm going to move it to EncoderConfig. It no longer makes sense to put it in Config since I'm using env vars to override the default.

PR #1231, opened 2019-04-22
Add versioning policy

tdaede:


doc/VERSIONING.md
```diff
@@ -0,0 +1,5 @@
+All changes that break the API require a minor-version bump.
```
rom1v: (while version < 1.0.0)

To be more precise :

> Before you reach 1.0.0, anything goes, but if you make breaking changes, increment the minor version. In Rust, breaking changes include adding fields to structs or variants to enums.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field

PR #1232, opened 2019-04-22
Add some comments around frame building.

tdaede:


src/encoder.rs
```diff
@@ -904,10 +914,10 @@ pub struct InterPropsConfig {
   pub reorder: bool,
   pub multiref: bool,
   pub pyramid_depth: u64,
-  pub group_src_len: u64,
-  pub group_len: u64,
+  pub group_src_len: u64, /// number of source frames in group
```
lu-zero: Comment before the variable or clippy (and rustdoc) would cry

PR #1234, opened 2019-04-22
Fix subgop variable name.

tdaede:
It should be "segment" as far as I can tell.


tdaede:
Actually really not sure about this now. It's pretty confusing.

PR #1235, opened 2019-04-22
Simplify frame idx

tdaede:


src/api.rs
```diff
@@ -795,25 +795,20 @@ impl<T: Pixel> ContextInner<T> {
...
+    }
 
-      if !self.needs_more_frames(self.frame_invariants[&idx].number) {
-        self.idx += 1;
-        return Err(EncoderStatus::LimitReached);
-      }
-      idx
-    };
+    if !self.needs_more_frames(self.frame_invariants[&self.idx].number) {
+      self.idx += 1;
```
lu-zero: If we hit the limit we should not have anything more to output and `idx`, `frames_processed`, `segment_start_*` and `frame_count` could be reset before continuing.

(note to me: `needs_more_frames` should be renamed if we keep using it here)

tdaede: Er, this particular line of code is not checking for the limit, is it? Also, why would we reset those?

lu-zero: It does, and that's why I'd rename that function:
``` rust
  !(self.limit == 0 || self.frame_invariants[&self.idx].number < self.limit)
```

tdaede: Bleh. Do you know why we are even incrementing self.idx at all here then?

lu-zero: Because originally it was returning a NeedsMoreFrames and when I changed it to avoid an infiniloop I forgot to remove it ^^;

tdaede: OK, I dropped it. Seems to still pass tests.

PR #1236, opened 2019-04-23
Reuse code for first keyframe and all other keyframes.

tdaede:


rom1v:
> Refactorsdfasdifjwse

:thinking:

rom1v:
> Reuse code for first keyframe and all other keyframes

Is it just to avoid a special case for idx == 0, or because this special case was harmful?

tdaede:
It was just to drop a special case, but it also changes the behavior a bit - the first frame was previously not inserted into the list of keyframes, but now it is. Nothing actually reads it, though.

lu-zero:
I guess it could land if you want, ideally we should stop keeping generations of frame invariants around (and copying them)

shssoichiro:
Is there a way to do lookahead without keeping multiple frames' worth of invariants in memory? The reason for making a Vec of FrameInvariants was to create a path toward implementing frame lookahead.

tdaede:
You need to keep some data structure around, at least. FrameInvariant works, but isn't a 1:1 match for the eventual data you want to keep around.

I think the main concern is the possible confusion of creating a "plan" of FrameInvariants ahead of time, but then having them actually change when encoding. At minimum it needs to be documented well what's the actual FI used and what's the "plan" from the lookahead.

PR #1239, opened 2019-04-23
Remove set_limit from the public API

lu-zero:
It is not really needed and if we want to continue encoding after a flush we should use a different mean.

### DIFF
--- a/src/bin/rav1e.rs
+++ b/src/bin/rav1e.rs
@@ -29,31 +29,43 @@ use crate::decoder::VideoDetails;
 use std::fs::File;
 use std::io::BufWriter;
 
-fn read_frame<T: Pixel, D: Decoder>(ctx: &mut Context<T>, decoder: &mut D, video_info: VideoDetails) {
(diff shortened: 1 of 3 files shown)


src/bin/rav1e.rs
```diff
@@ -29,31 +29,42 @@ use crate::decoder::VideoDetails;
...
+ limit: usize,
+ input: D,
+}
 
-      let _ = ctx.send_frame(Some(Arc::new(frame)));
-    }
-    _ => {
+impl<D: Decoder> Source<D> {
+  fn read_frame<T: Pixel>(&mut self, ctx: &mut Context<T>, video_info: VideoDetails) {
+    if self.limit == 0 {
```
barrbrain: This seems not to fit well with the default value of zero, which indicates to read until the end of the input stream. Unless I'm mistaken, in this case no frames are read.

lu-zero: Good catch!

PR #1253, opened 2019-04-29
Fix hack 2

tdaede:


lu-zero:
Looks prettier this way.

tdaede:
One of the motivations here is to actually not recreate the fi at all. I'm trying to refactor towards that.

kodabb:
can i has a better title for this PR? 🙏

PR #1257, opened 2019-05-01
Move frame away

lu-zero:
In preparation to have an external Frame with additional fields in `api.rs`.


tdaede:
Probably should update VERSIONING.md to include src/frame.rs when/if it's part of the public api.

PR #1259, opened 2019-05-06
Check code style in pre-commit and travis

xuguangxin:
add files with huge changes (> 100 lines) to ignore list.

now we have 20 files in ingore list.

ignore = [
    "src/api.rs",
    "src/bin/common.rs",
    "src/deblock.rs",
    "src/ec.rs",
    "src/encoder.rs",
    "src/mc.rs",
    "src/me.rs",
    "src/partition.rs",
    "src/plane.rs",
    "src/predict.rs",
    "src/rate.rs",
    "src/rdo.rs",
    "src/rdo_tables.rs",
    "src/test_encode_decode.rs",

    "build.rs",
    "src/context.rs", # Partial
    "src/lib.rs", # Did not clean yet to avoid conflicts with open PRs.
    "src/cdef.rs", # Did not try to clean yet.
    "src/lrf.rs", # Did not try to clean yet.
    "tests/aom.rs",
]


vladimir-kazakov:
The usage of the Nightly toolchain looks very suspicious (as in #1266), because there is a stable version of `rustfmt` for the stable version of Rust. It can be installed by executing the `rustup component add rustfmt` command.

xuguangxin:
@vladimir-kazakov , stable version has the same problem 
my fmt version is
rustfmt 1.0.3-stable (d6829d6 2019-02-14)

also, I found the following warnings when I run it with out -- check

```
$ cargo fmt --
Warning: can't set `comment_width = 79`, unstable features are only available in nightly channel.
Warning: can't set `match_arm_blocks = false`, unstable features are only available in nightly channel.
Warning: can't set `fn_args_density = Compressed`, unstable features are only available in nightly channel.
Warning: can't set `trailing_comma = Never`, unstable features are only available in nightly channel.
Warning: can't set `ignore = IgnoreList({"src/context.rs", "src/lib.rs", "build.rs", "src/cdef.rs", "tests/aom.rs", "src/lrf.rs"})`, unstable features are only available in nightly channel.
Warning: can't set `comment_width = 79`, unstable features are only available in nightly channel.
Warning: can't set `match_arm_blocks = false`, unstable features are only available in nightly channel.
Warning: can't set `fn_args_density = Compressed`, unstable features are only available in nightly channel.
Warning: can't set `trailing_comma = Never`, unstable features are only available in nightly channel.
Warning: can't set `ignore = IgnoreList({"src/cdef.rs", "src/lib.rs", "src/lrf.rs", "tests/aom.rs", "src/context.rs", "build.rs"})`, unstable features are only available in nightly channel.
Warning: can't set `comment_width = 79`, unstable features are only available in nightly channel.
Warning: can't set `match_arm_blocks = false`, unstable features are only available in nightly channel.
Warning: can't set `fn_args_density = Compressed`, unstable features are only available in nightly channel.
Warning: can't set `trailing_comma = Never`, unstable features are only available in nightly channel.
Warning: can't set `ignore = IgnoreList({"build.rs", "src/lrf.rs", "src/context.rs", "src/lib.rs", "tests/aom.rs", "src/cdef.rs"})`, unstable features are only available in nightly channel.
Warning: can't set `comment_width = 79`, unstable features are only available in nightly channel.
Warning: can't set `match_arm_blocks = false`, unstable features are only available in nightly channel.
Warning: can't set `fn_args_density = Compressed`, unstable features are only available in nightly channel.
Warning: can't set `trailing_comma = Never`, unstable features are only available in nightly channel.
Warning: can't set `ignore = IgnoreList({"tests/aom.rs", "src/lib.rs", "src/context.rs", "build.rs", "src/cdef.rs", "src/lrf.rs"})`, unstable features are only available in nightly channel.
Warning: can't set `comment_width = 79`, unstable features are only available in nightly channel.
Warning: can't set `match_arm_blocks = false`, unstable features are only available in nightly channel.
Warning: can't set `fn_args_density = Compressed`, unstable features are only available in nightly channel.
Warning: can't set `trailing_comma = Never`, unstable features are only available in nightly channel.
Warning: can't set `ignore = IgnoreList({"build.rs", "tests/aom.rs", "src/lib.rs", "src/context.rs", "src/lrf.rs", "src/cdef.rs"})`, unstable features are only available in nightly channel.

```

any suggestion?

thanks

xuguangxin:
I checked some changes. It seems reasonable.
take https://github.com/xiph/rav1e/blob/d7599d66b3a91a2a72be436e017b313b94d8a11c/benches/bench.rs#L154-L158
for example. It ident 4 spaces instead of 2.

thanks

vladimir-kazakov:
> I found the following warnings when I run it with out -- check
> any suggestion?

Well, as the warning message says, those configuration options are available only in Nightly. They don't exist in Stable. You can see available Stable options by running `cargo fmt -- --help=config`.

I'm afraid the existing config was created using the Nightly version of rustfmt, which, of course, cannot be used in Travis CI. I'd propose to remove unstable options from the config.

That said, some options are actually quite good, and formatting code in a different way, probably, would be not desirable. For example, "[trailing_comma = Never](https://rust-lang.github.io/rustfmt/#trailing_comma)". Perhaps, rav1e mainterners could clarify whether it would be okay to remove unstable options and format code a bit differently.

I would also like to propose to address the formatting problem from a different angle. The main issue of this PR is that these changes are not interesting and not useful. Even if they will be committed, it won't change anything at all. Other people will still continue committing using wrong formatting. You will be fixing it, and they will be breaking it. To solve this problem, rustfmt has to be added to Travis CI as soon as possible, breaking builds that contain wrong formatting. This will stop people committing wrongly formatted code, releasing you from the need to fix it over and over again. Of course, it's not that easy to add rustfmt to Travis CI right now, because almost everything is formatted incorrectly, according to rustfmt. So, in the beginning, there will be a need to ignore all this stuff. Since the `ignore` options is still unstable, the only option left is to tell `cargo fmt` which files to check specifically. Fixing multiple files at a time is probably a very big and undesirable change, so it would make sense doing it in a more gentle way, file by file. Let's say, in the beginning, you could check only one file. This would be a good start in adding rustfmt to Travis CI. Later, when more files are fixed, they could be added to the `cargo fmt` command, until eventually all files are fixed, which means that there is no need to specify any files in `cargo fmt` anymore.

xuguangxin:
> Other people will still continue committing using wrong formatting

@vladimir-kazakov, agree. 
I plan to send another patch to enforce the code style.
Since you are mentioned. I will combine it with this patch.
How about https://github.com/xiph/rav1e/pull/1259/commits/185df65d9454a0ac40489514670b7d5ed1eba56d? it will install git commit hook to enforce the code style.
thanks

vladimir-kazakov:
I think having rustfmt in Travis CI would be enough.

I don't like this Git hook, because:
1. It's more complex than just a check in Travis CI.
1. It won't work on Windows.
1. It won't work on any OS that doesn't have Bash.
1. It won't work without the Nightly toolchain.
1. I'm not sure that forcing proper formatting on every commit is humane :) It may change the way how people work with their local branches in an unpleasant for them way. This hook may be very annoying for them, without a good reason for that.

xuguangxin:
> I think having rustfmt in Travis CI would be enough.
> 
> I don't like this Git hook, because:
> 
> 1. It's more complex than just a check in Travis CI.
> 2. It won't work on Windows.
> 3. It won't work on any OS that doesn't have Bash.

This will not break windows or other OS. git will ignore the hook if we do not have bash. But you are right, maybe we need to add this check to travis too.
 
> 4. It won't work without the Nightly toolchain.

Firstly, Nightly just a temporary solution. We will move to stable if cargo fmt supports styles we wanted. 
Secondly, if we add format check in travis. Every contributor needs to install the nightly toolchain. else they cannot fix fmt issue reported by travis. 
Thirdly, the pre commit only check the code when you want to commit code, it will not bring an issue to normal people

> 5. I'm not sure that forcing proper formatting on every commit is humane :) It may change the way how people work with their local branches in an unpleasant for them way. This hook may be very annoying for them, without a good reason for that.

Force format for every commit is a good thing. If not, you have to append a code formatting patch for every pr. Someone wants to cherry-pick your bug fixing patches will have a problem, they may need to find and apply your formatting patch too.

thanks

xuguangxin:
@vladimir-kazakov , added the pre-commit check to travis.

vladimir-kazakov:
> This will not break windows or other OS. git will ignore the hook if we do not have bash.

What's the point in such check, if it won't work for everyone?

> Check it on one OS better than check it on none OS :)

I don't understand how it is better. The end result is the same - some code is not formatted properly.

> Nightly just a temporary solution. We will move to stable if cargo fmt supports styles we wanted.

What if not? What if they'll abandon those unstable ideas? I don't think it's worth depending on unstable things.

> Secondly, if we add format check in travis. Every contributor needs to install the nightly toolchain. else they cannot fix fmt issue reported by travis.

I wasn't proposing adding Nightly to Travis CI. I'd prefer to depend only on Stable, both locally and in Travis CI. Speaking of issues, it's not quite correct. If you know what's wrong, sure you can fix it yourself, without using rustfmt for that.

> the pre commit only check the code when you want to commit code, it will not bring an issue to normal people

Normal people don't commit? :)

> Force format for every commit is a good thing. If not, you have to append a code formatting patch for every pr. Someone wants to cherry-pick your bug fixing patches will have a problem, they may need to find and apply your formatting patch too.

Usually, when the PR is ready to be merged, it's rebased on top of the master branch and squashed into a single commit, which merges formatting changes with the rest of the changes. No need to cherry-pick anything.

xuguangxin:
@vladimir-kazakov , thanks for the detailed answer.

> I don't understand how it is better. The end result is the same - some code is not formatted properly.

At least on linux, it can help you make sure your code style is right for every commit.

> > Nightly just a temporary solution. We will move to stable if cargo fmt supports styles we wanted.
> 
> What if not? What if they'll abandon those unstable ideas? I don't think it's worth depending on unstable things.

We can remove most of  unstabled code styles. 
But one important thing missed in the stable toolchain, the [ignore](https://github.com/rust-lang/rustfmt/blob/531b2d9136c41908a44c0e45da6fdf0635c3e680/Configurations.md#ignore) list. Without it, we need to format the entire repo before we enforce the code style.


> > Force format for every commit is a good thing. If not, you have to append a code formatting patch for every pr. Someone wants to cherry-pick your bug fixing patches will have a problem, they may need to find and apply your formatting patch too.
> 
> Usually, when the PR is ready to be merged, it's rebased on top of the master branch and squashed into a single commit, which merges formatting changes with the rest of the changes. No need to cherry-pick anything.

Not every PR just has one commit only, take https://github.com/xiph/rav1e/pull/1126/commits as an example.  If you are not following the code style. You need to add a code formatting patch at the end of PR .
I know you have a concern about forcing everyone to fix very commit. It will introduce a burden to developers.  But, have a good style for every commit is important too.
How about this:
1. Switch to stable toolchain for fmt, and format entire repo.
2. Add git hooks to format code. It should work like git  [trailing-space ](https://stackoverflow.com/questions/591923/make-git-automatically-remove-trailing-whitespace-before-committing), All ill-style code will be silently corrected and combined with the original commit.
3. Add code style check to Travis. It's a gatekeeper for windows and mac. They can't get the benefit of 2.  When travis find their long PR has code style problem,  they may need append code style patch at end of commits.

xuguangxin:
@vladimir-kazakov , update. Please help review https://github.com/xiph/rav1e/pull/1259/commits
Instead of blocking check-in. this approach will show some wrong and automatically apply the right format for you.  The process just like this:

> $ git ci -a -m "fix"
> Diff in /home/thomas/codec/rav1e/build.rs at line 6:
>  #[allow(unused_imports)]
>  use std::env;
>  use std::fs;
> -    use std::path::Path;
> -    use std::path;
> +use std::path;
> +use std::path::Path;
> 
>  #[allow(dead_code)]
>  fn rerun_dir<P: AsRef<Path>>(dir: P) {
> [WARN]code style check failed. We will apply above changes with your commit
> 
> [localmaster 52a587f] fix
>  1 file changed, 1 insertion(+)

thanks

xuguangxin:
interesting, crago fmt's output conflict with cargo clippy...
https://travis-ci.org/xiph/rav1e/jobs/536116941#L3140

vladimir-kazakov:
> Instead of blocking check-in. this approach will show some wrong and automatically apply the right format for you.

Unfortunately, rustfmt is not that smart to understand what is more readable for people. For example: https://github.com/xiph/rav1e/pull/1259/commits/a77427d25d7cd04f4691bbcc4f7f9c838bdf2614#diff-1dc5c602cfb76d06b9a53e29defa44daR679.

Because of that, I'm against the idea of letting rustfmt to format code automatically. The code is written for people, not for compilers, so only people should be in control of how the code looks like.

I also don't think that it's a good idea to format all code in a single commit. It's very difficult to review so many changes.

> But one important thing missed in the stable toolchain, the ignore list. Without it, we need to format the entire repo before we enforce the code style.

You can specify names of files that has to be checked in the `cargo fmt` command. This will act as the ignore list (everything that is not specified is ignored).

> All ill-style code will be silently corrected and combined with the original commit.

There is a danger that these silent modification will be less readable. Because of that, it's very important to not do anything silently. The author of the code better knows how the code should look like. There may be suggestions from rustfmt, and they are welcome, but it should be up to the author whether these suggestions should be applied.

> a good style for every commit is important too

I agree, but only if the good style really improves readability. If not, it's annoying.

In my opinion, rustfmt, currently, has many issues with readability and formatting code adequately, which may be annoying. rav1e already has some code marked with `#[rustfmt::skip]`. Your commits also add trailing commas, which don't really add any value, since they don't make the code more readable. I'm not saying that rustfmt shouldn't be used. I'm trying to say that it should be used very carefully, and only under constant supervision of a human. rustfmt shouldn't do anything automatically.

.travis.yml
```diff
@@ -46,6 +46,10 @@ after_script:
 
 jobs:
   include:
+      - name: "Check Code Style"
```
vladimir-kazakov: This will cause a separate build, which is time consuming. I guess verification of code style could be done together with verification of code quality (Clippy).

.travis.yml
```diff
@@ -46,6 +46,10 @@ after_script:
...
+        script:
+         - rustup component add rustfmt
+         - cargo fmt -- --check
```
vladimir-kazakov: Before running this command, it would be also nice to output the version of rustfmt, just to better understand what Travis CI is currently using. This may help debugging potential issues.

src/lrf.rs
```diff
@@ -1082,22 +1082,24 @@ fn wiener_stripe_filter<T: Pixel>(
   let limit = (1 << (bit_depth + 1 + WIENER_BITS - round_h)) - 1;
 
   let mut work: [i32; MAX_SB_SIZE + 7] = [0; MAX_SB_SIZE + 7];
+  let vfilter3 = 128
```
vladimir-kazakov: This looks weird. I guess it would be better to exclude this piece of code from being formatted by rustfmt at all. It would be also nice to report this issue to Clippy authors.

xuguangxin:
yeah, seems rustfmt is not mature enough. 
Let me close this PR.
Thanks, @vladimir-kazakov  for helpful discussion

PR #1263, opened 2019-05-07
Add an activity mask

shreevari:


shreevari:
I'm not calculating distortion. I'm calculating spatial activity. What I'm doing is calculating the mean and variance of 8x8 blocks and normalizing these values and then storing the output as a subsampled(by a factor of 8 on both axes) plane as an activity mask.

src/encoder.rs
```diff
@@ -344,6 +346,53 @@ impl<T: Pixel> FrameState<T> {
...
   }
 
+  pub fn compute_activity_mask(&mut self) {
+    //activity mask is computed only for luma plane
+
+    let luma_plane = &self.input.planes[0];
+    let PlaneConfig { width, height, .. } = luma_plane.cfg;
+    let luma = &luma_plane.data;
+    let PlaneConfig { xdec, ydec, .. } = self.activity_mask.cfg;
+    let mut act_mask = vec![0 as u32; (width >> xdec) * (height >> ydec) ];
```
barrbrain: I mentally added the follwing to clarify that this is intended as an 8x8 mask:
```rust
assert!(xdec == 3);
assert!(ydec == 3);
```

src/encoder.rs
```diff
@@ -344,6 +346,53 @@ impl<T: Pixel> FrameState<T> {
...
+
+    for r in 0..(height >> ydec) {
+      for c in 0..(width >> xdec) {
+        let base = (r << xdec) * width + (c << ydec);
```
barrbrain: The factor `width` here in calculating the `base` should be the stride of the luma plane and padding should also be considered. `Plane::index` encapsulates both of these concerns.
```suggestion
        let base = luma_plane.index(r << xdec, c << ydec);
```

src/encoder.rs
```diff
@@ -344,6 +346,53 @@ impl<T: Pixel> FrameState<T> {
...
+
+        //Calculate mean
+        let mut mean: u32 = 0;
+        for x_offset in 0..(1 << xdec) {
+          for y_offset in 0..(1 << ydec) {
+            let offset = (x_offset << xdec) + y_offset;
```
barrbrain: Likewise, here `x_offset` need not be shifted as we are iterating the subsampled locations but `y_offset` should be scaled by the luma stride.
```suggestion
            let offset = x_offset + y_offset * stride; // luma_plane.cfg.stride
```

src/encoder.rs
```diff
@@ -344,6 +346,53 @@ impl<T: Pixel> FrameState<T> {
...
+            mean += temp;
+          }
+        }
+        mean >>= xdec + ydec;
+
+        //Calculate variance
+        let mut var: i32 = 0;
+        for x_offset in 0..(1 << xdec) {
+          for y_offset in 0..(1 << ydec) {
+            let offset = (x_offset << xdec) + y_offset;
```
barrbrain: The same applies to the second pass.

src/encoder.rs
shreevari: It is intended as an 8x8 mask. But if required, it could easily be changed to other block sizes by varying xdec and ydec in the PlaneCfg of the activity mask. It infers xdec and ydec from the `activity_mask` and computes accordingly.
Do you suggest that I add these assertions to be sure we're computing for 8x8 blocks?

src/encoder.rs
shreevari: Wow, that's really clean! I'll make the changes accordingly.

src/encoder.rs
shreevari: Changing his caused `test_encode_decode::tile_encoding_with_stretched_restoration_units_aom` to fail. I'm unsure why.

src/encoder.rs
```diff
@@ -344,6 +346,56 @@ impl<T: Pixel> FrameState<T> {
...
+    let luma_stride = luma_plane.cfg.stride;
+    let luma = &luma_plane.data;
+    let PlaneConfig { xdec, ydec, .. } = self.activity_mask.cfg;
+    assert!(xdec==3);
+    assert!(ydec==3);
+    let mut act_mask = vec![0 as u32; (width >> xdec) * (height >> ydec) ];
+
+    for r in 0..(height >> ydec) {
+      for c in 0..(width >> xdec) {
+        let base = luma_plane.index(r << xdec, c << ydec);
```
barrbrain: The identifiers `r` and `c` are transposed here. The failing test case is resolved by the following:
```suggestion
        let base = luma_plane.index(c << xdec, r << ydec);
```

shreevari: Changed it.

tdaede:
OK, I understand it now. I guess my main remaining concern with this approach is that Plane is really designed for storing pixels, and therefore it's both kinda a weird use of xdec/ydec, and also the resulting value we store is going to be only 8-bit for 8-bit video and 16-bit for 10/12 bit video. Have you verified that the range of the computed activity fits within this limitation for both cases?

Would it make more sense to store a higher resolution activity even for 8-bit video?

tdaede:
Also, the activity mask isn't used by anything, so it'll be really easy for this code to get broken. Are you going to follow up with a patch to use it somewhere? A small unit test that just calls it for an 8x8 plane and checks against a known value would help a lot in the meantime.

barrbrain:
> ... the resulting value we store is going to be only 8-bit for 8-bit video and 16-bit for 10/12 bit video. Have you verified that the range of the computed activity fits within this limitation for both cases?

If we wish to have the activity representation match the pixel range and compand the values, we could store [2σ] (twice the standard deviation rounded to the nearest integer.) (See [Popoviciu's inequality on variances](https://en.wikipedia.org/wiki/Popoviciu%27s_inequality_on_variances).)

shreevari:
> OK, I understand it now. I guess my main remaining concern with this approach is that Plane is really designed for storing pixels, and therefore it's both kinda a weird use of xdec/ydec, and also the resulting value we store is going to be only 8-bit for 8-bit video and 16-bit for 10/12 bit video. Have you verified that the range of the computed activity fits within this limitation for both cases?

I hadn't considered 10/12 bit video. Currently, it creates an 8-bit activity plane for all video.

> Would it make more sense to store a higher resolution activity even for 8-bit video?

Do you mean 16-bit activity plane for 8-bit video?
Well actually, I think that the activity needn't be any more precise than 8-bits even for 10-12 bit video since I plan on using it only for quantization. But if you feel that it may be useful in something chroma related, it would be better to have more precision in the activity plane.

Or maybe, we could have 2 precisions of activity masking and choose whichever is needed. But I don't have enough experience with activity masking to tell whether having more precision yields better results.


> > ... the resulting value we store is going to be only 8-bit for 8-bit video and 16-bit for 10/12 bit video. Have you verified that the range of the computed activity fits within this limitation for both cases?
> 
> If we wish to have the activity representation match the pixel range and compand the values, we could store [2σ] (twice the standard deviation rounded to the nearest integer.) (See [Popoviciu's inequality on variances](https://en.wikipedia.org/wiki/Popoviciu%27s_inequality_on_variances).)

Looks very interesting. Initially I thought of experimenting by taking a variance around median instead of mean, but I have no idea what may be the result. I'll check this out.

shreevari:
> Also, the activity mask isn't used by anything, so it'll be really easy for this code to get broken. Are you going to follow up with a patch to use it somewhere? A small unit test that just calls it for an 8x8 plane and checks against a known value would help a lot in the meantime.

I'll add a unit test in the next patch

src/encoder.rs
```diff
@@ -299,6 +299,7 @@ pub struct FrameState<T: Pixel> {
   pub input: Arc<Frame<T>>,
   pub input_hres: Plane<T>, // half-resolution version of input luma
   pub input_qres: Plane<T>, // quarter-resolution version of input luma
+  pub activity_mask: Plane<T>,
```
tdaede: On IRC, I think you said you wanted activity to be 16-bit always. So this should be a Plane<u16>.

Other than that lgtm.

shreevari: I've not pushed those changes yet 😅. I am looking at adding a function to retrieve the activity of a tile from this mask in TileStateMut. After that, I plan to add a unit test as discussed.

tdaede: Great, thanks!

shreevari:
Rebased it to the latest commit

src/encoder.rs
```diff
@@ -347,6 +349,51 @@ impl<T: Pixel> FrameState<T> {
...
+      }
+    }
+    var.iter_mut().for_each(|sum| *sum = (((*sum >> (xdec+ydec)) as f32).sqrt()) as u32 * 2);
+
+    let max = var.iter().max().unwrap();
+    let activity_mask: &mut [u16] = &mut *self.activity_mask.data;
+    activity_mask
+      .iter_mut()
+      .zip(var.iter())
+      .map(|(var1, var2)| *var1 = if *max!=0 {(1<<16 * var2 / max)} else {0} as u16 )
```
barrbrain: Clippy is rightly warning to be wary of operator precedence near a shift. This line is subtly different from the intended calculation. The odd suggestion given by Clippy demonstrates how it is being parsed.
```suggestion
      .map(|(var1, var2)| *var1 = if *max!=0 {(1<<16) * var2 / max} else {0} as u16 )
```

src/encoder.rs
```diff
@@ -2322,4 +2369,57 @@ mod test {
...
+    let actual_mask = &fs.activity_mask.data;
+    let activity = [
+      65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535,
+      65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535,
+      65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535,
+      65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535,
+      65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535,
+      65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535,
+      65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535,
+      65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535u16];
```
barrbrain: This can be simplified to `let activity = [65535u16; 64];` which may be inlined to the following line.

src/encoder.rs
```diff
@@ -347,6 +349,51 @@ impl<T: Pixel> FrameState<T> {
     }
   }
 
+  pub fn compute_activity_mask(&mut self) {
```
vibhoothiiaanand: If we could have documentation at this time itself, it will be very much helpful in future.
So it will be nice if we could add a few lines what this and subsequent function works in a nice way.

shreevari:
> Hi Sheervari,
> 
> One small nitpick from my side
> For Commit message, I would prefer to keep as "Add" instead of "Added" since you are going to add.

All nitpicks are welcome :) I understand what you're saying. I'll change that in the next commit amendment (where I add some documentation).
Thanks for the review :)

src/encoder.rs
```diff
@@ -2322,4 +2373,50 @@ mod test {
     assert_eq!(RAV1E_PARTITION_TYPES[RAV1E_PARTITION_TYPES.len() - 1],
                PartitionType::PARTITION_SPLIT);
   }
-}
+
+  fn create_frame(frame_width: usize, frame_height: usize) -> (Frame<u16>, FrameInvariants<u16>) {
+    let mut frame = Frame::<u16>::new(frame_width, frame_height, ChromaSampling::Cs420);
```
dwbuiten: Is it really correct to hardcode 4:2:0 here?

shreevari: Not really. Since we're dealing with only luma plane here, I think the chroma subsampling is irrelevant in this test.

dwbuiten: OK, make sense.

shreevari:
A test for TileStateMut::get_activity() has been added.

src/encoder.rs
```diff
@@ -347,6 +351,61 @@ impl<T: Pixel> FrameState<T> {
...
+    }
+    var.iter_mut().for_each(|sum| *sum = (((*sum >> (xdec+ydec)) as f32).sqrt()) as u32 * 2);
+
+    //Activity values are normalized to a 16-bit scale
+    let max = var.iter().max().unwrap();
+    let activity_mask: &mut [u16] = &mut *self.activity_mask.data;
+    activity_mask
+      .iter_mut()
+      .zip(var.iter())
+      .map(|(var1, var2)| *var1 = if *max!=0 {65535 * var2 / max} else {0} as u16 )
```
tdaede: Looking at your WIP patch for activity masking, I think you may actually want to skip the normalization step.

shreevari: Is there any use case for the normalized activity mask?
I have stored the mean variance of blocks in a frame in [another variable](https://github.com/xiph/rav1e/pull/1263/commits/0704c05356ff71dd0bb3a6ff4b9372adf88394be#diff-26006a0497afb96d876927e114a7b29eR303) in case I need to denormalize these values.
But if there are no use cases for a normalized mask, then I'll remove this step and the mean variance value.

shreevari:
@tdaede @barrbrain 
Activity is calculated factoring CSF and is no longer normalized.
Thanks to @lu-zero for #1342

shreevari:
The maximum variance of a block of n-bit pixels is 2^(2*(n-1)).
Bit-depth: 08 => Max variance: 1<<14
Bit-depth: 09 => Max variance: 1<<16
Bit-depth: 10 => Max variance: 1<<18
Hence, I think that we should boost/truncate the deviation values by 2 bits for bit-depths of 08/10 respectively before the forward transform.

shreevari:
I had to change the source branch.
Continues at #1393.

PR #1264, opened 2019-05-07
Turn off enable_filter_intra in sequence header

ycho:
So that "use_filter_intra = false" is not coded for every intra mode block.


ycho:
AWCY result will soon be available here [AWCY for -s 5](https://beta.arewecompressedyet.com/?job=test_enable_intra_flag_off_in_seq_header%402019-05-15T03%3A16%3A03.207Z&job=master_d1ba52e4%402019-05-15T03%3A17%3A00.149Z)

Issue #998, opened 2019-02-14
add a mp4 muxer


kodabb:
in order to favor usability of the command line encoder, we should add a mp4 (isomedia) muxer following the spec available https://aomediacodec.github.io/av1-isobmff/

writing a full mp4 muxer is a pretty big task, so it could be simpler to use a separate library. Some possible candidates are
* l-smash, https://github.com/l-smash/l-smash/compare/av1, currently not working
* libavformat, seems to work, but there still might be some compatibility problems
* gpac, currently used to mux the chimera streams

tdaede:
Another reason this is nice is because .ivf doesn't signal keyframes, requiring a remuxer to have an AV1 parser.

PR #1265, opened 2019-05-08
add mp4 muxer.

xuguangxin:
@kodabb and @tdaede 
This will fix  https://github.com/xiph/rav1e/issues/998.
I am not familiar with rust. Any suggestion is welcome.

Please ignore de2611a and 32b57af, I already sent them out at https://github.com/xiph/rav1e/pull/1259

thanks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ dav1d-sys = { version = "0.2", optional = true }
 aom-sys = { version = "0.1.2", optional = true }
 scan_fmt = { version = "0.2", optional = true }
 ivf = { path = "ivf/", optional = true }
+avformat-sys = { path = "avformat-sys/", optional = true }
 rayon = "1.0"
 bincode = "1.1"
 arrayvec = "0.4.10"
--- a/src/bin/muxer/avformat.rs
+++ b/src/bin/muxer/avformat.rs
@@ -1,0 +1,137 @@
+// Copyright (c) 2017-2019, The rav1e contributors. All rights reserved
+//
+// This source code is subject to the terms of the BSD 2 Clause License and
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+#[cfg(feature = "avformat-sys")]
+use avformat_sys::*;
+
+#[allow(unused_imports)]
+use super::Muxer;
+#[allow(unused_imports)]
+use std::ffi::CString;
+#[allow(unused_imports)]
+use std::io;
+#[allow(unused_imports)]
+use std::mem;
+#[allow(unused_imports)]
+use std::ptr;
+
+#[cfg(feature = "avformat-sys")]
+pub struct AvformatMuxer {
+  context: *mut AVFormatContext,
+  stream_time_base: AVRational ,  //time base get from container
+  time_base: AVRational ,         //set by muxer caller
+  duration: i64,
+}
+
+#[cfg(feature = "avformat-sys")]
+impl AvformatMuxer {
+  pub fn open(path: &str) -> Box<dyn Muxer> {
+    unsafe {
+      let mut context = ptr::null_mut();
+      let p = CString::new(path).expect("convert to cstring failed");
+      match avformat_alloc_output_context2(
+        &mut context,
+        ptr::null_mut(),
+        ptr::null(),
+        p.as_ptr() as *const i8
+      ) {
+        0 => {
+          let mut io = ptr::null_mut();
+          let ret =
+            avio_open(&mut io, p.as_ptr() as *const i8, AVIO_FLAG_WRITE as i32);
+          if ret < 0 {
+            panic!("open ouput file {} failed", path);
+          }
+          (*context).pb = io;
+          let default_time_base = AVRational {num: 1, den: 1};
+          Box::new(AvformatMuxer { context,
+            stream_time_base : default_time_base, time_base: default_time_base,
+            duration: 0,
+          })
+        }
+        e => panic!("open ouput failed, error = {}", e)
+      }
+    }
+  }
+}
+
+#[cfg(feature = "avformat-sys")]
+impl Muxer for AvformatMuxer {
+  fn write_header(
+    &mut self, width: usize, height: usize, framerate_num: usize,
+    framerate_den: usize
+  ) {
+    unsafe {
+      let stream = avformat_new_stream(self.context, ptr::null_mut());
+      if stream.is_null() {
+        panic!("new stream failed");
+      }
+      let param = (*stream).codecpar;
+      (*param).codec_type = AVMediaType::AVMEDIA_TYPE_VIDEO;
+      (*param).codec_id = AVCodecID::AV_CODEC_ID_AV1;
+      (*param).width = width as i32;
+      (*param).height = height as i32;
+      let ret = avformat_write_header(self.context, ptr::null_mut());
+      if ret < 0 {
+        panic!("write header failed error = {}", ret);
+      }
+      self.stream_time_base = (*stream).time_base;
+
+      //set timebase and duration base on fps.
+      self.time_base.den = framerate_num as i32;
+      self.time_base.num = framerate_den as i32;
+      self.duration = av_rescale_q(1, self.time_base, self.stream_time_base);
+    }
+  }
+
+  fn write_frame(&mut self, pts: u64, data: &[u8]) {
+    unsafe {
+      let mut pkt: AVPacket = mem::zeroed();
+      av_init_packet(&mut pkt);
+      pkt.data = data.as_ptr() as *mut _;
+      pkt.size = data.len() as i32;
+
+      let pts = av_rescale_q(pts as i64, self.time_base, self.stream_time_base);
+      pkt.pts = pts;
+      //no b frame, pts equals dts
+      pkt.dts = pts;
+      pkt.duration = self.duration;
+      let ret = av_write_frame(self.context, &mut pkt);
+      if ret < 0 {
+        panic!("write frame failed error = {}", ret);
+      }
+    }
+  }
+
+  fn flush(&mut self) -> io::Result<()> {
+    unsafe {
+      loop {
+        let ret = av_write_frame(self.context, ptr::null_mut());
+        if ret <0 {
+          panic!("write frame failed error = {}", ret)
+        }
+        if ret == 1 {
+          break;
+        }
+      }
+    }
+    Ok(())
+  }
+}
+
+#[cfg(feature = "avformat-sys")]
+impl Drop for AvformatMuxer {
+  fn drop(&mut self) {
+    unsafe {
+      av_write_trailer(self.context);
+      avio_close((*self.context).pb);
+      avformat_free_context(self.context);
+    }
+  }
+}
--- a/src/bin/muxer/mod.rs
+++ b/src/bin/muxer/mod.rs
@@ -10,10 +10,19 @@
 mod ivf;
 use self::ivf::IvfMuxer;
 
+mod mp4;
+use mp4::Mp4Muxer;
+
 mod y4m;
 pub use self::y4m::write_y4m_frame;
 
+#[cfg(feature = "avformat-sys")]
+mod avformat;
+
 use std::io;
+use std::ffi::OsStr;
+use std::path::Path;
+
 
 pub trait Muxer {
   fn write_header(
@@ -27,5 +36,25 @@ pub trait Muxer {
 }
 
 pub fn create_muxer(path: &str) -> Box<dyn Muxer> {
-  IvfMuxer::open(path)
+  if path == "-" {
+    return IvfMuxer::open(path);
+  }
+
+  let ext = Path::new(path)
+    .extension()
+    .and_then(OsStr::to_str)
+    .map(str::to_lowercase)
+    .expect("no extension");
+
+  match &ext[..] {
+    "mp4" => {
+      Mp4Muxer::open(path)
+    }
+    "ivf" => {
+      IvfMuxer::open(path)
+    }
+    _e => {
+      panic!("{} is not a supported extension, please change to .ivf", ext);
+    }
+  }
 }
--- a/src/bin/muxer/mp4.rs
+++ b/src/bin/muxer/mp4.rs
@@ -1,0 +1,29 @@
+// Copyright (c) 2017-2018, The rav1e contributors. All rights reserved
+//
+// This source code is subject to the terms of the BSD 2 Clause License and
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+use super::Muxer;
+
+#[cfg(feature = "avformat-sys")]
+use super::avformat::AvformatMuxer;
+
+pub struct Mp4Muxer {
+
+}
+
+#[allow(unused_variables)]
+impl Mp4Muxer {
+  pub fn open(path: &str) -> Box<dyn Muxer> {
+    #[cfg(feature = "avformat-sys")] {
+      AvformatMuxer::open(path)
+    }
+    #[cfg(not(feature = "avformat-sys"))] {
+      panic!("need avformat-sys for .mp4, please build with --features=\"avformat-sys\", or you can use .ivf extension");
+    }
+  }
+}
--- a/avformat-sys/Cargo.toml
+++ b/avformat-sys/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "avformat-sys"
+version = "0.1.0"
+authors = ["Luca Barbato <lu_zero@gentoo.org>"]
+license = "MIT"
+description = "FFI bindings to ffmpeg"
+edition = "2018"
+build = "build.rs"
+
+[package.metadata.pkg-config]
+libavformat = "58.18.102"
+libavcodec = "58.0.0"
+libavutil = "56.0.0"
+
+[features]
+build_sources = []
+
+[build-dependencies]
+bindgen = "0.49"
+metadeps = "1.1.2"
+
+[dependencies]
--- a/avformat-sys/LICENSE
+++ b/avformat-sys/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Luca Barbato
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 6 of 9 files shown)


xuguangxin:
to test this feature. you need run "cargo run --features="ffmpeg-sys" in.y4m --output out.mp4"

src/bin/muxer.rs
```diff
@@ -153,6 +163,129 @@ impl IvfMuxer {
   }
 }
 
+#[cfg(feature = "ffmpeg-sys")]
```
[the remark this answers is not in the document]
xuguangxin: thanks for the suggestion. I will prepare the patch.

xuguangxin:
@lu-zero fixed, please help review again. 
thanks

src/bin/mux/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+#[cfg(feature = "ffmpeg-sys")]
+mod avformatmuxer;
+#[cfg(feature = "ffmpeg-sys")]
+use avformatmuxer::AvformatMuxer;
+
+use std::ffi::OsStr;
+use std::path::Path;
+
+fn need_container(path: &str) -> bool {
+  let ext =
```
lu-zero: This function is still wrong. You do not want to produce an `ivf`  if the extension is not supported.

src/bin/mux/avformatmuxer.rs
```diff
@@ -0,0 +1,123 @@
...
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+#[cfg(feature = "ffmpeg-sys")]
+use ffmpeg_sys::*;
+
+#[allow(unused_imports)]
```
lu-zero: Why this?

src/bin/mux/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+      //need time to figure out.
+      "mp4" => true,
+      _ => false
+    },
+    _ => false
+  }
+}
+
+#[allow(unreachable_code)]
+pub fn create_muxer(path: &str) -> Box<dyn Muxer> {
```
lu-zero: This should panic if the extension isn't supported.

src/bin/rav1e.rs
```diff
@@ -7,16 +7,13 @@
...
 
-use y4m;
-
 #[macro_use]
 extern crate scan_fmt;
 
 mod common;
 mod decoder;
-mod muxer;
+mod mux;
```
lu-zero: I wouldn't change the name

src/bin/mux/mod.rs
xuguangxin: ok, I will change to this

src/bin/rav1e.rs
xuguangxin: We put the Muxer trait to muxer.rc, clippy will report an error if the directory also named muxer. 
any suggestion for fix this?
thanks

xuguangxin:
@lu-zero , everything changed except the mux namespace. 
thanks

xuguangxin:
rebased for https://github.com/xiph/rav1e/commit/bb8d569030b691a3f003e1133647484b98b467be, 
@kodabb , any comment for this pr?

thanks

kodabb:
it's on my list of code to review, i'll get to it this week

xuguangxin:
sure, take your time.
thanks

src/bin/rav1e.rs
lu-zero: The layout I had in mind is:
```
|- ./
   |- decoder/
      |- mod.rs
      |- y4m.rs
   |- muxer/
      |- ivf.rs
      |- mod.rs
      |- mp4.rs
      |- mkv.rs
   |- rav1e.rs
   |- common.rs
```

I just checked and this seems working :)

xuguangxin: two questions for this. 
1. avformat can support many extensions like .mp4,  .mkv, .webm. Do we really need different files for mp4 and mkv? What is the benefit?
2.  We have two major things in the muxer module. The muxer trait and the create_muxer factory function.  Where we put these two things?
Put them all in mod.rs may introduce cycle dependency. ivf and mp4 depend on mods because of Muxer trait.  mods depend on ivf, mp4 because of create_muxer.

thanks

lu-zero: > two questions for this.
> 
>     1. avformat can support many extensions like .mp4,  .mkv, .webm. Do we really need different files for mp4 and mkv? What is the benefit?

We do.

Avformat is not the only provider for muxing, e.g. l-smash is traditionally preferred to mux mp4 when available.

>     2. We have two major things in the muxer module. The muxer trait and the create_muxer factory function.  Where we put these two things?
>        Put them all in mod.rs may introduce cycle dependency. ivf and mp4 depend on mods because of Muxer trait.  mods depend on ivf, mp4 because of create_muxer.

There aren't problems with cycle dependencies in rust.

xuguangxin: good to know those, I will update according to your comment.
>Avformat is not the only provider for muxing, e.g. l-smash is traditionally preferred to mux mp4 when available.

> There aren't problems with cycle dependencies in rust.

thanks for suggestion

lu-zero: Thank you for your work :) I know how frustrating can be refining a patch for something that works well enough for you already.

xuguangxin: Do not worry. Your request is very reasonable. The way is more clear than my original patch. I am happy to do this.

xuguangxin: updated, please help review.
thanks

kodabb:
> thread 'main' panicked at 'need ffmpeg-sys for .mp4, please build with --features="ffmpeg-sys", or you can use .ivf extesion', src/bin/muxer/mp4.rs:26:7

should this module be enabled by default? i think that the error message could be scary but at least it explains how to do it which i like

>      handler_name    : VideoHandler

is it possible to set the name of the encoder to `rav1e` + version instead of the default videoHandler? or maybe there is a more specific place where this could be added

anyway all my major points were already covered by @lu-zero, so aside my two questions i think it looks good!

thanks for providing this functionality

xuguangxin:
> thread 'main' panicked at 'need ffmpeg-sys for .mp4, please build with --features="ffmpeg-sys", or you can use .ivf extesion', src/bin/muxer/mp4.rs:26:7
> 
> should this module be enabled by default? i think that the error message could be scary but at least it explains how to do it which i like
> 

ffmpeg-sys need ffmpeg dev env, and it may only work with the latest version of ffmpeg.
Enable it by default may introduce trouble if somebody does not care about the mp4.
I can enable it in travis if you think it's ok.


>  handler_name    : VideoHandler
> is it possible to set the name of the encoder to rav1e + version instead of the default videoHandler? or maybe there is a more specific place where this could be added
> 

This is is a new requirement. I can create and follow up it with a new issue.

> anyway all my major points were already covered by @lu-zero, so aside my two questions i think it looks good!
> 
> thanks for providing this functionality

welcome, I am willing to do any thing benefit the av1 eco system.

xuguangxin:
@lu-zero , @kodabb , any more comments on this? Can we merge it now?

thanks

xuguangxin:
Welcome, thanks for careful review.  Constructive feedback from you and kodabb helps me a lot.

kodabb:
@lu-zero all good on my side

src/bin/muxer/avformat.rs
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2017-2018, The rav1e contributors. All rights reserved
```
dwbuiten: Wrong copyright year?

src/bin/muxer/avformat.rs
```diff
@@ -0,0 +1,123 @@
...
+      match avformat_alloc_output_context2(
+        &mut context,
+        ptr::null_mut(),
+        ptr::null(),
+        p.as_ptr() as *const i8
+      ) {
+        0 => {
+          let mut io = ptr::null_mut();
+          let ret =
+            avio_open(&mut io, p.as_ptr() as *const i8, AVIO_FLAG_WRITE);
```
dwbuiten: Why are we using AVIO directly?

src/bin/muxer/avformat.rs
```diff
@@ -0,0 +1,123 @@
...
+      let framerate_num = framerate_num as i32;
+      let duration = (*stream).time_base.den * framerate_den
+        / ((*stream).time_base.num * framerate_num);
+      self.duration = duration as i64;
+    }
+  }
+
+  fn write_frame(&mut self, pts: u64, data: &[u8]) {
+    unsafe {
+      let mut pkt: AVPacket = mem::zeroed();
```
dwbuiten: `av_packet_alloc()`?

src/bin/muxer/avformat.rs
```diff
@@ -0,0 +1,123 @@
...
+      (*param).height = height as i32;
+      let ret = avformat_write_header(self.context, ptr::null_mut());
+      if ret < 0 {
+        panic!("write header failed error = {}", ret);
+      }
+      let framerate_den = framerate_den as i32;
+      let framerate_num = framerate_num as i32;
+      let duration = (*stream).time_base.den * framerate_den
+        / ((*stream).time_base.num * framerate_num);
+      self.duration = duration as i64;
```
dwbuiten: Since this looks to be used for PTS rescaling, storing this as an integer and just multiplying it later seems not-so-accurate (?). Is there a reason not to use proper rational scaling?

src/bin/muxer/avformat.rs
```diff
@@ -0,0 +1,123 @@
...
+    unsafe {
+      let mut pkt: AVPacket = mem::zeroed();
+      av_init_packet(&mut pkt);
+      pkt.data = data.as_ptr() as *mut _;
+      pkt.size = data.len() as i32;
+
+      let pts = pts as i64;
+      pkt.pts = pts * self.duration;
+      //no b frame, pts equals dts
+      pkt.dts = pts * self.duration;
```
dwbuiten: This abuses the fact that rav1e's timestamps will always increment by one. Is this OK?

src/bin/muxer/avformat.rs
```diff
@@ -0,0 +1,123 @@
...
+      if ret < 0 {
+        panic!("write frame failed error = {}", ret);
+      }
+    }
+  }
+
+  fn flush(&mut self) -> io::Result<()> {
+    unsafe {
+      let io = (*self.context).pb;
+      avio_flush(io);
```
dwbuiten: Shouldn't flushing also entail calling `av_write_frame` until it returns 1?

src/bin/muxer/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+use self::ivf::IvfMuxer;
+
+mod mp4;
+use mp4::Mp4Muxer;
+
+mod y4m;
+pub use self::y4m::write_y4m_frame;
+
+#[cfg(feature = "ffmpeg-sys")]
+mod avformat;
```
dwbuiten: What is `avformat` used for in this file?

src/bin/rav1e.rs
```diff
@@ -195,11 +193,23 @@ fn main() {
...
   if video_info.bit_depth == 8 {
     do_encode::<u8, y4m::Decoder<'_, Box<dyn Read>>>(
-      cfg, cli.verbose, progress, err, &mut cli.io.output, &mut source, y4m_enc
+      cfg,
+      cli.verbose,
+      progress,
+      err,
+      &mut *cli.io.output,
+      &mut source,
+      y4m_enc
```
dwbuiten: Unrelated cosmetic changes.

src/bin/rav1e.rs
```diff
@@ -195,11 +193,23 @@ fn main() {
...
   } else {
     do_encode::<u16, y4m::Decoder<'_, Box<dyn Read>>>(
-      cfg, cli.verbose, progress, err, &mut cli.io.output, &mut source, y4m_enc
+      cfg,
+      cli.verbose,
+      progress,
+      err,
+      &mut *cli.io.output,
+      &mut source,
+      y4m_enc
```
dwbuiten: Unrelated cosmetic changes.

Cargo.toml
```diff
@@ -36,6 +36,7 @@ dav1d-sys = { version = "0.2", optional = true }
 aom-sys = { version = "0.1.2", optional = true }
 scan_fmt = { version = "0.2", optional = true }
 ivf = { path = "ivf/", optional = true }
+ffmpeg-sys = { git = "https://github.com/meh/rust-ffmpeg-sys.git", branch = "master", optional = true }
```
dwbuiten: Is there no real crate?

src/bin/muxer/avformat.rs
xuguangxin: for mux, caller need set/close the avio. 

see
https://github.com/FFmpeg/FFmpeg/blob/819ed1df94dff335c2653a83e1d3001aa5797f9e/libavformat/avformat.h#L1374-L1386

src/bin/muxer/avformat.rs
xuguangxin: av_write_frame will not take ownership, 
av_init_packet is enough for our case. It's better if we can use uninitialized memory. But I did not find an easy way in rust.

src/bin/muxer/avformat.rs
xuguangxin: hmm, this is good catch. Maybe I need to use av_rescale_q. Let me check.

src/bin/muxer/avformat.rs
xuguangxin: no, the name is not accurate, you can think duration is time unit. PTS is a number of the unit.
In our special case, one time unit is 1/fps.

src/bin/muxer/avformat.rs
xuguangxin: yes, your are right.

src/bin/rav1e.rs
xuguangxin: it's formatted by cargo fmt.
I will change it.

Cargo.toml
xuguangxin: No, the crate only works with ffmpeg 3.x, we need  github repo to work with 4.x

xuguangxin:
@dwbuiten , thanks for the review. Seems we do have some issues.
Let me update it in the following days.
thanks

Cargo.toml
lu-zero: This could be problematic. Could you please split the set with the first 2 commit in another PR?

I'd land it immediately and then we can figure out how to get the libavformat muxing sorted out without git crates.

xuguangxin: sure, wait 2 mins

xuguangxin: done https://github.com/xiph/rav1e/pull/1304

src/bin/muxer/avformat.rs
dwbuiten: 'better'?

src/bin/muxer/avformat.rs
dwbuiten: Which is exactly what I said above; it's a special case from the way rav1e handles timestamps. I just want to make sure it's OK to assume this will always be the case.

src/bin/muxer/avformat.rs
xuguangxin: I mean the av_init_packet is enough for this case since av_write_frame will not take the packet ownership.
If we use av_packet_alloc, it introduces unnecessary malloc and we need to call av_packet_free at the end of the function. 
Maybe I missed some important things here. Could you explain why you prefer av_packet_alloc?

src/bin/muxer/mod.rs
xuguangxin: sorry for miss this. the mp4.rs depends on avformat, so  I need this to export avfrormat.rs for mp4.rs

src/bin/muxer/avformat.rs
xuguangxin: Just check more code. Actually flush is not so useful for container format.
If we did not call av_write_trailer. the entire stream is not parseable even we flush all things.
Anyway, I will change the code to follow av_write_frame document.

src/bin/muxer/avformat.rs
xuguangxin: Not clearly understand the question. Let explain more on the mechanism.
1. mp4 has a mp4_time_scale,  If it's 1000 it means 1s has 1000 mp4 time unit. 
2. rav1e application will give an fps to the muxer. 
3. We use fps as the video_time_scale. 
4. We set duration as mp4_time_scale/video_time_scale. suppose mp4_time_scale is 100, video_time_scale is 25. it means 1 video time unit has 40(1000/25) mp4 time unit.
5. Pts based on video timescale. we use pts*duration to convert it to mp4 timescale.
If we do not want to use fps as video_time_scale. We need to add new params to muxer and change step 3. But I do not think it's necessary, because use fps as the video time scale is perfectly well for video 
thanks

lu-zero: In certain applications could produce a stream with timestamps like this:
```
0 10 25 26 28 29 32 42
```

Right now it does not happen because our input source cannot behave like that.

Once we accept input with arbitrary timestamps we'll have to deal with it.

src/bin/muxer/avformat.rs
dwbuiten: Mostly just because it avoids uninitialized memory. I have no strong feelings on this.

src/bin/muxer/avformat.rs
dwbuiten: Yes, at some point we will be accepting proper variable framerate input and outputting it as such.

Cargo.toml
xuguangxin: @lu-zero , code rebased.
How to handle this problem?
thanks

tdaede:
Is this ready to be reviewed again?

xuguangxin:
@tdaede , yes, all issues addressed except this one.
I am waiting for feedback from @lu-zero .
thanks 



> No, the crate only works with ffmpeg 3.x, we need github repo to work with 4.x


> This could be problematic. Could you please split the set with the first 2 commit in another PR?
> 
> I'd land it immediately and then we can figure out how to get the libavformat muxing sorted out without git crates.

Cargo.toml
lu-zero: I hadn't time to check which are the crates available. 

Wrapping libavformat isn't this difficult, so worst case we could make the (N+1)th one with just that purpose and put it on crates.io.

xuguangxin: Hi @lu-zero , thanks for suggestion. Any document/guidance for this?

thanks

lu-zero: [here](https://rust-lang.github.io/rust-bindgen/tutorial-0.html) the tutorial, [here](https://github.com/rust-av/aom-rs/tree/master/aom-sys) a working example if you want to try yourself. Ping me on irc and I'll try to help.

xuguangxin: @lu-zero .
Thanks for offering the help. Seems it's not so difficult. I will try it by myself. And ping you if I have a hard issue.

xuguangxin:
@tdaede , I will follow lu-zero's guidance. and start working on it tomorrow. It may need 1~2 weeks to finish the work. 

thanks

xuguangxin:
@lu-zero , after referring to aom-sys. I have made the prototype runnable. Next step is to integrate it to rav1e. I have 2 questions:

1. Seems we have two ways to integrate. 
   1. checkin code to rav1e repo
   2. upload it crates.io.
I prefer i, because ii need cargo.io maintain effort.  We have [many ](https://crates.io/search?q=ffmpeg)ffmpeg packages in crates.io. No one is maintained in the previous half year. I guess if we upload it to cartes.io, we will repeat the same pattern. What is your suggestion?

2. Why name you perfered?  av-sys?


thanks
thanks

xuguangxin:
@lu-zero , done. please help review it.
thanks

av-sys/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "av-sys"
```
lu-zero: maybe call it `avformat-sys`

xuguangxin: sure, updated. please help review it again.
thanks

xuguangxin:
@kodabb and @tdaede any feedback?

thanks

src/bin/muxer/mp4.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+#[allow(unused_variables)]
+impl Mp4Muxer {
+  pub fn open(path: &str) -> Box<dyn Muxer> {
+    #[cfg(feature = "avformat-sys")] {
+      AvformatMuxer::open(path)
+    }
+    #[cfg(not(feature = "avformat-sys"))] {
+      panic!("need avformat-sys for .mp4, please build with --features=\"avformat-sys\", or you can use .ivf extesion");
```
tdaede: typo: extension

xuguangxin:
@tdaede , thanks, updated.

PR #1266, opened 2019-05-09
Do not use nightly tool chain in readme. Nightly tool chain will report misleading information.

xuguangxin:


tdaede:
This changes the entire build to Nightly, which we don't want.

@vladimir-kazakov do you know if we still need Nightly in the readme?

xuguangxin:
@tdaede thanks for the feedback.

Not matter we use "cargo clippy" or "cargo nightly+ clippy"
We still have an issue like this:

> error: casting from `*mut u16` to a more-strictly-aligned pointer (`*mut std::arch::x86_64::__m128i`)

 It's ignored by travis.

We have two options:
1. update the read me to ignore the issue.  like https://github.com/xiph/rav1e/pull/1266/commits/d8eb0873728d45c5c86ce37504ce3b088e2ba6a3
2. fix it and find out why travis ignored this sort of error.

Which you preferred? 

thanks

vladimir-kazakov:
> @vladimir-kazakov do you know if we still need Nightly in the readme?

@tdaede, no, it should work on Stable as well. I used it previously without any issues. The same is related to Travis CI.

> Not matter we use "cargo clippy" or "cargo nightly+ clippy", we still have an issue like this: error: casting from *mut u16 to a more-strictly-aligned pointer (*mut std::arch::x86_64::__m128i)
> It's ignored by travis.

@xuguangxin, are you sure it's really ignored? Previously it wasn't, and everything worked as expected. Maybe it's because of Nightly? Or maybe the current version of Clippy still doesn't know that `clippy::cyclomatic_complexity` was renamed to `clippy::cognitive_complexity`, and this somehow affects something else?

xuguangxin:
`@xuguangxin, are you sure it's really ignored? Previously it wasn't, and everything worked as expected. Maybe it's because of Nightly? Or maybe the current version of Clippy still doesn't know that clippy::cyclomatic_complexity was renamed to clippy::cognitive_complexity, and this somehow affects something else?`
@vladimir-kazakov , yes. I'm sure. Not reproduce the issue, you may need "cargo clean" before you do "cargo clippy"
Not sure it's related to my cargo version or not. My cargo version is "cargo 1.34.0 (6789d8a0a 2019-04-01)". ubuntu 18.04 built-in version.

thanks

vladimir-kazakov:
I tried executing the `cargo clean` command before the `cargo clippy` command from `.travis.yml`, but it still doesn't produce any warnings or errors. Everything works as expected. This is also confirmed by other recent PRs - if Clippy would produce any warnings or errors, Travis CI builds in recent PRs would fail, because currently Clippy is configured to treat warnings as errors.

My version of Clippy is `clippy 0.0.212 (726176e 2019-04-18)`. The same version is used by Travis CI. According to Clippy's [releases](https://github.com/rust-lang/rust-clippy/releases) on GitHub and [this page](https://rust-lang.github.io/rust-clippy/), this is the latest version. You can get it by using `rustup components add clippy`, as explained in Clippy's [README.md](https://github.com/rust-lang/rust-clippy/blob/master/README.md).

To get the version of Clippy, you have to execute `cargo clippy --version`. I assume that you've got a preview version of Clippy by following instructions in rav1e's [README.md](https://github.com/xiph/rav1e/blob/master/README.md), by executing `rustup component add clippy-preview --toolchain nightly`. These instructions are wrong. Clippy doesn't need the Nightly toolchain. It can work on the stable version of Rust. You also don't need to specify `+nightly` when running `cargo clippy`. In other words, the correct setup can be achieved by executing the following sequence of commands:

1. `rustup install stable`
1. `rustup component add clippy`
1. `cargo clippy`

So, everything works as expected, and I don't see a need in changes in this PR. Although, perhaps, the rav1e's README.md could be updated with correct instructions.

I also don't see a need to duplicate the exact Clippy command from `.travis.yml` in `README.md`, because then there will be two separate places that will have to be updated, when the list of ignored Clippy's lints will change. Perhaps it would make sense to tell in `README.md` to get the exact command from `.travis.yml`? This would reduce unnecessary duplication and would simplify the maintenance.

xuguangxin:
@vladimir-kazakov , thanks for the detailed explanation. 
I tried to use stable toolchain. And I checked with my Clippy version it's "clippy 0.0.212 (726176e 2019-04-18)", same as you. But I still have errors.
So I [refined](https://github.com/xiph/rav1e/pull/1266/commits/03f686c556b4c877ffb1c9c7f3294acc17ab9947) the current patch, make the travis call  "cargo clippy" directly. 
It's failed in travis now.  Could you help check why?
thanks

lu-zero:
> I also don't see a need to duplicate the exact Clippy command from `.travis.yml` in `README.md`, because then there will be two separate places that will have to be updated, when the list of ignored Clippy's lints will change. Perhaps it would make sense to tell in `README.md` to get the exact command from `.travis.yml`? This would reduce unnecessary duplication and would simplify the maintenance.

We can add a clippy.sh and call it from travis as well.

(incidentally I'd love to have the clippy toml configuration extended to support this)

xuguangxin:
@lu-zero , put them in clippy toml is a good idea. But seems cargo/clippy developers still working on the configuration.
https://github.com/rust-lang/cargo/issues/5034. 
it's not done yet.

vladimir-kazakov:
> So I refined the current patch, make the travis call "cargo clippy" directly.
> It's failed in travis now. Could you help check why?

Because there are warnings and errors in the code that are not fixed yet (and it's not even clear whether they should be fixed). The original Clippy command in `.travis.yml` disables some of the lints, so that everything can successfully build without any warnings and errors. You removed the list of disabled lints from the Clippy command, which enabled those lints, and as a result, they broke the build.

When I was mentioning the `cargo clippy` command, I actually meant the original command from `.travis.yml`. I didn't see a need to reference the full command, because it's quite long. So, you just need to revert all changes to this file.

> We can add a clippy.sh and call it from travis as well.

This is an option, but unfortunately not cross-platform. Adding such scripts for other platforms, in my opinion, would be more complicated than just telling people where to get the complete command. You would also need to duplicate the full command in every such script, which would cause the duplication.

> I'd love to have the clippy toml configuration extended to support this

Yeah, this would be the best option. I was also surprised that it's still impossible.

xuguangxin:
@vladimir-kazakov sorry for misunderstanding you.
The readme is very misleading. How about updating it like https://github.com/xiph/rav1e/pull/1266/files
thanks

README.md
```diff
@@ -100,40 +100,36 @@ https://arewecompressedyet.com/analyzer/?d=https://people.xiph.org/~mbebenita/an
 
 ## Coding style
 Check code formatting with [rustfmt](https://github.com/rust-lang-nursery/rustfmt) before submitting a PR.
-rav1e currently uses the nightly version of rustfmt.
+rav1e currently uses the stable version of rustfmt.
```
vladimir-kazakov: Just replacing "nightly" with "stable" is a bit confusing, because the stable toolchain is not something unique. I guess it would be better, if the entire related text would be reworded in a better way.

For example, what does "currently" mean in this sentence? Are there any plans to switch to a non-stable toolchain? Why? This word produces too much questions. It would be better, if it wouldn't exist at all.

I think it would be better to replace this sentence with something like "rav1e uses the stable version of Rust (the `stable` toolchain)", and then continue with instructions without constantly repeating "stable" everywhere. Something like "to install `rustfmt`, execute...", and so forth.

README.md
```diff
@@ -100,40 +100,36 @@ https://arewecompressedyet.com/analyzer/?d=https://people.xiph.org/~mbebenita/an
...
 ```
 
 then
 
 ```
-cargo +nightly fmt -- --check
+cargo fmt -- --check
 ```
 
 You should also try [clippy](https://github.com/rust-lang-nursery/rust-clippy).
```
vladimir-kazakov: "you should also try" sounds weird. It's quite important, so perhaps it could be a recommendation?

README.md
```diff
@@ -100,40 +100,36 @@ https://arewecompressedyet.com/analyzer/?d=https://people.xiph.org/~mbebenita/an
...
 
 To install clippy:
 
 ```
-rustup component add clippy-preview --toolchain nightly
+rustup component add clippy
 ```
 
-then
+then you can refer travis script for detailed cargo clippy [command](https://github.com/xiph/rav1e/blob/master/.travis.yml)
```
vladimir-kazakov: The meaning of "travis script" may not be very clear for people who don't know what it is. The exact file name (clickable, as a link) would be more understandable.

I also feel like something important is missing from this sentence. I mean it's not immediately clear what I should look for and why. Could it be somehow addressed as well?

xuguangxin:
@vladimir-kazakov , thanks for good suggestion.
How about the new commit?
thanks

README.md
```diff
@@ -98,42 +98,44 @@ https://arewecompressedyet.com/analyzer/?d=https://people.xiph.org/~mbebenita/an
 
 # Contributing
 
-## Coding style
-Check code formatting with [rustfmt](https://github.com/rust-lang-nursery/rustfmt) before submitting a PR.
-rav1e currently uses the nightly version of rustfmt.
-
-To install nightly:
+## Toolchain
+rav1e uses the stable version of Rust (the stable toolchain)
```
vladimir-kazakov: Missing dot at the end of the sentence.

README.md
```diff
@@ -98,42 +98,44 @@ https://arewecompressedyet.com/analyzer/?d=https://people.xiph.org/~mbebenita/an
...
+cargo fmt -- --check
 ```
 
-You should also try [clippy](https://github.com/rust-lang-nursery/rust-clippy).
-Rust also uses nightly for clippy.
+
+## Code Analysis
+The [clippy](https://github.com/rust-lang-nursery/rust-clippy) will help catch common mistakes and improve your Rust code.
+
+We also recommend you use it before submitting a PR.
```
vladimir-kazakov: The word "also" is not needed.

README.md
```diff
@@ -98,42 +98,44 @@ https://arewecompressedyet.com/analyzer/?d=https://people.xiph.org/~mbebenita/an
...
 
 To install clippy:
 
 ```
-rustup component add clippy-preview --toolchain nightly
+rustup component add clippy
 ```
 
-then
+then you can search "cargo clippy" in  [travis script](https://github.com/xiph/rav1e/blob/master/.travis.yml) for detailed command and run it in your local machine.
```
vladimir-kazakov: It would be nice to see the file name of the "travis script", without the need to hover over the link or click it. I'd personally replace "travis script" with just ".travis.yml".

"in your local machine" could be simplified to "on your machine".

xuguangxin:
@vladimir-kazakov , thanks for helpful suggestions. 
updated.

thanks

vladimir-kazakov:
Looks good to me. Thanks for your contribution.

PR #1269, opened 2019-05-10
Don't allocate in forward transforms.

tdaede:


tdaede:
@dwbuiten  this is the biggest memory usage improvement I've gotten so far

lu-zero:
On my testcases it goes down from `47.6 M` allocation to `14.2 M` allocations, the cumulative memory allocation goes down from `11.2 GB` to `6.2 GB`.

PR #1275, opened 2019-05-11
Enforce a lower bound of 1. Fixes #1211.

### LINKED ISSUE #1211: Better user experience when choosing `-I 0`
Encoder statistics says all re encoded as KEY, but only one frame is encoded repetitively.

rainliu:
cargo run -- nyan.y4m -I 0 --output test.ivf
    Finished dev [optimized + debuginfo] target(s) in 0.43s
     Running `target/debug/rav1e nyan.y4m -I 0 --output test.ivf`
thread 'main' panicked at 'Keyframe interval must be greater than 0', src/bin/common.rs:357:7

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -353,6 +353,10 @@ fn parse_config(matches: &ArgMatches<'_>) -> EncoderConfig {
     }
 
     // Validate arguments
+    if max_interval < 1 {
+      panic!("Keyframe interval must be greater than 0");
+    }
+
     if speed > 10 {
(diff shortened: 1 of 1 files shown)


barrbrain:
We encourage to squash and force-push pull requests to simplify history before merging. We will add some information about this to the contribution section of the README.

To be specific about the commit message, I suggest something like:
> Check if CLI option for keyframe interval is in range
>
> Enforce a lower bound of 1. Fixes #1211.

lu-zero:
Please keep the two-line format suggested.

PR #1279, opened 2019-05-16
Explicitly use the ChaChaRng crate

lu-zero:


tdaede:
Why?

lu-zero:
rand spun off the implementations to be lean so the inner ChaChaRng is deprecated.

PR #1280, opened 2019-05-16
Less allocations

lu-zero:
| | commit | peak memory usage | cumulative allocations |
|-|-|-:|-:|
| before | 465360f63d60b42bdd313a96abf94d376b89f11d | 63.0 MB | 20.1 M |
| after | f4133c6e184fb11438998162623991040b4bd4d4 | 78.3 MB | 5.7 M |

75% less memory allocations, and about 20% higher memory usage.

Data provided by [malt](https://github.com/memtt/malt):

```
malt target/debug/rav1e ../rush_field_cuts_1080p.y4m  --quantizer 85 -o /dev/null -l 2
```


src/context.rs
```diff
@@ -1374,14 +1374,16 @@ impl IndexMut<BlockOffset> for FrameBlocks {
   }
 }
 
+const MAX_B_SIZE: usize = 512;
```
tdaede: This value is the maximum tile width (4096) divided by the partition context granularity (8x8). Can you define it like that?

Actually also that doesn't really match the left contexts, which are MAX_MIB_SIZE. Either those should be dropped in half or this doubled. Would have to check the partition code to be sure.

lu-zero: Sure! the number like that is quite magic indeed  ^^;

lu-zero:
Changes done

src/context.rs
```diff
@@ -1374,14 +1374,17 @@ impl IndexMut<BlockOffset> for FrameBlocks {
   }
 }
 
+const PARTITION_CONTEXT_GRANULARITY: usize = 8;
+const MAX_B_SIZE: usize = MAX_TILE_WIDTH / PARTITION_CONTEXT_GRANULARITY;
```
tdaede: MAX_B_SIZE is still a bad name, normally a "B" is 4x4. How about "PARTITION_CONTEXT_MAX_WIDTH"? I know it's long but it's only used a few places.

lu-zero: Sure, let me change it now.

PR #1281, opened 2019-05-16
Do not use NeedMoreData to signal that a frame had been encoded

lu-zero:
This avoids some confusion @dwbuiten reported some time ago. (yet another cherry-pick from the segment-threading set)

(crav1e will be updated accordingly)

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -550,6 +550,8 @@ pub enum EncoderStatus {
   /// May be emitted by `Context::receive_packet` after a flush request had been processed
   /// or the frame limit had been reached.
   LimitReached,
+  /// A Frame had been encoded but not emitted yet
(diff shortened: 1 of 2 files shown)


src/api.rs
```diff
@@ -550,6 +550,8 @@ pub enum EncoderStatus {
   /// May be emitted by `Context::receive_packet` after a flush request had been processed
   /// or the frame limit had been reached.
   LimitReached,
+  /// A Frame had been encoded but not emitted yet
```
dwbuiten: has

dwbuiten:
Seems reasonable. The only other option I can think of (at 1am) is emitting a zero sized or null packet, but that would be uglier.

PR #1282, opened 2019-05-16
Add signal support

lu-zero:
So the cli can exit cleanly on sigterm.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ ivf = { path = "ivf/", optional = true }
 rayon = "1.0"
 bincode = "1.1"
 arrayvec = "0.4.10"
+signal-hook = { version = "0.1.9", optional = true }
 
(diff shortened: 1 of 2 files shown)


lu-zero:
there is a [ctrlc](https://crates.io/crates/ctrlc) that could be also used for this purpose.

tdaede:
Looks okay. Does this slow down exiting by a lot?

tdaede:
In particular, it may make sense to not catch the second Ctrl-C.

lu-zero:
It slows down by one frame encoding, I can check if I can refine the behavior

tdaede:
OK. Just turning off the handler after one Ctrl-C would be fine. I just predict it being hugely annoying at -s 0 to wait a minute for the encoder to exit.

lu-zero:
Updated accordingly.

PR #1285, opened 2019-05-20
RCStat initalization fixes

dwbuiten:


dwbuiten:
Dropped the kbps patch until I can ask @lu-zero Some Stuff.

tdaede:
Uhhh, yikes. That explains the persistent undershoot.... (does it behave a lot better now?)

dwbuiten:
Yes, from my limited testing. The first frame still looks real bad, however.

However:
* https://git.xiph.org/?p=users/tterribe/rav1e.git;a=commit;h=3f178a5cf0e8a18c5701438c0cd4d5df08ccdab7
* `guess_frame_subtype` is broken and returning the wrong number of subtypes, and @tterribe is working to fix that.
* I've been working with @barrbrain and @tterribe to try and make the initial quant numbers less bad, but have uh... let's call it hit stumblings blocks. See notebook: https://gist.github.com/barrbrain/e4304ec3d75ffebb50f8bc531282e98b

PR #1288, opened 2019-05-21
Add trial encodes for first frame of each type

barrbrain:
An optimal content-independent scale estimator is up to an order of magnitude off, depending on content complexity. Account for that by running a trial encode for the first frame of each type and replacing the estimated scale with the observed value. Note that the scale filter has a latency of one frame of the same type, so we need to update the initial scale value as well.

### DIFF
--- a/src/rate.rs
+++ b/src/rate.rs
@@ -802,6 +802,32 @@ impl RCState {
     }
     dropped
   }
+
+  pub fn needs_trial_encode(&self, fti: usize) -> bool {
+      self.target_bitrate > 0 && self.nframes[fti] == 0
+  }
+
+  pub fn record_trial_encode(
+    &mut self, bits: i64, fti: usize, log_target_q: i64
+  ) {
+    assert!(self.needs_trial_encode(fti));
+    assert!(bits > 0);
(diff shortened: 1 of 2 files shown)


src/rate.rs
```diff
@@ -802,6 +802,32 @@ impl RCState {
     }
     dropped
   }
+
+  pub fn needs_trial_encode(&self, fti: usize) -> bool {
+      self.target_bitrate > 0 && self.nframes[fti] == 0
```
tdaede: An obvious future improvement is to zero this for all frame types when we insert a detected scene change,  but that can be separate.

dwbuiten: Presumably this would work the same for force keyframes (supposedly forthcoming!).

tdaede: A keyframe that's not at a natural scene change might not need a recode, because its stats are similar to the last keyframe. Dunno if that's worth taking advantage of though.

dwbuiten: I'm not actually sure that would be strictly true. There's also the case where a keyframe may be forced on what would have been selected to be a "scene change" anyway.

PR #1289, opened 2019-05-21
rate: Replace imported Theora scale/exp values with properly trained ones

dwbuiten:
These were derived by running encodes of many clips at each quantizer, an logging their resolution, coded size, and log-space target quantizer. The relationship appeears to be linear, unlike Theora, so the bucketing by ibpp was removed in favour of a single value per frame type.

These values seem to give pretty good ballpark results; enough to base trial frame encodes on them.

This also includes proper values for both B frame depths, and thus removes the reuse of P frame scale/exp for B frames.

The data set used for these particular values are all of the video clips on https://media.xiph.org/video/derf/ as well as subset3 (for extra I frame data). Adding more clips from Vimeo's own corpus did not seem to affect the values much at all, anyway.

Thank you to @tterribe and @barrbrain for fielding my silly questions, and for the help aggregating the data. @barrbrain's notebook is here: https://gist.github.com/barrbrain/e4304ec3d75ffebb50f8bc531282e98b.

I have kept all the data collect in a GCS bucket if anyone wants to look, or it's useful for some other training in the future (though I can't imagine what). The infrastructure I wrote to automate dara collection on GCE should also be reusable for future training.

## BIG NOTE
**This change requires @tterribe's patch to fix the incorrect output from `guess_frame_subtype` is applied before it can work correctly. I have been testing by manually hacking in the correct amounts of frame subtypes.**

### DIFF
--- a/src/rate.rs
+++ b/src/rate.rs
@@ -499,34 +499,20 @@ impl RCState {
     // Start with a buffer fullness and fullness target of 50%.
     let reservoir_target = (reservoir_max + 1) >> 1;
     // Pick exponents and initial scales for quantizer selection.
-    let ibpp = npixels/bits_per_frame;
-    // All of these initial scale/exp values are from Theora, and have not yet
-    //  been adapted to AV1, so they're certainly wrong.
-    // The B-frame values especially are simply copies of the P-frame values.
-    let i_exp: u8;
-    let i_log_scale: i64;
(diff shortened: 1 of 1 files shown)


barrbrain:
> Would love if the .ipynb ended up in the repo :)

See PR #1293 for a version of the notebook cleaned up for inclusion.

PR #1290, opened 2019-05-21
Added --recons for -r RECONSTRUCTION option.

jnickelsburg:


src/bin/common.rs
```diff
@@ -248,6 +248,7 @@ pub fn parse_cli() -> CliOptions {
     .arg(
       Arg::with_name("RECONSTRUCTION")
         .help("Outputs a Y4M file containing the output from the decoder")
+        .long("recons")
```
tdaede: if you're going to go with a long name might as well go with "reconstruction"

jnickelsburg: Done

PR #1293, opened 2019-05-22
Document rate-control empirical analysis

barrbrain:
We performed a simple linear regression of the bitrate with respect to the quantizer, operating on the logarithm of both.
The data set used was all of the video clips on https://media.xiph.org/video/derf/ as well as subset3 (for extra I frame data).
To enable processing an arbitrarily large data set, an online regression algorithm was implemented. In practice, 440MB of text formatted data were sufficient.


barrbrain:
As it turns out, the white-space discrepancy and trust error were unrelated. Jupyter track user-trust of documents on an opt-in basis. Downloaded documents are considered untrusted by default.

PR #1295, opened 2019-05-23
Remove the max limit of reservoir_frame_delay 

dwbuiten:
With the default max keyint of 10s, and common length fades found in content, 256 frames is not sufficient to allow rate control to adapt properly; it is barely even past the GOP boundary.

This will introduce possible overflow problems with `log_scale` when 2-pass mode is intoduced.

# Notes

This fixes horrendous artifacting on some clips, particularily with fades,

Sample file can be found [here](http://chromashift.org/fades_make_me_sad.y4m.xz). It is in xz-compressed y4m form, for @tdaede's convenience.

It can be reproduced easily with: `rav1e --output out.ivf --bitrate 500 fades_make_me_sad.y4m`.

@tterribe says `inter_delay_target` should also be clamped to something like 2x the value of `reservoir_frame_delay`, but I cannot remember the reason (other than it's related to 2-ass), and thus have not included it. I'll loop back on that one...

### DIFF
--- a/src/rate.rs
+++ b/src/rate.rs
@@ -394,6 +394,9 @@ pub struct RCState {
   // The maximum buffer fullness (total size of the buffer).
   reservoir_max: i64,
   // The log of estimated scale factor for the rate model in Q57 format.
+  //
+  // TODO: Convert to Q23 or figure out a better way to avoid overflow
+  // once 2-pass mode is introduced, if required.
   log_scale: [i64; FRAME_NSUBTYPES],
   // The exponent used in the rate model in Q6 format.
   exp: [u8; FRAME_NSUBTYPES],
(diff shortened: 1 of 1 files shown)


src/rate.rs
```diff
@@ -394,6 +394,9 @@ pub struct RCState {
   // The maximum buffer fullness (total size of the buffer).
   reservoir_max: i64,
   // The log of estimated scale factor for the rate model in Q57 format.
+  //
+  // TODO: Convert to Q23 or figured out a better way to avoid overflow
+  // once 2-pass mode is introduced.
```
barrbrain: Typo with "figured"  rather than "figure". I recommend qualifying with an "if required" or similar  somewhere in this comment.

dwbuiten: Done.

tdaede:
Any reason for not going any higher, e.g. 1024? Where does the improvement fall off?

tdaede:
needs rebase

dwbuiten:
@tdaede From what I understood (or misunderstood) from @tterribe , I don't think anything past 1.5x max GOP length would be useful... although that does beg the question what happens on 50 or 60 fps clips!

dwbuiten:
@tdaede Following a talk with with Tim at the watercooler meeting yesterday, I've changed it to just remove the limit entirely.

src/rate.rs
```diff
@@ -477,16 +480,14 @@ impl RCState {
...
+    //  of 12.
     // The interval is short enough to allow reaction, but long enough to allow
     //  looking into the next GOP (avoiding the case where the last frames
     //  before an I-frame get starved).
     // The 12 frame minimum gives us some chance to distribute bit estimation
     //  errors in the worst case.
-    // The 256 frame maximum means we'll require 8-10 seconds of pre-buffering
-    // at 24-30 fps, which is not unreasonable.
-    let reservoir_frame_delay = clamp((max_key_frame_interval*3) >> 1, 12, 256);
+    let reservoir_frame_delay = (max_key_frame_interval*3).max(12);
```
barrbrain: The comment says "1.5x the keyframe interval" but the code says `max_key_frame_interval*3`.

dwbuiten: That'll teach me to do saturday night changes...

PR #1298, opened 2019-05-24
Add --tiles option to automatically assign tiles.

tdaede:
Use a simple algorithm that greedily splits, longer direction first.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -76,8 +76,15 @@ pub struct EncoderConfig {
   pub quantizer: usize,
   pub bitrate: i32,
   pub tune: Tune,
+  /// log2(tile columns). If tiles is also specified, this acts
+  /// as a minimum.
   pub tile_cols_log2: usize,
+  /// log2(tile rows). If tiles is also specified, this acts
+  /// as a minimum.
   pub tile_rows_log2: usize,
+  /// Number of tiles desired. The video is split automatically so that
+  /// it contains at least this many tiles.
+  pub tiles: usize,
   pub speed_settings: SpeedSettings,
   /// `None` for one-pass encode. `Some(1)` or `Some(2)` for two-pass encoding.
   pub pass: Option<u8>,
@@ -126,6 +133,7 @@ impl EncoderConfig {
       tune: Tune::default(),
       tile_cols_log2: 0,
       tile_rows_log2: 0,
+      tiles: 0,
       speed_settings: SpeedSettings::from_preset(speed),
       pass: None,
       show_psnr: false,
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -186,6 +186,14 @@ pub fn parse_cli() -> CliOptions {
         .takes_value(true)
         .default_value("0")
     )
+    .arg(
+      Arg::with_name("TILES")
+        .help("Number of tiles. Tile-cols and tile-rows are overridden\n\
+               so that the video has at least this many tiles.")
+        .long("tiles")
+        .takes_value(true)
+        .default_value("0")
+    )
     // MASTERING
     .arg(
       Arg::with_name("PIXEL_RANGE")
@@ -446,6 +454,8 @@ fn parse_config(matches: &ArgMatches<'_>) -> EncoderConfig {
 
   cfg.tile_cols_log2 = matches.value_of("TILE_COLS_LOG2").unwrap().parse().unwrap();
   cfg.tile_rows_log2 = matches.value_of("TILE_ROWS_LOG2").unwrap().parse().unwrap();
+
+  cfg.tiles = matches.value_of("TILES").unwrap().parse().unwrap();
 
   if cfg.tile_cols_log2 > 6 || cfg.tile_rows_log2 > 6 {
     panic!("Log2 of tile columns and rows may not be greater than 6");
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -500,13 +500,38 @@ impl<T: Pixel> FrameInvariants<T> {
     let w_in_b = 2 * config.width.align_power_of_two_and_shift(3); // MiCols, ((width+7)/8)<<3 >> MI_SIZE_LOG2
     let h_in_b = 2 * config.height.align_power_of_two_and_shift(3); // MiRows, ((height+7)/8)<<3 >> MI_SIZE_LOG2
 
-    let tiling = TilingInfo::new(
+    let mut tiling = TilingInfo::new(
       sequence.sb_size_log2(),
       config.width,
       config.height,
       config.tile_cols_log2,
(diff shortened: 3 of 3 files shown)


dwbuiten:
This should probably contain some information (in help, or an API comment) about how it interacts with the tile row/col log2 options.

tdaede:
Done. I'm actually wondering if I should delete the log2 options entirely in favor of this one.

lu-zero:
probably we could remove them for now and add col/rows back once the other tile mode is supported.

src/tiling/tiler.rs
```diff
@@ -73,17 +74,37 @@ impl TilingInfo {
...
       .max(Self::tile_log2(max_tile_area_sb, sb_cols * sb_rows));
 
-    let tile_cols_log2 =
-      tile_cols_log2.max(min_tile_cols_log2).min(max_tile_cols_log2);
-    let tile_width_sb = sb_cols.align_power_of_two_and_shift(tile_cols_log2);
+    let mut tile_cols_log2 =
+      tile_cols_log2_requested.max(min_tile_cols_log2).min(max_tile_cols_log2);
 
     let min_tile_rows_log2 = if min_tiles_log2 > tile_cols_log2 {
       min_tiles_log2 - tile_cols_log2
```
rom1v: `min_tile_rows_log2` is used to write the expected bitstream: https://github.com/xiph/rav1e/blob/a38059cf28bd75f7d42c9defee071a2b71d2c3c1/src/header.rs#L633

I did not check in detail, but here, it depends on `tile_cols_log2`, and `tile_cols_log2` is modified afterwards (below). So I suspect that under some conditions, `min_tile_rows_log2` may end up inconsistent with `tile_cols_log2`, leading to bitstream corruption.

rom1v: Btw, the line I quoted is incorrect:

```
diff --git a/src/header.rs b/src/header.rs
index c863a32c..502e6e54 100644
--- a/src/header.rs
+++ b/src/header.rs
@@ -630,7 +630,7 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
       self.write_bit(false);
     }
 
-    let rows_ones = ti.tile_rows_log2 - ti.min_tile_cols_log2;
+    let rows_ones = ti.tile_rows_log2 - ti.min_tile_rows_log2;
     for _ in 0..rows_ones {
       self.write_bit(true);
     }
```

tdaede: Indeed. I'll move the calculation of this to after the loop. (I think it breaks >4k wide video)

tdaede: I had also considered an alternate implementation that just creates entire new TilingInfos in a loop instead, if you think that would be cleaner.

src/api.rs
```diff
@@ -76,8 +76,15 @@ pub struct EncoderConfig {
   pub quantizer: usize,
   pub bitrate: i32,
   pub tune: Tune,
+  /// log2(tile columns). If tiles is also specified, this acts
+  /// as a minimum.
   pub tile_cols_log2: usize,
+  /// log2(tile rows). If tiles is also specified, this acts
+  /// as a minimum.
```
rom1v: IMO, either the user should provide `--tile-cols-log2` and `--tile-rows-log2, or `--tiles`, but never both.

src/api.rs
```diff
@@ -76,8 +76,15 @@ pub struct EncoderConfig {
...
   pub tile_rows_log2: usize,
+  /// Number of tiles desired. The video is split automatically so that
+  /// it contains at least this many tiles.
```
rom1v: I don't know why, but "at most" this many tiles would be less confusing. It seems weird to request 10 tiles and end up with 16.

Moreover, on small video (for example 3×3 SB), if you request 10 tiles, you'll get 9. (OK, this is a weak argument, you'll have the symmetrical behavior with big videos).

src/tiling/tiler.rs
```diff
@@ -73,17 +74,37 @@ impl TilingInfo {
...
+
+      let tiles = cols*rows;
+      if tiles >= tiles_requested { break };
+
+      if ((tile_height_sb >= tile_width_sb) && (tile_rows_log2 < max_tile_rows_log2))
+        || (tile_cols_log2 >= max_tile_cols_log2) {
+        tile_rows_log2 += 1;
+      } else {
+        tile_cols_log2 += 1;
+      }
```
rom1v: I like the idea :+1:

rom1v:
> or rounding to a power-of-two

That could be confusing, though. If requesting 8 tiles result in 6 tiles, requesting 6 tiles should yield the same result, which would bot be the case if we round it to 4.

src/api.rs
tdaede: OK. We can error out in this case, or just make one completely override the other.

tdaede:
I'm good with doing this outside TileInfo, I'll push a new patch that does that.

The logic of not having a tiles_log2 intermediate is that depending on the video resolution there are certain non-log2 combinations, e.g. 3 vertical tiles. Also log2 tiles is just super unintuitive.

My thought for rounding up is that you'd often set this for webrtc based on the number of cores, and rounding up is more likely to give you 100% CPU utilization. But rounding down is also OK.

tdaede:
I pushed a new version of the patch that makes --tiles totally override tile_cols_log2, and no longer touches TilingInfo::new() at all. Hopefull this makes it a lot less error prone.

dwbuiten:
Is there anything stopping this from being merged? This would be useful as a base for 'auto' tiles where we increase tiles for resolutons that are greater than the max tile w/h, such as 8k. Currently we just... crash.

lu-zero:
I'd land it now and refine after if needed.

dwbuiten:
If it lands, I can add 'auto'.

tdaede:
I would like for @rom1v to look at it again.

rom1v:
Looking at the code, it looks good to me. :+1:

---

As an improvement, since `max_tile_rows_log2` and `min_tile_rows_log2` depend only on `sb_size_log2`, `frame_width` and `frame_height`, they could be computed only once. Concretely:

https://github.com/xiph/rav1e/blob/09f5595bf78ef13d6d9d15d89b901ef1c09effaf/src/tiling/tiler.rs#L76-L90

could be extracted to a separate function, and we could loop over it (instead of the whole `TilingInfo`). But this is not very important (and could be done later if necessary).

src/api.rs
rzumer: It is hard to tell which would be cleaner. If returning an error is a viable option I think I would prefer that over automatically "fixing" the number of tiles.

tdaede: Returning an error is a viable option. The only downside is that it's not always obvious what #s of tiles are allowed

PR #1303, opened 2019-05-27
Do not lock stderr

lu-zero:
This fixes the problem @rom1v and @shreevari had.


lu-zero:
clippy should be happy now :)

PR #1304, opened 2019-05-28
refact: add muxer for ivf, this is a start point for other muxer like mp4, webm

xuguangxin:


src/bin/rav1e.rs
```diff
@@ -196,11 +192,11 @@ fn main() {
 
   if video_info.bit_depth == 8 {
     do_encode::<u8, y4m::Decoder<'_, Box<dyn Read>>>(
-      cfg, cli.verbose, progress, err, &mut cli.io.output, &mut source, y4m_enc
+      cfg, cli.verbose, progress, err, &mut *cli.io.output, &mut source, y4m_enc
```
lu-zero: This probably will need a manual rebase.

lu-zero: Let see if I managed to do it directly from the github UI

lu-zero: No, it failed, I manually redid properly

xuguangxin: Thanks for help.
Updated.

PR #1309, opened 2019-05-29
Adjust params of profile release to execution performance

Michael-Lfx:


lu-zero:
It seems to break coveralls (maybe it is a travis glitch?), could you please mention the speedup in the commit?

tdaede:
I think all of these are default on Release except for panic mode, but that one should only change code size?

Michael-Lfx:
Yeah, `panic = "abort"` reduces the package size.

tdaede:
I don't think the reduced binary size is justified by the reduced quality abort messages at this point in development.

PR #1310, opened 2019-05-29
Refine rate-control scale/exp values with training data

barrbrain:
These were derived by analyzing the data collected for the first iteration and applying a 3-segment piecewise-linear fit. The boundaries were aligned to integer values of pixels-per-bit, while optimizing for similarity to a cubic interpolation of the control points (log-quantizer as a function of log-bitrate.)
    
Testing against objective-1-fast, these values appear unbiased within a fifth of a standard deviation prior to trial encodes. After adjusting scale by trial encode, there is no significant bias and the standard deviation of initial frame sizes is on the order of 3%, for all frame types except P-frames.


src/rate.rs
```diff
@@ -500,19 +500,39 @@ impl RCState {
     // Start with a buffer fullness and fullness target of 50%.
     let reservoir_target = (reservoir_max + 1) >> 1;
     // Pick exponents and initial scales for quantizer selection.
-    //
+    let ibpp = npixels/bits_per_frame;
```
dwbuiten: You should update the comment below this, since its mention of a single bucket is no longer accurate.

PR #1317, opened 2019-05-30
Rate-control: double trial encodes for I and P frames

barrbrain:
Validating with objective-1-fast, after the second trial there is no statistically significant bias in the scale of initial frames for each type. Subsequent I and P frames also have no significant bias in scale prediction; B0 and B1 type frames are biased from the scale prediction on the order of 5% and 3%.

```python
from scipy import stats
stats.ttest_1samp(data, 0)
```

Initial frames:

frame type  | p-value
--|--
I  | 0.805331
P  | 0.096869
B0 | 0.329777
B1 | 0.037461

Subsequent frames:

frame type  | p-value
--|--
I  | 0.785931
P  | 0.993957
B0 | 0.000010
B1 | 0.000038

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -858,7 +858,15 @@ impl<T: Pixel> ContextInner<T> {
           let fi = self.frame_invariants.get_mut(&cur_idx).unwrap();
           fi.set_quantizers(&qps);
 
-          if self.rc_state.needs_trial_encode(fti) {
+          let trials = if !self.rc_state.needs_trial_encode(fti) {
+            0
+          } else if fti > 1 {
+            1
+          } else {
+            2
+          };
+          for _ in 0..trials {
+            let fi = self.frame_invariants.get_mut(&cur_idx).unwrap();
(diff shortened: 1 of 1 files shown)


barrbrain:
Another way of interpreting these results is that a single trial is sufficient to ensure that the first encoded frame of each type produces an unbiased scale estimate. As such, discarding the scale of the first trial and taking the scale of the first encoded frame as the initial estimate for subsequent frames should yield a similar results.

tdaede:
>As such, discarding the scale of the first trial and taking the scale of the first encoded frame as the initial estimate for subsequent frames should yield a similar results.
Isn't this effectively what master currently does?

barrbrain:
> > As such, discarding the scale of the first trial and taking the scale of the first encoded frame as the initial estimate for subsequent frames should yield a similar results.
> 
> Isn't this effectively what master currently does?

It is not what master is doing and that is a bug. #1320 fixes that.
Closing this since it is just working around that bug.

PR #1321, opened 2019-06-01
rate: Merge record_trial_encode into update_state

barrbrain:
This eliminates some code duplication and aligns with [Theora rate control](https://git.xiph.org/?p=theora.git;a=blob;f=lib/rate.c;h=bf2b1396a17ea3c7d7ac87e8b78771f7874a4a8a;hb=HEAD#l732) design.

### DIFF
--- a/src/rate.rs
+++ b/src/rate.rs
@@ -726,8 +726,13 @@ impl RCState {
   }
 
   pub fn update_state(
-    &mut self, bits: i64, fti: usize, log_target_q: i64, droppable: bool
+    &mut self, bits: i64, fti: usize, log_target_q: i64, trial: bool,
+    droppable: bool
   ) -> bool {
+    if trial {
+      assert!(self.needs_trial_encode(fti));
+      assert!(bits > 0);
+    }
     let mut dropped = false;
     // Update rate control only if rate control is active.
     if self.target_bitrate > 0 {
@@ -747,7 +752,7 @@ impl RCState {
         // If this is the first example of the given frame type we've seen, we
         //  immediately replace the default scale factor guess with the
         //  estimate we just computed using the first frame.
-        if self.nframes[fti] <= 0 {
+        if trial || self.nframes[fti] <= 0 {
           let f = &mut self.scalefilter[fti];
           let x = log_scale_q24;
           f.x[0] = x;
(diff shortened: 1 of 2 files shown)


src/rate.rs
```diff
@@ -786,54 +787,34 @@ impl RCState {
...
 
   pub fn needs_trial_encode(&self, fti: usize) -> bool {
       self.target_bitrate > 0 && self.nframes[fti] == 0
   }
-
-  pub fn record_trial_encode(
-    &mut self, bits: i64, fti: usize, log_target_q: i64
-  ) {
-    assert!(self.needs_trial_encode(fti));
-    assert!(bits > 0);
```
negge: These two assert!() statements did not make it.  Do we want to still check for them if trial is true?

src/rate.rs
```diff
@@ -726,7 +726,8 @@ impl RCState {
   }
 
   pub fn update_state(
-    &mut self, bits: i64, fti: usize, log_target_q: i64, droppable: bool
+    &mut self, bits: i64, fti: usize, log_target_q: i64, trial: bool,
+    droppable: bool
   ) -> bool {
     let mut dropped = false;
     // Update rate control only if rate control is active.
```
negge: Is it possible that self.target_bitrate <= 0 when update_state() is called with trial == true?  If so the else side of this block will not execute and your trial encode will not be recorded.  Do we want an assert!() for this to catch changed behavior?

barrbrain: So, if `trial` implies `needs_trial_encode()` then this cannot be the case. Let's keep the asserts that were dropped.

src/rate.rs
barrbrain: Yes, the above comment is related and where we should confirm the preconditions.

src/rate.rs
```diff
@@ -747,7 +752,7 @@ impl RCState {
         // If this is the first example of the given frame type we've seen, we
         //  immediately replace the default scale factor guess with the
         //  estimate we just computed using the first frame.
-        if self.nframes[fti] <= 0 {
+        if trial || self.nframes[fti] <= 0 {
```
dwbuiten: Is there ever a time when only one of these is true? i.e. can they be merged?

barrbrain: At present, trial implies `self.nframes[fti] <= 0` but my intention was to be explicit about the dual use of this block.

dwbuiten: That's what I thought. No complaint from me if you prefer it.

barrbrain:
> [02:45am KST] derf: barrbrain: Thanks for 1321. LGTM.

Tim also approves.

PR #1322, opened 2019-06-01
Enable the Clippy's "type_complexity" lint

vladimir-kazakov:
This PR enables the [type_complexity](https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity) lint, and fixes all warnings caused by it.

### DIFF
--- a/src/transform/inverse.rs
+++ b/src/transform/inverse.rs
@@ -1492,7 +1492,9 @@ fn av1_idct64(input: &[i32], output: &mut [i32], range: usize) {
   output[63] = clamp_value(stg10[0] - stg10[63], range);
 }
 
-static INV_TXFM_FNS: [[fn(&[i32], &mut [i32], usize); 5]; 4] = [
+type InvTxfmFn = fn(input: &[i32], output: &mut [i32], range: usize);
+
+static INV_TXFM_FNS: [[InvTxfmFn; 5]; 4] = [
   [av1_idct4, av1_idct8, av1_idct16, av1_idct32, av1_idct64],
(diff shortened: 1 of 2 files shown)


src/transform/inverse.rs
```diff
@@ -1492,7 +1492,9 @@ fn av1_idct64(input: &[i32], output: &mut [i32], range: usize) {
   output[63] = clamp_value(stg10[0] - stg10[63], range);
 }
 
-static INV_TXFM_FNS: [[fn(&[i32], &mut [i32], usize); 5]; 4] = [
+type InvTxfmFn = [fn(&[i32], &mut [i32], usize); 5];
```
tdaede:
```suggestion
type InvTxfmFn = fn(&[i32], &mut [i32], usize)
```

would be a better way to break it apart I think

src/transform/inverse.rs
```diff
@@ -1492,7 +1492,9 @@ fn av1_idct64(input: &[i32], output: &mut [i32], range: usize) {
   output[63] = clamp_value(stg10[0] - stg10[63], range);
 }
 
-static INV_TXFM_FNS: [[fn(&[i32], &mut [i32], usize); 5]; 4] = [
+type InvTxfmFn = fn(&[i32], &mut [i32], usize);
```
lu-zero:
```suggestion
type InvTxfmFn = fn(input: &[i32], output: &mut [i32], bits_size: usize);
```

While at it.

vladimir-kazakov: Done, but instead of `bits_size`, I used `range` to match signatures of inverse transform functions. I also think that it's a better name, because it's more clear that it's a clamping range.

PR #1323, opened 2019-06-02
[WIP] Bias RDO  based on activity

shreevari:
Biasing the distortion based on activity measured using activity calculated in #1263.
The scale for the distortion bias will be computed empirically to produce encodes of the same file size as without activity bias.

The plan is to use Newton-Raphson Method to compute the scale which minimizes MSE of file size as compared to encodes without any activity bias.
```
s' = s - (MSE/f'(s))
where f : s -> file_size
```

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1235,7 +1235,7 @@ pub fn encode_tx_block<T: Pixel>(
     w.add_bits_frac(estimated_rate as u32);
   }
   let bias =
-    compute_distortion_bias(fi, ts.to_frame_block_offset(tile_bo), bsize);
+    compute_distortion_bias(fi, ts.to_frame_block_offset(tile_bo), bsize, p);
   (has_coeff, RawDistortion::new(tx_dist) * bias * fi.dist_scale[p])
 }
 
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -10,6 +10,7 @@
 
 #![allow(non_camel_case_types)]
 
+use crate::api::FrameType;
 use crate::api::*;
 use crate::cdef::*;
 use crate::context::*;
@@ -355,6 +356,7 @@ fn compute_distortion<T: Pixel>(
             fi,
             input_region.subregion(bias_area).frame_block_offset(),
             bsize,
+            0,
           )
         },
       )
@@ -369,6 +371,7 @@ fn compute_distortion<T: Pixel>(
           fi,
           input_region.subregion(bias_area).frame_block_offset(),
           bsize,
+          0,
         )
       },
     ),
@@ -401,6 +404,7 @@ fn compute_distortion<T: Pixel>(
               fi,
               input_region.subregion(bias_area).frame_block_offset(),
               bsize,
+              p,
             )
           },
         ) * fi.dist_scale[p];
@@ -431,11 +435,14 @@ fn compute_tx_distortion<T: Pixel>(
           fi,
           input_region.subregion(bias_area).frame_block_offset(),
           bsize,
+          0,
         )
       },
     ) * fi.dist_scale[0]
   } else {
-    tx_dist
+    let bias =
+      compute_distortion_bias(fi, ts.to_frame_block_offset(tile_bo), bsize, 0);
+    ScaledDistortion((tx_dist.0 as f64 * bias) as u64)
   };
 
   if !luma_only && skip {
@@ -465,6 +472,7 @@ fn compute_tx_distortion<T: Pixel>(
               fi,
               input_region.subregion(bias_area).frame_block_offset(),
               bsize,
+              p,
             )
           },
         ) * fi.dist_scale[p];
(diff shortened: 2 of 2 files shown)


tdaede:
Have you tried running this through AWCY?

shreevari:
> Have you tried running this through AWCY?
Not yet. 
> ```
> s' = s - (MSE/f'(s))
> where f : s -> file_size
> ```
I'm trying to derive the function in order to perform Newton's method on Hydra.
I plan to run encodes to sample values of this function, plot a graph and try to approximate an equation for f (assuming that f does not change with quantizer value).

shreevari:
The maximum variance of a block of n-bit pixels is 1 << (n-1)*2

dwbuiten:
In case you missed it on IRC:

```
[08:51] < barrbrain_> Taking a look at s_p's WIP 1323, if using many tiles and --tune Psnr it may behave as intended.
[08:53] < barrbrain_> Otherwise, tile granularity is too coarse and --tune doubles up on biasing for activity.
[08:53] < barrbrain_> *--tune Pschovisual
```

This may be related to what we saw with the horse picture on Friday.

src/rdo.rs
```diff
@@ -352,9 +352,15 @@ fn compute_tx_distortion<T: Pixel>(
   distortion
 }
 
-fn compute_rd_cost<T: Pixel>(fi: &FrameInvariants<T>, rate: u32, distortion: u64) -> f64 {
+fn bias_distortion(actual_distortion: f64, activity: u16, mean_activity: u16) -> f64 {
+  let s_1: f64 = 1_f64;
+  let s_0 = 0_f64;
+  actual_distortion * (s_1 * (mean_activity / activity) as f64 + s_0)
```
barrbrain: Need a floor on `activity` to avoid divide-by-zero. Due to integer division, if `s_0 == 0` and `activity > mean_activity` then the distortion will be scaled by zero, which means rate is minimized at all costs.

shreevari: Yes, I observed that during the encodes and changed it locally. I'll push the changes soon.
```
actual_distortion * (s_1 * (mean_activity as f64 / activity as f64) + s_0)
```
This is what I'd done. I'll handle divide-by-zero error in the next revision.

src/encoder.rs
```diff
@@ -350,6 +356,89 @@ impl<T: Pixel> FrameState<T> {
...
+        
+        let tx_type = TxType::DCT_DCT;
+        let tx_size = TxSize::TX_8X8;
+        let mut tx_deviation = [0_i32; 64];
+        forward_transform(&deviation, &mut tx_deviation, tx_size.width(), tx_size, tx_type, 8);
+
+        // Applying CSF to the deviation on tx_domain
+        tx_deviation.iter_mut().zip(csf.iter().flatten()).for_each(|(d, &csf)| *d = (*d as f32 * csf).floor() as i32);
+
+        let mut cs_dev = Plane::<u16>::new(8, 8, 0, 0, 0, 0);
```
barrbrain: Note that the pixel values of `cs_dev` are initialized to 128.

src/encoder.rs
```diff
@@ -350,6 +356,89 @@ impl<T: Pixel> FrameState<T> {
...
+        let mut cs_deviation = PlaneRegionMut::new(
+          &mut cs_dev,
+          Rect{x: 0, y: 0, width: 8, height: 8});
+        inverse_transform_add(&tx_deviation, &mut cs_deviation, tx_size, tx_type, 8);
+
+        let act: f32 = cs_deviation.rows_iter().flatten().map(|dev| *dev as f32 * *dev as f32).sum();
```
barrbrain: The mean of `cs_dev`, 128, needs to be subtracted before squaring to calculate the activity.

src/encoder.rs
```diff
@@ -350,6 +356,89 @@ impl<T: Pixel> FrameState<T> {
...
+          x: (x << xdec) as isize,
+          y: (y << ydec) as isize,
+          width: 8,
+          height: 8,
+        };
+        
+        let block = luma.subregion(block_rect);
+        
+        let mut mean: i32 = block.rows_iter().flatten().map(|&val| {let int: i32 = CastFromPrimitive::cast_from(val); int}).sum();
+        mean >>= xdec + ydec;
```
barrbrain: We may skip subtracting the mean by directly zeroing the DC value after transforming to the frequency domain.

src/encoder.rs
barrbrain: We may avoid rapidly churning the allocator here by hoisting this `new()` outside the loops. We then need to explicitly fill in each iteration. The following worked for me but is a quick-hack:
```rust
cs_dev.data.copy_from_slice(&[128u16; 128]);
```

tdaede: Given it's a fixed size new, you should be able to replace it with AlignedAlloc instead.

src/encoder.rs
shreevari: How about I change the DC coeff multiplication factor of the CSF to 0?

src/encoder.rs
shreevari: Thanks for the hack. It looks cleaner than how I'd usually implement.

shreevari: > ```rust
> cs_dev.data.copy_from_slice(&[128u16; 128]);
> ```

Any reason for the size of the slice being 128? Won't 64 suffice? (For an 8x8 block)

src/encoder.rs
shreevari: I'll update this.

src/encoder.rs
barrbrain: It's due to the stride alignment of `PlaneData`, 32 bytes = 16 u16 pixels = 8 cols * 2; an internal detail that leaks here. We may avoid this altogether by accumulating in the frequency domain but then we need to be mindful of scale.

shreevari: > Given it's a fixed size new, you should be able to replace it with AlignedAlloc instead.

AlignedArray instead of Plane::new() for cs_dev?

shreevari:
The scale is tuned using Newton-Raphson's algorithm. Data gathered for tuning can be found [here](https://gist.github.com/shreevari/3aa043a35f87099263b4160769c465e8).

tdaede:
Do you have awcy results of this rebased on the new activity masking patch?

shreevari:
> Do you have awcy results of this rebased on the new activity masking patch?

The scale needs tuning. I am yet to figure out how to change the scale with respect to the base quantizer. I've to discuss this with @tterribe.

ycho:
Please check my relevant comment at https://github.com/xiph/rav1e/pull/1485#issuecomment-540082355
I probably should have stated this earlier, I think the activity based RDO would not help yielding PSNR gains by its nature but is more expected to have gains in the metrics that are similar to the methodology to obtain activity.

src/rdo.rs
```diff
@@ -475,16 +480,33 @@ fn compute_mean_importance<T: Pixel>(
...
   const FACTOR: f32 = 3.;
   const ADDEND: f64 = 0.65;
 
-  let bias = (mean_importance / FACTOR) as f64 + ADDEND;
+  let mut bias = (mean_importance / FACTOR) as f64 + ADDEND;
+
+  if pli == 0 {
+    const ACTIVITY_SCALE: f64 = 0.5;
+    let block_activity = compute_activity(fi, frame_bo, bsize);
+    bias /= ACTIVITY_SCALE * block_activity.unwrap_or(1.0_f64 / ACTIVITY_SCALE);
```
ycho: So, do you intend to apply activity scaler on top of block importance for temporal rdo?
Eventually that is the scenario how rav1e encoder will run, but in the beginning can you consider separate activity masking experiment from temporal rdo? I.e. disable temporal rdo, and your activity scalar is directly applied to the distortion. Otherwise, any possible inteferences btw those two scalars can mislead your experiment?
Also, is this PR still effective? I know that the 1st commit, adding activity masking, is now in separate PR.

Issue #1262, opened 2019-05-07
Very Bad Edge Artifacting on Sintel


dwbuiten:
Output file: http://chromashift.org/sintel.720.mp4 - look at ~11:52.
Encoding parameters: Encoded via crav1e API, with: `speed=2:tile_cols_log2=1:tile_rows_log2=1:quantizer=255:bitrate=2252800:key_frame_interval=72:min_key_frame_interval=72`

I can provide the exact lossless input sequence (with our zimg-based resizing, etc.) if that is useful. Just didn't yet because Big Files.

tdaede:
Looks like it's happening through the whole movie, should be able to reproduce with a much shorter sequence like Shields.

Kinda looks like the skip decision is broken.

tdaede:
So playing with this more I think it's actually a rate control issue. I didn't dump the quantizers in the mp4 to be sure, but I was unable to reproduce this without rate control enabled. The right edge looks bad because that's new content coming in at the old quantizer - the rest of the frame is just a pan so it's going to look good as it's just motion compensated old image data.

tdaede:
... which means, now that we have landed some rate control fixes, could you try encoding it again and see if it is still present?

dwbuiten:
@tdaede I will re-run it when I push the clmap fix, at least. Ideally I'd like the frame subtype fix in, but I do not know its status.

dwbuiten:
@tdaede will re-run a test, but i dont see how it can be rate control, since only edge blocks are affected; the rest of the frame looks fine...

dwbuiten:
@tdaede It did not go away. :(

dwbuiten:
http://chromashift.org/sintel.new.mp4

dwbuiten:
I can reproduce with a shorter (340 frame, 640x278) clip. Those specific artifacts are also visible on higher quantizer encodes (e.g. 190). The difference is the whole frame looks bad rather than just part of the frame, as it does with bitrate mode.

Relevant scroll back:
```
[15:01] < barrbrain> oh, it’s quite simple… those blocks are skip with motion vectors extending beyond the frame edge
[15:01] < Daemon404> o.o
[15:02] < Daemon404> could ratecontol really cause that?
[15:02] < barrbrain> the vertical lines are edge extension
[15:02] < barrbrain> rate control and/or RDO
[15:02] < Daemon404> i see
[15:02] < barrbrain> once RC is sane, probably just RDO
[15:02] < Daemon404> i mean
[...]

PR #1324, opened 2019-06-02
api: Expose the reservoir frame delay as a settable option

dwbuiten:
There's no reason to explicitly bind this to the max keyframe interval. It's useful to, for example, set it larger than 1.5x the keyframe interval for short intervals (1-3s) to prevent bitrate starvation after larger than expected (complex) keyframes.

### Notes

* Right now the default is 1.5x the max keyframe interval. Would it be useful to set it based on some threshold of big/small keyints, or would that be too confusing?
* Is there a better option name I could use?
* This should address most of #1262 by way of letting users set this value when using small GOPs. The weird edge artifacts at high quantziers remain.
* Since it will now be user settable, we need to definitely address `log_scale` overflow ([link](https://github.com/xiph/rav1e/blob/52f0bd7fc460e3f3a137d5b48ddfc075bbc78649/src/rate.rs#L398-L399)) when implementing two pass. 
* Should we enforce a minimum of 12 still when it is user set? (**EDIT: I did this and re-pushed.**)

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -74,6 +74,8 @@ pub struct EncoderConfig {
   pub min_key_frame_interval: u64,
   /// The *maximum* interval between two keyframes
   pub max_key_frame_interval: u64,
+  /// The number of frames over which to distribute the reservoir usage.
+  pub reservoir_frame_delay: Option<i32>,
   pub low_latency: bool,
   pub quantizer: usize,
   pub bitrate: i32,
@@ -122,6 +124,7 @@ impl EncoderConfig {
       time_base: Rational { num: 1, den: 30 },
       min_key_frame_interval: 12,
       max_key_frame_interval: 240,
+      reservoir_frame_delay: None,
       low_latency: false,
       quantizer: 100,
       bitrate: 0,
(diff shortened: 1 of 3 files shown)


dwbuiten:
I am thinking it should probaly hae a value sanity check (or the 12 minimum val applied) to prevent crazy things?

tdaede:
This looks okay but I'd also rather have the default match your use case.

For the default, would just a fixed number of seconds make more sense?

Also, for the API would it make sense to expose it in seconds rather than frames?

dwbuiten:
I think it would be odd to expose it as seconds, especially for low framerate content or, in the future, VFR content.

As for the default, what do you think would work? Default of `*3` for longer GOP content seems excessive. Threshold?

tdaede:
Maybe just a fixed number of frames for the default? Or min(fixed # of frames, 1.5*GOP) ?

dwbuiten:
Upated with a default of 1.5*GOP with a value at least 240 being enforced, and a minimum of 12 being enforced on user set values. Updated the CLI usage too.

PR #1325, opened 2019-06-02
rate: Do not enable frame drops by default

dwbuiten:
It wasn't supposed to be enabled, and we don't handle it correctly.


src/rate.rs
```diff
@@ -538,7 +538,7 @@ impl RCState {
       reservoir_frame_delay,
       maybe_ac_qi_max,
       // By default, enforce hard buffer constraints.
```
lu-zero: What about this comment?

PR #1326, opened 2019-06-03
Enable signal-hook in the supported platforms

lu-zero:
I might need @EwoutH to test it on windows as well.


lu-zero:
Now everything works, I tested locally by cross-building the whole lot.

PR #1328, opened 2019-06-03
Support cross compiling

lu-zero:
The build target in `build.rs` is available as environment variable at
runtime, not compile time.

While at it refactor the code to support the ARM assembly.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,8 +1,5 @@
 // build.rs
 
-#[cfg(all(target_arch = "x86_64", feature = "nasm"))]
-extern crate nasm_rs;
-
 #[allow(unused_imports)]
 use std::env;
 use std::fs;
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -21,45 +18,57 @@ fn rerun_dir<P: AsRef<Path>>(dir: P) {
...
-                "src/x86/sad_avx.asm"
-            ],
-            &[&config_include_arg, "-Isrc/"]
-        );
-        println!("cargo:rustc-link-lib=static=rav1easm");
-        rerun_dir("src/x86");
-        rerun_dir("src/ext/x86");
+    let arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+    let os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+    // let env = env::var("CARGO_CFG_TARGET_ENV").unwrap();
```
barrbrain: I think it's a fair compromise to use `#[warn(unused_variables)]` for future reference.

lu-zero: actually I'd consider allow so it does not get too verbose

barrbrain: I think we are trying to say the same thing but I was too terse. Avoiding unused variables would make this less neat.

PR #1329, opened 2019-06-03
Prohibit CFL if luma tx is partitioned

ycho:


src/rdo.rs
```diff
@@ -781,7 +781,7 @@ pub fn rdo_mode_decision<T: Pixel>(
     let mut probs = intra_mode_set.iter().map(|&a| (a, probs_all[a as usize])).collect::<Vec<_>>();
     probs.sort_by_key(|a| !a.1);
 
-    let mut modes = ArrayVec::<[_;7]>::new();
```
lu-zero: Why this?

ycho: Ooops, this patch mistakenly included WIP code from the other branch of mine.
Btw, later when we are able to enable the currently disabled directional intra predictions, it will need the change at line 784. I talked with TD-Linux regarding this. Please try w/o change, it will lack array! :)

ycho:
Has not actually causing any crashes. So, it is possible CFL can still be available when luma partition has multiple tx blocks. CFL just needs decoded luma samples anyway, right?

tdaede:
CfL is allowed regardless of the TX split. It should work in rav1e.

ycho:
> CfL is allowed regardless of the TX split. It should work in rav1e.

Great- I wanted this confirmed. Then closing this.

PR #1332, opened 2019-06-04
Use `dyn` for a dynamic trait

linkmauve:
`Muxer` is a trait, and passing dynamic traits without the `dyn` keyword is now deprecated.


linkmauve:
Maybe we should add this deny crate-wide?

lu-zero:
Could be a good idea let see what the team thinks :)

shssoichiro:
It will at the least be `warn` by default in an upcoming rustc (slated for 1.37): https://github.com/rust-lang/rust/pull/61203. But we may want to change it before then.

Issue #1202, opened 2019-04-10
Publish to crates.io


lu-zero:
Prerequisites:
- [x] Have a nasm-rs release we can use
- [x] Publish a `0.1.0` for `ivf` and `rav1e`

rzumer:
This seems like a duplicate of #491. If that's the case, I recommend editing the opening post in that issue with the checklist appended, and closing this one.

lu-zero:
It is not :)

rzumer:
Then can you detail the difference, besides the fact that this issue is specifying a particular repository?

PR #1336, opened 2019-06-04
ivf: Use ivf from crates.io

vibhoothiiaanand:
Removes the local version of ivf and use the published version of ivf
Ref #1202


lu-zero:
It is wrong, we want to keep the ivf sources alive.

PR #1342, opened 2019-06-05
Unbreak the row iterators

lu-zero:
The remaning field was not updated


rom1v:
:scream: good catch!

I wondered why it had not been detected before, but the reason is probably that we often call `take(Self::H)` afterwards.

PR #1344, opened 2019-06-08
Add version to the api and use it in the cli

lu-zero:


rzumer:
How would the output look for major/minor/patch versions for a weekly release? Assuming we compiled a weekly release dated today, would it result in a major version string of 20190610? Maybe it is not a concern, but I am wondering how it would look.

lu-zero:
We are using semver so a snapshot and a version are unrelated

current version:
```
rav1e 0.1.0 (20190430-62-gdcf74e5)
```

Release:
```
rav1e 0.1.0 (v0.1.0)
```

Some commit after:
```
rav1e 0.1.0 (v0.1.0-1-g743d464)
```

Snapshot after:
```
rav1e 0.1.1 (20190610)
```

rzumer:
Oh I see. That looks good to me.

src/lib.rs
```diff
@@ -61,6 +61,64 @@ pub use crate::util::{CastFromPrimitive, Pixel};
...
+  /// Patch version component
+  ///
+  /// It is increased every time a release provides only backwards-compatible bugfixes.
+  pub fn patch() -> usize {
+    env!("CARGO_PKG_VERSION_PATCH").parse().unwrap()
+  }
+
+  /// Version information as presented in CARGO_PKG_VERSION
+  ///
+  /// Can be parsed by [semver](https://crates.io/crates/semver).
```
dwbuiten: Markdown in headers is gross.

src/lib.rs
```diff
@@ -61,6 +61,64 @@ pub use crate::util::{CastFromPrimitive, Pixel};
...
+  pub fn short() -> String {
+    env!("CARGO_PKG_VERSION").to_string()
+  }
+
+  /// Version information as presented in CARGO_PKG_VERSION followed by the
+  /// short commit hash if present.
```
dwbuiten: This documentation isn't super useful to API users who just want to know what sort of string it outputs; it relies on specific Rust env knowledge. Same for the others.

dwbuiten:
Examples in the API docs would be useful.

src/lib.rs
rzumer: If I recall correctly it is encouraged to use Markdown in comments because `rustdoc` can parse it. Personally, I agree that it is ugly, but I can see the case for it.

src/lib.rs
rzumer: I assume that this corresponds to the `version` line in the `[package]` category in Cargo.toml. I think it may be clearer in those terms, as it is easier to locate for those unfamiliar with Cargo.

src/lib.rs
lu-zero: Once it is rendered is lovely thought :) (most people use either `cargo doc --open` or docs.rs)

lu-zero:
I'll do now.

lu-zero:
I added a doctest/example while at it.

PR #1346, opened 2019-06-11
Public API / Doc cleanup

lu-zero:
Use `cargo doc --open` to see the difference.

Would probably make sense to group all the configuration structs in some way e.g. all the color definition structs in a `color` module.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -11,14 +11,15 @@ mod predict;
 mod transform;
 mod me;
 
-use rav1e::*;
-use rav1e::cdef::cdef_filter_frame;
-use rav1e::context::*;
-use rav1e::partition::*;
-use rav1e::predict::*;
-use rav1e::rdo::rdo_cfl_alpha;
-use rav1e::rdo::RDOType;
-use rav1e::ec::*;
+use rav1e::bench::api::*;
+use rav1e::bench::encoder::*;
+use rav1e::bench::cdef::*;
+use rav1e::bench::context::*;
+use rav1e::bench::ec::*;
+use rav1e::bench::partition::*;
+use rav1e::bench::predict::*;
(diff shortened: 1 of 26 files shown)


benches/bench.rs
```diff
@@ -12,18 +12,19 @@ mod transform;
...
+use rav1e::bench::partition::*;
+use rav1e::bench::predict::*;
+use rav1e::bench::rdo::*;
+
 use crate::transform::transform;
 
 use criterion::*;
 use std::time::Duration;
 
+
```
kodabb: nit: extra linebreak

src/bin/common.rs
```diff
@@ -10,8 +10,8 @@
 use crate::muxer::{create_muxer, Muxer};
 use crate::{ColorPrimaries, MatrixCoefficients, TransferCharacteristics};
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand, Shell};
-use rav1e::partition::BlockSize;
 use rav1e::*;
+use scan_fmt::scan_fmt;
```
kodabb: this one is used only in the cli, is this the right place?

lu-zero: yes, we could rename the filename so it is less misleading later :)

benches/bench.rs
lu-zero: I'll remove it now. Thank you for spotting.

lu-zero:
> Massive improvement, thanks.
> 
> Context and ContextInner still both get exposed, can that be fixed?

Sure, I didn't touch that file to not collide with Tim's patch.

lu-zero:
Now it looks like this:

Screenshot_2019-06-13 rav1e - Rust

tdaede:
Perfect.

PR #1354, opened 2019-06-18
Fix deprecated function call in nasm-rs

rzumer:
Should fix the warning when building.


lu-zero:
I did that as part of https://github.com/xiph/rav1e/pull/1353

rzumer:
OK. There was no reference to the issue and the PR name does not seem related, so I did not notice.

PR #1355, opened 2019-06-18
Set Y4M pixel range to limited from CLI

### LINKED ISSUE #1341: Interpret y4m streams as limited range in the cli
The original spec specifies limited range. I think there is a ffmpeg metadata extension that allows full range content that we could also support.

rzumer:
Attempt to fix #1341. I don't think the `y4m` crate is parsing range in the header, so simply use the default value of limited.

### DIFF
--- a/src/bin/rav1e.rs
+++ b/src/bin/rav1e.rs
@@ -162,6 +162,13 @@ fn main() {
   cli.enc.bit_depth = video_info.bit_depth;
   cli.enc.chroma_sampling = video_info.chroma_sampling;
   cli.enc.chroma_sample_position = video_info.chroma_sample_position;
+
+  // If no pixel range is specified via CLI, assume limited,
+  // as it is the default for the Y4M format.
+  if cli.enc.pixel_range == PixelRange::Unspecified {
+    cli.enc.pixel_range = PixelRange::Limited;
+  }
+
   cli.enc.time_base = video_info.time_base;
   let cfg = Config {
     enc: cli.enc,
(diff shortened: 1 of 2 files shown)


lu-zero:
Do we have an override for it?

rzumer:
I don't think so. I will add an option for it in a bit.

rzumer:
There is in fact already a `--pixel_range` option. My approach was wrong though, I will force-push a fix that will correctly allow users to override the default.

src/bin/rav1e.rs
```diff
@@ -162,6 +162,15 @@ fn main() {
   cli.enc.bit_depth = video_info.bit_depth;
   cli.enc.chroma_sampling = video_info.chroma_sampling;
   cli.enc.chroma_sample_position = video_info.chroma_sample_position;
+
+  // If no pixel range is specified via CLI, assume limited,
```
tdaede: Do we want to be able to explicitly set the "Unspecified" value from the cli? (seems useless to me but we do lose that ability now)

rzumer: What is the behaviour in the encoder if it is unspecified? I also thought it was not worth adding an extra parameter to `CliOptions`, but we can do that if preferred.

tdaede: I think unspecified is actually a valid field option in the AV1 header.

tdaede: (then it's up to the player to decide. tbh this behavior seems useless to me)

rzumer: > I think unspecified is actually a valid field option in the AV1 header.

Indeed.

While it doesn't seem very useful, if someone did want to specify the range as "unspecified", it would be surprising for it to be changed to "limited". So for the sake of avoiding oddities for end users I have pushed a commit to honor a user's choice to leave the range as unspecified. Either way is fine by me though, so let me know if I should roll it back.

tdaede: Seems best this way.

PR #1356, opened 2019-06-18
Enforce minimum tiling requirements when encoding large frames

rzumer:
Note: right now this is using the fixed temporal ratio of 1.1 as in the specification for rate estimation. In practice we have to make sure that a single tile cannot have a rate higher than 4K60 with a `TemporalParallelDen / TemporalParallelNum` ratio of 1.1.

This sets a minimum number of tiles regardless of what is requested by the user. So if we need at least 2 tiles to encode a given clip and the user requests 1, 2 tiles will be coded (otherwise the stream will not be compliant). If the requested number of tiles is higher, there is no change in behavior.

Tested by `println!()`ing `min_tiles`. There are other issues when trying to encode very large frames.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -504,17 +504,33 @@ impl<T: Pixel> FrameInvariants<T> {
     let w_in_b = 2 * config.width.align_power_of_two_and_shift(3); // MiCols, ((width+7)/8)<<3 >> MI_SIZE_LOG2
     let h_in_b = 2 * config.height.align_power_of_two_and_shift(3); // MiRows, ((height+7)/8)<<3 >> MI_SIZE_LOG2
 
+    let frame_rate = (config.time_base.den / config.time_base.num) as usize;
+    let min_tile_cols = (config.width - 1) / 4096;
+    // The minimum number of tiles is determined based on the following requirements:
+    // - A tile cannot be wider than 4096 pixels
+    // - A tile cannot be larger than 4096x2304 pixels
+    // - The tile size * the frame rate * a temporal ratio cannot exceed a given fixed rate
+    //   corresponding to 4K60 resolution with a 1.1 ratio.
+    let min_tiles = 1 + min_tile_cols.max(
+      (config.width * config.height) / (4096 * 2304)).max(
+      (config.width * config.height * frame_rate) / (4096_f64 * 2176_f64 * 60_f64 * 1.1) as usize
+    );
+
     let mut tiling = TilingInfo::new(
       sequence.sb_size_log2(),
       config.width,
       config.height,
-      config.tile_cols_log2,
+      config.tile_cols_log2.min(min_tile_cols),
       config.tile_rows_log2
     );
 
-    if config.tiles > 0 {
-      let mut tile_rows_log2 = 0;
-      let mut tile_cols_log2 = 0;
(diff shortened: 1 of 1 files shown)


tdaede:
ohh I think TemporalParallel is related to temporal scalability which we don't use.

>If scalability_mode_idc is not present or equal to a reserved value, then TemporalParallelNum and TemporalParallelDen are defined to be equal to 1.

rzumer:
@tdaede That seems right. But I think it is also dynamic, so not something that can reliably be controlled when setting the tiles at the configuration stage. Do you want to remove this ratio (effectively setting it to 1)?

src/encoder.rs
```diff
@@ -504,17 +504,29 @@ impl<T: Pixel> FrameInvariants<T> {
...
       config.width,
       config.height,
-      config.tile_cols_log2,
+      config.tile_cols_log2.min(min_tile_cols),
       config.tile_rows_log2
     );
 
-    if config.tiles > 0 {
+    if config.tiles > 0 || tiling.rows * tiling.cols < min_tiles {
       let mut tile_rows_log2 = 0;
```
tdaede: It might make sense to adjust the behavior slightly here. This was written normally so that if you specify --tiles, tile_rows_log2 is ignored. However, if we failed to meet max tile size, we maybe don't want to totally ignore those, but treat them as a minimum. So maybe set tile_rows_log2 to config.tile_rows_log2 here, but only if (config.tiles == 0 && tiling.rows * tiling.cols < min_tiles)

tdaede: (Another option is just to fail here if the user uses tile_rows_log2/cols_log2 explicitly rather than --tiles)

rzumer: Agreed; should be fixed now.

tdaede:
> @tdaede That seems right. But I think it is also dynamic, so not something that can reliably be controlled when setting the tiles at the configuration stage.
> 
> Since it is not supported, I guess I will set the ratio to 1 (i.e. remove the scaling) and write a note there.

When/if we ever support this, we can move tiling to FrameInvariants::new_inter_frame etc. We're allowed to change it on a per frame basis.

src/encoder.rs
```diff
@@ -504,17 +504,33 @@ impl<T: Pixel> FrameInvariants<T> {
...
+    let frame_rate = (config.time_base.den / config.time_base.num) as usize;
+    let min_tile_cols = (config.width - 1) / 4096;
+    // The minimum number of tiles is determined based on the following requirements:
+    // - A tile cannot be wider than 4096 pixels
+    // - A tile cannot be larger than 4096x2304 pixels
+    // - The tile size * the frame rate * a temporal ratio cannot exceed a given fixed rate
+    //   corresponding to 4K60 resolution with a 1.1 ratio.
+    let min_tiles = 1 + min_tile_cols.max(
+      (config.width * config.height) / (4096 * 2304)).max(
+      (config.width * config.height * frame_rate) / (4096_f64 * 2176_f64 * 60_f64 * 1.1) as usize
```
tdaede: 1.0?

rzumer: The maximum rate given here is based on a temporal ratio of 1.1, so that half of the equation does need to be scaled. The left-hand side is unscaled.

tdaede: Oh OK.

PR #1357, opened 2019-06-18
Fix partition context granularity.

tdaede:
I think we can actually do with a granularity of 8 with a
bit more code rewriting.


tdaede:
Closed in favor of #1358

PR #1358, opened 2019-06-18
Shrink size of partition contexts by half.

tdaede:
We were only using every other value of these arrays.


rzumer:
Some comments would be nice to document why this is done, since it is not necessarily intuitive (at least to me).

Besides that, LGTM, assuming everything works, tests pass etc.

tdaede:
added comments

PR #1360, opened 2019-06-19
fix one potential crash issue when directional intra modes enabled

rainliu:
when directional intra modes are enabled, running "target\release\rav1e.exe nyan.y4m -o enc_file.ivf --limit 1 -s 3 -r rec_file.y4m" will cause crash due to hard coded ArrayVec::<_;7>.

This commit fixed this potential crash issue.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -782,7 +782,7 @@ pub fn rdo_mode_decision<T: Pixel>(
     let mut probs = intra_mode_set.iter().map(|&a| (a, probs_all[a as usize])).collect::<Vec<_>>();
     probs.sort_by_key(|a| !a.1);
 
-    let mut modes = ArrayVec::<[_;7]>::new();
+    let mut modes = ArrayVec::<[_;INTRA_MODES]>::new();
     probs
       .iter()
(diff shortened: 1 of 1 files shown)


lu-zero:
I would reword the commit message a little, but sounds a good idea to set the array to the upper bound to avoid surprises later. @tdaede do you agree?

tdaede:
This change is correct, but can we use .len() (is it a const fn?)

rainliu:
@tdaede, .len() is not const fn. 

I am trying to simplify this change by using PredictionMode::NEARESTMV as size of ArrayVec:

```
-let mut modes = ArrayVec::<[_;RAV1E_INTRA_MODES_LEN]>::new();
+let mut modes = ArrayVec::<[_;PredictionMode::NEARESTMV as usize]>::new();
```

tdaede:
I think the way the patch is currently written is more robust than doing that. Another constant you could use is INTRA_MODES - it's higher than the number of intra modes we currently test, but wasting a bit of space here isn't an issue.

yuryanliu:
just updated commit by using INTRA_MODES

PR #1362, opened 2019-06-19
Fix README command for running cargo bench.

tdaede:


rzumer:
I guess that will not work on Windows. Why is this environment variable needed? Can't it be a default?

tdaede:
It is for a pretty gross reason - the benches use non-pub functions, but can only "see" pub ones. So this exports everything pub.

We could make it a cargo --feature and that would make it work on windows. It's still hella gross though. I think there is a way to move the benches directly into the files and avoid this whole deal, still reading.

tdaede:
https://github.com/rust-lang/rust/issues/50297

tdaede:
I made it marginally less gross.

lu-zero:
I made it _this_ ugly intentionally so we do not forget and `--all-features` won't be a huge surprise.

The avformat muxer made it already sufficiently surprising though...

tdaede:
@lu-zero do you have a less bad solution?

lu-zero:
We can go your way since the patch is ready and just document that `--all-features` should not be used.

tdaede:
documented

PR #1364, opened 2019-06-19
Fix for deblocking desync on non-mult-of-8 crop frame sizes

xiphmont:
This patch fixes rav1e to the agreed deblocking crop behavior in the
spec.  It should correct several desyncs that were showing up in the
following (CDEF and LRF) filters that pull pixels out of the
non-displayed buffer area.

### DIFF
--- a/src/deblock.rs
+++ b/src/deblock.rs
@@ -1237,11 +1237,11 @@ fn sse_h_edge<T: Pixel>(
 
 // Deblocks all edges, vertical and horizontal, in a single plane
 pub fn deblock_plane<T: Pixel>(
-  deblock: &DeblockState, p: &mut Plane<T>, pli: usize, blocks: &FrameBlocks,
-  bd: usize
+  fi: &FrameInvariants<T>, deblock: &DeblockState, p: &mut Plane<T>, pli: usize, blocks: &FrameBlocks
 ) {
   let xdec = p.cfg.xdec;
   let ydec = p.cfg.ydec;
+  let bd = fi.sequence.bit_depth;
 
   match pli {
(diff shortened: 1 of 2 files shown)


xiphmont:
Clippy can get bent.

xiphmont:
I'd hope so, that was my primary test case.

src/deblock.rs
```diff
@@ -1350,25 +1359,30 @@ pub fn deblock_plane<T: Pixel>(
...
-  bd: usize
+  fi: &FrameInvariants<T>, rec: &Plane<T>, src: &Plane<T>, v_sse: &mut [i64; MAX_LOOP_FILTER + 2],
+  h_sse: &mut [i64; MAX_LOOP_FILTER + 2], pli: usize, blocks: &FrameBlocks
 ) {
   let xdec = rec.cfg.xdec;
   let ydec = rec.cfg.ydec;
 
+  // Deblocking happens in 4x4 (luma) units; luma x,y are clipped to
+  // the *crop frame* by 4x4 block.
+  let cols = cmp::min(blocks.cols, (fi.width + MI_SIZE - 1) >> MI_SIZE_LOG2);
```
ycho: So, we cannot just use : cols = (fi.width + MI_SIZE - 1) >> MI_SIZE_LOG2 ? Since it seems always <=  blocks.cols.
And same for rows?

xiphmont: *probably*  
Something in the back of my head said 'be paranoid'.  You're probably right.

xiphmont: Addressed by PR 1366

PR #1365, opened 2019-06-20
2-pass Rate Control from Tim

dwbuiten:
I did a preliminary review while this was on pastebin.

More comments welcome.

crav1e will need to be updated.

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -34,6 +34,8 @@ pub struct CliOptions {
   pub skip: usize,
   pub verbose: bool,
   pub threads: usize,
+  pub pass1file_name: Option<String>,
+  pub pass2file_name: Option<String>
 }
 
 pub fn parse_cli() -> CliOptions {
@@ -82,12 +84,16 @@ pub fn parse_cli() -> CliOptions {
     )
     // ENCODING SETTINGS
     .arg(
-      Arg::with_name("PASS")
-        .help("Specify first-pass or second-pass to run as a two-pass encode; If not provided, will run a one-pass encode")
-        .short("p")
-        .long("pass")
+      Arg::with_name("FIRST_PASS")
+        .help("Perform the first pass of a two-pass encode, saving the pass data to the specified file for future passes")
+        .long("first-pass")
         .takes_value(true)
-        .possible_values(&["1", "2"])
+    )
+    .arg(
+      Arg::with_name("SECOND_PASS")
+        .help("Perform the second pass of a two-pass encode, reading the pass data saved from a previous pass from the specified file")
+        .long("second-pass")
+        .takes_value(true)
     )
     .arg(
       Arg::with_name("LIMIT")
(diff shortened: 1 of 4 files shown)


dwbuiten:
Some of your suggestions are bloody terrible, Clippy.

gnafuthegreat:
It seems '--first-pass' and '--second-pass' require arguments (seems 'true' worked), but I was expecting them to be switches (like '--low_latency').  What is the expected behavior for these?

lu-zero:
I'm at the meetup, I'll try to see if there something we can do that is better than this +__+

tdaede:
We turn clippy on and off per lint, so if there's something that's really dumb we can just turn it off.

dwbuiten:
Most of its suggestions are kinda OK. Putting a `_` in the magic and e.g. `0x7FFFFFFF` -> `0x7FFF_FFFF` seems real silly.

src/bin/common.rs
```diff
@@ -82,12 +84,16 @@ pub fn parse_cli() -> CliOptions {
     )
     // ENCODING SETTINGS
     .arg(
-      Arg::with_name("PASS")
-        .help("Specify first-pass or second-pass to run as a two-pass encode; If not provided, will run a one-pass encode")
-        .short("p")
-        .long("pass")
+      Arg::with_name("FIRST_PASS")
```
tdaede: Maybe call this FIRST_PASS_FILE for more clarity in --help?

src/bin/common.rs
```diff
@@ -82,12 +84,16 @@ pub fn parse_cli() -> CliOptions {
...
+        .help("Perform the first pass of a two-pass encode, saving the pass data to the specified file for future passes")
+        .long("first-pass")
+        .takes_value(true)
+    )
+    .arg(
+      Arg::with_name("SECOND_PASS")
```
tdaede: SECOND_PASS_INPUT_FILE maybe?

src/bin/common.rs
dwbuiten: I thought it was silly to add new options in the first place instead of using `--stats` and `--pass`... Perhaps ask Tim?

src/bin/common.rs
dwbuiten: See other reply.

dwbuiten:
> Looks OK. We already have bit serialization code used for the uncompressed header, it might be nice to use that. But meh.

Which parts?

src/bin/common.rs
tdaede: Yeah that makes sense. The previous --stats option actually outputs a "human readable" json file rather than the machine-readable first pass file... we could rename that one to something else.

I think potentially the idea of the two options is to allow x many passes, by both having a pass input and output file at the same time.

shssoichiro: I would suggest there's no reason not to have them be the same JSON file. The intent behind the JSON output file is that someone could write a script to extract data they cared about from the first-pass file (e.g. keyframe numbers). I don't think it would conflict with that goal to have additional fields in that file, and having just one stats file would be simpler IMO.

dwbuiten: I'm strongly against having a JSON based stats file, especially when we introduce temportal RDO, which has a *lot* of data...

lu-zero: We already use bincode for the machine-readable file for the train-rdo, we can use it for the pass files.

lu-zero: I do not have a strong opinion. What works well for you works for me :)

dwbuiten:
> I think potentially the idea of the two options is to allow x many passes, by both having a pass input and output file at the same time.

Merging as-is right now then with a go-ahead from IRC. The CLI options aren't API, so we can revisit them if need be.

PR #1366, opened 2019-06-20
Address codeview's suggestion in PR #1364; not a bugfix

xiphmont:
Clarification suggested by Yushin; I agree I was being over-paranoid
in rows/cols setup, and the paranoia is just going to confuse someone
eventually.


ycho:
Yay- you know I never blamed though. I guess that Mar's Rover would have such code as well! Thanks for addressing this!

PR #1368, opened 2019-06-20
Clean up subsampled block size and chroma transform size functions

rzumer:
Requesting @ycho for review as the last person who touched this code, but if someone else wants to review please go ahead.

I found the format of the `Subsampled_Size` table (which is equivalent to `ss_size_lookup`) to be difficult to read, and there is unnecessary data for subsampling in the vertical but not horizontal direction (which does not correspond to any supported color space). It is replaced here with pattern matching. It is a bit longer than the LUT but more readable in my opinion.

There were also some errors in the `ss_size_lookup` tables, mostly block sizes that should be invalid that were not marked as such in 4:2:2. I used the specification as reference for this conversion.

### DIFF
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -236,8 +236,52 @@ impl BlockSize {
     }
   }
 
+  /// Source: Subsampled_Size (AV1 specification section 5.11.38)
+  pub fn subsampled_size(self, xdec: usize, ydec: usize) -> BlockSize {
+    match (xdec, ydec) {
+      (0, 0) /* 4:4:4 */ => self,
+      (1, 0) /* 4:2:2 */ => match self {
+        BLOCK_4X4 | BLOCK_8X4 => BLOCK_4X4,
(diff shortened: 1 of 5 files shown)


src/partition.rs
```diff
@@ -226,8 +226,52 @@ impl BlockSize {
     }
   }
 
+  /// Source: Subsampled_Size (AV1 specification section 5.11.38)
+  pub fn subsampled_size(self, xdec: usize, ydec: usize) -> BlockSize {
+    match (xdec, ydec) {
+      (0, 0) /* 4:4:4 */ => self,
+      (1, 0) /* 4:2:2 */ => match self {
+        BLOCK_4X4 | BLOCK_8X4 => BLOCK_4X4,
```
ycho: Right, this is rustic way!
The whole ss_size_lookup[ ] table and how ss_size is lookup-ed has been hastly copied from libaom.

rzumer:
Just force-pushed to rebase on top of master.

PR #1369, opened 2019-06-20
Disable unreadable literal lint.

tdaede:
It works OK for decimals but splits hex unnecessarily. If it
gets an option for controlling how sensitive it is, we can turn
it back on.


dwbuiten:
LGTM, I suppose.

Does it really not have a way to confine it to decimals?

tdaede:
https://github.com/rust-lang/rust-clippy/issues/2842

PR #1372, opened 2019-06-20
Fix vertical partitions being coded for 4:2:2 content

rzumer:
One branch was allowing vertical partitions to be added to the list.

Also panic with a more friendly error message instead of trying to access an array out of bounds when trying to code such an invalid partition.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1780,7 +1780,13 @@ fn encode_partition_topdown<T: Pixel, W: Writer>(
       if split_horz { partition_types.push(PartitionType::PARTITION_HORZ); };
       if split_vert { partition_types.push(PartitionType::PARTITION_VERT); };
     } else if bsize.width_log2() == fi.min_partition_size.width_log2() + 1 {
-      partition_types.extend_from_slice(RAV1E_PARTITION_TYPES);
+      partition_types.push(PartitionType::PARTITION_NONE);
+      partition_types.push(PartitionType::PARTITION_SPLIT);
+      partition_types.push(PartitionType::PARTITION_HORZ);
+
+      if fi.sequence.chroma_sampling != ChromaSampling::Cs422 {
+        partition_types.push(PartitionType::PARTITION_VERT);
+      }
     } else {
       partition_types.push(PartitionType::PARTITION_NONE);
       partition_types.push(PartitionType::PARTITION_SPLIT);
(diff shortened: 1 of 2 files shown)


src/partition.rs
```diff
@@ -280,6 +280,10 @@ impl BlockSize {
 
   pub fn largest_chroma_tx_size(self, xdec: usize, ydec: usize) -> TxSize {
     let plane_bsize = self.subsampled_size(xdec, ydec);
+    if plane_bsize == BLOCK_INVALID {
+      panic!("invalid block size for this subsampling mode");
```
tdaede: Can we put this panic in subsampled_size instead? I'd love to get rid of the BLOCK_INVALID enum value.

rzumer: There are other areas in the code where the output of `subsampled_size()` is checked, so it is not doable without bigger changes.

tdaede: OK. We can leave that for some other time.

(or right now, I might do it)

rzumer: Sounds good.

src/encoder.rs
```diff
@@ -1780,7 +1780,13 @@ fn encode_partition_topdown<T: Pixel, W: Writer>(
       if split_horz { partition_types.push(PartitionType::PARTITION_HORZ); };
       if split_vert { partition_types.push(PartitionType::PARTITION_VERT); };
     } else if bsize.width_log2() == fi.min_partition_size.width_log2() + 1 {
-      partition_types.extend_from_slice(RAV1E_PARTITION_TYPES);
+      partition_types.push(PartitionType::PARTITION_NONE);
+      partition_types.push(PartitionType::PARTITION_SPLIT);
```
ycho: I'd like to share that changing the order of partition type causes a change in bitstream thus in AWCY bd-rate results (ignorable, < 0.02%, presumably due to equality condition). https://beta.arewecompressedyet.com/?job=master-c84872d70f7fd741d7c3c03cf44cf5e4cf1ea21d&job=master-bbb8e24938cf5fc047ce43f4fd55b95e62e0c953

Please note that changing the order of partition type test is fine with current topdown strategy but not allowed with bottomup fashion.

rzumer: Thank you for pointing this out, will fix.

rzumer: It seems in the `must_split` case if we use the same order as `RAV1E_PARTITION_TYPES` there is actually a small loss in average BD-rate (0.14).

PR #1373, opened 2019-06-21
Disable intra tx partition

ycho:
Disable transform partition for intra mode block
    
    Until the two functons has_tr() and has_bl() is fixed to use partition type info
    or add new two separate versions, disable intra tx partition.
    
    Our existing has_tr() and has_bl() have been shared by both mvref and intra prediction uses,
    and it works if partition size == tx block size.
    However, it can fail if multiple tx blocks exist in a partition, due to missing logic
    to decide the availability of top right and bottom left pixels based on partition type.
    
    Then, enable the disabled directional intra prediction
    since it works if no tx partition is done.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -635,7 +635,8 @@ impl<T: Pixel> FrameInvariants<T> {
       fi.ref_frames[i] = 0;
     }
 
-    fi.tx_mode_select = fi.config.speed_settings.rdo_tx_decision;
+    // Until has_tr() and has_bl() is fixed to use partition info, disable intra tx partition
+    fi.tx_mode_select = false;
 
     fi
   }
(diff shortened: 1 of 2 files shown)


ycho:
AWCY runs the commits, so I can post the results soon.

ycho:
This PR does not affect speed level 4~10, since those does not use tx_size RDO.

[AWCY result, disabled tx partition + enable back directional intra prediction at default speed 5](https://beta.arewecompressedyet.com/?job=master-72c47ecfeba33b4d839c9f10584574c1ef892a9c&job=enable_back_dir_intra_pred%402019-06-21T21%3A57%3A37.757Z)

| -0.5290 | -0.6753 | -0.7057 |  -0.6093 | -0.7094 | -0.6987 |    -0.6883 |

Issue #1286, opened 2019-05-20
Desync since LRF refactor


ycho:
Desync between encoder and decoder occurs with the use of 2x2 tiles.

git checkout 465360f63d60b42bdd313a96abf94d376b89f11d -b 465360f6
(Older version seems having same issue
git checkout 8058856aa6950e827c930aadb350f77823893156 -b 8058856aa)

./target/release/rav1e ~/sequences/objective-1-fast/ducks_take_off_1080p50_60f.y4m -o test.ivf -r test_rec.y4m --quantizer 60 --speed=5 --limit=2 --tile-rows-log2 1 --tile-cols-log2 1 --tune Psnr --low_latency

cmp test_rec.yuv test_dec.yuv
test_rec.yuv test_dec.yuv differ: char 11923177, line 1
==> It seems 4th frame has mismatch. (Since frame size = 1920x1080x1.5)

If the tile options "--tile-rows-log2 0 --tile-cols-log2 0", there is no desync.

ycho:
Update, if --quantizer 10 (with 2x2 tiles), the desync has gone.
"--quantizer 100" still has desync

ycho:
It turns out, using the same enc options above but w/o tiles, older commit before TILE has landed desyncs as well.

tmatth:
```
e7ac4a56bd3f7141bf5f7e29e05a04d02ba7bc81 is the first bad commit
commit e7ac4a56bd3f7141bf5f7e29e05a04d02ba7bc81
Author: Monty Montgomery <xiphmont@gmail.com>
Date:   Tue Jan 22 02:52:46 2019 -0500

    Ongoing optimization and refactoring work for LRF RDO
    
    Unroll various boundary cases to get a better feel for expression flow.
```

ycho:
> According to git bisect:
> 
> ```
> e7ac4a56bd3f7141bf5f7e29e05a04d02ba7bc81 is the first bad commit
> commit e7ac4a56bd3f7141bf5f7e29e05a04d02ba7bc81
> Author: Monty Montgomery <xiphmont@gmail.com>
> Date:   Tue Jan 22 02:52:46 2019 -0500
> 
>     Ongoing optimization and refactoring work for LRF RDO
>     
>     Unroll various boundary cases to get a better feel for expression flow.
> ```
> 
[...]

PR #1375, opened 2019-06-21
Fix for #1286: desync in LRF--- not desunc anymore

xiphmont:
Correct subtle bug: LRF filters pixel source (cdef or deblock) is
chosen according to y coordinate after clamping to crop-frame, not
before.  Bug only appears at bottom-of-frame.

### DIFF
--- a/src/lrf.rs
+++ b/src/lrf.rs
@@ -131,32 +131,30 @@ fn sgrproj_box_sum_slow<T: Pixel>(a: &mut u32, b: &mut u32,
                                   stripe_y: isize, stripe_h: usize,
                                   x: isize, y: isize,
                                   r: usize, n: u32, one_over_n: u32, s: u32, bdm8: usize,
-                                  backing: &PlaneSlice<T>, backing_w: usize, backing_h: usize,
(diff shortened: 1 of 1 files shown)


src/lrf.rs
```diff
@@ -255,23 +252,21 @@ fn sgrproj_box_sum_fastx_r2<T: Pixel>(a: &mut u32, b: &mut u32,
...
                                       s: u32, bdm8: usize,
-                                      backing: &PlaneSlice<T>, backing_h: usize,
-                                      cdeffed: &PlaneSlice<T>, cdeffed_h: usize) {
+                                      backing: &PlaneSlice<T>, cdeffed: &PlaneSlice<T>,
+                                      crop_h: usize) {
   let mut ssq = 0;
   let mut sum = 0;
   for yi in y - 2..=y + 2 {
+    // clamp to crop rectangle before deciding source.
+    let clamp_y = clamp(yi, -cdeffed.y, crop_h as isize - 1);
```
lu-zero: Maybe this logic could be factored out since it is used in two places (and then we could put a note on it to remember why it exists).

PR #1379, opened 2019-06-23
Move plane into the frame module

lu-zero:


lu-zero:
Is it fine for you as well @barrbrain ?

PR #1381, opened 2019-06-24
Import crav1e lib.rs in rav1e and make it buildable using cargo-c

lu-zero:
Follow [the short instructions](https://github.com/lu-zero/cargo-c/wiki#install) to test it.

- [x] import all the history from lib.rs
- [x] make it build with the current [cargo-c](https://github.com/lu-zero/cargo-c)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,4 +91,4 @@ codegen-units = 1
 lto = true
 
 [workspace]
-members = ["ivf", "crates/nasm_rs", "crates/avformat-sys"]
+members = [".", "ivf", "crates/nasm_rs", "crates/avformat-sys"]
--- a/src/capi.rs
+++ b/src/capi.rs
@@ -17,7 +17,7 @@ use std::os::raw::c_int;
 /// Raw video Frame
 ///
 /// It can be allocated throught rav1e_frame_new(), populated using rav1e_frame_fill_plane()
-/// and freed using rav1e_frame_drop().
+/// and freed using rav1e_frame_unref().
 pub struct Frame(Arc<rav1e::Frame>);
 
 type EncoderStatus=rav1e::EncoderStatus;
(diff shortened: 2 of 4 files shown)


tdaede:
YES

Cargo.toml
```diff
@@ -91,4 +91,4 @@ codegen-units = 1
 lto = true
 
 [workspace]
-members = ["ivf", "crates/nasm_rs", "crates/avformat-sys"]
+members = [".", "ivf", "crates/nasm_rs", "crates/avformat-sys"]
```
tdaede: wut

src/capi.rs
```diff
@@ -0,0 +1,502 @@
...
+
+/// Raw video Frame
+///
+/// It can be allocated throught rav1e_frame_new(), populated using rav1e_frame_fill_plane()
+/// and freed using rav1e_frame_unref().
+pub struct Frame(Arc<rav1e::Frame<u16>>);
+
+#[repr(C)]
+#[derive(Copy, Clone, Debug, FromPrimitive)]
+pub enum EncoderStatus {
```
tdaede: Are you happy with an enum here? AFAIK it's more "C-like" just to use ints for errors.

Cargo.toml
lu-zero: It is a shortcoming of the toolchain, cargo-metadata does not have the concept of main crate and cargo-metadata does not put the main crate as first otherwise.

src/capi.rs
lu-zero: It did work well enough for the last months so I wouldn't change it now :)

dwbuiten: I'm fine with enums. I like to be explicit, myself

Cargo.toml
tdaede: have you filed an upstream bug?

src/capi.rs
tdaede: k

Cargo.toml
lu-zero: Yes, but cargo-metadata does not have a mean to provide the information I need.

Issue #970, opened 2019-02-08
Full 4:2:2/4:4:4 support


rzumer:
4:2:2 and 4:4:4 chroma sampling is supported in a degraded configuration. The following features should be fixed and re-enabled:
- [ ] inter frames
- [ ] restoration
- [ ] CfL prediction

ycho:
I think these are tasks 'TODO', not bugs?
Also, the terminology 'restoration' in video or image compression/processing generally means restoring from lost signal, or even create new info that does not exist (in super-resolution realm).
I believe you meant 'reconstruction' by 'restoration'.

rzumer:
@ycho:
Features are disabled because of bugs when enabling 4:2:2/4:4:4, hence the label. I added the "enhancement" tag because it does not affect usage with 4:2:0 and they (4:2:2/4:4:4) are still functional or supposed to be functional in the degraded state, so it depends on your perspective.

As you can see in the encoder/sequence parameters, restoration is disabled when not using 4:2:0.

ycho:
Well, yes- it depends on perspective.
I see that "// FIXME: inter unsupported with 4:2:2 and 4:4:4 chroma sampling" is from you, so it is the bug that you desire to be fixed (I mean, by someone who is able). :)
I thought 422 and 444 should not work since I have never agreed on enabling 422 and 444, but there are already bunch of code and even testers are there already!
Regarding 'restoration', yes, I think I misread at 2nd use (i.e. restoration (enabling causes decoding failure") of that term in to-do. But in "inter frames (enabling causes faulty restoration output... ", I doubt what you meant.

rzumer:
>  I see that "// FIXME: inter unsupported with 4:2:2 and 4:4:4 chroma sampling" is from you, so it is the bug that you desire to be fixed (I mean, by someone who is able). :)

I don't think it is a high priority so I am not expecting someone else to look at it. There are likely multiple things missing to enable inter frames with non-4:2:0 sampling. In my commit history you can see I fixed a few issues related to 4:2:2/4:4:4, but I have not had time to look at it since then to fix the remaining ones, including inter frames (even intra-only does not work), restoration and disabled prediction modes.

>  I thought 422 and 444 should not work since I have never agreed on enabling 422 and 444, but there are already bunch of code and even testers are there already!

Do you mean that you would prefer to disable 4:2:2/4:4:4 completely because you disagree with enabling them? They were enabled in this degraded configuration back in February.

[...]

PR #1382, opened 2019-06-24
Fix and enable non-4:2:0 inter frames (excluding sub-8x8 blocks)

rzumer:
Some work towards #970.

This also explicitly disables blocks of size under 8x8 for non-4:2:0 inter frames via ugly condition sets in the `encoder.rs` partitioning functions.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1473,17 +1473,27 @@ pub fn write_tx_tree<T: Pixel>(
 
     for p in 1..3 {
       ts.qc.update(qidx, uv_tx_size, false, fi.sequence.bit_depth, fi.dc_delta_q[p], fi.ac_delta_q[p]);
-      let tx_bo = BlockOffset {
-        x: tile_bo.x  - ((bw * tx_size.width_mi() == 1) as usize),
-        y: tile_bo.y  - ((bh * tx_size.height_mi() == 1) as usize)
-      };
 
-      let po = tile_bo.plane_offset(&ts.input.planes[p].cfg);
-      let (_, dist) =
-        encode_tx_block(fi, ts, cw, w, p, tx_bo, luma_mode, uv_tx_size, uv_tx_type,
+      for by in 0..bh_uv {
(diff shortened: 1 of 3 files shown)


tdaede:
You mean 8x8 and smaller?

rzumer:
@tdaede Yes, thanks, just fixed the description.

src/encoder.rs
```diff
@@ -1473,17 +1473,27 @@ pub fn write_tx_tree<T: Pixel>(
...
       ts.qc.update(qidx, uv_tx_size, false, fi.sequence.bit_depth, fi.dc_delta_q[p], fi.ac_delta_q[p]);
-      let tx_bo = BlockOffset {
-        x: tile_bo.x  - ((bw * tx_size.width_mi() == 1) as usize),
-        y: tile_bo.y  - ((bh * tx_size.height_mi() == 1) as usize)
-      };
 
-      let po = tile_bo.plane_offset(&ts.input.planes[p].cfg);
-      let (_, dist) =
-        encode_tx_block(fi, ts, cw, w, p, tx_bo, luma_mode, uv_tx_size, uv_tx_type,
+      for by in 0..bh_uv {
```
tdaede: Ohh is this for 32x32 chroma tx limitation on 64x64 txs?

rzumer: I guess so. I naively copied it from `write_tx_blocks()` in case that was the missing detail and it seems it was.

tdaede:
You could also update the README if you like.

rzumer:
@tdaede Sorry, I forgot to ask about the readme; what would you like to change in there? Currently it does not provide any details about disabled features for 4:2:2 and 4:4:4, so the bullet point saying the support for these sampling formats is limited is still valid.

tdaede:
I guess Limited is still okay. This vastly improves the usability of 4:4:4 and 4:2:2 though, to where many would be happy using it as is.

rzumer:
OK, if the remaining issues are hard to fix I will think of a better way to express that line.

PR #1384, opened 2019-06-26
capi: Add two pass API

dwbuiten:
Currently, it uses heap allocations to pass the returned slice to C-land. This isn't much a perf hit, give the infrequent use (once per frame), and is a lot simpler than having callers provide alloc callbacks.


dwbuiten:
Currently in the process of testing, since I need to add support to either $dayjob transcoder or the FFmpeg patch, to test.

After this lands, I'll be upstreaming the FFmpeg patch for good.

PR #1387, opened 2019-06-27
set key frame flag for container format.

xuguangxin:
container format will use this information to build seek index.
This will make the mp4 seekable

### DIFF
--- a/src/bin/muxer/ivf.rs
+++ b/src/bin/muxer/ivf.rs
@@ -1,5 +1,6 @@
 use super::Muxer;
 use ivf::*;
+use rav1e::prelude::*;
 use std::fs::File;
 use std::io;
 use std::io::Write;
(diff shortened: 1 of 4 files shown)


src/bin/muxer/ivf.rs
```diff
@@ -25,7 +26,8 @@ impl Muxer for IvfMuxer {
     );
   }
 
-  fn write_frame(&mut self, pts: u64, data: &[u8]) {
+  #[allow(unused_variables)]
+  fn write_frame(&mut self, pts: u64, data: &[u8], frame_type: FrameType) {
```
lu-zero:
```suggestion
  fn write_frame(&mut self, pts: u64, data: &[u8], _frame_type: FrameType) {
```

Is simpler.

xuguangxin: thanks for suggestion. 
fixed.

PR #1393, opened 2019-06-29
 Add an activity mask

shreevari:
I had to change the source branch.
Continued from #1263.

### DIFF
--- a/src/activity.rs
+++ b/src/activity.rs
@@ -1,0 +1,117 @@
+// Copyright (c) 2017-2019, The rav1e contributors. All rights reserved
+//
+// This source code is subject to the terms of the BSD 2 Clause License and
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+use crate::frame::*;
+use crate::tiling::*;
+use crate::util::*;
+
+#[derive(Debug, Default, Clone)]
+pub struct ActivityMask {
+  variances: Vec<Vec<f64>>,
+  // Width and height of the original frame that is masked
+  width: usize,
+  height: usize,
+  // Side of unit (square) activity block in log2
+  granularity: usize,
+}
+
+impl ActivityMask {
+  pub fn from_plane<T: Pixel>(luma_plane: &Plane<T>) -> ActivityMask {
+    let PlaneConfig { width, height, .. } = luma_plane.cfg;
+
+    let granularity = 3;
+
+    let aligned_luma = Rect {
+      x: 0_isize,
+      y: 0_isize,
+      width: (width >> granularity) << granularity,
+      height: (height >> granularity) << granularity,
+    };
+    let luma = PlaneRegion::new(luma_plane, aligned_luma);
+
+    let mut variances = Vec::new();
+
+    for y in 0..height >> granularity {
+      let mut row = Vec::new();
+      for x in 0..width >> granularity {
+        let block_rect = Area::Rect {
+          x: (x << granularity) as isize,
+          y: (y << granularity) as isize,
+          width: 8,
+          height: 8,
+        };
+
+        let block = luma.subregion(block_rect);
+
+        let mean: f64 = block
+          .rows_iter()
+          .flatten()
+          .map(|&pix| {
+            let pix: i16 = CastFromPrimitive::cast_from(pix);
+            pix as f64
+          })
+          .sum::<f64>()
+          / 64.0_f64;
+        let variance: f64 = block
+          .rows_iter()
+          .flatten()
+          .map(|&pix| {
+            let pix: i16 = CastFromPrimitive::cast_from(pix);
+            (pix as f64 - mean).powi(2)
+          })
+          .sum::<f64>();
+        row.push(variance);
+      }
+
+      variances.push(row)
+    }
+    ActivityMask { variances, width, height, granularity }
+  }
+
+  pub fn variance_at(&self, x: usize, y: usize) -> Option<f64> {
+    if x > self.width >> self.granularity
+      || y > self.height >> self.granularity
+    {
+      None
+    } else {
+      Some(*self.variances.get(x).unwrap().get(y).unwrap())
+    }
+  }
+
+  pub fn mean_activity_of(&self, rect: Rect) -> Option<f64> {
+    let Rect { x, y, width, height } = rect;
+    let (x, y) = (x as usize, y as usize);
+    let granularity = self.granularity;
+    let (dec_x, dec_y) = (x >> granularity, y >> granularity);
+    let (dec_width, dec_height) =
+      (width >> granularity, height >> granularity);
+
+    if x > self.width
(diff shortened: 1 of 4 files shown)


src/encoder.rs
```diff
@@ -351,6 +358,72 @@ impl<T: Pixel> FrameState<T> {
     }
   }
 
+  /// Computes the variance of 8x8 blocks factoring Contrast Sensitivity
+  /// Function in the tx_domain.
+  ///
+  /// # Panics
+  ///
+  /// Panics if the activity mask is not decimated by 8 on both axes.
+  pub fn compute_activity_mask(&mut self) {
```
lu-zero: From what I can see it could be computed independently from the rest of the rest of encoding process, could you please make so that the function takes in the Plane + it own context data?

I would try to have it as a pre-process stage since it would be quite simple to run this in parallel over a group of frames before we run some inherently serial process down the road.

src/tiling/tile_state.rs
```diff
@@ -50,6 +50,8 @@ pub struct TileStateMut<'a, T: Pixel> {
   pub input_tile: Tile<'a, T>, // the current tile
   pub input_hres: &'a Plane<T>,
   pub input_qres: &'a Plane<T>,
+  pub activity_mask: &'a Plane<u16>,
+  pub mean_activity: u16,
```
lu-zero: Do we want a mean activity per-tile or per frame or have it with a finer granularity?

src/tiling/tile_state.rs
```diff
@@ -144,4 +168,4 @@ impl<'a, T: Pixel> TileStateMut<'a, T> {
       y: self.sbo.y + tile_sbo.y,
     }
   }
-}
+}
```
lu-zero: Please configure your editor or run rustfmt before pushing.

src/encoder.rs
shreevari: Yes, I'll try it out.

src/tiling/tile_state.rs
shreevari: I don't think we need any mean tile activity stored since the activity mask is not normalized anymore. We can calculate the mean in any context when required.

src/tiling/tile_state.rs
shreevari: I manually check for these kinds of things right now. I will run rustfmt from now on.

src/encoder.rs
shreevari: I have moved the computation of the activity mask to its own library. But I have not moved the call to a pre-processing stage.

lu-zero:
And since you'll have to rebase it you might update the year while at it :)

src/activity.rs
```diff
@@ -0,0 +1,145 @@
+// Copyright (c) 2017-2018, The rav1e contributors. All rights reserved
```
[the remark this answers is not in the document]
shreevari: Done

src/encoder.rs
tdaede: This looks OK. That said this single-line function is a bit superfluous.

src/quantize.rs
```diff
@@ -17,6 +17,8 @@ use std::convert::Into;
 use std::mem;
 use std::ops::AddAssign;
 
+pub const QSCALE: usize = 3;
```
tdaede: Can you please define this constant with a comment? It's a bit far removed from where it's actually used.

src/activity.rs
```diff
@@ -0,0 +1,145 @@
...
+      x: 0_isize,
+      y: 0_isize,
+      width: (width >> xdec) << xdec,
+      height: (height >> ydec) << ydec,
+    };
+    let luma = PlaneRegion::new(luma_plane, aligned_luma);
+    
+    let mut mask = Vec::new();
+
+    for x in 0 .. width >> xdec {
```
ycho: I think you want to switch "for x" and "for y", so that y loop is outer side that mostly benefit to consume cache-loaded line of bytes ( >= 64)?

src/quantize.rs
shreevari: Found out that this is actually a duplicate.
I'll use [this](https://github.com/xiph/rav1e/blob/master/src/rate.rs#L50) instead.

src/activity.rs
shreevari: I'm sorry, my browser was showing the cached version of this page and didn't load your comment till I pushed changes.
I'll try to make the switch.

shreevari:
> Please consider change my commented request.

@ycho I have made the changes you have requested. Thanks for noticing the mixed up iteration sequence.

src/encoder.rs
shreevari: I agree. I'll remove the function.

ycho:
> > Please consider change my commented request.
> 
> @ycho I have made the changes you have requested. Thanks for noticing the mixed up iteration sequence.

Yes, I see that this late and wanted to mark it as resolved but seems cannot do that?

src/activity.rs
```diff
@@ -0,0 +1,126 @@
...
+      width: 0_usize,
+      height: 0_usize,
+      granularity: 0_usize,
+    }
+  }
+
+  pub fn new_from_plane<T: Pixel>(luma_plane: &Plane<T>) -> ActivityMask {
+    let PlaneConfig { width, height, .. } = luma_plane.cfg;
+
+    let granularity = 3;
```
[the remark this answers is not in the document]
shreevari: Yes, granularity in both axes is the same and is 2^3 = 8 pixels.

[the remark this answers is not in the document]
shreevari: > If so, some comment next to "granularity: 0_usize,"will be helpful to dev.

Do I justify why granularity is 0 at this point or mention what granularity is when it is first defined?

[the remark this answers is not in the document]
ycho: First of all, we want to know what is the unit for 'granularity'. Is it pixel or pixels in log2, etc?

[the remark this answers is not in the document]
shreevari: > First of all, we want to know what is the unit for 'granularity'. Is it pixel or pixels in log2, etc?

I've mentioned more about granularity over [here](https://github.com/xiph/rav1e/pull/1393/files#diff-5a664c95e33c6765f7b5588364c6fcb2R20). The unit is pixels in log_2.

src/activity.rs
```diff
@@ -0,0 +1,126 @@
...
+use crate::tiling::*;
+use crate::util::*;
+
+#[derive(Debug, Default, Clone)]
+pub struct ActivityMask {
+  variances: Vec<Vec<f64>>,
+  // Width and height of the original frame that is masked
+  width: usize,
+  height: usize,
+  // Side of unit (square) activity block in log2
```
ycho: Oh, there it was already, good work. I missed this with seeing just impl { } below.

shreevari: Thanks :smile:

src/activity.rs
```diff
@@ -0,0 +1,126 @@
...
+  granularity: usize,
+}
+
+impl ActivityMask {
+  pub fn new() -> ActivityMask {
```
barrbrain: I think this no-args `new()` could be `Default::default()` instead. See other examples of `impl Default for`.

shreevari: I've derived default on ActivityMask anyway. What do you think about chucking this whole new() function?
https://github.com/xiph/rav1e/pull/1393/files#diff-5a664c95e33c6765f7b5588364c6fcb2R14

PR #1394, opened 2019-06-30
Move luma_chroma_mode_rdo to a stand-alone function

shreevari:
luma_chroma_mode_rdo was a huge closure.
It is moved to a function of it's own.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -447,6 +447,121 @@ impl Default for EncodingSettings {
     }
   }
 }
+
+#[inline]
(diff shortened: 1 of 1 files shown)


shreevari:
> You should inline it.
> 
> Ideally we should try to reduce the arguments, when possible.

#[inline] or #[inline(always)]?

shreevari:
> Ideally we should try to reduce the arguments, when possible.

How do I go about reducing the arguments? Since a closure could capture variables from its environment, there were fewer arguments before.

lu-zero:
Let's land it with simple inline

PR #1396, opened 2019-07-01
Use integral images in the self guided filter

KyleSiefring:
How does this look?


lu-zero:
Needs a rebase, and looks awesome :)

xiphmont:
Yeah, needs a rebase but I'm pretty sure additional work past that should happen after push to master.  This is a great chunk of code.

xiphmont:
Yeah, needs a rebase but I'm pretty sure additional work past that should happen after push to master.  This is a great chunk of code.

Issue #1376, opened 2019-06-22
Inconsistent spacing style in CLI options


rzumer:
Some options use dashes to separate words (e.g. tiling options), others use underscores (e.g. low latency). I would like to decide on one and change the options that do not correspond.

lu-zero:
we could go with dashes, IIRC clap could match both so we do not have issues of compatibility.

PR #1398, opened 2019-07-02
replaced underscores with dashes in cli options

khodzha:
underscored options left as aliases for backward compatibility

this should fix https://github.com/xiph/rav1e/issues/1376 (assuming dashes is decided to be the spacing style)


lu-zero:
thank you!

PR #1402, opened 2019-07-03
header: write CDEF strengths in order

tmatth:
This avoids a desync when e.g., CDEF bits == 2

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -851,11 +851,10 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
       assert!(fi.cdef_bits < 4);
       self.write(2, fi.cdef_bits)?; // cdef bits
       for i in 0..(1 << fi.cdef_bits) {
-        let j = i << (3 - fi.cdef_bits);
-        assert!(fi.cdef_y_strengths[j] < 64);
-        assert!(fi.cdef_uv_strengths[j] < 64);
-        self.write(6, fi.cdef_y_strengths[j])?; // cdef y strength
-        self.write(6, fi.cdef_uv_strengths[j])?; // cdef uv strength
+        assert!(fi.cdef_y_strengths[i] < 64);
+        assert!(fi.cdef_uv_strengths[i] < 64);
+        self.write(6, fi.cdef_y_strengths[i])?; // cdef y strength
+        self.write(6, fi.cdef_uv_strengths[i])?; // cdef uv strength
       }
     }
     Ok(())


tmatth:
Alternately, the mismatch between strengths being signalled and strengths being applied could be resolved by something like:
```diff
diff --git a/src/cdef.rs b/src/cdef.rs
index d181a09..610a07d 100644
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -342,8 +342,9 @@ pub fn cdef_filter_superblock<T: Pixel>(
 ) {
   let coeff_shift = fi.sequence.bit_depth as i32 - 8;
   let cdef_damping = fi.cdef_damping as i32;
-  let cdef_y_strength = fi.cdef_y_strengths[cdef_index as usize];
-  let cdef_uv_strength = fi.cdef_uv_strengths[cdef_index as usize];
+  let cdef_sparse_index = cdef_index << (3 - fi.cdef_bits);
+  let cdef_y_strength = fi.cdef_y_strengths[cdef_sparse_index as usize];
+  let cdef_uv_strength = fi.cdef_uv_strengths[cdef_sparse_index as usize];
   let cdef_pri_y_strength = (cdef_y_strength / CDEF_SEC_STRENGTHS) as i32;
   let mut cdef_sec_y_strength = (cdef_y_strength % CDEF_SEC_STRENGTHS) as i32;
   let cdef_pri_uv_strength = (cdef_uv_strength / CDEF_SEC_STRENGTHS) as i32;
```

but to me this solution is clearer.

tmatth:
0 change in metrics.

https://beta.arewecompressedyet.com/?job=master-7f7209e95861e4d15211cb465f1d49dbef86029e&job=rav1e-strengths-in-order-header%402019-07-03T14%3A53%3A21.289Z

src/header.rs
```diff
@@ -851,11 +851,10 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
       assert!(fi.cdef_bits < 4);
       self.write(2, fi.cdef_bits)?; // cdef bits
       for i in 0..(1 << fi.cdef_bits) {
-        let j = i << (3 - fi.cdef_bits);
```
ycho: So, this has been wrong? Use (1 << fi.cdef_bits), instead of "i << (3 - fi.cdef_bits)" that is wrong.

tmatth: @ycho the problem is that this order is not what was being used when the encoder was actually applying CDEF.

PR #1404, opened 2019-07-03
Add an unused satd and move sad

KyleSiefring:


src/diff.rs
```diff
@@ -0,0 +1,524 @@
+// Copyright (c) 2019, The rav1e contributors. All rights reserved
```
tdaede: maybe dist for distortion?

src/diff.rs
```diff
@@ -0,0 +1,524 @@
...
+    hadamard2d(data, (4, 4));
+  }
+
+  fn hadamard8x8(data: &mut [i32]) {
+    hadamard2d(data, (8, 8));
+  }
+
+  #[allow(unused)]
+  #[inline(always)]
+  pub fn get_satd<T: Pixel>(
```
tdaede: Needs to be documented about what it is (4x4 or 8x8 hadamard, larger sizes tiled)

src/diff.rs
```diff
@@ -0,0 +1,524 @@
...
+    get_sad_same_inner::<u8>();
+  }
+
+  #[test]
+  fn get_sad_same_u16() {
+    get_sad_same_inner::<u16>();
+  }
+
+  fn get_satd_same_inner<T: Pixel>() {
+    let blocks: Vec<(BlockSize, u32)> = vec![
```
tdaede: Some sort of test to show the scaling is correct would be nice. I can sorta see that within these numbers. A field of 1's might be nice.

src/diff.rs
```diff
@@ -0,0 +1,524 @@
...
+      (BLOCK_16X16, 9984),
+      (BLOCK_16X32, 13824),
+      (BLOCK_32X16, 13760),
+      (BLOCK_32X32, 27952),
+      (BLOCK_32X64, 37168),
+      (BLOCK_64X32, 45104),
+      (BLOCK_64X64, 84176),
+      (BLOCK_64X128, 127920),
+      (BLOCK_128X64, 173680),
+      (BLOCK_128X128, 321456),
```
tdaede: Also relative to SAD. These look sorta out of whack.

tdaede: Though I guess for a randomly generated frame, it's plausible.

src/diff.rs
```diff
@@ -0,0 +1,534 @@
...
+      (BLOCK_4X16, 3136),
+      (BLOCK_16X4, 2632),
+      (BLOCK_8X32, 7056),
+      (BLOCK_32X8, 6624),
+      (BLOCK_16X64, 18432),
+      (BLOCK_64X16, 21312),
+    ];
+
+    let bit_depth: usize = 8;
+    let (input_plane, rec_plane) = setup_planes::<T>();
```
tdaede: Oh it's different data too. Nevermind.

src/diff.rs
KyleSiefring: This is likely due to how the two planes are set up.

PR #1409, opened 2019-07-05
capi: Support u8 pixel format

lu-zero:
The hard part is done, mapping the capi::Packet can be made it work with a From impl

### DIFF
--- a/src/capi.rs
+++ b/src/capi.rs
@@ -25,7 +25,23 @@ use crate::prelude as rav1e;
 ///
 /// It can be allocated throught rav1e_frame_new(), populated using rav1e_frame_fill_plane()
 /// and freed using rav1e_frame_unref().
-pub struct Frame(Arc<rav1e::Frame<u16>>);
+#[derive(Clone)]
+pub enum Frame {
+  U8(Arc<rav1e::Frame<u8>>),
+  U16(Arc<rav1e::Frame<u16>>),
+}
+
+impl From<Arc<rav1e::Frame<u8>>> for Frame {
+  fn from(f: Arc<rav1e::Frame<u8>>) -> Frame {
+    Frame::U8(f)
+  }
+}
+
+impl From<Arc<rav1e::Frame<u16>>> for Frame {
+  fn from(f: Arc<rav1e::Frame<u16>>) -> Frame {
+    Frame::U16(f)
+  }
+}
 
 #[repr(C)]
 #[derive(Copy, Clone, Debug, FromPrimitive)]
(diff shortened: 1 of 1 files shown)


src/capi.rs
```diff
@@ -78,6 +94,53 @@ pub struct Config {
...
+impl EncoderContext for rav1e::Context<u8> {
+  fn new_frame(&self) -> Frame {
+    rav1e::Context::<u8>::new_frame(self).into()
+  }
+  fn send_frame(&mut self, frame: Option<Frame>) -> Result<(), rav1e::EncoderStatus> {
+    if let Some(frame) = frame {
+      if let Frame::U8(f) = frame {
+        self.send_frame(f)
+      } else {
+        Err(rav1e::EncoderStatus::Failure)
```
dwbuiten: Won't this end up (once the other funcs are done) basically be two whole separate copied implementations?

That seems not ideal compared to having some sort of macro'd `match` inside each function, and having only *one* implementation...

lu-zero: Ideally we can go with the same pattern I applied to `rav1e_frame_fill_plane()`, I wanted to try the dyn Trait way as well and see what works better all in all.

``` rust
match (ctx, frame) {
    (Context::U8(c), Frame::U8(f)) => foo_internal(...),
    (Context::U16(c), Frame::U16(f)) => foo_internal(...),
    _ => SomeError
}
```

Probably is compact enough

dwbuiten: My preference is the latter, yeah - much more compact and only in one location.

src/capi.rs
```diff
@@ -78,6 +94,53 @@ pub struct Config {
...
+trait EncoderContext {
+  fn new_frame(&self) -> Frame;
+  fn send_frame(&mut self, frame: Option<Frame>) -> Result<(), rav1e::EncoderStatus>;
+/*
+  fn receive_packet(&mut self) -> Packet;
+  fn container_sequence_header(&self) -> Vec<u8>;
+
+  fn twopass_bytes_needed(&self) -> usize;
+  fn twopass_in(&mut self, buf: &[u8]) -> Result<usize, EncoderStatus>;
+  fn twopass_out(&self) -> Vec<u8>;*/
```
kodabb: nit: extra comment cruft?

src/capi.rs
```diff
@@ -435,13 +498,13 @@ pub unsafe extern "C" fn rav1e_twopass_in(ctx: *mut Context, buf: *mut u8, buf_s
 #[no_mangle]
 pub unsafe extern "C" fn rav1e_send_frame(ctx: *mut Context, frame: *const Frame) -> EncoderStatus {
     let frame = if frame.is_null() {
-        None
+      None
```
kodabb: nit: whitespaces

src/capi.rs
lu-zero: It is what's left to be done :P

lu-zero:
I added `receive_packet()` and I used the Trait to make so I the other functions can be still converted one by one and at the end we can swap the `Context<u16>` with the `enum` and everything should work properly.

dwbuiten:
Tested and works, output is good, speedup is there.

src/capi.rs
lu-zero: rustfmt will be run after that.

lu-zero:
I squashed the whole thing and addressed the nit @kodabb, now we just have to wait for travis

PR #1411, opened 2019-07-06
Revert "Use integral images in the self guided filter (#1396)"

KyleSiefring:
This reverts commit 16ea18ea6eaf3f0e8549a6804ebb45ea37f659fb.

AWCY shows this commit changing stats. It should not.


KyleSiefring:
Yeah, I think I reintroduced the desync fixed here. https://github.com/xiph/rav1e/commit/3a7c7d12bdf5d61bd939139ba646ab68e3a2b212

PR #1412, opened 2019-07-07
Add fixed version of "Use integral images in the self guided filter (…

KyleSiefring:
…#1396)"

### DIFF
--- a/src/frame/plane.rs
+++ b/src/frame/plane.rs
@@ -303,7 +303,7 @@ impl<T: Pixel> Plane<T> {
   }
 
   #[inline]
-  fn row_range(&self, x: isize, y: isize) -> Range<usize> {
+  pub fn row_range(&self, x: isize, y: isize) -> Range<usize> {
     debug_assert!(self.cfg.yorigin as isize + y >= 0);
     debug_assert!(self.cfg.xorigin as isize + x >= 0);
     let base_y = (self.cfg.yorigin as isize + y) as usize;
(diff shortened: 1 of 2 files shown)


KyleSiefring:
```
diff --git a/src/lrf.rs b/src/lrf.rs
index 1acb584..4958b99 100644
--- a/src/lrf.rs
+++ b/src/lrf.rs
@@ -247,17 +247,19 @@ impl<'a, T: Pixel> Iterator for VertPaddedIter<'a, T> {
   #[inline(always)]
   fn next(&mut self) -> Option<Self::Item> {
     if self.end > self.y {
+      // clamp before deciding the source
+      // clamp vertically to storage at top and passed-in height at bottom
+      let cropped_y = clamp(self.y, 0, self.crop - 1);
+      // clamp vertically to stripe limits
+      let ly = clamp(cropped_y, self.stripe_begin - 2, self.stripe_end + 1);
+
       // decide if we're vertically inside or outside the strip
       let src_plane =
-        if self.y >= self.stripe_begin && self.y < self.stripe_end as isize {
+        if ly >= self.stripe_begin && ly < self.stripe_end as isize {
           self.cdeffed
         } else {
           self.deblocked
         };
-      // clamp vertically to storage at top and passed-in height at bottom
-      let cropped_y = clamp(self.y, 0, self.crop - 1);
-      // clamp vertically to stripe limits
-      let ly = clamp(cropped_y, self.stripe_begin - 2, self.stripe_end + 1);
       // cannot directly return self.ps.row(row) due to lifetime issue
       let range = src_plane.row_range(self.x, ly);
       self.y += 1;
```

PR #1414, opened 2019-07-08
capi: Add some missing "C"

dwbuiten:
It apparently worked without them... somehow.

Spotted by @jamrial.


lu-zero:
extern defaults to "C", I guess being consistent is good.

tdaede:
Indeed C is the default, but I prefer it explicit.

PR #1416, opened 2019-07-08
Add output_frameno & end_of_subgop tests

YaLTeR:
These are tests for `output_frameno` to `input_frameno` correspondence and values of `end_of_subgop`.

The first set of two tests is for low-latency mode (no reordering, nothing interesting).

The second set of 5 tests is for chopping off 0 to 4 frames from the end of the video (5 being our subgop size). These seem to work correctly.

The third set of 5 tests is for splitting a subgop of 5 with a scene change at the 0 to 4th frame. They are marked as `#[ignore]` because this logic doesn't seem to be working correctly, as I outlined in the comment:
```rust
// Ignored because (TODO):
// 1. The scene change detector is fed frames in the wrong order (in the
//    output order rather than in the input order).
// 2. The output_framenos are incorrect for some reason.
```
These tests' ground truth is what I believe should be the correct output.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1404,6 +1404,16 @@ mod test {
   }
   */
 
+  fn fill_frame_const<T: Pixel>(frame: &mut Frame<T>, value: T) {
+    for plane in frame.planes.iter_mut() {
+      let stride = plane.cfg.stride;
+      for row in plane.data.chunks_mut(stride) {
+        for pixel in row {
+          *pixel = value;
+        }
+      }
+    }
+  }
 
   #[interpolate_test(low_latency_no_scene_change, true, true)]
   #[interpolate_test(reorder_no_scene_change, false, true)]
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -1485,4 +1495,316 @@ mod test {
...
+      } else {
+        fill_frame_const(&mut input, T::max_value());
+      }
+
+      let _ = ctx.send_frame(Arc::new(input));
+    }
+
+    ctx.flush();
+
+    (0..)
```
lu-zero: `ctx.inner.frame_invariants.values()` maybe does the same in a simpler way.

YaLTeR: Eh, I'm not sure. Since here I'm calling `set_frame_properties` every step (which actually fills `frame_invariants` one by one) and also collecting the `end_of_subgop` values it returns (which are not saved anywhere else) in addition to the frame invariants themselves. So to use `frame_invariants.values()` I'd have to collect all `end_of_subgop` values separately using basically the same code, and then additionally zip them with `frame_invariants.values()` or something.

lu-zero: We can play code golf on it later :)

PR #1422, opened 2019-07-09
Add frame_invariants.pyramid_level tests

YaLTeR:
Also add missing comments to the `output_frameno` tests since I now understand what's going on there.

The scene change tests are once again commented out because I don't believe that part of the code is currently working correctly.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1588,6 +1588,35 @@ mod test {
 
   #[interpolate_test(0, 0)]
   #[interpolate_test(1, 1)]
+  fn pyramid_level_low_latency_minus(missing: u64) {
+    // Test pyramid_level configurations when there are <missing> less frames
+    // than the perfect subgop size, in no-reorder mode.
+
+    let ctx = setup_encoder::<u8>(
+      64,
+      80,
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -1612,7 +1641,9 @@ mod test {
     let limit = 10 - missing;
 
     // data[output_frameno] = (input_frameno, end_of_subgop)
-    let data = encode_frames(&mut ctx, limit, 0);
+    let data = encode_frames(ctx, limit, 0)
+      .map(|(fi, end_of_subgop)| (fi.input_frameno, end_of_subgop))
+      .collect::<Vec<_>>();
```
tdaede: might be nice for an assert on pyramid depth

YaLTeR: Done.

src/api.rs
```diff
@@ -1811,4 +1957,108 @@ mod test {
       }
     );
   }
+
+  #[interpolate_test(0, 0)]
```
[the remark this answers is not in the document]
YaLTeR: Done.

PR #1430, opened 2019-07-10
Make guess_frame_subtypes() work without frame_invariants

dwbuiten:
Currently it won't be populated in the first call in twopass_out(), before receive_packet() is called.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -577,6 +577,9 @@ impl InterConfig {
   // When re-ordering is disabled, this always returns 0.
   pub(crate) fn get_idx_in_group_output(&self,
    output_frameno_in_gop: u64) -> u64 {
+    // The first frame in the GOP should be a keyframe and is not re-ordered,
+    //  so we should not be calling this function on it.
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -1257,24 +1264,38 @@ impl<T: Pixel> ContextInner<T> {
...
           // We do not currently use forward keyframes, so they should always
-          //  end the current TU.
+          //  end the current TU (thus we always increment ntus below).
           debug_assert!(fi.show_frame);
-          output_frameno += 1;
-          ntus += 1;
-          nframes_total += 1;
-          continue;
+          true
+        }
```
lu-zero: Could I nag you to fix the coding style? ^^ Otherwise we could run a pass of rustfmt over the whole file.

dwbuiten: The coding style is consistent with the rest of the file...

lu-zero: I'll pass rustfmt once you land this, the part consistent with this is another patch from Tim :)

PR #1432, opened 2019-07-10
Rustfmt api and capi

lu-zero:
`capi.rs` is an import from a project with the default rust coding style, `api.rs` piled up discrepancies over time.


tdaede:
is this using nightly rustfmt still?

lu-zero:
yes.

PR #1436, opened 2019-07-10
Move scene change detection to lookahead

YaLTeR:
https://beta.arewecompressedyet.com/?job=master-47cea49ada2684410e250602b07176977d8f0cbc&job=lookahead-scd-2

This introduces a lookahead computation function and moves scene change detection there. The effect is that it's now done in the correct frame order, and therefore the tests from the previous PR are now fixed.

I also changed `SceneChangeDetector` to not update the last frame on `send_frame` as I'm always updating it manually.

I think this logic still has a potential of breaking if the lookahead size is less than the subgop size, but I haven't tried that, and also it's currently at 10 and the plan is to increase it further.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -675,7 +675,9 @@ pub(crate) struct ContextInner<T: Pixel> {
   pub(crate) config: EncoderConfig,
   rc_state: RCState,
   maybe_prev_log_base_q: Option<i64>,
-  pub first_pass_data: FirstPassData
+  pub first_pass_data: FirstPassData,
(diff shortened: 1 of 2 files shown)


YaLTeR:
Rebased.

YaLTeR:
Removed a left-over debug message.

YaLTeR:
Removed a now-redundant line.

lu-zero:
> 18:23 < YaLTeR> lu_zero: do you want me to do another AWCY? I don't think the changes I force pushed affected anything

It is fine then.

PR #1442, opened 2019-07-12
capi: Add RaData

lu-zero:


lu-zero:
> 12:05 < lu_zero> Daemon404: ok to land the capi fix?
> 13:48 < Daemon404> lu_zero: ok

PR #1443, opened 2019-07-12
Move FrameInvariants computation into lookahead

YaLTeR:
[AWCY](https://beta.arewecompressedyet.com/?job=master-2e127a6af412347ee3903ad391c2fabc60c4c9fe&job=lookahead-mv-2_5_check_that_nothing_changed%402019-07-12T23%3A12%3A55.064Z)

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -998,6 +998,7 @@ impl<T: Pixel> ContextInner<T> {
             output_frameno_in_gop,
             next_keyframe_input_frameno
           );
+          assert!(fi.invalid);
           return Ok((fi, false));
         }
       }
@@ -1028,6 +1029,7 @@ impl<T: Pixel> ContextInner<T> {
         self.seq,
         self.gop_input_frameno_start
       );
+      assert!(!fi.invalid);
       Ok((fi, true))
     } else {
       let next_keyframe_input_frameno =
@@ -1040,8 +1042,10 @@ impl<T: Pixel> ContextInner<T> {
         next_keyframe_input_frameno
       );
       if input_frameno >= next_keyframe_input_frameno {
+        assert!(fi.invalid);
         return Ok((fi, false));
       }
+      assert!(!fi.invalid);
       Ok((fi, true))
     }
   }
--- a/src/api.rs
+++ b/src/api.rs
@@ -2269,4 +2269,243 @@ mod test {
       }
     );
   }
+
+  #[interpolate_test(0, 0)]
+  #[interpolate_test(1, 1)]
+  #[interpolate_test(2, 2)]
+  #[interpolate_test(3, 3)]
+  #[interpolate_test(4, 4)]
+  fn output_frameno_incremental_reorder_minus(missing: u64) {
+    // Test output_frameno configurations when there are <missing> less frames
+    // than the perfect subgop size, computing the lookahead data incrementally.
+
+    let mut ctx = setup_encoder::<u8>(
+      64,
+      80,
+      10,
+      100,
+      8,
+      ChromaSampling::Cs420,
+      5,
+      5,
+      0,
+      false,
+      true
+    );
+
+    // TODO: when we support more pyramid depths, this test will need tweaks.
+    assert_eq!(ctx.inner.inter_cfg.pyramid_depth, 2);
+
+    let limit = 10 - missing;
+    for _ in 0..limit {
+      send_frames(&mut ctx, 1, 0);
+      ctx.inner.compute_lookahead_data();
+    }
+    ctx.flush();
+    ctx.inner.compute_lookahead_data();
+
+    // data[output_frameno] = (input_frameno, !invalid)
+    let data = get_frame_invariants(ctx)
+      .map(|fi| (fi.input_frameno, !fi.invalid))
+      .collect::<Vec<_>>();
+
+    assert_eq!(
+      &data[..],
+      match missing {
+        0 =>
+          &[
+            (0, true), // I-frame
+            (4, true), // P-frame
+            (2, true), // B0-frame
+            (1, true), // B1-frame (first)
+            (2, true), // B0-frame (show existing)
+            (3, true), // B1-frame (second)
+            (4, true), // P-frame (show existing)
+            (5, true), // I-frame
+            (9, true), // P-frame
+            (7, true), // B0-frame
+            (6, true), // B1-frame (first)
+            (7, true), // B0-frame (show existing)
+            (8, true), // B1-frame (second)
+            (9, true)  // P-frame (show existing)
+          ][..],
+        1 =>
+          &[
+            (0, true),  // I-frame
+            (4, true),  // P-frame
+            (2, true),  // B0-frame
+            (1, true),  // B1-frame (first)
+            (2, true),  // B0-frame (show existing)
+            (3, true),  // B1-frame (second)
+            (4, true),  // P-frame (show existing)
+            (5, true),  // I-frame
+            (5, false), // Last frame (missing)
+            (7, true),  // B0-frame
+            (6, true),  // B1-frame (first)
+            (7, true),  // B0-frame (show existing)
+            (8, true),  // B1-frame (second)
+            (8, false)  // Last frame (missing)
+          ][..],
+        2 =>
+          &[
+            (0, true),  // I-frame
+            (4, true),  // P-frame
+            (2, true),  // B0-frame
+            (1, true),  // B1-frame (first)
+            (2, true),  // B0-frame (show existing)
+            (3, true),  // B1-frame (second)
+            (4, true),  // P-frame (show existing)
+            (5, true),  // I-frame
+            (5, false), // Last frame (missing)
+            (7, true),  // B0-frame
+            (6, true),  // B1-frame (first)
+            (7, true),  // B0-frame (show existing)
+            (7, false), // 2nd last (missing)
+            (7, false)  // Last frame (missing)
+          ][..],
+        3 =>
+          &[
+            (0, true),  // I-frame
+            (4, true),  // P-frame
+            (2, true),  // B0-frame
+            (1, true),  // B1-frame (first)
+            (2, true),  // B0-frame (show existing)
+            (3, true),  // B1-frame (second)
+            (4, true),  // P-frame (show existing)
+            (5, true),  // I-frame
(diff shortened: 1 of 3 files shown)


src/api.rs
```diff
@@ -683,8 +683,10 @@ pub(crate) struct ContextInner<T: Pixel> {
   keyframes: BTreeSet<u64>,
   /// A storage space for reordered frames.
   packet_data: Vec<u8>,
-  gop_output_frameno_start: u64,
-  pub(crate) gop_input_frameno_start: u64,
+  /// Maps `output_frameno` to `gop_output_frameno_start`.
+  gop_output_frameno_start: BTreeMap<u64, u64>,
+  /// Maps `output_frameno` to `gop_input_frameno_start`.
+  pub(crate) gop_input_frameno_start: BTreeMap<u64, u64>,
```
tdaede: This structure is technically redundant with the keyframe BTreeSet, isn't it?

src/encoder.rs
```diff
@@ -479,6 +479,10 @@ pub struct FrameInvariants<T: Pixel> {
   pub pyramid_level: u64,
   pub enable_early_exit: bool,
   pub tx_mode_select: bool,
+  /// If true, this `FrameInvariants` corresponds to a skipped frame and should
+  /// be ignored. Skipped frames occur when a subgop is prematurely ended, for
+  /// example, by a key frame or the end of the video.
+  pub skip: bool,
```
tdaede: This is inelegant at best. I really want to get rid of writing these frameinvariants at all... I guess I'm okay adding this temporarily (it's basically the same as the bool we were returning first)

Also, "skip" is a bitstream feature, so having another concept named skip is confusing. Maybe "dropped" or "invalid" is better.

src/api.rs
```diff
@@ -976,9 +976,18 @@ impl<T: Pixel> ContextInner<T> {
             == 0
             && prev_input_frameno == (next_keyframe_input_frameno - 1))
         {
+          input_frameno = next_keyframe_input_frameno;
+
+          // If we'll return early, do it before modifying the state.
+          match self.frame_q.get(&input_frameno) {
```
tdaede: frame_q.contains() ?

src/encoder.rs
tdaede: Also, now that this is a fi member, maybe an assert at the beginning of the encode_tile function would make sense to make sure we never try to actually encode one of these frames?

src/api.rs
YaLTeR: Flush is recorded in this queue as `None`, and we're not interested in the flush, so we need `Some(Some(_))`.

src/api.rs
```diff
@@ -1123,6 +1126,12 @@ impl<T: Pixel> ContextInner<T> {
     if let Some((last_frame, _)) = frames_to_process.iter().last() {
       self.next_lookahead_frame = last_frame + 1;
     }
+
+    // Compute the frame invariants.
```
tdaede: This comment is no longer accurate?

src/api.rs
```diff
@@ -1230,6 +1244,23 @@ impl<T: Pixel> ContextInner<T> {
...
+            .frame_invariants
+            .iter_mut()
+            .skip_while(|(&output_frameno, _)| {
+              output_frameno <= cur_output_frameno
+            })
+            .filter(|(_, fi)| !fi.skip)
+            .map(|(_, fi)| fi)
+          {
+            subsequent_fi.rec_buffer = rec_buffer.clone();
+            subsequent_fi.set_ref_frame_sign_bias();
```
tdaede: It would be really nice to track that these are invalid somehow, to avoid future bugs. Maybe tests are enough, I dunno. Or maybe they shouldn't even be a part of fi?

src/api.rs
YaLTeR: There's some other stuff using this (inside `select_qi`) which I haven't looked too thoroughly into.

src/api.rs
```diff
@@ -1244,16 +1244,17 @@ impl<T: Pixel> ContextInner<T> {
 
           update_rec_buffer(fi, fs);
 
-          // Copy the `rec_buffer` into subsequent frame invariants.
+          // Copy persistent fields into subsequent frame invariants.
```
tdaede: Are these fields actually used by anything in the lookahead? Technically for everything except the next frame to be encoded, they are set to incorrect values.

src/encoder.rs
YaLTeR: Renamed to `invalid`, added an assert in `encode_frame`.

src/api.rs
tdaede: What I mean is, this is basically the same info as the keyframe set. i.e. this number changes each time the keyframe = true. Arguably it's still worth having as either you need to search for the last keyframe, or compare the current and last gop_input_frameno_start values. It's important that it stays in sync with the keyframes though. It would be nice if it was clearly set where keyframes are set.

src/api.rs
tdaede: ok

src/api.rs
YaLTeR: Wdym? This is the lookahead code that computes the frame invariants.

src/api.rs
YaLTeR: The encode-decode tests caught the `rec_buffer` part but not the `ref_frame_sign_bias` part (which was caught by everything on AWCY). It'd be nice to remove them from FI (and potentially look into refactoring everything in FI that depends on cloning the previous frame's FI), but I can't really say how good of an idea it is because I'm not very familiar with the rest of the code yet.

src/api.rs
YaLTeR: I think in the lookahead I will actually be setting them to different values (original frames) to find MVs.

src/api.rs
tdaede: ah whoops was reading wrong line

src/api.rs
tdaede: I thought maybe to make them Option, but they are used quite a bit. I guess we can leave it for the future.

src/api.rs
YaLTeR: There's some non-obvious semantics on how and when it changes with respect to e.g. invalid frames. I'd wait for further simplification of `build_frame_properties` to fix this.

src/api.rs
tdaede: Yeah OK.

Also technically I think you only need to set one fi in the future for correct encoding?

YaLTeR: Yeah, I thought about this. Let me try it.

src/api.rs
tdaede: OK

src/api.rs
YaLTeR: Well, turns out this breaks `test_encode_decode::reordering_aom`.

tdaede: Probably because we aren't writing into the correct fi ahead.

This isn't computationally expensive, so I guess maybe it's okay to leave it until these get split out by lookahead.

tdaede: Though they are in order by output_frameno, so it's kinda weird.

YaLTeR: What I did was taking this `for subsequent_fi in` and making it into an `if let Some(subsequent_fi) = chain.next() { ... }`.

tdaede: OK I *think* that's correct (it should receive frame_invariants[output_frameno + 1]
Might have to do with not getting past invalid fis or something?

YaLTeR:
Okay, I think I rebased everything correctly.

YaLTeR:
Github says nothing changed so yeah should be good.

YaLTeR:
[Updated AWCY](https://beta.arewecompressedyet.com/?job=master-cafa6f7d0419b6d71c54b3c5480702ca91de1445&job=lookahead-mv-2_6_check_that_nothing_changed%402019-07-13T01%3A08%3A37.608Z)

src/api.rs
```diff
@@ -1006,41 +1047,50 @@ impl<T: Pixel> ContextInner<T> {
...
         output_frameno_in_gop,
         next_keyframe_input_frameno
       );
       if input_frameno >= next_keyframe_input_frameno {
-        return Ok((fi, false));
+        assert!(fi.invalid);
+        return Ok(fi);
       }
-      Ok((fi, true))
+      assert!(!fi.invalid);
```
lu-zero: move the assert to the caller, you are checking basically every branch.

src/api.rs
```diff
@@ -1245,6 +1332,8 @@ impl<T: Pixel> ContextInner<T> {
     }
     for i in 0..(self.output_frameno - 1) {
       self.frame_invariants.remove(&i);
+      self.gop_output_frameno_start.remove(&i);
```
lu-zero: if you have more fields indexed by the same value maybe it is time to consider a struct.

src/api.rs
YaLTeR: In the subsequent commit I removed that bool return value so I'm not sure I can do that.

lu-zero: wouldn't be simpler to check for `fi.invalid` after you call `set_frame_properties`?

YaLTeR: It would be, before I removed the `bool` return value. Unless I'm missing something, I believe that now this is the only way to check the correctness of `fi.invalid` according to the branches taken in `build_frame_properties` (and it has already caught one bug during the development of this PR).

src/api.rs
YaLTeR: I think the eventual plan was to get rid of at least one of those two items once `build_frame_properties` is further simplified? @tdaede what do you think?

tdaede: Yeah I'd like to remove at least input/output_frameno_start but am not 100% sure I can remove both. They could be combined into a struct.

Currently frame_invariants is inserted at a different time than gop_*start, so it can't be put in a struct. Maybe once we drop the redundant frame_invariant writes then it will be ok.

src/api.rs
tdaede: @lu-zero also note that the fi.invalid condition is opposite for some branches

lu-zero: I see, I guess trying to refactor the checks is not really worth it then ^^

YaLTeR:
[AWCY](https://beta.arewecompressedyet.com/?job=master-c00f1556294867eb848ec9f681900932f04cbe32&job=lookahead-mv-2_9_check_that_nothing_changed) showed no differences. Should be good for merge.

PR #1446, opened 2019-07-15
Add assembly for satd

KyleSiefring:


KyleSiefring:
Code review and replies from tim.

```
+  pub mod tiling { pub use crate::tiling::*; }

Can this be pub(crate)?

+pub fn get_sad

This should be pub(crate)

+pub fn get_satd

This should be pub(crate).

+pub use self::nasm::get_satd;

pub(crate)?

> All of these are copied from the surrounding lines or similiar code. I don't know why they are the way they are and I don't really want to change them in this patch.



+fn bench_get_satd

Benchmarks are nice and all, but what I actually want to see are correctness tests. SAD has a test against "known good" values, but even better would be a way to run the whole encoder where the SIMD is always checked against the "native" version (a la CHECK_ASM in Daala).

> We have tests for satd like sad already in the code. I will add a check_asm mode when changing to static dispatch.



+    let org_stride = plane_org.plane_cfg.stride as libc::ptrdiff_t;
+    let ref_stride = plane_ref.plane_cfg.stride as libc::ptrdiff_t;

I think you should probably just use isize instead.

> Yeah, I just copy and pasted from sad. I can change this. I would like to switch to isize myself since I don't see a difference.



+    ; 0, 1, 0, 1, 0, 1, 0, 1
+    ; 2, 3, 2, 3, 2, 3, 2, 3

These comments pretty much make sense to you, and no one else.
I get what you mean here, for the first one, but beyond this what you call "0", "1", etc., becomes unclear, as the inputs are a two dimensional array which you are partially transposing. I.e., it is not easy to verify where all of the values from an input column go, because (a) you name all of the values from different columns the same thing and (b) you're mixing multiple values together without changing their names. I only figured that you were transforming columns and then rows (instead of, for example, transforming 2x2 blocks and then groups of 2x2 blocks) by tracing through the instructions manually. From the comments this was completely opaque.

> Thanks for the input. I was wondering if the comments would be erudite.



+    shufps             tmp, m0, m1, q2020
+    shufps              m0, m1, q3131

Mixing floating point and integer operations on the same registers is something that Intel threatened to make slow in the past, followed through on for some processors, and then made fast again. It is the sort of thing they are likely to mess with in the future as well, so I would avoid it where we can. In this case I do not think that you need it at all, especially since you do not care about the order of the final coefficients in the registers.

> I'm not super concerned about intel breaking a floating point shuffle instruction in the future. I will write it with your method since it looks faster.



+    ; Reduce horizontally early instead of fully transposing
+    ; 2, X, 2, X

I'm not sure this is saving any work, since you are only using half of the resulting registers.

If you use the sequence bfly -> punpck[lh]wd -> bfly -> punpck[lh]dq -> bfly -> punpck[lh]qdq -> bfly+abs+add then you wind up with everything in m0 ready to be summed horizontally, and do not require shufps (to be avoided for the above reasons) or pshufd (slow on some processors, because it requires moving things on arbitrary sides of the 64-bit boundary).

Here is an example of comments decribing what happens in the above sequence that someone else might actually be able to follow. It uses the convention that the first lane is written on the left, which is in my experience the more common one.

m0  d2 c2 b2 a2 d0 c0 b0 a0
m1  d3 c3 b3 a3 d1 c1 b1 a1

bfly ->
m0  d2+d3 c2+c3 b2+b3 a2+a3 d0+d1 c0+c1 b0+b1 a0+a1
m1  d2-d3 c2-c3 b2-b3 a2-a3 d0-d1 c0-c1 b0-b1 a0-a1

punpck[lh]wd
m0  d0-d1 d0+d1 c0-c1 c0+c1 b0-b1 b0+b1 a0-a1 a0+a1
m1  d2-d3 d2+d3 c2-c3 c2+c3 b2-b3 b2+b3 a2-a3 a2+a3

bfly ->
m0  d0-d1+d2-d3 d0+d1+d2+d3 c0-c1+c2-c3 c0+c1+c2+c3 b0-b1+b2-b3 b0+b1+b2+b3 a0-a1+a2-a3 a0+a1+a2+a3 
m1  d0-d2-d2+d3 d0+d1-d2-d3 c0-c1-c2+c3 c0+c1-c2-c3 b0-b1-b2+b3 b0+b1-b2-b3 a0-a1-a2-a3 a0+a1-a2-a3

->
m0  s2 s0 r2 r0 q2 q0 p2 p0
m1  s3 s1 r3 r1 q3 q1 p3 p1

punpck[lh]dq ->
m0  q3 q1 q2 q0 p3 p1 p2 p0
m1  s3 s1 s2 s0 r3 r1 r2 r0

bfly ->
m0  q3+s3 q1+s1 q2+s2 q0+s0 p3+r3 p1+r1 p2+r2 p0+r0
m1  q3-s3 q1-s1 q2-s2 q0-s0 p3-r3 p1-r1 p2-r2 p0-r0

punpck[lh]qdq ->
m0  p3-r3 p1-r1 p2-r2 p0-r0 p3+r3 p1+r1 p2+r2 p0+r0
m1  q3-s3 q1-s1 q2-s2 q0-s0 q3+s3 q1+s1 q2+s2 q0+s0

bflyabs ->
m0  p3-r3+q3-s3  p1-r1+q1-s1 p2-r2+q2-s2 p0-r0+q0-s0 p3+r3+q3+s3 p1+r1+q1+s1 p2+r2+q2+s2 p0+r0+q0+s0
m1  p3-r1-q3+s3  p1-r1-q1+s1 p2-r2-q2+s2 p0-r0-q0+s0 p3+r3-q3-s3 p1+r1-q1-s1 p2+r2-q2-s2 p0+r0-q0-s0

> Will try to implement this.


+    ; Utilize the equality (abs(a+b)+abs(a-b))/2 = max(abs(a),abs(b)) to merge
+    ;  the final butterfly stage, calculation of absolute values, and the
+    ;  first stage of accumulation.

Can't do this.

> Got it.


+    movhlps             m1, m0

Mixing float and int ops again. movhlps is also creating a false dependency here (the high 64 bits of m1 are left unmodified, but you do not care what they are). You can just use punpckhqdq.

+    movhlps            xm1, xm0

I believe this usage can be replaced by punpckhqdq as well.

> I would have not caught the false dependancy. Will avoid this instruction in the future. It's in many places in dav1d so I thought it was good.



+; Two Hadamard transforms share a lane.

I'm not sure what this means (here and elsewhere). "Lane" usually refers to a single element of a SIMD register (i.e., a 16-bit word in a 128-bit register), so it would not make any sense to share one. Do you just mean register?

> lane means an 128-bit part of an avx register. This is actually pretty common parlance. I will be more specific. Use 128-bit lane instead.


+    ; transpose

libtheora's sse2trans.h includes an implementation of an 8x8 transpose that also only uses one temporary but does not require any register renaming, incidentally. It is the sort of thing that should probably be in its own macro, as well. I'm sure we'll want to use it in more places eventually.

> This looks like something that doesn't translate well to avx2's 3 op instructions.


+    ; 0->8, 1->0, 2->1, 3->2, 4->3, 5->4, 6->5, 7->6, 8->7

I think this comment is backwards? (m0, m1) is an active output pair from the previous stage, but sending m0 -> m8 and m1 -> m0 both (a) splits it up and (b) discards the value of of m1 (since m0 is immediately overwritten in the next block).

> Oh, wow. These comments were giving me headaches when I wrote them. Will look into fixing them.


+    ; Reduce horizontally and convert to 32 bits
+    pxor                m2, m2
+    punpcklwd           m1, m0, m2
+    punpckhwd           m0, m2
+    paddd               m0, m1

Any reason not to use pmaddwd with a register of all 1's?

+    pshufd             xm1, xm0, q1111

If only the low 32 bits matter, any reason not to use pshuflw instead? It should be faster.

> Will do.
```

lu-zero:
Squash and merge anytime I guess :)

KyleSiefring:
I'm going to wait for Tim to respond. He may have time to review code on the trip back.

KyleSiefring:
Somewhere in this the strides are getting swapped :/.
Should modify the tests to actually test for this.

PR #1449, opened 2019-07-15
Add sequence to ContextInner and build keyframes directly from seq.

tdaede:


tdaede:
This means that keyframes now start out with a cleared rec_buffer, so qres/hres motion search isn't run on them (and they aren't used for motion vector candidates for future frames). This will save some computation but will affect AWCY results.

https://beta.arewecompressedyet.com/?job=keyframe_from_seq%402019-07-16T00%3A21%3A24.837Z&job=master-5079a62a69ec62578452ead6b9d5e5022b37996e

ycho:
Interesting that the change (gain) only happens with
Netflix_TunnelFlag_1920x1080_60fps_8bit_420_60f.y4m | -0.0819

PR #1450, opened 2019-07-15
Fix inverted condition for min_tile_cols.

### LINKED ISSUE #1447: --tile-cols-log2 not working as expected
I'm running `./target/release/rav1`, built with the latest code. When I use `--tiles 4`, the encoding has two rows and two columns, as expected. But when I use ` --tile-cols-log2 1 --tile-rows-log2 1`, the encoding has two rows but only one column.

I expect it to be the same as `--tiles 4`. Is ` --tile-cols-log2 1 --tile-rows-log2 1` the right way to specify two rows and two columns?

tdaede:
Fixes #1447

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -528,7 +528,7 @@ impl<T: Pixel> FrameInvariants<T> {
       sequence.sb_size_log2(),
       config.width,
       config.height,
-      config.tile_cols_log2.min(min_tile_cols),
+      config.tile_cols_log2,
       config.tile_rows_log2
     );
 


src/encoder.rs
```diff
@@ -524,7 +524,7 @@ impl<T: Pixel> FrameInvariants<T> {
       sequence.sb_size_log2(),
       config.width,
       config.height,
-      config.tile_cols_log2.min(min_tile_cols),
+      config.tile_cols_log2.max(min_tile_cols),
```
rom1v: IMO, the caller should only provide `config.tile_cols_log2`, it's the resposability of `TilingInfo` internals to clip the requested value if necessary (it already does that for other constraints).

Moreover, it seems wrong to min (or max) a number of cols in log2 with a number of cols (not in log2).

tdaede: yeah it looks like that part of the condition was always wrong. I'll move it inside TilingInfo and fix it.

tdaede: In fact you already wrote the exact same condition in TilingInfo, so I just deleted this one entirely.

tdaede:
rebased

Issue #1441, opened 2019-07-12
Scenecut Tweaks


dwbuiten:
See: https://code.videolan.org/videolan/x264/blob/master/encoder/slicetype.c#L1426-1455

Now that we have lookahead for scene cuts, this should be easier, although require some refactoring.

Related to, but not part of https://github.com/xiph/rav1e/issues/794, technically.

CC: @YaLTeR @lu-zero

shssoichiro:
I think this is similar enough to #794 that one of them could be closed, unless you feel there's something extra that's not included in the other ticket. I've picked up work on this again after the recent lookahead changes have made implementing this easier.

dwbuiten:
I opened this because adding these tweaks is markedly different from switching to a cost-based keyframe placement algorithm, to me. I should have have two different issues in the first place...

shssoichiro:
That makes sense. In that case, I can get a PR open for this ticket today and open a separate PR for #794 when that's ready.

shssoichiro:
I found that the lookahead wasn't functioning the way it needed to for this to work correctly, so I've been doing some debugging to try and fix that. It was only doing the full lookahead at the very start of the video, and afterwards would only read one frame past the current. So, I will instead have this PR up Soon (TM).

tdaede:
OK. The one thing I'm not sure about is if it would be better to look at multiple frames in the future (e.g. reading from frame_q) or rather mutating fi's in the past, and rewriting multiple fi's if a keyframe is inserted. One reason to do the latter is the fi's will have qres and hres motion search data on them from @YaLTeR's patches, which could be used instead of the current full-frame diff.

shssoichiro:
That sounds like a possibility. The other option is to do qres and hres motion search during `compute_lookahead_data`, which would avoid having to edit FIs after they've been set. Either way will require a decent amount of refactor. For now, I've opened a PR to move forward with the change as-is (looking only ahead).

YaLTeR:
> do qres and hres motion search during `compute_lookahead_data`

This is actually exactly what I do for temporal RDO, and I'm planning to PR it later today. I need to point out though that this is motion search on _original_ frames rather than _reconstructed_ frames, so the usual qres and hres motion search is still present too.

PR #1452, opened 2019-07-16
Move scene change detection into ContextInner

shssoichiro:
It will need access to these fields for lookahead scene detection


YaLTeR:
Why move it here, again? Also I'm not sure if we always have the last frame in `frame_q` to fetch it from there.

shssoichiro:
This is related to #1441 / #794. When we use lookahead for scene change detection, it will need access to the whole lookahead queue instead of just the current and last frame. This also avoids having the scenechange detector keep an extra reference to the last frame, when we can look that up from the frame queue. Originally I hoped to change the ScenechangeDetector struct to hold a reference to the frame_q, but that wouldn't be possible without unsafe code or a major refactor.

I considered that the detect_scene_change function could just be a standalone function in the scenechange module, but it seemed unnecessary to keep that module for one small function.

Maybe it just makes more sense to close this until the full new scenechange algorithm is implemented. I wanted to keep the refactor separate from changing the algorithm, but maybe that doesn't make sense here.

tdaede:
#1443 just landed, which by nature of constructing fi's in lookahead also effectively moved scenecut detection to lookahead. So maybe this refactor isn't necessary?

shssoichiro:
I think some change still needs to be made here, because `SceneChangeDetector` still doesn't have access to all the frames. However, after looking through the AOM code more, this PR might also not do what we need without more refactoring. I'll hold off on this change for now until we can be more certain what the lookahead-based scenecut detection looks like.

PR #1454, opened 2019-07-17
Constrain input y and stripe size to fit in frame

KyleSiefring:
Doesn't fix any bugs. Might give a tiny speedup.

Results: https://beta.arewecompressedyet.com/?job=master-5079a62a69ec62578452ead6b9d5e5022b37996e&job=test_new_bounds_lrf2%402019-07-17T19%3A46%3A51.020Z

### DIFF
--- a/src/lrf.rs
+++ b/src/lrf.rs
@@ -962,9 +962,15 @@ impl RestorationState {
       let crop_h = (fi.height + (1 << ydec >> 1)) >> ydec;
 
       for si in 0..stripe_n {
-        // stripe y pixel locations must be able to overspan the frame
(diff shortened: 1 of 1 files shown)


KyleSiefring:
Are either of the two comments still relevant? I think the first on is definitely not but I'm not sure about the first one.

xiphmont:
....which two?  I see one in the diff...

KyleSiefring:
The other is trailing a line of code.
This seems to have problems with odd sized videos. Hmm.

KyleSiefring:
Okay, this should work.

PR #1455, opened 2019-07-17
WIP: Remove log2 from tile api.

tdaede:


src/api.rs
```diff
@@ -84,13 +84,14 @@ pub struct EncoderConfig {
...
   pub tune: Tune,
-  /// log2(tile columns). Unused if tiles is specified.
-  pub tile_cols_log2: usize,
-  /// log2(tile rows). Unused if tiles is specified.
-  pub tile_rows_log2: usize,
-  /// Number of tiles desired. The video is split automatically so that
-  /// it contains at least this many tiles. Overrides tile_cols_log2
-  /// and tile_rows_log2.
+  /// Number of tiles horizontally. Encoder may round this number up
+  /// due to bitstream limitations.
```
rom1v: up or down. For example, if user passes `--tile-rows 6`, it will convert internally to use `tile_rows_log2 = 3`.

If the input is 600px, there will be 5 tiles.
If the input is 700px, there will be 6 tiles.
If the input is 800px, there will be 7 tiles.

To avoid confusion, maybe we should force `--tile-rows` and `--tile-cols` to accept only powers of 2?

src/api.rs
```diff
@@ -84,13 +84,14 @@ pub struct EncoderConfig {
...
+  pub tile_cols: usize,
+  /// Number of tiles vertically. Encoder may round this number up
+  /// due to bitstream limitations.
+  pub tile_rows: usize,
+  /// Total number of tiles desired. Encoder will try to optimally split
+  /// to reach this number of tiles, rounded up.
```
dwbuiten: Needs to document which takes precedence if both are set.

lu-zero:
`RUSTFLAGS="--cfg=cargo_c" cargo build` or `check` to test w/out using cargo cbuild.

dwbuiten:
@tdaede @lu-zero I spot https://github.com/xiph/rav1e/pull/1469 - presumably this should also be done pre-release.

lu-zero:
yes.

src/api.rs
tdaede: ehhhhh yeah. Though technically it can still go up with >4k video.

src/api.rs
tdaede: Do you have a preference for the precedence?

dwbuiten: Nope... whichever floats your boat.

src/api.rs
tdaede: Updated it to reject non powers of 2.

src/api.rs
tdaede: Added.

tdaede:
Updated. Note that this makes context creation fallible (can return null) in the capi.

Alternatively we could make the setter fail.

rom1v:
> Note that this makes context creation fallible

Passing a value which is not a power-of-2 could have the same behavior as passing a log2 value greater than 6 (invalid too). It did not required to make the context creation fallible.

(I don't know whether it is better or not to make it fallible, bit IMO it is independant of passing non-log2 values.)

Also, the choice to pass tiles instead of tiles-log2 is just a UI decision (more user-friendly). For simplicity, we could still use log2 values internally for the context config.

What do you think?

src/capi.rs
```diff
@@ -441,23 +441,30 @@ pub unsafe extern fn rav1e_config_parse_int(
   }
 }
 
+fn context_new_internal(cfg: &rav1e::Config) -> Result<Context, rav1e::EncoderStatus> {
```
dwbuiten: Is there a reason this needs to be a separate function?

src/capi.rs
```diff
@@ -441,23 +441,30 @@ pub unsafe extern fn rav1e_config_parse_int(
...
-      _ => EncContext::U16(cfg.new_context())
-    },
-    last_err: None
-  };
-
-  Box::into_raw(Box::new(ctx))
+  if let Ok(ctx) = context_new_internal(cfg) {
+    Box::into_raw(Box::new(ctx))
+  } else {
+    std::ptr::null_mut()
```
dwbuiten: Does this properly work with last_err?

src/capi.rs
```diff
@@ -383,8 +383,8 @@ unsafe fn option_match(
     "threads" => (*cfg).cfg.threads = value.parse().map_err(|_| ())?,
 
     "tiles" => enc.tiles = value.parse().map_err(|_| ())?,
-    "tile_rows_log2" => enc.tile_rows_log2 = value.parse().map_err(|_| ())?,
-    "tile_cols_log2" => enc.tile_cols_log2 = value.parse().map_err(|_| ())?,
+    "tile_rows" => enc.tile_rows = value.parse().map_err(|_| ())?,
+    "tile_cols" => enc.tile_cols = value.parse().map_err(|_| ())?,
```
dwbuiten: Is it preferrable to fail during parsing?

Will the `new_context` error info be properly propagated/surfaced to the user (i.e. they know what failed)?

src/capi.rs
lu-zero: You can `map()` and `unwrap_or()` to the same effect but it would be less readable.

src/capi.rs
lu-zero: Probably it would be a good idea to fail on setting in the `capi`.

src/capi.rs
tdaede: I don't think so. I can make it work.

src/capi.rs
tdaede: Sure, I can do that.

src/capi.rs
```diff
@@ -464,23 +489,42 @@ pub unsafe extern fn rav1e_config_parse_int(
...
 /// Multiple contexts can be generated through it.
+/// Returns Null if context creation failed, e.g. by passing
+/// an invalid Config.
 #[no_mangle]
 pub unsafe extern fn rav1e_context_new(cfg: *const Config) -> *mut Context {
   let cfg = &(*cfg).cfg;
   let enc = &cfg.enc;
 
-  let ctx = Context {
+  let ctx_maybe = Result::<Context> {
```
lu-zero: This should call `context_new_internal` I guess.

dwbuiten: `context_new_internal` Seems to be added and never used?

tdaede: Yeah I meant to drop context_new_internal entirely.

src/capi.rs
tdaede: @lu-zero I don't actually know how to fix this. Suggestion?

lu-zero: Either we add a last_err to Config and more api surface or it is fine to just return NULL.

I would just return NULL.

PR #1457, opened 2019-07-18
api: Remove the old and never-used stats/pass API

dwbuiten:
This was added before Rav1e even had rate control, and was never used, since Tim's pass API is buffer based.

Sending this now since it needs to be cleaned up before Crate Day™.

# PLEASE READ

So this builds and runs fine, but now the build has a bunch of weird and (entirely unrelated?) dead code warnings like:

```
warning: method is never used: `y`
   --> src/tiling/tile_motion_vectors.rs:79:7
    |
79  |       pub fn y(&self) -> usize {
    |       ^^^^^^^^^^^^^^^^^^^^^^^^
...
113 | tile_motion_vectors_common!(TileMotionVectorsMut, mut);
    | ------------------------------------------------------- in this macro invocation
```

Also from `rdo.rs`, `lrf.rs`, and a few others. I don't get it... can someone help?

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -26,7 +26,6 @@ use crate::util::Pixel;
 
 use std::collections::BTreeMap;
 use std::collections::BTreeSet;
-use std::path::PathBuf;
 use std::sync::Arc;
 use std::{cmp, fmt, io};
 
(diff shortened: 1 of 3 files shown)


tdaede:
One goal of this pass API not addressed by the new stats API was to indicate to higher level tools where keyframes happened, for purposes of splitting / packaging. Don't know if we want to provide a replacement for that or not.

tdaede:
Actually because Packet has this information in the api, we don't need this API at all. If someone still wants this it could be resurrected in the CLI only.

PR #1458, opened 2019-07-18
Warnings cleanup

lu-zero:
As uncovered by @dwbuiten patch.

PR #1459, opened 2019-07-18
Do not run benchmarks on Travis CI

rzumer:
As far as I can tell, we are not using CI benchmark results and they may be unreliable. With this change, they will still be built and return compilation errors, but they will not run on Travis.


rzumer:
Compared to #1458, this cuts the total build time by over 10 minutes.

PR #1462, opened 2019-07-19
Constrain input stripe height of lrf to fit in frame

KyleSiefring:
Might fix an rdo problem with odd frame sizes. No tests were run on
this.

### DIFF
--- a/src/lrf.rs
+++ b/src/lrf.rs
@@ -450,7 +450,8 @@ pub fn sgrproj_stripe_filter<T: Pixel>(set: u8, xqd: [i8; 2], fi: &FrameInvarian
       // Move left to encompass all the used data
       &cdeffed.go_left(left_uniques),
       &deblocked.go_left(left_uniques),
-      stripe_h,
+      // since r2 uses every other row, we need an extra row if stripe_h is odd
+      stripe_h + if s_r2 > 0 { stripe_h & 1 } else { 0 },
       crop_h,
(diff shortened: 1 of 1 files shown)


src/lrf.rs
```diff
@@ -964,11 +966,12 @@ impl RestorationState {
       for si in 0..stripe_n {
         let (stripe_start_y, stripe_size) = if si == 0 {
           (0, (64 - 8) >> ydec)
-        }  else {
+        } else {
+          let start = (si * 64 - 8) >> ydec;
           (
-            (si as isize * 64 - 8) >> ydec,
+            start as isize,
             // one past, unlike spec
```
KyleSiefring: I'm unsure what this comment means.

KyleSiefring:
https://beta.arewecompressedyet.com/?job=master-193cf46fa97f3c6cd872861add23409f57221236&job=test_height_bounds_lrf1%402019-07-19T18%3A58%3A27.911Z

src/lrf.rs
xiphmont: I'm using a C-style fencepost so you can check it against foo < end unlike the spec where they use foo <= end

xiphmont: The spec uses end fenceposts set to the last valid entry and checks that accesses are <= that value.
I used a C-style fencepost value that's one past the end and check that accesses are < that value.

PR #1463, opened 2019-07-19
Flip the axes sgr is looped on

KyleSiefring:
Precursor to some form of vectorization.


lu-zero:
Clippy is [sad](https://travis-ci.org/xiph/rav1e/jobs/561249463) could you fix that either by allowing the lint or removing the +0 ?

lu-zero:
Thank you!

Issue #1464, opened 2019-07-20
use const-fn when we can


lu-zero:
It is an issue to track the language [metaissue](https://github.com/rust-lang/rust/issues/57563) as more support lands in stable.

shssoichiro:
Clippy has a lint for this which is disabled by default: https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn

However, the reason it's disabled by default is because it is a WIP and has false positives if used on stable. We should keep an eye on this to see if it becomes better at some point in the future.

shssoichiro:
We've implemented this in the majority of places where it makes sense

PR #1465, opened 2019-07-20
Make ilog64 const

lu-zero:
leading_zeroes and this kind of functions are const-ready since 1.34.0
We neeed https://github.com/rust-lang/rust/issues/57349 to constantize
further.

(first step in #1464)


src/rate.rs
```diff
@@ -86,7 +86,7 @@ const DQP_Q57: &[i64; FRAME_NSUBTYPES] = &[
 // This is the number of bits that would be required to represent v in two's
 //  complement notation with all of the leading zeros stripped.
 // TODO: Mark const once leading_zeros() as a constant function stabilizes.
```
shssoichiro: We can remove this TODO as part of this

rzumer:
Closing in favor of #1466 until `const fn` is supported for trait functions.

PR #1466, opened 2019-07-20
Remove rate::ilog64()

rzumer:
It is superseded by `util.rs::ILog::ilog()`.

### DIFF
--- a/src/rate.rs
+++ b/src/rate.rs
@@ -13,8 +13,7 @@ use crate::quantize::ac_q;
 use crate::quantize::dc_q;
 use crate::quantize::select_ac_qi;
 use crate::quantize::select_dc_qi;
-use crate::util::clamp;
-use crate::util::Pixel;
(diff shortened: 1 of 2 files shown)


lu-zero:
It exists because we cannot have const-fn traits in stable and that could be the case for a while =/

rzumer:
Right, so one of #1465 and #1466 can be chosen, or add const fn functions for all required `PrimInts` besides i64.

rzumer:
After discussing with @lu-zero he feels indifferent about it (correct me if I'm wrong) and I prefer this solution, since less duplicated code seems a bigger benefit than constifying a non-critical function. I will squash and push after CI succeeds.

lu-zero:
It is better to have less lines of code around until the const-fn lands completely.

Keep in mind that it has an effect if the function arguments are constant themselves, otherwise it doesn't do anything :)

PR #1467, opened 2019-07-20
Optimize tile.rs::tile_log2()

rzumer:
Not that useful since tile computation is only done once as far as I know, but this turns `tile_log2()` into a one-liner, and in test benchmarks it is several times faster.

### DIFF
--- a/src/tiling/tiler.rs
+++ b/src/tiling/tiler.rs
@@ -111,11 +111,10 @@ impl TilingInfo {
   ///
   /// <https://aomediacodec.github.io/av1-spec/#tile-size-calculation-function>
   fn tile_log2(blk_size: usize, target: usize) -> usize {
-    let mut k = 0;
-    while (blk_size << k) < target {
-      k += 1;
-    }
-    k
+    debug_assert!(blk_size != 0 && target > blk_size &&
+      blk_size.ilog() == (blk_size - 1).ilog() + 1 && // Block size is a power of 2
+      target.ilog() == (target - 1).ilog() + 1); // Target is a power of 2
+    if target <= blk_size { 0 } else { target.ilog() - blk_size.ilog() }
   }
 
   #[inline(always)]


rzumer:
Test results for reference (this PR is of course, the fastest version):
```
tiles/normal            time:   [4.3147 ns 4.3292 ns 4.3466 ns]                          
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe
tiles/smart             time:   [930.76 ps 932.84 ps 935.71 ps]                         
  3 (3.00%) high mild
  2 (2.00%) high severe
tiles/smarter           time:   [764.55 ps 765.84 ps 767.24 ps]                           
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe
```

rzumer:
Tests caught the inverted comparison in the if statement, hopefully fixed...

rzumer:
It looks like the tests include `block_size` and `target` parameters that are not powers of 2. Do you know if this is valid, @rom1v (or anyone else)? According to @lu-zero resolutions should be restricted to powers of 2. I am seeing a block size of 2304, which seems oddly large, and a target of 3.

rzumer:
After looking at how this function is used I think non-powers of 2 are legal, and supporting them in this PR might not be worth the additional effort, so I will close this for now.

lu-zero:
I misunderstood the question, the tiles are in number of power of 2, not their dimension ^^;

lu-zero:
This should do the right thing.

``` rust
fn tile_log2(blk_size: usize, target: usize) -> usize {
    let t = target.leading_zeros() as usize;
    let b = blk_size.leading_zeros() as usize;
    
    if t > b {
        0
    } else  {
        1 + b - t
    }
}
```

rzumer:
@lu-zero [Here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a32b89bd3f28294006ef816b081011d0) is a Playground link comparing the two.

rzumer:
If someone has a working post-condition to fix off-by-one errors, feel free to leave a comment and re-open, but I have not found the right one and do not want to spend a lot of time on a function that is only called a few times at the start of an encode.

PR #1468, opened 2019-07-20
Convert two functions to const fn

rzumer:
Work towards #1464.

### DIFF
--- a/src/tiling/tiler.rs
+++ b/src/tiling/tiler.rs
@@ -119,7 +119,7 @@ impl TilingInfo {
   }
 
   #[inline(always)]
-  pub fn tile_count(&self) -> usize {
+  pub const fn tile_count(&self) -> usize {
     self.cols * self.rows
   }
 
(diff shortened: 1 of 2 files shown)


src/tiling/tiler.rs
```diff
@@ -118,8 +118,7 @@ impl TilingInfo {
     k
   }
 
-  #[inline(always)]
-  pub fn tile_count(&self) -> usize {
+  pub const fn tile_count(&self) -> usize {
```
lu-zero: cols and rows aren't constant, the compiler shouldn't allow this.

rzumer: It does though.

tdaede: I think it just means you can construct a const TilingInfo, and then call this method on it and the result is also const.

rzumer: That makes sense.

rzumer: @lu-zero I added back the inline attribute to this function, let me know if that's fine by you.

lu-zero: Sure, thank you :)

PR #1470, opened 2019-07-20
Deduplicate tiling configuration validation and adjustment

### LINKED ISSUE #1451: Duplicate MAX_AREA_SB
https://github.com/xiph/rav1e/pull/1356 duplicates some logic inside TilingInfo::new() for maximum area computation. We should only do it once.

rzumer:
This reverts commit 95c773470fdb6bd0e6126e85ae834bdec72494bf and adds a rate restriction to the `TilingInfo` constructor.

Fixes #1451.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -154,6 +154,10 @@ impl EncoderConfig {
       train_rdo: false
     }
   }
+
+  pub fn frame_rate(&self) -> f64 {
+    Rational::from_reciprocal(self.time_base).as_f64()
+  }
 }
 
 /// Contains all the speed settings
--- a/src/api.rs
+++ b/src/api.rs
@@ -43,6 +43,17 @@ impl Rational {
   pub fn new(num: u64, den: u64) -> Self {
     Rational { num, den }
   }
+
+  pub fn from_reciprocal(reciprocal: Self) -> Self {
+    Rational {
+      num: reciprocal.den,
+      den: reciprocal.num,
+    }
+  }
+
+  pub fn as_f64(self) -> f64 {
+    self.num as f64 / self.den as f64
+  }
 }
 
 #[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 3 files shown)


dwbuiten:
Doesn't this then risk producing invalid streams?

rzumer:
@dwbuiten All restrictions except the rate bound are checked in `TilingInfo::new()` as far as I can tell. There is a small window where a stream that meets those requirements can exceed the rate limit.

It is probably better to leave that check in, but the question is whether it should be validated in the `TilingInfo` constructor as with everything else (in which case the frame rate much be added as an argument), or left outside of it (which splits the responsibility for validating `TilingInfo` and is not ideal).

tdaede:
I think it'd be better to add it in TilingInfo::new(). Another copy of the splitting loop (as used outside TilingInfo) could be put inside.

tdaede:
Would you rather add the check in here or open a separate pr to re-add the check inside TilingInfo?

rzumer:
> Would you rather add the check in here or open a separate pr to re-add the check inside TilingInfo?

I can add it to this PR.

rzumer:
I added validation for the rate and tested the addition in the Rust Playground to make sure numbers are accurate.

src/api.rs
```diff
@@ -143,6 +154,10 @@ impl EncoderConfig {
       train_rdo: false
     }
   }
+
+  pub fn frame_rate(&self) -> f64 {
```
lu-zero: This is often not true, probably we should add a field for that (and consider having a tileinfo builder in the frame invariants and not the other way round. ^^

rzumer: I understand that it will not remain accurate, but isn't it always true *currently*, or can rav1e produce VFR streams, or streams with a time base whose reciprocal is not the frame rate?

PR #1471, opened 2019-07-20
Rename TilingInfo constructor

rzumer:
Renamed based on my comment in #1451.

The rationale is that `new()` is normally reserved for "general constructors" that set fields based on the values provided. It is expected that the parameters passed to such constructors are mapped to struct fields "as-is".

I believe that the `TilingInfo` is closer to a "conversion constructor", hence the `from_` prefix. I hope that it will make it clearer that the arguments are not mapped to the struct fields directly.

My main reference is https://rust-lang-nursery.github.io/api-guidelines/naming.html.

If anyone disagrees, or wants to suggest a different name, please feel free.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -524,7 +524,7 @@ impl<T: Pixel> FrameInvariants<T> {
       (config.width * config.height * frame_rate) / (4096_f64 * 2176_f64 * 60_f64 * 1.1) as usize
     );
 
-    let mut tiling = TilingInfo::new(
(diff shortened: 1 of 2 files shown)


rzumer:
@lu-zero I started on this path (`from_frame_invariants()`), but `FrameInvariants` itself needs to create `TilingInfo` in its constructor, introducing a cyclic dependency which is probably more trouble than it saves.

lu-zero:
That means that FrameInvariants should be split.

Land this change and we'll look at FrameInvariants later.

PR #1473, opened 2019-07-21
Make Sequence::sb_size[_log2]() const fn

rzumer:
Work towards #1464.

The same cannot be done with corresponding `FrameInvariants` functions, because trait bounds other than `Sized` are not supported with `const fn` currently (`FrameInvariants` has a `Pixel` bound).

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -290,8 +290,8 @@ impl Sequence {
   }
 
   #[inline(always)]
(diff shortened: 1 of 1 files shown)


rzumer:
> Keep the `#[inline(always)]`, it might still be needed (we should doublecheck).

Done and force-pushed (also to remerge properly/fix Clippy warning)

PR #1475, opened 2019-07-21
Frame params

lu-zero:
- [x] send_frame API
- [x] keyframe placement logic
- [ ] capi/crav1e to expose it? (@dwbuiten do you want to give it a go?)

(It sits on top of the API cleanup)

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -701,6 +701,8 @@ pub(crate) struct ContextInner<T: Pixel> {
   ///  memory for the whole life of the encode.
   // TODO: Is this needed at all?
   keyframes: BTreeSet<u64>,
+  // TODO: Is this needed at all?
+  keyframes_forced: BTreeSet<u64>,
   /// A storage space for reordered frames.
   packet_data: Vec<u8>,
   /// Maps `output_frameno` to `gop_output_frameno_start`.
@@ -814,7 +816,9 @@ impl<T: Pixel> Context<T> {
   /// let mut ctx: Context<u8> = cfg.new_context();
   /// let f1 = ctx.new_frame();
   /// let f2 = f1.clone();
-  /// let info = FrameParameters { keyframe : true };
+  /// let info = FrameParameters {
+  ///   frame_type_override: FrameTypeOverride::Key
+  /// };
   ///
   /// // Send the plain frame data
   /// ctx.send_frame(f1);
@@ -832,7 +836,7 @@ impl<T: Pixel> Context<T> {
   where
     F: IntoFrame<T>
   {
-    let (frame, _) = frame.into();
+    let (frame, params) = frame.into();
 
     if frame.is_none() {
       if self.is_flushing {
@@ -844,7 +848,7 @@ impl<T: Pixel> Context<T> {
       return Err(EncoderStatus::EnoughData)
     }
 
-    self.inner.send_frame(frame)
+    self.inner.send_frame(frame, params)
   }
 
   /// Retrieve the first-pass data of a two-pass encode for the frame that was
@@ -948,6 +952,7 @@ impl<T: Pixel> ContextInner<T> {
       frame_q: BTreeMap::new(),
       frame_invariants: BTreeMap::new(),
       keyframes: BTreeSet::new(),
+      keyframes_forced: BTreeSet::new(),
       packet_data,
       gop_output_frameno_start: BTreeMap::new(),
       gop_input_frameno_start: BTreeMap::new(),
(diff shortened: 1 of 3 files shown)


dwbuiten:
Looks easy enough to expose.

lu-zero:
Now the feature is implemented and tested.

src/api.rs
```diff
@@ -754,17 +784,47 @@ impl<T: Pixel> Context<T> {
...
+  /// It could return `EncoderStatus::EnoughData` if the incoming
+  /// frame queue is full or `EncoderStatus::Failure` if the per-frame
+  /// parameters set an impossible constraint.
   pub fn send_frame<F>(&mut self, frame: F) -> Result<(), EncoderStatus>
   where
-    F: Into<Option<Arc<Frame<T>>>>
+    F: IntoFrame<T>
   {
-    let frame = frame.into();
+    let (frame, props) = frame.into();
```
dwbuiten: We should be consistent in the use of props vs params.

src/api.rs
```diff
@@ -674,6 +679,8 @@ pub(crate) struct ContextInner<T: Pixel> {
   ///  memory for the whole life of the encode.
   // TODO: Is this needed at all?
   keyframes: BTreeSet<u64>,
+  // TODO: Is this needed at all?
+  keyframes_forced: BTreeSet<u64>,
```
dwbuiten: Maybe related to the old code but... why is this a tree?

src/api.rs
```diff
@@ -745,6 +753,28 @@ impl<T: Pixel> fmt::Display for Packet<T> {
...
+  fn into(self) -> (Option<Arc<Frame<T>>>, Option<FrameParameters>) {
+    (Some(self), None)
+  }
+}
+
+impl <T: Pixel> IntoFrame<T> for (Arc<Frame<T>>, FrameParameters) {
+  fn into(self) -> (Option<Arc<Frame<T>>>, Option<FrameParameters>) {
+    (Some(self.0), Some(self.1))
+  }
+}
```
dwbuiten: All of these impls are kind why I think rust is too 'clever' sometimes...

src/api.rs
```diff
@@ -891,16 +952,26 @@ impl<T: Pixel> ContextInner<T> {
...
     self.frame_q.insert(input_frameno, frame);
+
+    if let Some(props) = props {
+      if props.keyframe {
+        self.keyframes_forced.insert(input_frameno);
+      }
+    }
+
+    if !self.needs_more_lookahead() {
+      self.compute_lookahead_data();
```
dwbuiten: Is this related to or on top of the other PR (lookahead into send_frame)?

src/frame/mod.rs
```diff
@@ -20,6 +20,12 @@ pub use plane::*;
 
 const FRAME_MARGIN: usize = 16 + SUBPEL_FILTER_SIZE;
 
+/// Optional per-frame encoder parameters
+pub struct FrameParameters {
+  /// Force the frame to be emitted as keyframe
```
dwbuiten: Is this sufficient (the bool)?

In the future we may want to force frames to be switch frames or intra-only frames.

src/api.rs
lu-zero: naming things is hard: I let you pick properties vs params.

src/api.rs
lu-zero: HashSet seemed overkill.

src/api.rs
lu-zero: It is not related, there should be another pr open I can move this away but I needed to make sure it works with it.

src/api.rs
dwbuiten: Params makes more sense to me, since we're setting how to encode it, not the inherent properties of the frame (like colorspace).

src/api.rs
dwbuiten: Not a simple vec/heap array/queue?

src/api.rs
dwbuiten: Well, as long as it's not misleading or broken...

src/frame/mod.rs
dwbuiten: A frame-type enum makes sense to me (for now it will only have one entry).

src/api.rs
shssoichiro: A set makes sense in my opinion for the O(1) contains checks. We're doing contains checks much more frequently than inserts on both keyframes fields.

shssoichiro:
I think this shouldn't conflict with #1481. `compute_lookahead_data` is self contained, so moving the call to that into `send_frame` shouldn't cause any issues.

src/frame/mod.rs
lu-zero: Done.

src/api.rs
lu-zero: Done.

src/frame/mod.rs
```diff
@@ -20,6 +20,24 @@ pub use plane::*;
 
 const FRAME_MARGIN: usize = 16 + SUBPEL_FILTER_SIZE;
 
+/// Override the frame type decision
+///
+/// Only certain frame types can be selected.
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum FrameTypeOverride {
+  /// Do not force any decision.
+  No,
```
dwbuiten: Is there a point in having `No` instead of making `frame_type_override` an Option?

lu-zero: `Some(FrameTypeOverride::Key)` is fairly verbose.

rzumer: Seems like the idiomatic way to do it though.

lu-zero:
Discussing with @tdaede this takes the priority over the scenechange improvements since looks like it needs further iterations and testing.

I'd land it today and wait for the other issues pending in [the project](https://github.com/xiph/rav1e/projects/11) to be completed.

src/frame/mod.rs
tdaede: One issue is that I think overriding to a Key or S-frame should be allowed (or maybe even intra only), but overriding to a inter frame shouldn't. So that doesn't fit the Option idiom very well.

PR #1477, opened 2019-07-22
Cleanup luma_chroma_mode_rdo.

tdaede:


src/rdo.rs
```diff
@@ -465,97 +465,97 @@ fn luma_chroma_mode_rdo<T: Pixel> (luma_mode: PredictionMode,
   luma_mode_is_intra: bool,
   mode_context: usize,
   mv_stack: &ArrayVec<[CandidateMV; 9]>) {
-    let (tx_size, mut tx_type) = rdo_tx_size_type(
-      fi, ts, cw, bsize, tile_bo, luma_mode, ref_frames, mvs, false,
-    );
+  let (tx_size, mut tx_type) = rdo_tx_size_type(
```
rzumer: Why is the indentation gone?

tdaede: It was erroneous - indented one level too far.

rzumer: Unless the GitHub view is wrong, it looks like there is no indentation in the inner scope of the function now.

rzumer: Unless the GitHub "Files changed" view is wrong, it looks like there is no indentation in the inner scope of the function now.

rzumer: I see, the argument list indentation made it appear as if the body was not indented properly in the diff view.

src/rdo.rs
```diff
@@ -465,97 +465,97 @@ fn luma_chroma_mode_rdo<T: Pixel> (luma_mode: PredictionMode,
...
+  // Find the best chroma prediction mode for the current luma prediction mode
+  let mut chroma_rdo = |skip: bool| {
+    for &chroma_mode in mode_set_chroma.iter() {
+      let wr = &mut WriterCounter::new();
+      let tell = wr.tell_frac();
 
-        if bsize >= BlockSize::BLOCK_8X8 && bsize.is_sqr() {
-          cw.write_partition(wr, tile_bo, PartitionType::PARTITION_NONE, bsize);
-        }
+      if skip { tx_type = TxType::DCT_DCT; };
```
ycho: So, this has been a bug then..., since 'skip' is passed to chroma_Rdo closure.

tdaede:
Yes, plus I changed the type of for loop.

PR #1480, opened 2019-07-22
Move the lookahead in send frame

lu-zero:
In preparation of moving all of it in a separate thread.


YaLTeR:
I believe this renders `_incremental_` tests obsolete as, effectively, now it's always incremental. So they should probably be removed.

tdaede:
please rebase and rerun tests

lu-zero:
Done :)

PR #1481, opened 2019-07-22
Use lookahead to avoid setting short flashes as keyframes

### LINKED ISSUE #1441: Scenecut Tweaks

shssoichiro:
Also fixes the current flow of lookahead so it will always look `LOOKAHEAD_FRAMES` ahead, instead of only at the start of the video.

[AWCY](https://beta.arewecompressedyet.com/?job=master-94b5fa2e61b4e63e79cf456a0db05c55c95b0cc0&job=scenechange%402019-07-22T15%3A10%3A15.475Z) shows significant improvement on one clip, no change on other clips, which seems to match the intent of this change.

Closes #1441

### DIFF
--- a/src/scenechange/mod.rs
+++ b/src/scenechange/mod.rs
@@ -7,69 +7,197 @@
 // Media Patent License 1.0 was not distributed with this source code in the
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
+use crate::api::EncoderConfig;
+use crate::api::InterConfig;
 use crate::frame::Frame;
-use crate::util::{CastFromPrimitive, Pixel};
+use crate::util::CastFromPrimitive;
+use crate::util::Pixel;
 
+use std::cmp;
+use std::collections::BTreeSet;
 use std::sync::Arc;
 
-/// Detects fast cuts using changes in colour and intensity between frames.
-/// Since the difference between frames is used, only fast cuts are detected
-/// with this method. This is probably fine for the purpose of choosing keyframes.
-pub struct SceneChangeDetector<T: Pixel> {
+/// Runs keyframe detection on frames from the lookahead queue.
+pub(crate) struct SceneChangeDetector {
   /// Minimum average difference between YUV deltas that will trigger a scene change.
   threshold: u8,
-  /// Frame number and frame reference of the last frame analyzed
-  last_frame: Option<(usize, Arc<Frame<T>>)>,
+  /// Frames that cannot be marked as keyframes due to the algorithm excluding them.
+  /// Storing the frame numbers allows us to avoid looking back more than one frame.
+  excluded_frames: BTreeSet<u64>,
 }
 
-impl<T: Pixel> Default for SceneChangeDetector<T> {
-  fn default() -> Self {
+impl SceneChangeDetector {
+  pub fn new(bit_depth: u8) -> Self {
+    // This implementation is based on a Python implementation at
+    // https://pyscenedetect.readthedocs.io/en/latest/reference/detection-methods/.
+    // The Python implementation uses HSV values and a threshold of 30. Comparing the
+    // YUV values was sufficient in most cases, and avoided a more costly YUV->RGB->HSV
+    // conversion, but the deltas needed to be scaled down. The deltas for keyframes
+    // in YUV were about 1/3 to 1/2 of what they were in HSV, but non-keyframes were
+    // very unlikely to have a delta greater than 3 in YUV, whereas they may reach into
+    // the double digits in HSV. Therefore, 12 was chosen as a reasonable default threshold.
+    // This may be adjusted later.
+    const BASE_THRESHOLD: u8 = 12;
     Self {
-      // This implementation is based on a Python implementation at
-      // https://pyscenedetect.readthedocs.io/en/latest/reference/detection-methods/.
-      // The Python implementation uses HSV values and a threshold of 30. Comparing the
-      // YUV values was sufficient in most cases, and avoided a more costly YUV->RGB->HSV
-      // conversion, but the deltas needed to be scaled down. The deltas for keyframes
-      // in YUV were about 1/3 to 1/2 of what they were in HSV, but non-keyframes were
-      // very unlikely to have a delta greater than 3 in YUV, whereas they may reach into
-      // the double digits in HSV. Therefore, 12 was chosen as a reasonable default threshold.
-      // This may be adjusted later.
-      threshold: 12,
-      last_frame: None,
+      threshold: BASE_THRESHOLD * bit_depth / 8,
+      excluded_frames: BTreeSet::new(),
     }
   }
-}
 
-impl<T: Pixel> SceneChangeDetector<T> {
-  pub fn new(bit_depth: usize) -> Self {
-    let mut detector = Self::default();
-    detector.threshold = detector.threshold * bit_depth as u8 / 8;
-    detector
-  }
-
-  pub fn set_last_frame(&mut self, ref_frame: Arc<Frame<T>>, frame_num: usize) {
-    self.last_frame = Some((frame_num, ref_frame));
-  }
-
-  pub fn detect_scene_change(&self, curr_frame: &Frame<T>, frame_num: usize) -> bool {
-    match self.last_frame {
-      Some((last_num, ref last_frame)) if last_num == frame_num - 1 => {
-        let len = curr_frame.planes[0].cfg.width * curr_frame.planes[0].cfg.height;
-        let delta_yuv = last_frame.iter().zip(curr_frame.iter())
-          .map(|(last, cur)| (
-            (i16::cast_from(cur.0) - i16::cast_from(last.0)).abs() as u64,
-            (i16::cast_from(cur.1) - i16::cast_from(last.1)).abs() as u64,
-            (i16::cast_from(cur.2) - i16::cast_from(last.2)).abs() as u64
-          )).fold((0, 0, 0), |(ht, st, vt), (h, s, v)| (ht + h, st + s, vt + v));
-        let delta_yuv = (
-          (delta_yuv.0 / len as u64) as u16,
-          (delta_yuv.1 / len as u64) as u16,
-          (delta_yuv.2 / len as u64) as u16
-        );
-        let delta_avg = ((delta_yuv.0 + delta_yuv.1 + delta_yuv.2) / 3) as u8;
-        delta_avg >= self.threshold
+  /// Runs keyframe detection on the next frame in the lookahead queue.
+  ///
+  /// This function requires that a subset of input frames
+  /// is passed to it in order, and that `keyframes` is only
+  /// updated from this method. `input_frameno` should correspond
+  /// to the first frame in `frame_set`.
+  ///
+  /// This will gracefully handle the first frame in the video as well.
+  pub fn analyze_next_frame<T: Pixel>(
+    &mut self,
+    previous_frame: Option<Arc<Frame<T>>>,
+    frame_set: &[Arc<Frame<T>>],
+    input_frameno: u64,
+    config: &EncoderConfig,
+    inter_cfg: &InterConfig,
+    keyframes: &mut BTreeSet<u64>,
+    keyframes_forced: &BTreeSet<u64>,
+  ) {
+    let frame_set = match previous_frame {
+      Some(frame) => [frame].iter().chain(frame_set.iter()).cloned().collect::<Vec<_>>(),
+      None => {
+        // The first frame is always a keyframe.
+        keyframes.insert(0);
+        return;
       }
-      _ => panic!("Incorrect frame order in SceneChangeDetector")
+    };
+
+    self.exclude_scene_flashes(&frame_set, input_frameno, inter_cfg);
+
+    if self.is_key_frame(
+      &frame_set,
+      input_frameno,
+      config,
+      keyframes,
+      keyframes_forced,
+    ) {
+      keyframes.insert(input_frameno);
     }
+  }
+
+  /// Determines if the second frame in `frame_subset` should be a keyframe.
+  ///
+  /// The fact that the first frame in `frame_subset` refers to the previous frame
+  /// is an implementation detail and would make the public API more confusing,
+  /// which is why this method should remain private.
+  fn is_key_frame<T: Pixel>(
+    &self,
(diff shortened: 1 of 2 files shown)


src/api.rs
```diff
@@ -928,10 +929,26 @@ impl<T: Pixel> ContextInner<T> {
...
+      .unwrap_or(0);
+    self.needs_more_frames(lookahead_end)
+      && lookahead_end < last_frame_processed + LOOKAHEAD_FRAMES
+  }
+
+  /// Indicates whether the lookahead queue is ready for frames to be
+  /// analyzed for scene changes.
+  pub(crate) fn lookahead_ready(&self) -> bool {
+    let lookahead_end = self.frame_q.keys().last().cloned().unwrap_or(0);
+    self.next_lookahead_frame + LOOKAHEAD_FRAMES < lookahead_end ||
```
dwbuiten: Unrelated: Do we intend LOOKAHEAD_FRAMES to always be hardcoded?

src/scenechange/mod.rs
```diff
@@ -7,69 +7,202 @@
...
+    inter_cfg: &InterConfig,
+  ) -> bool {
+    if self.excluded_frames.contains(&frameno) {
+      // This frame has already been excluded as a candidate for keyframe selection.
+      return false;
+    }
+
+    // Lookahead to avoid coding short flashes as scenecuts
+    let lookahead_distance = cmp::min(
+      cmp::max(1, inter_cfg.pyramid_depth as usize) + 1,
```
dwbuiten: Is `inter_cfg.pyramid_depth` always correct? I seem to remember some funny business related to our pyramid dpeth stuff. I could be mistaken.

src/scenechange/mod.rs
```diff
@@ -7,69 +7,202 @@
...
+    // If each of BB ... EE are shorter than (lookahead_distance-p0), they are
+    // detected as flashes and not considered scenecuts.
+    // Instead, the first F frame becomes a scenecut.
+    // If the video ends before F, no frame becomes a scenecut.
+    for i in 1..=lookahead_distance {
+      if i < lookahead_distance
+        && self
+          .has_scenecut(&frame_subset[i], &frame_subset[lookahead_distance])
+      {
+        // If cur_p0 is the p0 of a scenecut, it cannot be the p1 of a scenecut.
```
dwbuiten: What do p0 and p1 mean here in rav1e's case?

src/api.rs
lu-zero: We could make it a variable and check what breaks if we set it to 2.

shssoichiro: In the future it will probably be a CLI option. We have access to `self.config` from here, I believe, so that should be fine.

src/scenechange/mod.rs
shssoichiro: That's a good question. In the x264 code it's setting the lookahead distance to the max number of b-frames in the pyramid. This seemed to be the equivalent in rav1e, but does it make sense to use something else here?

src/scenechange/mod.rs
shssoichiro: Good catch, will clarify.

src/scenechange/mod.rs
tdaede: It can only be shorter, not longer. (In particular pyramid_depth never changes, but we sometimes drop frames from the pyramid structure to create shorter subgop lengths)

shssoichiro: Resolved via IRC

src/scenechange/mod.rs
```diff
@@ -7,69 +7,202 @@
...
 use std::sync::Arc;
 
-/// Detects fast cuts using changes in colour and intensity between frames.
-/// Since the difference between frames is used, only fast cuts are detected
-/// with this method. This is probably fine for the purpose of choosing keyframes.
-pub struct SceneChangeDetector<T: Pixel> {
+/// Runs keyframe detection on frames from the lookahead queue.
+pub(crate) struct SceneChangeDetector {
   /// Minimum average difference between YUV deltas that will trigger a scene change.
+  /// of parameters.
```
YaLTeR: This line doesn't seem to belong here?

shssoichiro:
Do not merge yet, this seems to have some conflict with the lookahead RDO changes (is my assumption): https://beta.arewecompressedyet.com/?job=master-e56f405a6c64d651598c9a3b087d1decd704d81c&job=scenechange-lookahead%402019-07-24T17%3A08%3A15.457Z

src/scenechange/mod.rs
```diff
@@ -7,69 +7,202 @@
...
+    config: &EncoderConfig,
+    inter_cfg: &InterConfig,
+    keyframes: &mut BTreeSet<u64>,
+  ) {
+    let frame_set = match previous_frame {
+      Some(frame) => [frame].iter().chain(frame_set.iter()).cloned().collect(),
+      None => {
+        // The first frame is always a keyframe.
+        keyframes.insert(0);
+        input_frameno += 1;
```
YaLTeR: Shouldn't this just return? It seems to me that currently the second frame is analyzed twice (first time through here and second time normally).

src/api.rs
```diff
@@ -2977,4 +2984,51 @@ mod test {
       }
     );
   }
+
+  #[interpolate_test(1, 1)]
+  #[interpolate_test(2, 2)]
+  #[interpolate_test(3, 3)]
+  fn output_frameno_no_scene_change_at_flashes(flash_at: u64) {
```
YaLTeR: Could you also add a test for the second behavior (AAAAABBCCDDEEFFFFFF)? And it would be super nice to have these tests additionally check the boundary conditions for the flash detection (so the flash that spans the maximum number of frames for it to still be considered a flash, and then a flash that spans one more frame so it is considered a legit scene cut).

src/api.rs
```diff
@@ -1347,45 +1369,49 @@ impl<T: Pixel> ContextInner<T> {
...
+        .iter()
+        .filter_map(|(&input_frameno, frame)| {
+          if input_frameno >= self.next_lookahead_frame {
+            frame.clone()
+          } else {
+            None
+          }
+        })
+        .take(LOOKAHEAD_FRAMES as usize + 1)
+        .collect::<Vec<_>>();
```
YaLTeR: This is recomputing all lookahead frames to process every loop iteration, it seems that only the first one is needed.

YaLTeR: Ah, no, you're passing them to the keyframe detector... Maybe it's possible to compute them once, then, before the `while self.lookahead_ready()` loop?

src/scenechange/mod.rs
shssoichiro: The reason this doesn't return here is because it still needs to call `self.is_key_frame` to see if any of the frames after it can be excluded as flashes. But I'm realizing the names of these functions make it really unclear as to what is going on. I'll rename them to see if that makes things clearer.

shssoichiro: Actually, maybe you are right that it should just return...

lu-zero:
Is the PR in working condition now?

src/api.rs
```diff
@@ -3031,4 +3025,178 @@ mod test {
...
+        (2, true), // B0-frame
+        (1, true), // B1-frame (first)
+        (2, true), // B0-frame (show existing)
+        (3, true), // B1-frame (second)
+        (4, true), // P-frame (show existing)
+        (4, false), // invalid
+        (6, true), // B0-frame
+        (5, true), // B1-frame (first)
+        (6, true), // B0-frame (show existing)
+        (7, true), // P-frame
```
YaLTeR: That would be B1-frame (second).

src/api.rs
```diff
@@ -3031,4 +3025,178 @@ mod test {
...
+      .map(|fi| (fi.input_frameno, !fi.invalid))
+      .collect::<Vec<_>>();
+
+    assert_eq!(
+      &data[..],
+      &[
+        (0, true), // I-frame
+        (0, false), // invalid
+        (0, false), // invalid
+        (1, true), // P-frame
```
YaLTeR: B1-frame (first)

src/api.rs
```diff
@@ -3031,4 +3025,178 @@ mod test {
       ]
     );
   }
+
+  #[test]
+  fn output_frameno_no_scene_change_at_max_len_flash() {
+    // Test output_frameno configurations when there's a flash at the
+    // <flash_at>th frame.
```
YaLTeR: The comments need an update?

src/scenechange/mod.rs
```diff
@@ -114,30 +111,22 @@ impl SceneChangeDetector {
       return false;
     }
 
-    self.detect_scene_change(frame_subset, frameno, inter_cfg)
+    let excluded = self.excluded_frames.contains(&frameno);
+    !excluded && self.has_scenecut(&frame_subset[0], &frame_subset[1])
```
YaLTeR: I'd write this as
```rust
if (self.excluded_frames.contains(&frameno)) {
  return false;
}

self.has_scenecut(&frame_subset[0], &frame_subset[1])
```

src/scenechange/mod.rs
```diff
@@ -57,26 +56,26 @@ impl SceneChangeDetector {
...
       None => {
         // The first frame is always a keyframe.
         keyframes.insert(0);
-        input_frameno += 1;
-        frame_set.to_owned()
+        return;
       }
     };
 
+    self.exclude_scene_flashes(&frame_set, input_frameno, inter_cfg);
```
YaLTeR: Correct me if I'm wrong, but I think right now if the max GOP size happens to land exactly in the middle of a flash, then the middle of a flash would get an I-frame, and then the end of the flash wouldn't? Do we want any special logic for this, e.g. inserting an I-frame right before the flash in this case (what happens if that's below min GOP size though)?

shssoichiro: I believe you're correct. I did not see any special code for that in x264 either (where this optimization is adapted from), but I think it would be valuable to add it. Worth doing in this PR, or making a separate issue to track it?

src/api.rs
shssoichiro: It might be possible. But I think this is not likely to incur a large performance penalty, and I think the code is cleaner this way. The `clone` initially throws red flags, but it's only cloning an `Arc` pointer, which only means incrementing the ref count.

shssoichiro:
This should be ready, except for the open question on what to do at the GOP limit. Re-running AWCY to be sure. Tested locally and found slight improvement in PSNR--the RDO lookahead now is more consistent in how many frames it looks ahead, which is probably the cause of that.

shssoichiro:
Okay, that fix did not completely work, and will result in your computer running out of memory in the middle of long clips. Another change incoming soon.

src/api.rs
YaLTeR: Well, I'm mainly worried about the `.collect::<Vec<_>>()`

shssoichiro: It would require a counter variable, but I think that's more performant than collecting each time, you're right. I'll push that.

shssoichiro:
PR should be ready now. Running AWCY again.

shssoichiro:
[AWCY](https://beta.arewecompressedyet.com/?job=master-e7a0d6ec60e0086c0775f14fb1f6f8462f4d0b29&job=scenechange-lookahead%402019-07-26T03%3A21%3A39.166Z)

src/api.rs
```diff
@@ -1079,6 +1101,10 @@ impl<T: Pixel> ContextInner<T> {
       output_frameno_in_gop,
       self.gop_input_frameno_start[&output_frameno]
     );
+    let frames_needed = input_frameno + self.inter_cfg.keyframe_lookahead_distance();
+    if frames_needed > self.next_lookahead_frame && self.needs_more_frames(frames_needed) {
```
lu-zero: What happens when we are in flushing state?

src/api.rs
```diff
@@ -1407,45 +1433,53 @@ impl<T: Pixel> ContextInner<T> {
...
+          }
+        })
+        .take(self.inter_cfg.keyframe_lookahead_distance() as usize + 1)
+        .collect::<Vec<_>>();
+    let mut lookahead_idx = 0;
+
+    while self.lookahead_ready() {
+      // Process the next unprocessed frame
+      // Start by getting that frame and all frames after it in the queue
+      let lookahead_frames = &lookahead_frames[lookahead_idx..];
```
lu-zero: This looks suspicious and confusing. Please do not reslice `lookahead_frames` this way.

src/api.rs
```diff
@@ -1710,12 +1745,23 @@ impl<T: Pixel> ContextInner<T> {
...
-    if !self.needs_more_frames(
-      self.frame_invariants[&self.output_frameno].input_frameno
-    ) {
+    let input_frameno = self.frame_invariants[&self.output_frameno].input_frameno;
+    if !self.needs_more_frames(input_frameno) {
       return Err(EncoderStatus::LimitReached);
     }
 
+    // Block importance computation needs to have FI's already processed `LOOKAHEAD_FRAMES` frames into the future.
+    let ready_frames = self
```
lu-zero: The iterator seems the same as the one used in line 1494, please factor it out.

src/api.rs
shssoichiro: @YaLTeR had asked me to make this change for performance reasons, instead of running over the iterator inside the loop. I agree it would be clearer what is happening without this reslice.

src/api.rs
shssoichiro: The borrow checker doesn't like that, because it's referencing `self.output_frameno` within the iterator. I could make it its own function if I make it collect into a `Vec`, but that would be less performant, because in one of the usages we only `count` the items in the iterator without needing to collect.

src/api.rs
lu-zero: Changing the local variable name would be enough to avoid the confusion, possibly we can rework the loop later.

src/api.rs
shssoichiro: This works as if we reached the end of the video, because the `limit` gets set at the start of flushing.

src/scenechange/mod.rs
lu-zero: I'd consider this later, but we really should document somewhere how this code works.

PR #1482, opened 2019-07-22
Reverse skip/no-skip search for chroma mode.

tdaede:
This makes skip win ties over no-skip.

https://beta.arewecompressedyet.com/?job=master-e82003643aa440c02d3c6669cd49b529567ef277&job=chroma_reverse_order%402019-07-22T01%3A47%3A48.466Z

master-e82003643aa440c02d3c6669cd49b529567ef277 -> chroma_reverse_order@2019-07-22T01:47:48.466Z

-0.1993 | -0.1852 | -0.3532 |  -0.1811 | -0.1777 | -0.1736 |    -0.1683

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -465,9 +465,6 @@ fn luma_chroma_mode_rdo<T: Pixel> (luma_mode: PredictionMode,
   luma_mode_is_intra: bool,
   mode_context: usize,
   mv_stack: &ArrayVec<[CandidateMV; 9]>) {
-  let (tx_size, mut tx_type) = rdo_tx_size_type(
-    fi, ts, cw, bsize, tile_bo, luma_mode, ref_frames, mvs, false,
-  );
 
   let PlaneConfig { xdec, ydec, .. } = ts.input.planes[1].cfg;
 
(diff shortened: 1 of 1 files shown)


ycho:
I am trying to figure out where the 0.2% gain comes from. We see that rdo_tx_size_type() is now called for each chroma mode being tested (with given luma mode), instead of each luma mode only. We wanted the function called for each luma mode only since we believed that only luma pixels/info are required to decide tx_type at that time.

tdaede:
I tried moving *just* rdo_tx_size_type(), but there is no awcy difference: https://beta.arewecompressedyet.com/?job=master-e82003643aa440c02d3c6669cd49b529567ef277&job=just-tx-rdo-fix%402019-07-22T19%3A35%3A08.035Z

tdaede:
If I don't move it, then I get an assertion failure because tx_size is not set correctly for the skip case. I think with the other order, tx_size got set correctly at the end of the skip=false search by chance.

tdaede:
Actually, with this change I can probably remove the if skip { tx_type = DCT } line because rdo_tx_size_type already sets that.

ycho:
> Actually, with this change I can probably remove the if skip { tx_type = DCT } line because rdo_tx_size_type already sets that.

yes, that is good.

tdaede:
updated

PR #1483, opened 2019-07-22
Change ilog() return type to usize

rzumer:
Thanks to @tterribe for pointing out the usually unnecessary conversion.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -39,10 +39,8 @@ use arg_enum_proc_macro::ArgEnum;
 use bitstream_io::{BitWriter, BigEndian};
 use bincode::{serialize, deserialize};
 use rayon::iter::*;
-use std;
 use std::{fmt, io, mem};
-use std::io::Write;
-use std::io::Read;
+use std::io::{Read, Write};
 use std::sync::Arc;
 use std::fs::File;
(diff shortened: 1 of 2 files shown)


src/encoder.rs
```diff
@@ -2095,7 +2094,7 @@ fn encode_tile_group<T: Pixel>(fi: &FrameInvariants<T>, fs: &mut FrameState<T>)
   fs.cdfs = cdfs[idx_max];
   fs.cdfs.reset_counts();
 
-  let max_tile_size_bytes = ((max_len as u32).ilog() + 7) / 8;
+  let max_tile_size_bytes: u32 = ((max_len.ilog() + 7) / 8).try_into().unwrap();
```
lu-zero: ilog() usize -> u32 must never fail, you might use the `utils.rs` converter instead.

rzumer: Do you mean we should silently downcast and clamp the value? That seems like it would only create errors further down the line if it ever happened.

lu-zero: The maximum value of that expression fits in `u8`.

rzumer: I guess your point is that the `try_into().unwrap()` is inefficient compared to the direct cast. I will change it.

PR #1485, opened 2019-07-22
Initial temporal RDO implementation

YaLTeR:
This PR adds original-frame motion vector computation in lookahead, and block importance computation in lookahead, which is heavily based on x264's MBTree, as described in [this paper](https://pdfs.semanticscholar.org/032f/1ab7d9db385780a02eb2d579af8303b266d2.pdf). Additionally, the distortion is biased in the RD-cost computation according to the computed block importance. [Three-way AWCY](https://beta.arewecompressedyet.com/?job=master-5de2d05cbdc78af88402ee3d7b74261cb11c90ad&job=lookahead-mv-3-16-2019-07-23_131331-dbb9d2f&job=lookahead-mv-3-16-INVERTED-2019-07-23_131446-8df6d94) (original; this PR; this PR with inverted bias to show that it works).

The block importance computation is far from final. For instance, right now it only uses half-res MVs and a single intra prediction mode.

Two new Python scripts are included to visualize the computed MVs and block importances respectively, which are dumped when the corresponding flag is set in `compute_lookahead_data()`.

Things that might be desirable to do with this PR before merging:
- Add tests for the computed MVs and block importance values. An actual real clip with, say, two subgops worth of frames would be very valuable for this, but I guess it can be done on random pixels with a set seed?
- Play with the constants (`4` and `0.8` in the RD-cost biasing, lookahead frame queue size, etc.). However it's not really worth it to spend too much time on this until the block importance computation is closer to being finalized, and until we can affect the actual quantizers.
- ~~Hide `WRITE_DEBUG_FILES` behind a feature or something because it pulls two extra dependencies.~~
- Make a speed setting to enable or disable the block importance computation because it could take a while, especially later on when full MVs are employed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,13 @@ quick_test = []
 desync_finder = []
 bench = []
 
+# Enables debug dumping of lookahead computation results, specifically:
+# - i-qres.png: quarter-resolution luma planes,
+# - i-hres.png: half-resolution luma planes,
+# - i-mvs.bin: motion vectors,
+# - i-imps.bin: block importances.
+dump_lookahead_data = ["byteorder", "image"]
+
 [dependencies]
 arg_enum_proc_macro = "0.1.1"
 bitstream-io = "0.8"
@@ -48,6 +55,8 @@ bincode = "1.1"
 arrayvec = "0.4.10"
 better-panic = "0.1"
 err-derive = "0.1"
+image = { version = "0.21.2", optional = true }
+byteorder = { version = "1.3.2", optional = true }
 
 [build-dependencies]
 nasm-rs = { version = "0.1", path = "crates/nasm_rs/", optional = true }
--- a/src/api.rs
+++ b/src/api.rs
@@ -1192,7 +1192,8 @@ impl<T: Pixel> ContextInner<T> {
           cdfs: fs.cdfs,
           // TODO: can we set MVs here? We can probably even compute these MVs
           // right now instead of in encode_tile?
-          frame_mvs: fs.frame_mvs
+          frame_mvs: fs.frame_mvs,
+          output_frameno: self.next_lookahead_output_frameno - 1
         });
         for i in 0..(REF_FRAMES as usize) {
           if (fi.refresh_frame_flags & (1 << i)) != 0 {
@@ -1280,7 +1281,8 @@ impl<T: Pixel> ContextInner<T> {
         input_hres: fs.input_hres,
         input_qres: fs.input_qres,
         cdfs: fs.cdfs,
-        frame_mvs: fs.frame_mvs
+        frame_mvs: fs.frame_mvs,
+        output_frameno: self.next_lookahead_output_frameno - 1
       });
       for i in 0..(REF_FRAMES as usize) {
         if (fi.refresh_frame_flags & (1 << i)) != 0 {
@@ -1401,7 +1403,7 @@ impl<T: Pixel> ContextInner<T> {
           // TODO avoid the clone by having rec Arc.
           let rec = if fi.show_frame { Some(fs.rec.clone()) } else { None };
 
-          update_rec_buffer(fi, fs);
+          update_rec_buffer(self.output_frameno, fi, fs);
 
           // Copy persistent fields into subsequent FrameInvariants.
           let rec_buffer = fi.rec_buffer.clone();
--- a/src/api.rs
+++ b/src/api.rs
@@ -1300,9 +1300,16 @@ impl<T: Pixel> ContextInner<T> {
         self.next_lookahead_output_frameno - 1
       );
     }
+  }
 
-    // Compute the block importances for all frames down to the current output
-    // frame. First, initialize them all with zeros.
+  /// Computes the block importances for the current output frame.
+  fn compute_block_importances(&mut self) {
+    // SEF don't need block importances.
+    if self.frame_invariants[&self.output_frameno].show_existing_frame {
+      return;
+    }
+
+    // Get a list of output_framenos that we want to propagate through.
     let output_framenos = self
       .frame_invariants
       .iter()
@@ -1311,6 +1318,10 @@ impl<T: Pixel> ContextInner<T> {
       .map(|(&output_frameno, _)| output_frameno)
       .collect::<Vec<_>>();
 
+    // The first one should be the current output frame.
+    assert_eq!(output_framenos[0], self.output_frameno);
+
+    // First, initialize them all with zeros.
     for output_frameno in output_framenos.iter() {
       let fi = self.frame_invariants.get_mut(output_frameno).unwrap();
       for x in fi.block_importances.iter_mut() {
@@ -1672,6 +1683,9 @@ impl<T: Pixel> ContextInner<T> {
     ) {
       return Err(EncoderStatus::LimitReached);
     }
+
+    // Compute the block importances for the current output frame.
+    self.compute_block_importances();
 
     let cur_output_frameno = self.output_frameno;
 
--- a/src/api.rs
+++ b/src/api.rs
@@ -8,22 +8,27 @@
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
 use arg_enum_proc_macro::ArgEnum;
+use arrayvec::ArrayVec;
 use bitstream_io::*;
 use num_derive::*;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use serde_derive::{Deserialize, Serialize};
 
 use crate::context::{FrameBlocks, SuperBlockOffset};
+use crate::dist::get_satd;
 use crate::encoder::*;
-use crate::frame::Frame;
+use crate::frame::{AsRegion, Frame, PlaneOffset};
 use crate::metrics::calculate_frame_psnr;
 use crate::partition::*;
+use crate::predict::PredictionMode;
 use crate::rate::RCState;
 use crate::rate::FRAME_NSUBTYPES;
 use crate::rate::FRAME_SUBTYPE_I;
 use crate::rate::FRAME_SUBTYPE_P;
 use crate::rate::FRAME_SUBTYPE_SEF;
 use crate::scenechange::SceneChangeDetector;
+use crate::tiling::{Area, TileRect};
+use crate::transform::TxSize;
 use crate::util::Pixel;
 
 use std::collections::BTreeMap;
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -46,6 +46,8 @@ avformat-sys = { version = "0.1", path = "crates/avformat-sys/", optional = true
 rayon = "1.0"
 bincode = "1.1"
 arrayvec = "0.4.10"
+image = "0.21.2"
```
lu-zero: Make it feature-optional please.

src/api.rs
```diff
@@ -1101,6 +1108,13 @@ impl<T: Pixel> ContextInner<T> {
   }
 
   fn compute_lookahead_data(&mut self) {
+    /// Enable to make this code output:
+    /// - i-qres.png: quarter-resolution luma planes,
+    /// - i-hres.png: half-resolution luma planes,
+    /// - i-mvs.bin: motion vectors,
+    /// - i-imps.bin: block importances.
+    const WRITE_DEBUG_FILES: bool = false;
```
lu-zero: make it a `#[cfg(feature=dump_image)]` or such.

Cargo.toml
```diff
@@ -46,6 +46,8 @@ avformat-sys = { version = "0.1", path = "crates/avformat-sys/", optional = true
...
+byteorder = "1.3.2"
```
tdaede: For this dep, we ended up just using bitstream-io everywhere, which is basically a superset of what byteorder offers.

Cargo.toml
tdaede: Yeah you definitely want this one behind a feature.

src/api.rs
```diff
@@ -1133,6 +1147,503 @@ impl<T: Pixel> ContextInner<T> {
...
+
+      // TODO: some of this work, like downsampling, could be reused in the
+      // actual encoding.
+      let mut fs = FrameState::new_with_frame(fi, frame.clone());
+      fs.input_hres.downsample_from(&frame.planes[0]);
+      fs.input_hres.pad(fi.width, fi.height);
+      fs.input_qres.downsample_from(&fs.input_hres);
+      fs.input_qres.pad(fi.width, fi.height);
+
+      if WRITE_DEBUG_FILES {
```
lu-zero: as above.

src/api.rs
```diff
@@ -1133,6 +1147,503 @@ impl<T: Pixel> ContextInner<T> {
...
+              let tile_sbo = SuperBlockOffset { x: sbx, y: sby };
+              build_half_res_pmvs(fi, ts, tile_sbo, &tile_pmvs);
+            }
+          }
+        });
+
+      // Save the motion vectors to FrameInvariants.
+      fi.mvs = fs.frame_mvs.clone();
+
+      if WRITE_DEBUG_FILES {
```
lu-zero: same.

src/context.rs
```diff
@@ -1152,7 +1152,7 @@ mod test {
   }
 }
 
-const SUPERBLOCK_TO_PLANE_SHIFT: usize = MAX_SB_SIZE_LOG2;
+pub const SUPERBLOCK_TO_PLANE_SHIFT: usize = MAX_SB_SIZE_LOG2;
```
lu-zero: `pub(crate)` just in case.

src/rdo.rs
```diff
@@ -355,9 +355,48 @@ fn compute_tx_distortion<T: Pixel>(
   distortion
 }
 
-pub fn compute_rd_cost<T: Pixel>(fi: &FrameInvariants<T>, rate: u32, distortion: u64) -> f64 {
+pub fn compute_rd_cost_biased_by_importance<T: Pixel>(
```
tdaede: You only ever use this version, right? Why not just call it compute_rd_cost?

src/rdo.rs
```diff
@@ -355,9 +355,48 @@ fn compute_tx_distortion<T: Pixel>(
...
+  fi: &FrameInvariants<T>, rect: Rect, rate: u32, distortion: u64
+) -> f64 {
+  assert!(rect.x >= 0);
```
tdaede: This function is hot (called multiple times per block) so this needs to be debug_assert!().

src/rdo.rs
```diff
@@ -355,9 +355,48 @@ fn compute_tx_distortion<T: Pixel>(
...
+  assert!(rect.y >= 0);
+  assert!(rect.x % 4 == 0);
+  assert!(rect.y % 4 == 0);
+  assert!(rect.width % 4 == 0);
+  assert!(rect.height % 4 == 0);
+
+  // eprintln!("{:?}", rect);
+  // eprintln!("{}×{}", fi.w_in_b, fi.h_in_b);
```
tdaede: Drop debugging printfs.

src/encoder.rs
```diff
@@ -487,6 +488,19 @@ pub struct FrameInvariants<T: Pixel> {
   /// should be ignored. Invalid frames occur when a subgop is prematurely
   /// ended, for example, by a key frame or the end of the video.
   pub invalid: bool,
+  /// Motion vectors to the _original_ reference frames (not reconstructed).
+  /// Used for lookahead purposes.
+  pub mvs: Vec<FrameMotionVectors>,
```
lu-zero: I'd make it a separate struct. FrameInvariant will be split sooner or later and having already self-contained struct inside will make our life easier.

src/rdo.rs
```diff
@@ -355,9 +355,48 @@ fn compute_tx_distortion<T: Pixel>(
...
+  let x1 = rect.x as usize / 4;
```
tdaede: Pass a BlockOffset / BlockSize in instead so you don't need to divide?

src/rdo.rs
```diff
@@ -355,9 +355,48 @@ fn compute_tx_distortion<T: Pixel>(
...
+  let mut mean_importance = 0.;
+  let mut count = 0;
+  for y in y1..y2 {
+    for x in x1..x2 {
+      mean_importance += fi.block_importances[y * fi.w_in_b + x];
+      count += 1;
+    }
+  }
+  mean_importance /= count as f32;
+  // eprintln!("{}", mean_importance);
```
tdaede: drop debug printf

src/encoder.rs
```diff
@@ -2155,64 +2252,7 @@ fn encode_tile<'a, T: Pixel>(
...
-                save_block_motion(ts, BlockSize::BLOCK_32X32, tile_sbo.block_offset(0, 8), i, mv);
-              }
-              if let Some(mv) = pmvs[4][r] {
-                save_block_motion(ts, BlockSize::BLOCK_32X32, tile_sbo.block_offset(8, 8), i, mv);
-              }
-            }
-          }
-        }
-      }
+      let mut pmvs = build_half_res_pmvs(fi, ts, tile_sbo, &tile_pmvs);
```
lu-zero: This is something I wanted to do since a while, thank you :)

src/rdo.rs
```diff
@@ -355,9 +355,48 @@ fn compute_tx_distortion<T: Pixel>(
...
+
+  let bias = (mean_importance / 4.) as f64 + 0.8;
```
tdaede: comment describing where these numbers come from, please

src/rdo.rs
```diff
@@ -1037,7 +1082,10 @@ pub fn rdo_tx_type_decision<T: Pixel>(
         true
       )
     };
-    let rd = compute_rd_cost(fi, rate, distortion);
+    let PlaneOffset { x, y } = ts.to_frame_block_offset(tile_bo).to_luma_plane_offset();
```
lu-zero: those two lines should be factored in a method in ts.

src/encoder.rs
tdaede: This variable name is confusing, even with the comments. Consider we have ReferenceFrame::frame_mvs and FrameInvariants::mvs. lookahead_mvs or something would be better, even if more verbose.

src/encoder.rs
```diff
@@ -612,6 +626,15 @@ impl<T: Pixel> FrameInvariants<T> {
       tx_mode_select : false,
       default_filter: FilterMode::REGULAR,
       invalid: false,
+      mvs: {
+        let mut vec = Vec::with_capacity(REF_FRAMES);
```
tdaede: Any reason this is a vec rather than a boxed slice like block_importances?

src/rdo.rs
```diff
@@ -355,9 +355,48 @@ fn compute_tx_distortion<T: Pixel>(
...
+
+  let mut mean_importance = 0.;
+  let mut count = 0;
+  for y in y1..y2 {
+    for x in x1..x2 {
+      mean_importance += fi.block_importances[y * fi.w_in_b + x];
+      count += 1;
+    }
+  }
+  mean_importance /= count as f32;
```
tdaede: Divides are expensive. Take advantage of the fact that block areas are always a power of 2. (see BlockSize::width_log2 and height_log2)

src/api.rs
YaLTeR: Leave the `const` or have enabling the feature dump the files unconditionally? I think I prefer the latter.

src/context.rs
YaLTeR: All the other stuff there is `pub` rather than `pub(crate)`. I think someone needs to make a separate PR to change all unnecessary `pub` to `pub(crate)`.

src/encoder.rs
YaLTeR: Which fields do you want to split into a separate struct?

src/context.rs
tmatth: > All the other stuff there is `pub` rather than `pub(crate)`. I think someone needs to make a separate PR to change all unnecessary `pub` to `pub(crate)`.

For existing code, such a PR makes sense, but for newly added code we should start with limited scope from the outset.

src/api.rs
lu-zero: You can use an `env` var to enable/disable dumping at runtime, but all behind a feature guard please :)

src/encoder.rs
lu-zero: All that you are adding I think, you could fit them in a `FrameLookahead` struct maybe?

src/api.rs
tdaede: Unconditional is fine otherwise I think.

src/encoder.rs
tdaede: Well, in this case they *are* FrameInvariant, as in once the lookahead is finished they are baked in place, so it's not so bad right now.

YaLTeR: It would only make sense to split off `mvs` and `lookahead_rec_buffer` as being completely lookahead-specific. However, `mvs` could later, for example, get used as candidates for the real MV search, thus making them not lookahead-specific any more. So idk.

lu-zero: I guess you can land now as-is and we can refine later.

src/rdo.rs
YaLTeR: It would be easier if I didn't have to `min()` the end coordinates against `w_in_b` and `h_in_b`, respectively. Not doing that leads to asserts, for instance in the flushing tests.

src/rdo.rs
YaLTeR: So if I remove the original `compute_rd_cost`, call this function `compute_rd_cost`, then what should I call `compute_rd_cost_biased`?

tdaede: compute_rd_cost_biased could just be merged in, right? After all you're only ever using the biased version.

src/rdo.rs
tdaede: Ah right, I think although partitioning is forced on the edges, it's allowed to be split vertically / horizontally, which can allow quite a few blocks out of the frame. One option here is to add additional padding to your importance structure so you can drop the checks.

tdaede: Also here I think you want "count" to cover the whole area that was requested, not just the available area.

src/rdo.rs
YaLTeR: Yeah. I suppose I'll just do that.

src/rdo.rs
YaLTeR: So the out-of-bounds blocks would count as zero importance?

tdaede: Or whatever your "default" importance is (a block not referenced by any other frame)

YaLTeR: Right now I compute importance in `f32`s, I think I would need to compute it in `u64`s or `u32`s like x264 does to replace this with a shift.

lu-zero: It would be optimal.

tdaede: OK. Let's wait for a future patch to replace it with a shift. Then we can show there's no quality loss for doing so.

src/api.rs
```diff
@@ -1133,6 +1140,506 @@ impl<T: Pixel> ContextInner<T> {
     while self.set_frame_properties(self.next_lookahead_output_frameno).is_ok()
     {
       self.next_lookahead_output_frameno += 1;
+
+      // Compute the lookahead motion vectors.
```
shssoichiro: I'm concerned by the size of this function now. Is it possible to break this into smaller functions?

tdaede:
needs rebase

ycho:
@tdaede and @lu-zero , Hey- I wonder if you two reviewers had chance to watch actual decoded videos, in addition to metric changes in awcy?

lu-zero:
I reviewed only the code part, did you see something strange in the videos?

ycho:
> I reviewed only the code part, did you see something strange in the videos?

No, but that is why, when we review the patch that can particularly affect the subjective qualities, I hope us to check the decoded video frames as well. In order to to avoid any unwalnted artifacts. Temporal RDO is proven to give a gain so less change to give unexpected artifacts in motion but what I stressed in the Tuesday, Oct 8, weekly meeting was focused on activity masking. Because it is devised to help subjective quality not a PSNR. @shreevari 

I had chance to see just few clips using awcy's AB comparisons and was able to observe that the quality on stationary area like road signs in 
[Netflix_Driving_Pov](https://beta.arewecompressedyet.com/analyzer/?maxFrames=4&decoder=https://people.xiph.org/~mbebenita/analyzer/inspect.js&decoderName=master-5de2d05cbdc78af88402ee3d7b74261cb11c90ad&file=https://beta.arewecompressedyet.com/runs/master-5de2d05cbdc78af88402ee3d7b74261cb11c90ad/objective-1-fast/Netflix_DrivingPOV_1280x720_60fps_8bit_420_60f.y4m-172.ivf&decoder=https://people.xiph.org/~mbebenita/analyzer/inspect.js&decoderName=lookahead-mv-3-16-2019-07-23_131331-dbb9d2f&file=https://beta.arewecompressedyet.com/runs/lookahead-mv-3-16-2019-07-23_131331-dbb9d2f/objective-1-fast/Netflix_DrivingPOV_1280x720_60fps_8bit_420_60f.y4m-172.ivf) can be claimed better. Playing them and toggle also can show us the differences when frames are in motion.

PR #1488, opened 2019-07-23
Cli error reporting

lu-zero:
Make panic traces nicer and setup the infrastructure to report errors in a nicer way.

### DIFF
--- a/src/bin/rav1e.rs
+++ b/src/bin/rav1e.rs
@@ -193,6 +193,8 @@ fn do_encode<T: Pixel, D: Decoder>(
 }
 
 fn main() {
+  better_panic::install();
+
   let mut cli = parse_cli();
   let mut y4m_dec = y4m::decode(&mut cli.io.input).expect("input is not a y4m file");
   let video_info = y4m_dec.get_video_details();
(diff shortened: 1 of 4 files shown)


src/bin/rav1e.rs
```diff
@@ -139,17 +145,17 @@ fn process_frame<T: Pixel, D: Decoder>(
...
     }
-    Err(EncoderStatus::Failure) => {
-      panic!("Failed to encode video");
+    e @ Err(EncoderStatus::Failure) => {
+      let _ = e.map_err(|e| e.context("Failed to encode video"))?;
     }
-    Err(EncoderStatus::NotReady) => {
-      panic!("Mis-managed handling of two-pass stats data");
+    e @ Err(EncoderStatus::NotReady) => {
+      let _ = e.map_err(|e| e.context("Mis-managed handling of two-pass stats data"))?;
```
rzumer: No need for the hyphen in "mismanaged".

lu-zero: I can conflate the fix in this patch if you want.

lu-zero: Done.

Issue #1399, opened 2019-07-02
Disabling CDEF causes a > 80% regression on metrics


tmatth:
Using this patch:
```
diff --git a/src/encoder.rs b/src/encoder.rs
index 8734473..4f3f130 100644
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -213,7 +213,7 @@ impl Sequence {
       enable_ref_frame_mvs: false,
       enable_warped_motion: false,
       enable_superres: false,
-      enable_cdef: config.speed_settings.cdef && config.chroma_sampling != ChromaSampling::Cs422,
+      enable_cdef: false,
       enable_restoration: config.chroma_sampling != ChromaSampling::Cs422 &&
         config.chroma_sampling != ChromaSampling::Cs444, // FIXME: not working yet
       operating_points_cnt_minus_1: 0,
```

leads to this mess:

https://beta.arewecompressedyet.com/?job=master-f9ad26f55ebd5e3a827213b18f5b2534550ead39&job=master-no-cdef%402019-06-28T13%3A12%3A08.894Z

This probably points to a bug in our CDEF disabled codepath.

tdaede:
Did you run cargo test --features=decode_test?

tmatth:
```
running 85 tests
test api::test::flush_unlimited_reorder_scene_change_detection ... ok
test cdef::test::check_max_element ... ok
test cdef::test::test_padded_frame_copy ... ok
test cdef::test::test_padded_frame_copy_outside_input ... ok
test context::test::cdf_map ... ok
test context::test::cfl_joint_sign ... ok
test ec::test::booleans ... ok
test ec::test::cdf ... ok
test ec::test::mixed ... ok
test ec::test::update_cdf_4_sse2 ... ok
test encoder::test::check_partition_types_order ... ok
test frame::plane::test::copy_from_raw_u8 ... ok
test frame::plane::test::test_plane_pad ... ok
test api::test::flush_low_latency_scene_change_detection ... ok
test me::test::get_sad_same_u16 ... ok
test predict::test::pred_matches_u8 ... ok
test predict::test::pred_max ... ok
test quantize::test::gen_divu_table ... ok
test me::test::get_sad_same_u8 ... ok
test quantize::test::test_tx_log_scale ... ok
test rate::test::bexp64_vectors ... ok
test api::test::flush_unlimited_low_latency_no_scene_change ... ok
test rate::test::blog64_vectors ... ok
[...]

PR #1490, opened 2019-07-24
Disable self-guided filter when CDEF is disabled

rzumer:
This leads to about -73% to BD rate when CDEF is not used (at `-s 10`).

This should be reverted when #1399 is fixed.


rzumer:
Will merge this now since 4:4:4 now uses restoration (and no CDEF for inter frames IIRC).

The AWCY run validating this issue is [here](https://beta.arewecompressedyet.com/?job=master-no-cdef%402019-07-23T19%3A16%3A53.957Z&job=master-no-cdef-lrf-test%402019-07-24T02%3A24%3A35.622Z).

PR #1492, opened 2019-07-24
Enable loop restoration filter for 4:4:4 input

rzumer:
I re-tested it and it works with `nyan.y4m` converted to 4:4:4. If decoding tests succeed in CI I assume it is working. There are still desyncs when enabling it with 4:2:2, including out-of-bounds memory access in the self-guided filter and further desyncs causing decoding errors.


lu-zero:
The CI survive, we can land it :)

PR #1495, opened 2019-07-24
Split the api.rs file in multiple files

lu-zero:
And document the struct and enums while at it.

- [x] Color
- [ ] Config
- [ ] Lookahead


rzumer:
I think it is better not to punctuate doc comments for enum members.

lu-zero:
> The added comments seem fine to me, but I think it is better not to punctuate doc comments for enum members.

That's how it is documented commonly, so we should probably stay with the punctuation.

Issue #1000, opened 2019-02-14
Use reduced_still_picture_header when still_picture flag is set


tdaede:

rzumer:
#1498 implements necessary header changes, but we need to implement `disable_frame_end_update_cdf`.

cryptomilk:
I've just enabled still_picture in [libavif](https://github.com/AOMediaCodec/libavif/) and the encoded picture then can't be decoded by dav1d.

barrbrain:
From [AV1 Image File Format (AVIF), 2.1 AV1 Image Item](https://aomediacodec.github.io/av1-avif/#image-item):
* The AV1 Image Item Data should have its `still_picture flag` set to 1.
* The AV1 Image Item Data should have its `reduced_still_picture_header` flag set to 1.

This issue impacts producing conformant sequences for libavif.

cryptomilk:
We have added support for importing and exporting AVIF files in [darktable](https://darktable.org) which will be released with version 3.0.1. It would be great to have this fixed in rav1e soon.

rzumer:
@cryptomilk Sorry for the delay in addressing this, I implemented still picture mode a few months back and it seems like other changes in the encoder broke it since then. I will add some tests to avoid future regressions and try to solve the issues found (#2112 and #2113) in a timely manner. In the meantime, #2108 should solve errors when encoding a single image frame, which I think should cover your use case.

cryptomilk:
Awesome thanks. I'm travelling this week and will test it next week once i'm back.

rzumer:
Upon further testing there's some extra work to do before the reduced still picture header is supported, hopefully soon. #2108 fixes the still picture mode for any number of frames, without the reduced still picture header (which should be legal AVIF, but not recommended).

cryptomilk:
@rzumer Thank you very much for fixing this! Now there is one bug/feature left for AVIF and that's lossless support.

rzumer:
@cryptomilk I can't promise that support will be added soon, but I have tagged the corresponding issue (#151) for tracking.

PR #1498, opened 2019-07-24
Add still picture support

### LINKED ISSUE #999: Add API to set still_picture flag

rzumer:
* Add a `still_picture` parameter in `EncoderConfig` (fixes #999?)
* Add a CLI flag `--still-picture`
* (Attempt to) fix reduced still picture header handling in sequence header OBU
* Adjust sequence/FI default parameters to code valid still picture frames

Tested with two clips locally.

To do: fix the reduced still picture header, which currently desyncs if enabled (#1000).

Please advise if more API changes are needed.


rzumer:
I force pushed to fix the `else {` and added a line to handle the option in the C API.

lu-zero:
It needs a rebase

PR #1499, opened 2019-07-25
Remove Unspecified from the PixelRange

lu-zero:
The bitstream signals the information using a boolean.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -139,7 +139,7 @@ impl EncoderConfig {
       bit_depth: 8,
       chroma_sampling: ChromaSampling::Cs420,
       chroma_sample_position: ChromaSamplePosition::Unknown,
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -392,14 +392,13 @@ impl Default for ChromaSamplePosition {
 #[derive(ArgEnum, Debug, Clone, Copy, PartialEq, FromPrimitive)]
 #[repr(C)]
 pub enum PixelRange {
-  Unspecified = 0,
```
dwbuiten: Should you set `Limited` to 0 now?

lu-zero: It is 0 by default. I can be explicit if you think it is helpful.

lu-zero:
`APIChange: Breaking` added

PR #1502, opened 2019-07-25
Fix block importance propagation MV top left block

YaLTeR:
[AWCY](https://beta.arewecompressedyet.com/?job=master-5ea7a15c0a7a60c37be637851f2b594a9a6b79e4&job=fix-block-imp-mv-search-2) where I set `--range full`, not sure if it affects anything.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1529,10 +1529,17 @@ impl<T: Pixel> ContextInner<T> {
 
               // Coordinates of the top-left corner of the block intersecting the
               // reference block from the top-left.
-              let top_left_block_x =
-                reference_x / MI_SIZE_IN_MV_UNITS * MI_SIZE_IN_MV_UNITS;
-              let top_left_block_y =
-                reference_y / MI_SIZE_IN_MV_UNITS * MI_SIZE_IN_MV_UNITS;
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -1529,11 +1529,21 @@ impl<T: Pixel> ContextInner<T> {
 
               // Coordinates of the top-left corner of the block intersecting the
               // reference block from the top-left.
-              let top_left_block_x =
+              let mut top_left_block_x =
```
lu-zero: Something along the lines of:
``` rust
let top_left_block_x = (reference_x - if (reference_x < 0) { MI_SIZE_IN_MV_UNITS - 1 } else { 0 }) / MI_SIZE_IN_MV_UNITS * MI_SIZE_IN_MV_UNITS
```
Wouldn't be equivalent?

rzumer:
> not sure if it affects anything
It should not.

src/api.rs
YaLTeR: It should be I think.

lu-zero: Please update that way if it works :)

YaLTeR:
Done, awcy showed no change from before force push.

PR #1508, opened 2019-07-30
Add another IntoFrame variant

lu-zero:
It can be used to simplify the c-api.


lu-zero:
``` irc
19:39 < lu_zero> Daemon404: can I land the IntoFrame patch?
19:45 < Daemon404> lu_zero, yes
```

PR #1510, opened 2019-07-30
Don't factor the tile rate limit into min_tile_rows_log2.

### LINKED ISSUE #1509: Desync with large size image, for ex: 3840x2160, dav1d error: "Overrun in OBU bit buffer into next OBU"
commit: 5554cfe2f897d27cc64a8964f355e0d7982f081a

It seems master desync with crowd_run-1frame.y4m, 3840x2160.
Cannot decode with either dav1d or aomdec.

https://people.xiph.org/~yushin/data/crowd_run-1frame.y4m

./target/release/rav1e /Users/yushin/Downloads/crowd_run-1frame.y4m -o test.ivf -r test_rec.y4m --quantizer=100 --speed=10 --limit=1 --tune=Psnr

lassen:rav1e yushin$ ../dav1d/build/tools/dav1d -i test.ivf -o test_dec_dav1d.y4m
dav1d 0.3.1-25-g63eef33 - by VideoLAN
Overrun in OBU bit buffer into next OBU
Error decoding frame: Invalid argument
No data decoded

tdaede:
Fixes #1509.

### DIFF
--- a/src/tiling/tiler.rs
+++ b/src/tiling/tiler.rs
@@ -66,15 +66,20 @@ impl TilingInfo {
     let sb_cols = frame_width.align_power_of_two_and_shift(sb_size_log2);
     let sb_rows = frame_height.align_power_of_two_and_shift(sb_size_log2);
 
+    // these are bitstream-defined values and must not be changed
     let max_tile_width_sb = MAX_TILE_WIDTH >> sb_size_log2;
     let max_tile_area_sb = MAX_TILE_AREA >> (2 * sb_size_log2);
     let min_tile_cols_log2 = Self::tile_log2(max_tile_width_sb, sb_cols);
     let max_tile_cols_log2 = Self::tile_log2(1, sb_cols.min(MAX_TILE_COLS));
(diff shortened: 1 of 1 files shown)


rom1v:
What is the purpose of a "max tile rate"? This is not a restriction from the spec?

tdaede:
It's on page 643 of the spec: https://aomediacodec.github.io/av1-spec/av1-spec.pdf

Technically it is only enforced when we encode a level, which we don't right now. We'll need it eventually, though.

rzumer:
Thank you for fixing this, could you take the opportunity to comment the calculations? It is getting a bit complicated.

src/tiling/tiler.rs
```diff
@@ -86,8 +88,15 @@ impl TilingInfo {
     } else {
       0
     };
+    let min_tile_rows_ratelimit_log2 = if min_tiles_ratelimit_log2 > tile_cols_log2 {
+      min_tiles_ratelimit_log2 - tile_cols_log2
```
lu-zero: isn't that a `saturating_sub()`?

tdaede: yes but I'd rather have it match the other case right below

tdaede:
added comments

PR #1518, opened 2019-08-01
WIP: Signal interp_filter in the bitstream.

tdaede:


tdaede:
about a 0.04% loss https://beta.arewecompressedyet.com/?job=interpfilter-2019-08-09_175924-f3c43a4&job=master-ec0c7787dd1598f4cecaedda2a120ff4ab0eb6d0

tdaede:
going to actually stack all of the interpfilter commits here before landing

ycho:
Can we land this?

barrbrain:
A first step toward this was landed in #1866.

ycho:
I might have seen this PR, but was not able to recall this much progress is already there. I hope we want not to waste this PR and land of most of it please.

src/encoder.rs
```diff
@@ -1635,6 +1635,11 @@ pub fn encode_block_post_cdef<T: Pixel>(
           assert_eq!(mvs[0].col, mv_stack[0].this_mv.col);
         }
       }
+      if fi.is_filter_switchable
+        && ContextWriter::needs_interp_filter(bsize, luma_mode)
+      {
+        cw.write_interp_filter(w, tile_bo, bsize, 0, FilterMode::REGULAR);
```
ycho: I think the reason why there was desync in your test is that the decoder does not think it needs to decode the filter for each inter mode block if a frame header does not have is_filter_switchable = true, i.e. instead decoder already has decoded default filter. In spec, https://aomediacodec.github.io/av1-spec/#inter-block-mode-info-syntax, there is if block by " if ( interpolation_filter == SWITCHABLE ) {", if true it decodes the filter.

tdaede: That's the fi.is_filter_switchable flag here I think.

ycho: Right, I mean it should sync with 'is_filter_switchable' written in frame header. Otherwise, I think the decoder does not decode the info by cw.write_interp_filter(w, tile_bo, bsize, 0, FilterMode::REGULAR);  ?

tdaede: Right. I think it is synced, unless I'm missing something?

ycho: I landed https://github.com/xiph/rav1e/commit/485d179206e3a777db95157a5b25a1a3b1991a8e#diff-e59bfbbede3bbaad0b099696e5ec5716L635 , and if fi.is_filter_switchable is true, encoder does not write defalut filter. Then, decoder knows it needs to decode filter from each inter mode block.

ycho:
When you rebased on master today, was there still desync? I think it should not be there.

tdaede:
No, there is no desync.

ycho:
> No, there is no desync.

Great!

ycho:
@tdaede Would you have plan to follow up the build errors (mismatch) soon, or might be later after your current other work?

tdaede:
@ycho I will look at it a bit today, but if I don't find a fix I'd appreciate it if you could also look.

ycho:
> @ycho I will look at it a bit today, but if I don't find a fix I'd appreciate it if you could also look.

Sure, I wanted to help on this. Just I got completely taken by my son during thx giving holidays. Also, I think I plan move on this after trying fixing desync one #1598.

ycho:
With local testing, this causes corruption from 2nd frame, either speed 0 or 10. I expect speed 10 does not encode filter type per blocks, since its block size is  64x64 >= 32x32 checked by needs_interp_filter(), other than right or bottom frame borders.

ycho:
> No, there is no desync.

Which input and speed have you tested with?

PR #1522, opened 2019-08-01
Add newtypes for {Plane,Tile}{Super,}BlockOffset

YaLTeR:
Helps ensure type safety. `Plane` offsets are relative to the plane, and `Tile` offsets are relative to the tile, with methods on `TileState` providing conversions.

This is an idea I had ever since I bumped into this while implementing mean block importance computation in `compute_rd_cost()`. While I would like to say that error would've been entirely prevented had this been in-place, it wouldn't, because in that code I extracted raw `.x` and `.y` anyway. However, I clearly needed the frame block offset and I could've figured it out by looking at the type names.

[AWCY just in case (all zeros)](https://beta.arewecompressedyet.com/?job=block-offset-refactor&job=master-52074d09a3563fe7c2ea0730c13ec919052b847c)

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -192,8 +192,8 @@ fn adjust_strength(strength: i32, var: i32) -> i32 {
 pub fn cdef_analyze_superblock<T: Pixel>(
   in_frame: &Frame<T>,
   blocks: &TileBlocks<'_>,
-  sbo: SuperBlockOffset,
-  sbo_global: SuperBlockOffset,
+  sbo: TileSuperBlockOffset,
+  sbo_global: TileSuperBlockOffset,
   bit_depth: usize,
 ) -> CdefDirections {
   let coeff_shift = bit_depth as usize - 8;
@@ -206,7 +206,7 @@ pub fn cdef_analyze_superblock<T: Pixel>(
       // single-SB copy 'frame' that represents some superblock
       // in the main frame.
       let global_block_offset = sbo_global.block_offset(bx<<1, by<<1);
-      if global_block_offset.x < blocks.cols() && global_block_offset.y < blocks.rows() {
+      if global_block_offset.0.x < blocks.cols() && global_block_offset.0.y < blocks.rows() {
         let skip = blocks[global_block_offset].skip
           & blocks[sbo_global.block_offset(2*bx+1, 2*by)].skip
           & blocks[sbo_global.block_offset(2*bx, 2*by+1)].skip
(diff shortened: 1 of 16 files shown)


src/context.rs
```diff
@@ -1695,10 +1803,10 @@ impl<'a> BlockContext<'a> {
       let mut left: u8 = 0;
 
       for k in 0..txb_w_unit {
-        top |= self.above_coeff_context[plane][(bo.x >> xdec) + k];
+        top |= self.above_coeff_context[plane][(bo.0.x >> xdec) + k];
       }
       for k in 0..txb_h_unit {
-        left |= self.left_coeff_context[plane][(bo.y_in_sb() >> ydec) + k];
+        left |= self.left_coeff_context[plane][(bo.0.y_in_sb() >> ydec) + k];
```
tdaede: extraneous 0.

src/context.rs
```diff
@@ -1676,12 +1784,12 @@ impl<'a> BlockContext<'a> {
...
 
         for k in 0..txb_w_unit {
-          top |= self.above_coeff_context[0][(bo.x >> xdec) + k];
+          top |= self.above_coeff_context[0][(bo.0.x >> xdec) + k];
         }
         top &= COEFF_CONTEXT_MASK as u8;
 
         for k in 0..txb_h_unit {
-          left |= self.left_coeff_context[0][(bo.y_in_sb() >> ydec) + k];
+          left |= self.left_coeff_context[0][(bo.0.y_in_sb() >> ydec) + k];
```
tdaede: same

src/context.rs
```diff
@@ -1633,14 +1741,14 @@ impl<'a> BlockContext<'a> {
...
-      let sign = self.above_coeff_context[plane][(bo.x >> xdec) + k]
+      let sign = self.above_coeff_context[plane][(bo.0.x >> xdec) + k]
         >> COEFF_CONTEXT_BITS;
       assert!(sign <= 2);
       dc_sign += signs[sign as usize] as i16;
     }
 
     for k in 0..txb_h_unit {
-      let sign = self.left_coeff_context[plane][(bo.y_in_sb() >> ydec) + k]
+      let sign = self.left_coeff_context[plane][(bo.0.y_in_sb() >> ydec) + k]
```
tdaede: same

src/cdef.rs
```diff
@@ -485,8 +485,11 @@ pub fn cdef_filter_frame<T: Pixel>(fi: &FrameInvariants<T>, rec: &mut Frame<T>,
   // Perform actual CDEF, using the padded copy as source, and the input rec vector as destination.
   for fby in 0..fb_height {
     for fbx in 0..fb_width {
-      let sbo = SuperBlockOffset { x: fbx, y: fby };
+      let sbo = PlaneSuperBlockOffset(SuperBlockOffset { x: fbx, y: fby });
       let cdef_index = blocks[sbo.block_offset(0, 0)].cdef_index;
+
+      // In this particular instance we're operating on the whole frame.
```
tdaede: "CDEF application operates on the whole frame as if it were one tile".

YaLTeR:
Addressed the comments.

lu-zero:
It needs a rebase.

YaLTeR:
Done.

PR #1524, opened 2019-08-02
Fix excessive memory usage issue in lookahead

### LINKED ISSUE #1523: Using gigabytes of memory when encoding 720p long clips
(Sorry for my bad English)
using commit 914fe8df0df9cdb87bd0f56fa8a57c8a3eab948
Footprint increases slowly and serially(about 8MB/s).
I closed it when 3 GB RAM was used by rav1e.Not a frame was encoded before closing it.
(Can encode with --limit 60, but it takes a long time before the first frame was encoded)
Is it expected?
(or anything wrong with my command?)
system: xubuntu 18.04 with sse4.1
```
ffmpeg -i tears_of_steel_720p.mov -an -f yuv4mpegpipe - | rav1e - --quantizer 120 -o output.ivf
```

shssoichiro:
The lookahead would read all frames in the video into memory
before processing any of them. This would lead to huge amounts of
memory usage on longer videos.

Fixes #1523.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.35.0
+  - 1.36.0
 env:
   - RUST_BACKTRACE=1
 addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.36.0
+  - 1.35.0
 env:
   - RUST_BACKTRACE=1
 addons:
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -1782,13 +1782,13 @@ impl<T: Pixel> ContextInner<T> {
     if cur_input_frameno == 0 {
       return;
     }
-    for i in self.frame_q.keys().next().cloned().unwrap_or(0)..cur_input_frameno {
+    for i in 0..cur_input_frameno {
       self.frame_q.remove(&i);
     }
     if self.output_frameno < 2 {
       return;
     }
-    for i in self.frame_invariants.keys().next().cloned().unwrap_or(0)..(self.output_frameno - 1) {
+    for i in 0..(self.output_frameno - 1) {
       self.frame_invariants.remove(&i);
       self.gop_output_frameno_start.remove(&i);
       self.gop_input_frameno_start.remove(&i);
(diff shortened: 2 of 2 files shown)


src/api/mod.rs
```diff
@@ -1588,7 +1595,7 @@ impl<T: Pixel> ContextInner<T> {
       return Err(EncoderStatus::LimitReached);
     }
 
-    if self.needs_more_lookahead() {
+    if self.needs_more_fi_lookahead() {
```
shssoichiro: This is using the logic that was on 1609.

src/api/mod.rs
```diff
@@ -1781,13 +1782,13 @@ impl<T: Pixel> ContextInner<T> {
     if cur_input_frameno == 0 {
       return;
     }
-    for i in 0..cur_input_frameno {
+    for i in self.frame_q.keys().next().cloned().unwrap_or(0)..cur_input_frameno {
```
shssoichiro: This change and 1791 aren't related to memory usage, but should provide a minor performance improvement especially on long videos.

shssoichiro:
I will create an automated test to check for this in a future PR. Since it's a high priority, I want to get the fix in first.

lu-zero:
It is not a memory leak.

src/api/mod.rs
lu-zero: Please split them in a second patch.

dwbuiten:
This also fixes the issue where the entire clip gets buffered and all encoded during flushing, for me, FYI.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.35.0
+  - 1.36.0
```
tdaede: What change requires this?

shssoichiro: I was getting CI failures because the cargo-clippy component couldn't be found. It's possible it was an intermittent failure from Travis, although bumping the Rust version did fix it.

tdaede: I think it was intermittent. Could you drop it?

(in general bumping rust versions is fine, but preferably not as an unrelated change)

shssoichiro: Agreed, I'll roll it back and hopefully we get a green.

PR #1527, opened 2019-08-02
Implement quick and dirty simd intrinsics for sgr

KyleSiefring:
https://beta.arewecompressedyet.com/?job=master-52074d09a3563fe7c2ea0730c13ec919052b847c&job=sgr_simd_0%402019-08-02T00%3A11%3A25.533Z

~10% speedup.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ dump_ivf = ["ivf"]
 quick_test = []
 desync_finder = []
 bench = []
+check_asm = []
 
 # Enables debug dumping of lookahead computation results, specifically:
 # - i-qres.png: quarter-resolution luma planes,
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -22,6 +22,7 @@ dump_ivf = ["ivf"]
 quick_test = []
 desync_finder = []
 bench = []
+check_asm = []
```
tdaede: If you're going to go for checkasm style testing rather than unit tests, you need to turn checkasm on for ci.

src/lrf_simd.rs
```diff
@@ -0,0 +1,481 @@
...
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+use crate::lrf::*;
+use crate::frame::PlaneSlice;
+use crate::util::Pixel;
+
+const X_BY_XPLUS1: [u32; 256] = [
```
tdaede: You might want static rather than const here.

Issue #1513, opened 2019-07-31
"The application panicked (crashed)" when trying to run first pass


gnafuthegreat:
When running a first pass using the current master AppVeyor build on Windows, I get the following (after setting RUSTBACKTRACE=full):

```
Backtrace (most recent call first):
  File "<unknown>", line 0, in BaseThreadInitThunk
  File "<unknown>", line 0, in RtlUserThreadStart

The application panicked (crashed).
  no entry found for key
in src\libcore\option.rs, line 1036
thread: main
```

This is when using the following command line:

`rav1e.exe -o tfe-1080p-rav1e-750k-5-2pass.ivf -b 750 --first-pass tfe-1080p.stats tfe-1080p.y4m`

I'm going to see if I can find a build that works, and then find when it stopped working.

gnafuthegreat:
I was able to confirm when it stopped working.  c00f1556 works, d8b60703 fails.  So the offending commit appears to be:

https://github.com/xiph/rav1e/commit/d8b60703b2557bd937530625593675185504e291

gnafuthegreat:
Crash also happens on Linux, and I was able to get this output from a debug build:

```
Backtrace (most recent call first):
  File "<unknown>", line 0, in core::option::expect_failed
  File "/builddir/build/BUILD/rustc-1.36.0-src/src/libcore/option.rs", line 314, in core::option::Option<T>::expect
  File "/builddir/build/BUILD/rustc-1.36.0-src/src/liballoc/collections/btree/map.rs", line 1807, in <alloc::collections::btree::map::BTreeMap<K,V> as core::ops::index::Index<&Q>>::index
  File "/home/gideon/rav1e/src/api/mod.rs", line 1814, in rav1e::api::ContextInner<T>::guess_frame_subtypes
    self.gop_input_frameno_start[&self.output_frameno];
  File "/home/gideon/rav1e/src/rate.rs", line 908, in rav1e::rate::RCState::select_qi
    ctx.guess_frame_subtypes(&mut nframes, self.reservoir_frame_delay);
  File "/home/gideon/rav1e/src/rate.rs", line 1237, in rav1e::rate::RCState::get_twopass_out_params
    self.select_qi(ctx, output_frameno, FRAME_SUBTYPE_I, None).log_base_q;
  File "/home/gideon/rav1e/src/api/mod.rs", line 732, in rav1e::api::Context<T>::twopass_out
    let params = self
[...]

PR #1529, opened 2019-08-02
Add review policy.

tdaede:


README.md
```diff
@@ -186,6 +186,10 @@ Run regular benchmarks with:
 cargo bench --features=bench
 ```
 
+## Reviews
+
+PRs must be reviewed by at least one other collaborator. Once they are reviewed and pass CI, anyone may merge (though it's convention for the author to if they have permissions).
```
YaLTeR: Mention AWCY?

rzumer:
I am a bit uneasy about including URLs in commit messages, since they will probably eventually become inaccessible (and AWCY URLs can be quite long). Why are they needed specifically?

YaLTeR:
Probably enough to have them in the PR

tdaede:
It's so you can jump and look at more detailed results. I could make it optional I suppose.

shssoichiro:
In addition to rebasing, should there be a comment about squashing commits? Do we enforce squashing commits?

lu-zero:
We do not enforce squashing, sometimes makes sense if you push additional
refinements as commits instead of reading and pushing an update.

tdaede:
I originally enforced squashing, everyone hated it so I stopped. The main issue is that CI doesn't run on each individual commit, so by not squashing we have some revs that were never CI tested.

YaLTeR:
On the one hand yes, but on the other hand I just had bisection pinpoint me to the exact small change leading to #1513 and that's super helpful to have.

tdaede:
@lu-zero please re-review

barrbrain:
To clarify, the **Contributing** chapter has moved to [CONTRIBUTING.md](https://github.com/xiph/rav1e/blob/master/CONTRIBUTING.md). This section should be added to that document.

barrbrain:
This might need to be manually reapplied to the new file but it would be great to finally merge this.

barrbrain (approved):

PR #1532, opened 2019-08-05
Add test for amount of lookahead consumed

shssoichiro:
This required moving the RDO lookahead frames from a constant into `EncoderConfig` (it is not yet configurable from the CLI). This also cleans up some behaviors in lookahead, so it is now reading no more than the needed number of frames at any time and behaves correctly via the API as well as the CLI.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -3296,219 +3296,135 @@ mod test {
     );
   }
 
-  #[derive(Debug, Clone, Copy)]
+  #[derive(Clone, Copy)]
   struct LookaheadTestExpectations {
-    pre_receive_frame_q_len: usize,
-    pre_receive_fi_len: usize,
-    post_receive_frame_q_len: usize,
-    post_receive_fi_len: usize,
+    pre_receive_frame_q_lens: [usize; 60],
+    pre_receive_fi_lens: [usize; 60],
+    post_receive_frame_q_lens: [usize; 60],
+    post_receive_fi_lens: [usize; 60],
   }
 
   #[test]
   fn lookahead_size_properly_bounded_8() {
     const LOOKAHEAD_SIZE: u64 = 8;
-    const EXPECTATIONS: [LookaheadTestExpectations; 60] = [
-      LookaheadTestExpectations { pre_receive_frame_q_len: 1, pre_receive_fi_len: 0, post_receive_frame_q_len: 1, post_receive_fi_len: 0 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 2, pre_receive_fi_len: 0, post_receive_frame_q_len: 2, post_receive_fi_len: 0 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 3, pre_receive_fi_len: 0, post_receive_frame_q_len: 3, post_receive_fi_len: 0 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 4, pre_receive_fi_len: 0, post_receive_frame_q_len: 4, post_receive_fi_len: 0 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 5, pre_receive_fi_len: 0, post_receive_frame_q_len: 5, post_receive_fi_len: 0 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 6, pre_receive_fi_len: 0, post_receive_frame_q_len: 6, post_receive_fi_len: 0 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 7, pre_receive_fi_len: 1, post_receive_frame_q_len: 7, post_receive_fi_len: 1 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 8, pre_receive_fi_len: 1, post_receive_frame_q_len: 8, post_receive_fi_len: 1 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 9, pre_receive_fi_len: 1, post_receive_frame_q_len: 9, post_receive_fi_len: 1 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 10, pre_receive_fi_len: 1, post_receive_frame_q_len: 10, post_receive_fi_len: 1 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 11, pre_receive_fi_len: 7, post_receive_frame_q_len: 11, post_receive_fi_len: 7 },
-      LookaheadTestExpectations { pre_receive_frame_q_len: 12, pre_receive_fi_len: 7, post_receive_frame_q_len: 12, post_receive_fi_len: 7 },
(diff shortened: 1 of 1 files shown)


src/api/mod.rs
```diff
@@ -3266,4 +3288,92 @@ mod test {
...
+    let total_lookahead_len = rdo_lookahead_len + scenechange_lookahead_len + 1;
+    for frameno in rdo_lookahead_len..total_lookahead_len {
+      let input = ctx.new_frame();
+      let _ = ctx.send_frame(input);
+      let res = ctx.receive_packet();
+      assert!(res.is_err());
+      assert_eq!(ctx.inner.frame_q.len(), frameno + 1);
+      // The exact logic gets complicated here because of frame pyramids.
+      assert!(ctx.inner.frame_invariants.len() > 0);
+      assert!(ctx.inner.frame_invariants.len() < rdo_lookahead_len * 6 / 4);
```
YaLTeR: Where does the `6 / 4` come from? Could you comment this?

src/api/mod.rs
```diff
@@ -1297,37 +1298,44 @@ impl<T: Pixel> ContextInner<T> {
     }
   }
 
-  fn compute_lookahead_data(&mut self) {
+  fn compute_lookahead_data(&mut self) {// Compute the frame invariants.
```
YaLTeR:
```suggestion
  fn compute_lookahead_data(&mut self) {
    // Compute the frame invariants.
```

src/api/mod.rs
```diff
@@ -964,8 +964,9 @@ impl<T: Pixel> ContextInner<T> {
       output_frameno_in_gop,
       self.gop_input_frameno_start[&output_frameno]
     );
-    let frames_needed = input_frameno + self.inter_cfg.keyframe_lookahead_distance();
-    if frames_needed > self.next_lookahead_frame && self.needs_more_frames(frames_needed) {
+
+    self.compute_scenechange_lookahead(input_frameno);
```
YaLTeR: I think the general idea for the MV-based scene cut detection was to:
1. Get rid of the "global mutable state" in `build_frame_properties()` (particularly `gop_input_frameno_start` and `gop_output_frameno_start`);
2. Compute frame invariants and the lookahead motion vectors, then run the scenecut detection, and then if a scenecut is detected just recompute the affected frame invariants.

I'm not sure how much moving the logic here affects the amount of refactoring that will need to be done later for 2. At the very least, I think that respecting minimum and maximum GOP sizes was a little more obvious when it was outside of `build_frame_properties`, but I might be wrong.

src/api/mod.rs
```diff
@@ -3204,7 +3221,12 @@ mod test {
         (4, true), // B0-frame (show existing)
         (5, true), // B1-frame (second)
         (6, true), // P-frame (show existing)
+        (6, false), // invalid
+        (6, false), // invalid
         (7, true), // I-frame
```
YaLTeR: That should be the B1-frame (first).

src/api/mod.rs
```diff
@@ -3265,4 +3288,92 @@ mod test {
...
+      let input = ctx.new_frame();
+      let _ = ctx.send_frame(input);
+      while ctx.receive_packet().is_ok() {
+        // Receive packets until lookahead consumed, due to pyramids receiving frames in groups
+      }
+      // The exact logic gets complicated here because of frame pyramids.
+      assert!(ctx.inner.frame_q.len() > total_lookahead_len);
+      assert!(ctx.inner.frame_q.len() <= total_lookahead_len + scenechange_lookahead_len + 1);
+      assert!(ctx.inner.frame_invariants.len() > rdo_lookahead_len);
+      assert!(ctx.inner.frame_invariants.len() <= (rdo_lookahead_len + ctx.inner.inter_cfg.group_input_len as usize) * 6 / 4 + 1);
```
YaLTeR: Same here about `6 / 4`.

src/api/mod.rs
shssoichiro: The reason I moved it the way I did was to remove one place we depend on `self.next_lookahead_frame`. That piece of global mutable state was/still is a pain point. There's definitely still significant refactor needed to move to the way we want lookahead to work.

shssoichiro: Never mind, I have to revert this, I found a subtle bug where I had forgotten to remove one of the references to `self.next_lookahead_frame`.

shssoichiro:
I can split the constant to variable change, but the test won't pass without the additional refactor.

shssoichiro:
Marked WIP until #1535 merged.

shssoichiro:
@lu-zero Rebased and removed all unnecessary changes from this PR.

shssoichiro:
@lu-zero @YaLTeR ping

src/api/mod.rs
```diff
@@ -3281,4 +3295,94 @@ mod test {
...
+    let rdo_lookahead_len = ctx.config.rdo_lookahead_frames as usize;
+    for frameno in scenechange_lookahead_len..rdo_lookahead_len {
+      let input = ctx.new_frame();
+      let _ = ctx.send_frame(input);
+      let res = ctx.receive_packet();
+      assert!(res.is_err());
+      assert_eq!(ctx.inner.frame_q.len(), frameno + 1);
+      // The exact logic gets complicated here because of frame pyramids,
+      // but the size of the FI queue should always hold to this rule.
+      assert!(ctx.inner.frame_invariants.len() < rdo_lookahead_len);
```
YaLTeR: Perhaps an array like `let frame_invariants_len_at_step = [...];` that you compare with every step of this loop would be better?

shssoichiro: Yeah, we could use that to make the whole test simpler as well. Removes the need for multiple loops. I'll do that.

src/api/mod.rs
```diff
@@ -3296,10 +3296,219 @@ mod test {
...
+    pre_receive_fi_len: usize,
+    post_receive_frame_q_len: usize,
+    post_receive_fi_len: usize,
+  }
+
+  #[test]
+  fn lookahead_size_properly_bounded_8() {
+    const LOOKAHEAD_SIZE: u64 = 8;
+    const EXPECTATIONS: [LookaheadTestExpectations; 60] = [
+      LookaheadTestExpectations { pre_receive_frame_q_len: 1, pre_receive_fi_len: 0, post_receive_frame_q_len: 1, post_receive_fi_len: 0 },
```
YaLTeR: Doing this in a structure-of-arrays fashion rather than array-of-structures would allow to not repeat the same long field name every line. Something like:
```rust
let pre_receive_frame_q_len = [1, 2, 3, 4, ...];
let pre_receive_fi_len = [0, 0, 0, 0, ...];
...
```
Those arrays could also be inside the structure if you prefer.

shssoichiro:
Since I've addressed @lu-zero's PR splitting comment, and @YaLTeR has approved, I'm going to go ahead and merge.

PR #1534, opened 2019-08-06
Actually use the min keyframe interval from CLI

### LINKED ISSUE #1497: Rav1e doesn't respect MIN_KEYFRAME_INTERVAL
I have a 50 frame sequence and I'd like the first frame only to be a keyframe.
-i 100 and -I defaulting to 240 should work but I get 4 keyframes

```
cargo run --release --bin rav1e -- in.y4m -i 100 -o out.ivf
    Finished release [optimized] target(s) in 0.06s
     Running `target/release/rav1e in.y4m -i 100 -o out.ivf`
1280x720 @ 25/1 fps
encoded 54 frames, 0.207 fps, 1096.35 Kb/s                    
Key Frames:      4    avg size:   48216 B
Inter:          50    avg size:    2063 B

```

shssoichiro:
Fixes #1497

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -423,7 +423,7 @@ fn parse_config(matches: &ArgMatches<'_>) -> EncoderConfig {
       .unwrap_or_default();
 
     let mut cfg = EncoderConfig::with_speed_preset(speed);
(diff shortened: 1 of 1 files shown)


tmatth:
Nice catch.

Issue #1533, opened 2019-08-06
Malformed chroma output on videos with odd dimensions


shssoichiro:
Originally reported via IRC:

```
<soichiro> https://condor3280.startdedicated.com/files/blade.240.y4m.tar.gz
<soichiro> It plays fine in mpv
<soichiro> But the output from rav1e looks corrupted
<derf> Okay. Where is the broken version?
<soichiro> https://condor3280.startdedicated.com/files/blade.ivf
<derf> And what are you using to decode it?
<soichiro> Tested with mpv (on Linux and Windows), ffplay, and AOMAnalyzer
<derf> soichiro: What version of the y4m crate are you using?
<soichiro> 0.3.3
<derf> Okay, that is the version that is supposed to fix this problem.
<derf> Chagall: ^
<derf> (get_plane_sizes() for 0.3.2 and earlier was just busted)
<soichiro> I was going to check the reconstruction from rav1e but trying to get a reconstruction actually makes it crash
<derf> Great!
<soichiro> Rerunning in debug mode to get a useful stack trace
<soichiro>  File "src/bin/muxer/y4m.rs", line 91, in rav1e::muxer::y4m::write_y4m_frame
<soichiro>     y4m_enc.write_frame(&rec_frame).unwrap();
<soichiro> Which means there was an error in `write_frame` in the y4m crate
<soichiro> Cool!
<derf> Yeah, because the chroma plane sizes don't match.
<derf>   let (pitch_uv, height_uv) = (
<derf>     pitch_y / chroma_sampling_period.0,
<derf>     y4m_details.height / chroma_sampling_period.1
<derf>   );
<derf> At the top of write_y4m_frame().
<derf> Which is also just busted.
<Chagall> just needs to be rounded up the same way I guess
<derf> Well, I have the feeling this is not the only problem.
<soichiro> Yeah, I think there's some issue in the decoder when the width or height is odd
<soichiro> Which is not causing a crash but is causing malformed output
<derf> s/decoder/encoder/
<soichiro> Probably in both
<soichiro> but I say decoder because of how it's manifesting in that 240p test file, rav1e decodes it with the y4m crate
<soichiro> and I'm guessing the output from the decoder has the malformed chroma
<derf> As I said, Chagall already fixed this in the y4m crate.
<derf> That doesn't mean rav1e is using that crate correctly.
<derf> I don't know what he did to test it.
<soichiro> True
<derf> read_frame() in bin/decoder/y4m.rs is also obviously busted.
[...]

PR #1536, opened 2019-08-06
Fix chroma subsampling with odd luma sizes.

rzumer:
```
None of the existing input/output code was accounting for odd luma
 sizes when subsampling chroma correctly.
This replaces the previous sampling_period() API, which encouraged
 divisions and was used incorrectly everywhere it was invoked, with
 a get_chroma_dimensions() API.
Since this is what everyone was doing after calling
 sampling_period() anyway, we can now implement the logic once,
 correctly, in a central place.

There is still an API for getting the chroma decimation shifts,
 since these are used internally in by our Planes and tiling logic.
But we now use an API that actually returns the shifts
 (gauranteeing that the decimation can be expressed in terms of
 shifts, instead of blindly converting by subtracting 1 from the
 period), and correctly identifies that there is no sensible return
 value for Cs400.

Cs400 still will not work, because as soon as we try to copy the
 frame data out of the y4m buffers, we'll call chunks() with a
 chunk size of 0 (on a slice of size 0), which will panic!().
But we do not support Cs400 yet anyway.
The requisite special cases can be added in a future patch by
 whoever tries to get it to work.
```

### DIFF
--- a/src/frame/plane.rs
+++ b/src/frame/plane.rs
@@ -256,8 +256,8 @@ impl<T: Pixel> Plane<T> {
     let yorigin = self.cfg.yorigin;
     let stride = self.cfg.stride;
     let alloc_height = self.cfg.alloc_height;
-    let width = w >> self.cfg.xdec;
-    let height = h >> self.cfg.ydec;
+    let width = w + self.cfg.xdec >> self.cfg.xdec;
+    let height = h + self.cfg.ydec >> self.cfg.ydec;
 
     if xorigin > 0 {
       for y in 0..height {
--- a/src/frame/plane.rs
+++ b/src/frame/plane.rs
@@ -256,8 +256,8 @@ impl<T: Pixel> Plane<T> {
     let yorigin = self.cfg.yorigin;
     let stride = self.cfg.stride;
     let alloc_height = self.cfg.alloc_height;
-    let width = w + self.cfg.xdec >> self.cfg.xdec;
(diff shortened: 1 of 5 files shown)


src/frame/plane.rs
```diff
@@ -256,8 +256,8 @@ impl<T: Pixel> Plane<T> {
     let yorigin = self.cfg.yorigin;
     let stride = self.cfg.stride;
     let alloc_height = self.cfg.alloc_height;
-    let width = w >> self.cfg.xdec;
-    let height = h >> self.cfg.ydec;
+    let width = w + self.cfg.xdec >> self.cfg.xdec;
```
lu-zero: Isn't that a `round_shift`?

tmatth: @lu-zero wouldn't a `round_shift` be `w + (1 << self.cfg.xdec >> 1) >> self.cfg.xdec`

rzumer: Technically the same, but also doesn't work directly since the function in `util` accepts only `i32`. If it were documented and generic it could be worth it.

shssoichiro:
This fixes #1533.

src/frame/plane.rs
lu-zero: @tmatth yes, given the range of xdec it behaves the same way.

rzumer:
Reopening to trigger a new CI build...

PR #1537, opened 2019-08-06
capi: Expose RDO lookahead frames

dwbuiten:
This was missed in 2db1f9c81f033fcd1eb421236b52036ee74b03f2.

Maybe we should have some sort of test or checklist for API changes, since this seems to get missed often.


shssoichiro:
What is the effect of missing this change? Is it only that the option is not configurable from the CAPI, or does it cause the CAPI to stop working? Sorry, I'm unfamiliar with the CAPI entirely. I agree some sort of test would be great to have.

dwbuiten:
@shssoichiro It just means it would not be accessible by the C API.

PR #1541, opened 2019-08-07
Fix submit_awcy.py with empty -prefix

YaLTeR:
Decode the output of subprocess, otherwise it's a bytes and you can't concatenate that with a string below.


tdaede:
This is a python3 specific change - please also change the first line of the script to call python3, not python.

YaLTeR:
:white_check_mark:

Issue #1315, opened 2019-05-30
ColorDescription enum values don't match with the spec


sdroege:
E.g. https://aomediacodec.github.io/av1-spec/ mentions that the primaries are

| 1  | CP_BT_709 | BT.709 |
| -- | -- | -- |
| 2  | CP_UNSPECIFIED | Unspecified |
| 4  | CP_BT_470_M | BT.470 System M (historical) |
| 5  | CP_BT_470_B_G | BT.470 System B, G (historical) |
| 6  | CP_BT_601 | BT.601 |
| 7  | CP_SMPTE_240 | SMPTE 240 |
| 8  | CP_GENERIC_FILM | Generic film (color filters using illuminant C) |
| 9  | CP_BT_2020 | BT.2020, BT.2100 |
| 10 | CP_XYZ | SMPTE 428 (CIE 1921 XYZ) |
| 11 | CP_SMPTE_431 | SMPTE RP 431-2 |
| 12 | CP_SMPTE_432 | SMPTE EG 432-1 |
| 22 | CP_EBU_3213 | EBU Tech. 3213-E |

while the rav1e enum is (see comments for mismatches)

```rust
pub enum ColorPrimaries {
    BT709 = 1,
    Unspecified,
    BT470M = 4,
    BT470BG,
    ST170M, // BT_601
    ST240M,
    Film,
    BT2020,
    ST428,  // XYZ
    P3DCI, // SMPTE_431
    P3Display, // SMPTE_432
    Tech3213 = 22, // EBU_3213
}
```

Similar problems are with the other enums (transfer characteristics, matrix).

kodabb:
sorry what is the problem here? the names are different because it's a different project but the values and the semantics are identical

sdroege:
It's confusing if the spec and the API disagree about the names

sdroege:
Yeah and in GStreamer we have yet another name for some of these. Too many standards.

If you think following some other projects' naming scheme for consistency is more important than using the naming in the AV1 spec that's fine with me. Either way someone will have to translate names, I just thought it might be less confusing if the encoder API was at least consistent in naming with the spec it implements :)

PR #1544, opened 2019-08-08
Rename color config enum variants to match the AV1 specification

rzumer:


lu-zero:
Adding variants mapping requires to manually implement `FromStr`. I'd argue that the original names for the matrix coeffs are less confusing.

What about adding compat matches in the FromStr instead?

rzumer:
I actually find the current variants to be more confusing. The references listed in the comments/docs do not even use the same naming scheme for some of them, e.g. variants 11 and 12 of primaries. In addition some variants are named according to their SMPTE standard number, while others that have SMPTE names used in the AV1 spec/H.273 are renamed to something else. It is inconsistent. At least with the ffmpeg convention, there is some consistency with a reference.

I would rather keep further improvements like mapping from multiple alternatives to a different PR.

lu-zero:
I'd just follow the AV1 specification names assuming we can map it so they are sane enum values.

You already did the work documenting them so adding an extended FromStr could be an option (probably the macro in use could be extended so aliases are supported as custom `#[arg_alias("", "", "")]`)

rzumer:
I am definitely not against using the AV1 spec names instead, though I think some prefer the ffmpeg conventions. Awaiting comments from the other requested reviewers...

dwbuiten:
I agree with @lu-zero - we should follow the AV1 spec names.

rzumer:
It has matched the spec for 4 days now.

sdroege:
See also https://github.com/xiph/rav1e/issues/1315 . Thanks for taking care of this!

PR #1359, opened 2019-06-18
Fixes #1292

TheSirC:
Both `cargo test` and `cargo build` were run without errors.

PR #1549, opened 2019-08-12
Don't init memory for iimgs in sgr

KyleSiefring:


lu-zero:
What speedup does it give? You might want to use `MaybeUninit::uninit().assume_init()`

KyleSiefring:
It's about 2.5% faster.

KyleSiefring:
```
0.0000 |  0.0000 |  0.0000 |   0.0000 | 0.0000 |  0.0000 |     0.0000
```

KyleSiefring:
The new way of getting uninitialized memory is unstable in the version ci is using.

rzumer:
There is another PR (#1359) that includes that change, so I suggest fixing the CI issue on that end instead.

PR #1551, opened 2019-08-14
Corrected some common misspellings.

figgis:
Some minor things found when browsing through the code.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -150,7 +150,7 @@ pub struct Sequence {
   pub operating_point_idc: [u16; MAX_NUM_OPERATING_POINTS],
   pub display_model_info_present_flag: bool,
   pub decoder_model_info_present_flag: bool,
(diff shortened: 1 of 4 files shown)


src/encoder.rs
```diff
@@ -150,7 +150,7 @@ pub struct Sequence {
   pub operating_point_idc: [u16; MAX_NUM_OPERATING_POINTS],
   pub display_model_info_present_flag: bool,
   pub decoder_model_info_present_flag: bool,
-  pub level: [[usize; 2]; MAX_NUM_OPERATING_POINTS],	// minor, major
+  pub level: [[usize; 2]; MAX_NUM_OPERATING_POINTS],    // minor, major
```
dwbuiten: Stray change.

figgis: changed a single tab to spaces instead.

PR #1552, opened 2019-08-14
Give an intuitive error message if the Rust version is too old

shssoichiro:
The addition of a minimum Rust version also means we should
test against both the minimum version and the latest stable
version in .travis.yml. This MR adds the CI tests against the
latest stable, and mentions that the minimum version in .travis.yml
should match the minimum version in build.rs.


shssoichiro:
WIP while travis runs, want to verify the CI changes before requesting comments.

PR #1561, opened 2019-08-19
Wrap CLI error: ParseIntError for quantizer

barrbrain:
Initially, just wrap ParseIntError for quantizer.

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -382,18 +382,18 @@ impl MatchGet for ArgMatches<'_> {
 }
 
 fn parse_config(matches: &ArgMatches<'_>) -> Result<EncoderConfig, CliError> {
-  let maybe_quantizer = matches.value_of("QP").map(|qp| qp.parse().unwrap());
+  let maybe_quantizer = matches.value_of_int("QP");
   let maybe_bitrate = matches.value_of_int("BITRATE");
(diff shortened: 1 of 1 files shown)


src/bin/common.rs
```diff
@@ -382,18 +382,18 @@ impl MatchGet for ArgMatches<'_> {
...
-  let maybe_quantizer = matches.value_of("QP").map(|qp| qp.parse().unwrap());
+  let maybe_quantizer = matches.value_of_int("QP");
   let maybe_bitrate = matches.value_of_int("BITRATE");
   let quantizer = maybe_quantizer.unwrap_or_else(|| {
     if maybe_bitrate.is_some() {
       // If a bitrate is specified, the quantizer is the maximum allowed (e.g.,
       //  the minimum quality allowed), which by default should be
       //  unconstrained.
-      255
+      Ok(255)
```
shssoichiro: This is very much unrelated to this PR, but I'm realizing `EncoderConfig::quantizer` and `bitrate` would make more sense as an enum, something like:

```rust
enum Quality {
  Quantizer(usize),
  Bitrate(usize),
}
```

barrbrain: It would be good to tidy up the plumbing from here to `RCState::new()`, although that likely involves public API changes.

PR #1562, opened 2019-08-19
Travis: split aom and dav1d decode tests

barrbrain:
- [x] ~~For now choose `decode_test_dav1d`, it should be faster.~~
- [ ] ~~Ensure that the skipped variant can be built.~~
- [x] Just split the ignored tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,12 +80,21 @@ jobs:
         - cargo test --features=decode_test,decode_test_dav1d,quick_test --verbose
       env:
         - CACHE_NAME=MIN_RUSTC
-    - name: "Ignored Tests"
+    - name: "Ignored Tests (aom)"
       rust: stable
       script:
-        - travis_wait 60 cargo test --release --features=decode_test,decode_test_dav1d --verbose --color=always -- --color=always --ignored
+        - travis_wait 60 cargo test --release --features=decode_test --verbose --color=always -- --color=always --ignored
       env:
-        - CACHE_NAME=IGNORED_TESTS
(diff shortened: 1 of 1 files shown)


shssoichiro:
I think we need a discussion on this. I was asked specifically to run both, I believe by @YaLTeR (although I'm bad at remembering who tells me things).

barrbrain:
The purpose of the first patch is to gauge the impact of running both. If it is significant, then we may need to find the balance between CI responsiveness and avoiding silent breakage.

YaLTeR:
Another thing we could try is run aom and dav1d ignored tests in separate jobs. This way they should run in parallel, potentially giving a speed-up?

shssoichiro:
There's a little bit of waste from having to build rav1e twice. But hopefully not too bad, especially if the cache is used. That would be faster (based on wall time).

lu-zero:
Appveyor had a timeout.

.travis.yml
```diff
@@ -68,7 +68,7 @@ jobs:
         - which cargo-kcov || cargo install cargo-kcov
         # We need to run cargo clean to avoid caching issues
         - cargo clean
-        - RUSTFLAGS="-C link-dead-code" cargo build --features=decode_test,decode_test_dav1d,quick_test --tests --verbose
+        - RUSTFLAGS="-C link-dead-code" cargo build --features=decode_test_dav1d,quick_test --tests --verbose
```
shssoichiro: Do we want to keep both enabled for the quick tests?

barrbrain: Yes, I think I'll reduce this down to just splitting the ignored tests job.

PR #1569, opened 2019-08-21
Added Copyright message.

figgis:
These files lack a Copyright message:

build.rs
examples/simple_encoding.rs
ivf/src/lib.rs
crates/avformat-sys/build.rs
src/bin/error.rs
crates/avformat-sys/src/lib.rs
crates/nasm_rs/src/lib.rs
src/bin/decoder/mod.rs
src/bin/decoder/y4m.rs
src/bin/muxer/ivf.rs

Unsure if these should be updated as well...

### DIFF
--- a/ivf/src/lib.rs
+++ b/ivf/src/lib.rs
@@ -1,3 +1,13 @@
+// Copyright (c) 2001-2016, Alliance for Open Media. All rights reserved
+// Copyright (c) 2017-2018, The rav1e contributors. All rights reserved
+//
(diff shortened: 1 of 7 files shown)


lu-zero:
The front-matter is needed once you mix and match files with different but compatible licenses, so might be a good idea to add them in.

The files in `crates/` are vendored and could stay as they are IMHO.

figgis:
so, everything except files under crates/ should have a copyright message?

lu-zero:
Sounds good :)

PR #1576, opened 2019-08-22
Bias distortion per-block rather than with mean

YaLTeR:
[AWCY](https://beta.arewecompressedyet.com/?job=master-db0da3c97efb269cb23571f2c1f68afdf1ad9813&job=per-block-distortion-4-2019-08-21_141456-89544e8)

Previously the distortion was computed for a given block, then biased with the mean importance of this block. Now the distortion is biased granularly as it is computed which allows the biasing to be more accurate (as importance is computed on a small block size basis).

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -200,6 +200,11 @@ pub fn estimate_rate(qindex: u8, ts: TxSize, fast_distortion: u64) -> u64 {
 fn cdef_dist_wxh_8x8<T: Pixel>(
   src1: &PlaneRegion<'_, T>, src2: &PlaneRegion<'_, T>, bit_depth: usize,
 ) -> u64 {
+  debug_assert!(src1.plane_cfg.xdec == 0);
+  debug_assert!(src1.plane_cfg.ydec == 0);
+  debug_assert!(src2.plane_cfg.xdec == 0);
+  debug_assert!(src2.plane_cfg.ydec == 0);
+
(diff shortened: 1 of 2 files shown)


src/rdo.rs
```diff
@@ -333,13 +398,22 @@ fn compute_tx_distortion<T: Pixel>(
...
+      &rec_region,
       bsize.width(),
       bsize.height(),
+      |bias_area, bsize| {
+        compute_distortion_bias(
+          fi,
+          input_region.subregion(bias_area).frame_block_offset(),
+          bsize,
+        )
+      },
```
tdaede: This biases skip blocks, but for non-skip blocks we just use tx_dist. You'll still need to bias tx_dist with the mean.

src/rdo.rs
```diff
@@ -405,8 +488,14 @@ pub fn compute_rd_cost<T: Pixel>(
   let bias = (mean_importance / FACTOR) as f64 + ADDEND;
   debug_assert!(bias.is_finite());
 
+  bias
+}
+
+pub fn compute_rd_cost<T: Pixel>(
+  fi: &FrameInvariants<T>, rate: u32, distortion: u64,
+) -> f64 {
```
barrbrain: 💯 Overall, separating the `compute_distortion_bias` from `compute_rd_cost` appears to simplify the RDO functions. I think the implicit contract is that distortion functions need return the biased distortion.

PR #1579, opened 2019-08-23
Fixes to sgr projection solving math

xiphmont:
Though adapted from libaom, the existing LRF sgrproj_solve math is
incorrect in several places.  This patch corrects the errors and is
verified against libaom to give correct results.  Note that this code
is float and libaom is integer; this affects the exact solution.


xiphmont:
cdef-lrf-skip-fix-baseline@2019-08-22T23:25:34.279Z -> cdef-lrf-sgrproj-correct-hash@2019-08-23T05:26:32.866Z

-0.2764 | -0.0452 |  0.0418 |  -0.2853 | -0.3972 | -0.2990 |    -0.2681

xiphmont:
We easily could if we wanted to-- I did not test BD-rate in comparison.

PR #1580, opened 2019-08-23
Faster Chroma-from-Luma RDO

barrbrain:
[AWCY result at default speed](https://beta.arewecompressedyet.com/?job=master-5ab6b1c67071d5859b1643b81eb52caf28409f0b&job=cfl-early-termination%402019-08-23T11%3A32%3A08.294Z) with up to 1.4% reduction in encoding time:

| -0.0406 | -0.1360 | -0.0170 |  -0.0317 | -0.0844 | -0.0652 |    -0.0339 |

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -1218,7 +1218,7 @@ pub fn rdo_cfl_alpha<T: Pixel>(
         bit_depth,
         Some(PredictionMode::UV_CFL_PRED),
       );
-      let alpha_cost = |alpha: &i16| -> u64 {
+      let mut alpha_cost = |alpha: i16| -> u64 {
         let mut rec_region =
           rec.subregion_mut(Area::BlockStartingAt { bo: tile_bo.0 });
         PredictionMode::UV_CFL_PRED.predict_intra(
@@ -1227,7 +1227,7 @@ pub fn rdo_cfl_alpha<T: Pixel>(
           uv_tx_size,
           bit_depth,
           &ac.array,
-          *alpha,
+          alpha,
           &edge_buf,
         );
         sse_wxh(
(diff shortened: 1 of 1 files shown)


src/rdo.rs
```diff
@@ -1196,6 +1196,7 @@ pub fn rdo_cfl_alpha<T: Pixel>(
 ) -> Option<CFLParams> {
   let PlaneConfig { xdec, ydec, .. } = ts.input.planes[1].cfg;
   let uv_tx_size = bsize.largest_chroma_tx_size(xdec, ydec);
+  debug_assert!(bsize.subsampled_size(xdec, ydec) == uv_tx_size.block_size());
```
ycho: Was CfL enabled for a partition smaller than a certain size, like 32x32? Forgot that. 
(Because, I aware the cases where uv_tx_size can be smaller than subsampled bsize in the future,
such as with 128x128 SB added, and with that 4:4:4 chroma format will have 128x128 chroma bsize that can not have 128x128 tx, since the largest tx size is 64x64)

ycho: Oh, apparently cfl_allowed( ) says "self <= BlockSize::BLOCK_32X32"!

ycho: Also, I realized that the above debug_assert( ) seems indeed started from me! :)

tdaede:
> Probably Ok, is it possible that `cost.1 < cost.0` though?

I think this case is handled correctly. Could you elaborate a bit?

src/rdo.rs
barrbrain: Keeping this as it documents one of the core assumptions of AV1 CfL.

barrbrain:
> Probably Ok, is it possible that `cost.1 < cost.0` though?

Yes, hence the 2 if blocks and no else-if. I would like to rewrite this loop more expressively; help welcome.

Issue #1572, opened 2019-08-21
Use MaybeUninit properly


YaLTeR:
Now that we switched to `MaybeUninit` in #1292 we might as well start using it properly; right now as far as I can tell it's equivalent to `mem::uninitialized()` in our usage. See [the `MaybeUninit` docs](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html).

shssoichiro:
It appears the only remaining place where this needs to be fixed is in `AlignedArray::uninitialized`.

kornelski:
How to handle `PlaneRegionMut`? It's used in lots of places for `dst: &mut PlaneRegionMut<T>`.

It's relatively easy to remove `T: Pixel` bound on it, and make `PlaneRegionMut<MaybeUninit<T>>` work.

Alternatively, maybe there could be a simpler type that is purely `{width, height, stride}` that could be created ad-hoc from `PlaneRegionMut` and slices? It seems that many places that take `PlaneRegionMut` don't actually use all of the plane config complexity, and just redundantly compute things like `let xsize = (8 >> xdec) as isize;` to reduce it to back basic width/height/stride.

lu-zero:
When do we have a PlaneRegion that is pointing to something not initialized though?

kornelski:
`subpel_diamond_search` makes `PlaneRegionMut` from `let mut buf: Aligned<[T; 128 * 128]> = unsafe { Aligned::uninitialized() }` and then hopes `get_subpel_mv_rd` is initializing it.

PR #1582, opened 2019-08-23
Fix some MaybeUninit usage, move sgrproj arrays to the heap

YaLTeR:
This resolves a part of #1572. There's one notable omission: `UninitializedAlignedArray()` which is used in quite a few places and would need a lot of work to refactor.

AWCY results pending (should be zeros with minimal to no performance drop).


src/tiling/tile_state.rs
```diff
@@ -117,6 +129,13 @@ impl<'a, T: Pixel> TileStateMut<'a, T> {
         })
         .collect(),
       rdo: RDOTracker::new(),
+      stripe_filter_integral_image: vec![0; STRIPE_FILTER_INTEGRAL_IMAGE_SIZE],
```
KyleSiefring: Maybe it would be better to store and pass these in some sort of struct.

YaLTeR: That requires solving the hardest problem in computer science

KyleSiefring: SelfGuidedRestorationScratch maybe

tdaede: Sure, or IntegralImageBuffer. There are only 4 so a struct isn't super critical.

YaLTeR:
Done; gonna do another AWCY just in case.

YaLTeR:
All good.

PR #1392, opened 2019-06-29
WIP: Make flushing idempotent

lu-zero:
Ignore any other input once you start flushing.

Needs to be tested on real files.

PR #1587, opened 2019-08-24
Document Context

### LINKED ISSUE #1075: Document Context

YaLTeR:
Closes #1075.


YaLTeR:
Why does `Context::new_frame()` return `Arc<Frame>`? Wouldn't it make more sense for it to return just `Frame`? As far as I can tell, there's not much useful you can do with it without unwrapping it from the `Arc` first.

src/api/mod.rs
```diff
@@ -702,9 +740,8 @@ impl<T: Pixel> Context<T> {
   /// ctx.send_frame(None);
   /// ```
   ///
-  /// It could return `EncoderStatus::EnoughData` if the incoming
-  /// frame queue is full or `EncoderStatus::Failure` if the per-frame
-  /// parameters set an impossible constraint.
+  /// [`EncoderStatus::EnoughData`]: enum.EncoderStatus.html#variant.EnoughData
+  /// [`EncoderStatus::Failure`]: enum.EncoderStatus.html#variant.Failure
   pub fn send_frame<F>(&mut self, frame: F) -> Result<(), EncoderStatus>
```
shssoichiro: Probably makes sense to `#[inline]` this as well.

lu-zero:
> Why does `Context::new_frame()` return `Arc<Frame>`? Wouldn't it make more sense for it to return just `Frame`? As far as I can tell, there's not much useful you can do with it without unwrapping it from the `Arc` first.

[make_mut](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut) works fine for the purpose.

Ideally we could rework the whole thing so that the Frame is pool-backed and refcounted without an external Arc, but this is sort of future proof enough for now :)

YaLTeR:
> `make_mut` works fine for the purpose.

Eh, I'm not a big fan of this. Us returning an `Arc` kind of implies that we might've stored a reference ourselves, in which case `make_mut` making a clone of the underlying value would potentially break us. It just feels like adding un-clarity to the API for no particular reason.

Adding an `IntoFrame` impl for `Frame` would make much more sense in my opinion.

> Ideally we could rework the whole thing so that the Frame is pool-backed and refcounted without an external Arc

Any particular reason for this over the regular `Arc`?

src/api/mod.rs
```diff
@@ -588,32 +588,52 @@ pub(crate) struct ContextInner<T: Pixel> {
...
 pub struct Context<T: Pixel> {
   inner: ContextInner<T>,
   config: EncoderConfig,
   pool: rayon::ThreadPool,
   is_flushing: bool,
 }
 
+/// Status that can be returned by [`Context`] functions.
+///
+/// [`Context`]: struct.Context.html
```
dwbuiten: Is there a way to cross-link without specifying as HTML-only?

src/api/mod.rs
```diff
@@ -672,6 +692,18 @@ impl<T: Pixel> IntoFrame<T> for (Arc<Frame<T>>, Option<FrameParameters>) {
 }
 
 impl<T: Pixel> Context<T> {
+  /// Allocates and returns a new frame.
+  ///
+  /// # Examples
```
dwbuiten: `s/Examples/Example/`

src/api/mod.rs
```diff
@@ -672,6 +692,18 @@ impl<T: Pixel> IntoFrame<T> for (Arc<Frame<T>>, Option<FrameParameters>) {
...
+  ///
+  /// ```
+  /// use rav1e::prelude::*;
+  ///
+  /// let cfg = Config::default();
+  /// let ctx: Context<u8> = cfg.new_context();
```
dwbuiten: Note: After the #1455 this will need an `unwrap()`.

src/api/mod.rs
```diff
@@ -672,6 +692,18 @@ impl<T: Pixel> IntoFrame<T> for (Arc<Frame<T>>, Option<FrameParameters>) {
...
+  /// let frame = ctx.new_frame();
+  /// ```
+  #[inline]
```
dwbuiten: Why does a documentation PR contain functional (well, to the compiler) changes?

src/api/mod.rs
```diff
@@ -680,7 +712,13 @@ impl<T: Pixel> Context<T> {
     ))
   }
 
-  /// Send the information to the encoder
+  /// Sends the frame for encoding.
+  ///
+  /// It could return [`EncoderStatus::EnoughData`] if the incoming frame queue
+  /// is full or [`EncoderStatus::Failure`] if the per-frame parameters set an
+  /// impossible constraint.
```
dwbuiten: This is not accurate right now, since `EnoughData` is literally never returned, despite existing as a possible status.

src/api/mod.rs
```diff
@@ -680,7 +712,13 @@ impl<T: Pixel> Context<T> {
...
+  ///
+  /// # Examples
```
dwbuiten: `s/Examples/Example/`

src/api/mod.rs
```diff
@@ -725,18 +762,26 @@ impl<T: Pixel> Context<T> {
...
-  /// It is still safe to call this function when receive_packet() returns any
-  /// other error. It will return *None* instead of returning a duplicate copy
-  /// of the previous frame's data.
+  /// It is still safe to call this function when [`receive_packet`] returns
+  /// any other error. It will return `None` instead of returning a duplicate
+  /// copy of the previous frame's data.
+  ///
+  /// [`receive_packet`]: #method.receive_packet
+  /// [`EncoderStatus::Encoded`]: enum.EncoderStatus.html#variant.Encoded
+  /// [`EncoderStatus::LimitReached`]:
```
dwbuiten: The documentation references these statuses being returned by receive_packet, not this function.

src/api/mod.rs
```diff
@@ -745,38 +790,83 @@ impl<T: Pixel> Context<T> {
...
+  /// [`receive_packet`]: #method.receive_packet
+  /// [`twopass_bytes_needed`]: #method.twopass_bytes_needed
+  #[inline]
   pub fn twopass_in(&mut self, buf: &[u8]) -> Result<usize, EncoderStatus> {
     self.inner.rc_state.twopass_in(Some(buf)).or(Err(EncoderStatus::Failure))
   }
 
+  /// Returns the encoded data.
+  ///
+  /// # Examples
```
dwbuiten: `s/Examples/Example/`

src/api/mod.rs
```diff
@@ -745,38 +790,83 @@ impl<T: Pixel> Context<T> {
...
+  /// [`twopass_bytes_needed`] returns `0`.
+  ///
+  /// [`receive_packet`]: #method.receive_packet
+  /// [`twopass_bytes_needed`]: #method.twopass_bytes_needed
+  #[inline]
   pub fn twopass_in(&mut self, buf: &[u8]) -> Result<usize, EncoderStatus> {
     self.inner.rc_state.twopass_in(Some(buf)).or(Err(EncoderStatus::Failure))
   }
 
+  /// Returns the encoded data.
```
dwbuiten: This isn't entirely accurate... or at least, not the full story. Maybe something to mention this call is where the bulk of the encoding work is done.

src/api/mod.rs
```diff
@@ -745,38 +790,83 @@ impl<T: Pixel> Context<T> {
...
+  ///
+  /// # Examples
+  ///
+  /// ``` no_run
+  /// use rav1e::prelude::*;
+  /// #
+  /// # fn main() -> Result<(), EncoderStatus> {
+  /// let cfg = Config::default();
+  /// let mut ctx: Context<u8> = cfg.new_context();
```
dwbuiten: Will need `unwrap()` as noted above, after the tiles PR.

src/api/mod.rs
```diff
@@ -745,38 +790,83 @@ impl<T: Pixel> Context<T> {
...
+  /// let frame = ctx.new_frame();
+  ///
+  /// ctx.send_frame(frame)?;
+  /// ctx.flush();
+  ///
+  /// loop {
+  ///     match ctx.receive_packet() {
+  ///         Ok(packet) => { /* ... */ },
+  ///         Err(EncoderStatus::Encoded) => (),
+  ///         Err(EncoderStatus::LimitReached) => break,
```
dwbuiten: This is missing a bunch of possible stauses.

src/api/mod.rs
```diff
@@ -745,38 +790,83 @@ impl<T: Pixel> Context<T> {
...
     let inner = &mut self.inner;
     let pool = &mut self.pool;
 
     pool.install(|| inner.receive_packet())
   }
 
+  /// Flushes the encoder.
+  ///
+  /// Flushing signals the end of the video. After the encoder has been
+  /// flushed, no additional frames are accepted.
```
dwbuiten: Is this stuff true? I think @lu-zero wanted to make mid-stream flushing work.

src/api/mod.rs
```diff
@@ -745,38 +790,83 @@ impl<T: Pixel> Context<T> {
...
+  #[inline]
   pub fn flush(&mut self) {
     self.send_frame(None).unwrap();
   }
 
+  /// Produces a sequence header matching the current encoding context.
+  ///
+  /// Its format is compatible with the AV1 Matroska and ISOBMFF specification.
```
dwbuiten: You should add a note that it does not contain the needed config OBUs for e.g. color, which are required for some uses as part of the Matroska and ISOBMFF specifications, but are technically optional.

src/api/mod.rs
YaLTeR: Not that I know of. This is the way recommended by the API guidelines: https://rust-lang-nursery.github.io/api-guidelines/documentation.html#prose-contains-hyperlinks-to-relevant-things-c-link

src/api/mod.rs
lu-zero: I do not have a strong opinion on this, supporting it shouldn't be much code but the documentation is matching the current status.

src/api/mod.rs
YaLTeR: It is always `# Examples` and `# Panics`, even if one example. Check the standard library docs, e.g. https://doc.rust-lang.org/std/vec/struct.Vec.html#method.capacity

src/api/mod.rs
dwbuiten: Ah, that sucks...

src/api/mod.rs
dwbuiten: Sounds pretty stupid to me - is this actully documted as "you need to do this for rustdocs" anywhere, or is it just "stdlib does this"?

src/api/mod.rs
dwbuiten: If it matches current behavior, it's fine.

src/api/mod.rs
YaLTeR: As far as I can tell, these two are the "success" statuses, and all other ones are "error", so I return them (similarly to how you use `?` in the examples to return errors).

src/api/mod.rs
YaLTeR: What do you mean? It returns `EnoughData` right here:
https://github.com/xiph/rav1e/blob/dc34123f5316c66547d3cfbfb86855991db1a1ec/src/api/mod.rs#L722

src/api/mod.rs
dwbuiten: I wouldn't call `NeedMoreData` an error; it's a normal part of the encoding flow.

src/api/mod.rs
YaLTeR: I was under the impression that he PRd the opposite not too long ago: https://github.com/xiph/rav1e/pull/1392

src/api/mod.rs
YaLTeR: Could you elaborate? I'm not familiar with this, I just copied the capi docs.

src/api/mod.rs
dwbuiten: Ah, it certainly never used to until a few weeks back.

However... now it seems that `EnoughData` is used during flushing, and still not for when the incoming frame queue is full (which to my knowledge, cannot happen at the moment).

src/api/mod.rs
YaLTeR: Not after a `flush`, which I did in this example.

src/api/mod.rs
YaLTeR: Not sure it's spelled out anywhere as a guideline, but as far as I know it's pretty standard. Besides, it means no changing the line if examples are added or removed, and (provided that you can link to subsections, never tried that actually) consistent linking if you want to link to an example.

src/api/mod.rs
YaLTeR: I should split the inlines into a separate PR then.

src/api/mod.rs
dwbuiten: As per: https://aomediacodec.github.io/av1-isobmff/#av1codecconfigurationbox

The last entry, `configOBUs` contains zero or more, well, config OBUs (sequence header with color config, for example). rav1e currently doesnt return any of these (so, zero) as part of this function. Other things, like libavcodec/libavformat put them in there.

They're needed for e.g. generating a portable MSE codec string for use in a DASH manifest.

Just think it's worth a note that we don't include them.

src/api/mod.rs
dwbuiten: OK. They're not wrong, I was just wondering why they're in this PR.

src/api/mod.rs
dwbuiten: You're right, but that's not a particularily good example then.

YaLTeR: I should add another example which demonstrates `NeedMoreData`. This one can stay because is probably the simplest way you can use `receive_packet()`, and also it's ok for something like image encoding.

dwbuiten: That sounds fine to me.

src/api/mod.rs
lu-zero: > > `make_mut` works fine for the purpose.
> 
> Eh, I'm not a big fan of this. Us returning an `Arc` kind of implies that we might've stored a reference ourselves, in which case `make_mut` making a clone of the underlying value would potentially break us. It just feels like adding un-clarity to the API for no particular reason.
> 

We could add explicit setters/getters.

> Adding an `IntoFrame` impl for `Frame` would make much more sense in my opinion.
> 

We could add it easily, but you have to support Arc<Frames> from the users as well anyway.

> > Ideally we could rework the whole thing so that the Frame is pool-backed and refcounted without an external Arc
> 
> Any particular reason for this over the regular `Arc`?

Not many beside maybe a different ergonomy compared to what Arc provides.

I'm not against adding the `Frame` variant for IntoFrame and simplify the frame factory since it is not pool backed right now and probably it won't be in a long while.

dwbuiten: @tdaede I mentioned it above:

> They're needed for e.g. generating a portable MSE codec string for use in a DASH manifest.

That is, the *full* MSE codec string, which makes sure color is set properly.

PR #1588, opened 2019-08-24
Unbreak building with --no-default-features

### LINKED ISSUE #1586: cargo build --no-default-features failing
When trying to build without any features, it fails with a few errors:

```
error: A non-empty glob must import something with the glob's visibility
  --> src/cdef.rs:23:9
   |
23 | pub use self::native::*;
   |         ^^^^^^^^^^^^^^^

error[E0425]: cannot find value `CDEF_FILTER_FNS` in this scope
  --> src/asm/cdef.rs:76:13
   |
76 |       match CDEF_FILTER_FNS[cpu.as_index()][decimate_index(xdec, ydec)] {
   |             ^^^^^^^^^^^^^^^ not found in this scope
```

and some more that are similar. Looks like this was introduced with 79019d32fc101341ca840e0208a07fdb17b1023b. Seems like we need to add some more `cfg` checks to some parts of that code but I don't have enough knowledge of this code to do it myself.

Found it while building gst-plugins-rs that uses this.

lu-zero:
Fixes #1586


KyleSiefring:
lgtm

PR #1589, opened 2019-08-24
[WIP] Add metrics from AWCY into rav1e

shssoichiro:
This includes the following:
- PSNR (correct calculation)
- APSNR (this is what we were previously display as PSNR)
- PSNR HVS
- SSIM
- MSSSIM
- CIEDE2000

This does not include:
- Fast SSIM (because it's a bad metric)
- VMAF (because this would either require FFI or significant
time investment to port to Rust)

This is enabled by a `--metrics` CLI flag. Without the flag, it performs no calculations, so there is no speed penalty unless explicitly requested. For backwards compatibility, the previous `--psnr` flag remains, which will compute only the PSNR.


shssoichiro:
Current benchmarks using the included clip (1 frame 480x272):

```
metrics_psnr_bench      time:   [923.42 us 926.54 us 930.07 us]                               
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  5 (5.00%) high severe

metrics_psnrhvs_bench   time:   [3.6841 ms 3.6990 ms 3.7190 ms]                                   
  3 (3.00%) high mild
  2 (2.00%) high severe

metrics_ssim_bench      time:   [13.918 ms 13.942 ms 13.968 ms]                                
Found 7 outliers among 100 measurements (7.00%)
  5 (5.00%) high mild
  2 (2.00%) high severe

metrics_msssim_bench    time:   [20.813 ms 20.837 ms 20.859 ms]                                  
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe

metrics_ciede_bench     time:   [110.53 ms 110.63 ms 110.74 ms]                                 
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
```

dwbuiten:
1. Is the plan to use these in AWCY
2. Why was @KyleSiefring's CIEDE stuff not used (it even has SIMD I think)?
3. Why is `calculate_frame_metrics` exposed as an API in what will be the rav1e crate? This is general DSP stuff and does not belong as an API in an encoding library.

shssoichiro:
1. No, but if split into external crates they could be.
2. I didn't know a Rust implementation of CIEDE existed.
3. It's exposed so that the rav1e binary can use it. Possibly it makes more sense to move all the metrics code to the `bin` side of rav1e?

lu-zero:
>     1. No, but if split into external crates they could be.

If you want help we could have them in the @rust-av org, I'm sure I'll use them there.

>     2. I didn't know a Rust implementation of CIEDE existed.

https://github.com/KyleSiefring/dump_ciede2000 but it also should be make a library crate.

>     3. It's exposed so that the rav1e binary can use it. Possibly it makes more sense to move all the metrics code to the `bin` side of rav1e?

This is also an option, splitting the project in a workspace and have lib and cli crates clearly separated might be something we should think about again.

> The goal was to have a way of exposing these metrics locally, directly within rav1e, for quick comparison of speed levels, tunings, or code changes without having to upload to AWCY every time. This is not meant to replace AWCY either. AWCY gives more detailed comparisons, but it's slow, so not always ideal when you're tuning something.

I do like a lot the idea.

dwbuiten:
> 3. It's exposed so that the rav1e binary can use it. Possibly it makes more sense to move all the metrics code to the bin side of rav1e?

Either that, or its own crate, IMO. They're generally useful.

shssoichiro:
Will reopen with the new approach once ready

PR #1591, opened 2019-08-24
[NFC] Attach different lifetimes to the tile block struct and its data.

DiamondLovesYou:
So the tile block reference doesn't need to live as long as the data it "holds".


rom1v:
In theory, it releases some unnecessary constraints. In practice, the lifetime parameters are exposed to the type, so every usage of the type needs to pass all of them (at least as `'_`).

So in the general case, instead of:
```rust
struct MyView<'a> {
    pub w: &'a W,
    pub x: &'a X,
    pub y: &'a Y,
    pub z: &'a Z,
}
```

You would get:
```rust
struct MyView<'a, 'b, 'c, 'd> {
    pub w: &'a W,
    pub x: &'b X,
    pub y: &'c Y,
    pub z: &'d Z,
}
```

And passing that type to a function:
```rust
fn f(v: &MyView<'_, '_, '_, '_>) { … }
```

Just for information, I initially implemented it with 2 lifetimes (with less constraints though) in `tiling.78` (70bd1e5571a5069a7bba827a7530fcc73b216f53). I decided not to, because I considered it was not worth it, and change it in `tiling.79` (2883c68120f27e810f66a367335fae556bf87c7b).

But of course, if there is a benefit, we could change it.

PR #1593, opened 2019-08-26
Pre compute the simd functions 

lu-zero:
- [ ] Create a type alias for each `fn`
- [ ] Create a struct to hold them
- [ ] Add a `Default` and a `new(cpu: CpuFeatureLevel)`
- [ ] Put the struct someplace reachable (Config)
- [ ] Add inline helpers to hide the manual dispatch

### DIFF
--- a/src/asm/mod.rs
+++ b/src/asm/mod.rs
@@ -9,5 +9,5 @@
 
 #[cfg(all(target_arch = "x86_64", feature = "nasm"))]
 pub mod cdef;
-#[cfg(all(target_arch = "x86_64", feature = "nasm"))]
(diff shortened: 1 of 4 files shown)


lu-zero:
We should check if [pin](https://doc.rust-lang.org/std/pin/struct.Pin.html) can be used to give some warranty in that direction. 

I'm not keen in having too many branches over the hot path, x86 might or might not soak them away, but other architectures might have serious degradation because of it. (obviously we should benchmark first, see what happens and document it)

lu-zero:
It does something different but I think what we have is good enough

PR #1599, opened 2019-08-26
Fix doc tests.

tdaede:


lu-zero:
shouldn't it be unwrap?

lu-zero:
Let's land it with the typo fixed :)

PR #1600, opened 2019-08-27
[Don't merge yet] Avoid duplicated Computation of integral images for Self Guided Filte…

ycho:
…r RDO

Avoid a duplicated computation of integral images for
RDO of Self Guided Filter in Loop Restoration Filter.


lu-zero:
Looks nice :) it needs a rebase though.

ycho:
#1697 has resolved this issue in a better shape, so I am happily closing this! Thanks, @xiphmont !

PR #1602, opened 2019-08-27
Unbreak capi

YaLTeR:


dwbuiten:
@nox @lu-zero just want to double check this doesnt make stuff the ref counting stuff break again

PR #1605, opened 2019-08-28
Update Gitlab CI script with stages similar to Travis

shssoichiro:
This is more optimized than the travis script, because Gitlab:
- Allows us to use a custom docker image which already has all our dependencies
- Runs the build in stages, allowing us to reuse the results from one stage in a future stage or fail fast

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,76 +1,164 @@
 variables:
-    CARGO_HOME: $CI_PROJECT_DIR/.cargo
-    DOCKER_IMAGE_TAG: '1.36'
+  CARGO_HOME: $CI_PROJECT_DIR/.cargo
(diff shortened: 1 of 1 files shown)


lu-zero:
Note: to use this somebody should ask to have @xiph to enter the beta program.

shssoichiro:
The Gitlab pipeline is now working again: https://gitlab.xiph.org/xiph/rav1e/pipelines/1240

With a few limitations:
1. It does not currently push to Coveralls. The majority of the setup is in place. When we are ready for Gitlab to push to Coveralls, we will need to find the Coveralls repo_token and add it as a secret in Gitlab for the rav1e repository.
1. Our Gitlab instance currently only has one runner active at a time. This means there is no parallel running of jobs, which causes the whole pipeline to take... a while.

PR #1609, opened 2019-08-29
Map an input maximum keyframe interval of 0 to "infinity" (u64::MAX)

rzumer:
Opening this for discussion.

The goal of this PR is to provide a simple interface to set an infinite keyframe interval (i.e. keyframes inserted only on scene changes), as opposed to having users pass the maximum value of an unsigned 64-bit value themselves.

The alternative solution would be to add a new option that overrides the keyframe interval values provided, but in order to support this through the C API a new encoder config parameter would be needed, so overall it seems unnecessarily complex to me.

In order to avoid duplicating logic between the C API and the CLI, a function is added to the encoder config. It takes both minimum and maximum keyframe interval values in case we want to map more special cases in the future (e.g. auto interval when the minimum interval input value is 0).

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -209,6 +209,17 @@ impl EncoderConfig {
     }
   }
 
+  /// Sets the minimum and maximum keyframe interval, handling special cases as needed.
+  pub fn set_key_frame_interval(
+    &mut self, min_interval: u64, max_interval: u64,
+  ) {
+    self.min_key_frame_interval = min_interval;
+
+    // Map an input value of 0 to an infinite interval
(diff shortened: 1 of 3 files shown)


tdaede:
I think this is OK as long as 0 is separate from the default. Accidentally setting the default to infinity makes really obnoxiously broken files that plagued early versions of libvpx.

rzumer:
> I think this is OK as long as 0 is separate from the default. Accidentally setting the default to infinity makes really obnoxiously broken files that plagued early versions of libvpx.

The default is 240 (in default `EncoderConfig` implementation and in the CLI parser).

PR #1619, opened 2019-08-30
Handle integer overflows and other stuff

### LINKED ISSUE #1616: Flushing without any frames results in NeedMoreData
```rust
let config = Config::default();
let mut ctx: Context<u8> = config.new_context().unwrap();
ctx.flush();
let rv = ctx.receive_packet();
```
Here `rv` ends up being `Err(EncoderStatus::NeedMoreData)`. I'm not sure what exactly should happen here (I'm guessing it should return `LimitReached`? Is there some kind of an "empty packet" it should produce in this case?), but it's definitely not `NeedMoreData`.

YaLTeR:
I fuzzed things. Things broke.

Fixes #1616.

This changes the type of `rdo_lookahead_frames` in `EncoderConfig` from `u64` to `usize`. API change; merge this before 0.1 if possible?


src/api/mod.rs
```diff
@@ -459,8 +459,8 @@ pub struct Config {
   pub threads: usize,
 }
 
-fn check_tile_log2(n: usize) -> bool {
-  ((1 << TilingInfo::tile_log2(1, n)) - n) == 0 || n == 0
+fn check_tile_log2(n: usize) -> Option<bool> {
+  Some(((1 << TilingInfo::tile_log2(1, n)?) - n) == 0 || n == 0)
```
shssoichiro: What is the distinction between `None` and `Some(false)` here? If they mean the same thing, I'd suggest keeping this a `bool` and returning `false` instead of `None`.

YaLTeR: Good point

src/api/mod.rs
```diff
@@ -3862,4 +3872,74 @@ mod test {
     }
     assert_eq!(ctx.inner.frames_processed, LIMIT as u64);
   }
+
+  #[test]
```
shssoichiro: I love seeing new tests

YaLTeR: Find crash => add test => add fix :smiley:

src/api/mod.rs
```diff
@@ -494,10 +494,19 @@ impl Config {
 
     let mut config = self.enc.clone();
 
-    if !check_tile_log2(config.tile_cols) {
+    if config.width == 0 || config.height == 0 {
       return Err(EncoderStatus::Failure);
     }
-    if !check_tile_log2(config.tile_rows) {
+    if config.rdo_lookahead_frames == usize::max_value() {
```
lu-zero: I'd use a `max(MAX_RDO_LOOKAHEAD_FRAMES)`

src/api/mod.rs
```diff
@@ -3862,4 +3872,74 @@ mod test {
     }
     assert_eq!(ctx.inner.frames_processed, LIMIT as u64);
   }
+
+  #[test]
+  fn zero_frames() {
```
lu-zero: We should spin out a `api/test.rs`

YaLTeR:
Fixed, rebased, etc.

src/encoder.rs
```diff
@@ -160,11 +160,12 @@ pub struct Sequence {
 }
 
 impl Sequence {
-  pub fn new(config: &EncoderConfig) -> Sequence {
+  pub fn new(config: &EncoderConfig) -> Option<Sequence> {
```
lu-zero: I'd keep the assert and move the check (using ilog maybe) where the configuration is validated. The sequence is an internal detail and should never fail no matter the input.

PR #1621, opened 2019-08-31
Cpu feature level rework

lu-zero:
- [x] Use Default
- [x] Hide completely the cpu feature from the API
   - [x] Move the field in FrameInvariants

(as discussed with Tim on IRC)

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -21,7 +21,6 @@ use serde_derive::{Deserialize, Serialize};
 
 use crate::context::*;
 use crate::context::{FrameBlocks, SuperBlockOffset, TileSuperBlockOffset};
-use crate::cpu_features::*;
 use crate::dist::get_satd;
 use crate::encoder::*;
 use crate::frame::*;
@@ -100,8 +99,6 @@ pub struct EncoderConfig {
   pub still_picture: bool,
 
   // encoder configuration
-  /// Target CPU feature level.
-  pub cpu_feature_level: CpuFeatureLevel,
   /// Video time base.
   pub time_base: Rational,
   /// The *minimum* interval between two keyframes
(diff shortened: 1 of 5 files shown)


KyleSiefring:
I suggest you discuss this change with derf, since he is the one who mentioned using environmental variables in theora. I'm not entirely opposed to this change since it makes it possible to use with AWCY, but I don't actually want to average user to have access to this function.

One concern I have is with the ABI. In the future, we want to be able to add and remove different cpu feature levels. Will there be breakage if we do this?

lu-zero:
> I suggest you discuss this change with derf, since he is the one who mentioned using environmental variables in theora. I'm not entirely opposed to this change since it makes it possible to use with AWCY, but I don't actually want to average user to have access to this function.

Currently we have a situation in which the env approach seems to not work and we want the user to report back if something works or not with avx2 turned off.

Not sure which is the function you are wary of.

> One concern I have is with the ABI. In the future, we want to be able to add and remove different cpu feature levels. Will there be breakage if we do this?

Until we have dynamic libraries, our concern is on the API stability :)

If you want to mix and match multiple implementations at runtime then we'd either have to extend the enum with all the supported combinations or we'd have to use a struct of booleans.

In general I guess I'll have to discuss with Tim a lot since if we want to do that it would mean multiple branches in hot paths and that feels more problematic that what is already present.

moisesmcardona:
Unfortunately, it seems to not be helping. Both Native and AVX2 fails.

image

lu-zero:
> I suggest you discuss this change with derf, since he is the one who mentioned using environmental variables in theora. I'm not entirely opposed to this change since it makes it possible to use with AWCY, but I don't actually want to average user to have access to this function.

Derf wants to have a debug-only facility so I removed it from the API completely.

PR #1620, opened 2019-08-31
Enhanced CLI stats

shssoichiro:
Closes #1595

Example of verbose output:
```
[INFO  rav1e] Using y4m decoder: 1280x720p @ 60/1 fps, 4:2:0, 8-bit
[INFO  rav1e] CPU Feature Level: AVX2
[INFO  rav1e] Encoding settings: keyint_min=12 keyint_max=240 quantizer=100 bitrate=0 min_quantizer=0 low_latency=false tune=Psychovisual tiles=0 tile_rows=0 tile_cols=0 rdo_lookahead_frames=40 min_block_size=4x4 multiref=true fast_deblock=false reduced_tx_set=false tx_domain_distortion=true tx_domain_rate=false encode_bottomup=true rdo_tx_decision=true prediction_modes=Complex-All include_near_mvs=true no_scene_detection=false diamond_me=true cdef=true quantizer_rdo=true use_satd_subpel=true
[INFO  rav1e] Input Frame 0 - Key frame - 59872 bytes - encoded 1 frames, 0.003 fps, 28738.56 Kb/s
[INFO  rav1e] Input Frame 1 - Inter frame - 27367 bytes - encoded 2 frames, 0.003 fps, 20937.36 Kb/s
...
[INFO  rav1e] Input Frame 59 - Inter frame - 2068 bytes - encoded 60 frames, 0.011 fps, 3265.33 Kb/s
[INFO  rav1e::stats] ----------
[INFO  rav1e::stats] Key frame:             1 | avg QP:  79.00 | avg size:   59872 B
[INFO  rav1e::stats] Inter frame:          59 | avg QP: 117.29 | avg size:    5903 B
[INFO  rav1e::stats] Intra only frame:      0 | avg QP:   0.00 | avg size:       0 B
[INFO  rav1e::stats] Switching frame:       0 | avg QP:   0.00 | avg size:       0 B
[INFO  rav1e::stats] ----------
[INFO  rav1e::stats] bsize I:   x128    x64    x32    x16     x8     x4
[INFO  rav1e::stats]    128x:   0.0%   0.0%                              
[INFO  rav1e::stats]     64x:   0.0%   0.0%   0.4%   0.0%
[INFO  rav1e::stats]     32x:          0.3%   0.1%   1.5%   0.0%
[INFO  rav1e::stats]     16x:          0.0%   1.3%   1.1%   6.5%   0.0%
[INFO  rav1e::stats]      8x:                 0.0%   5.6%   6.8%  21.4%
[...]

PR #1622, opened 2019-08-31
Use the log crate instead of eprintln! statements

shssoichiro:
Split from #1620

The log level can be toggled using the RUST_LOG
environment variable, e.g. set to `RUST_LOG=debug`
for local debugging. This also enables us to add
e.g. debug! statements throughout the crate that
we wouldn't want users to see in production, but
would generally be useful to a rav1e developer.

### DIFF
--- a/src/bin/rav1e.rs
+++ b/src/bin/rav1e.rs
@@ -28,12 +28,12 @@ use rav1e::prelude::*;
 use crate::decoder::Decoder;
 use crate::decoder::VideoDetails;
 use crate::muxer::*;
+use std::env;
 use std::fs::File;
 use std::io::Read;
 use std::io::Seek;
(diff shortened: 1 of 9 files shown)


src/bin/rav1e.rs
```diff
@@ -208,27 +211,40 @@ fn do_encode<T: Pixel, D: Decoder>(
...
+  init_logger();
 
   match run() {
     Ok(()) => {}
     Err(e) => error::print_error(&e),
   }
 }
 
+fn init_logger() {
+  if env::var("RUST_LOG").is_err() {
```
shssoichiro: @YaLTeR pretty_env_logger doesn't provide a facility for setting a default log level internally--it's intended to work with the env variable.

PR #1623, opened 2019-09-02
Fix incorrect reg count for satd assembly

KyleSiefring:
Not really sure if this will fix the bug we have on windows.


moisesmcardona:
Unfortunately, the issue is still present.

PR #1624, opened 2019-09-02
Correct number of registers used in satd8x8

KyleSiefring:
I probably should have held out on the previous patch but eh.

I suspect this will fix the bug.


moisesmcardona:
I'm already building it. Will report back.

moisesmcardona:
Yes, this PR fixes the problem on Windows.

Issue #1015, opened 2019-02-19
Document & distribute speed functions


EwoutH:
I encoded 120 frames of the [Morocco](https://www.youtube.com/watch?v=hVvEISFw9w0&t=2s) clip with rav1e f2aa4c6 from speed 6 to 0 and found some [strange results](https://docs.google.com/spreadsheets/d/1JpPz-Tvj8OO9VGyv-0_VQQRlMVpChuwYeRwr3L3Ss1E/edit#gid=1747162432):

- Speed 6 and 5 are completely identical in result, while the latter is somewhat slower (probably measurement error).
- Speed 2 produces larger files than speed 3 while 4% slower.
- Speed 1 is about equal size to speed 3 while 14% slower.

![rav1e-fps-size](https://user-images.githubusercontent.com/15776622/53012233-b1623500-3442-11e9-8e38-9819d6d755ea.png)

This was of course a single, short clip, tonight I will try some more with different scene's and motion patterns. 

I think two main changes are needed:
- Document speed features. I would propose a large table with the different speed settings on the X-axis and the different speed features on the Y-axis, where each cell states if it's enabled if not
- Better distribute speed functions over the different settings. The gap between 1 and 0 and between 5 and 4 is way to large, while the others are very small.

EwoutH:
Based on [api.rs](https://github.com/xiph/rav1e/blob/master/src/api.rs) I've composed the [following table](https://docs.google.com/spreadsheets/d/1Z_nkVZqc6CZeFBtFptugzspX-7-qf9nC1pHqfJ4mE60/edit?usp=sharing). Can someone confirm this is correct?

|                      |    -s 0    |    -s 1    |       -s 2       |       -s 3       |  -s 4  |  -s 5  |  -s 6  |  -s 7  |  -s 8  |  -s 9  |  -s 10 |
|----------------------|:----------:|:----------:|:----------------:|:----------------:|:------:|:------:|:------:|:------:|:------:|:------:|:------:|
| min_block_size       |     4X4    |     4X4    |        8X8       |       16X16      |  32X32 |  64X64 |  64X64 |  64X64 |  64X64 |  64X64 |  64X64 |
| multiref             |    TRUE    |    TRUE    |       TRUE       |       FALSE      |  FALSE |  FALSE |  FALSE |  FALSE |  FALSE |  FALSE |  FALSE |
[...]

PR #1628, opened 2019-09-03
Fix documentation for speed levels 9 and 10

shssoichiro:
The speed settings had reduced TX set enabled, but it was not
documented as such.


EwoutH:
Also updated [the table](https://docs.google.com/spreadsheets/d/1Z_nkVZqc6CZeFBtFptugzspX-7-qf9nC1pHqfJ4mE60/edit?usp=sharing) (from #1015)

PR #1630, opened 2019-09-04
Fix some more things found by fuzzing

YaLTeR:
Leaving this as a draft because I want to collect more potential fixes from further fuzzing... Although currently both of my fuzzing runs are blocked on #1611.

Fixes #1631.

Issue #1611, opened 2019-08-29
`assertion failed: sbo.0.x % (1 << sb_shift) == 0` since earlier this week


sdroege:
If you run `cargo test` in
https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/tree/master/gst-plugin-rav1e
it will fail with panics

```
thread 'test_encode_y444' panicked at 'assertion failed: sbo.0.x % (1 << sb_shift) == 0', /home/slomo/.cargo/git/checkouts/rav1e-14e16d25dd5cd9c0/aa4271e/src/tiling/tile_restoration_state.rs:369:1
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:200
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:477
   7: std::panicking::begin_panic
             at /rustc/eae3437dfe991621e8afdc82734f4a172d7ddf9b/src/libstd/panicking.rs:411
   8: rav1e::tiling::tile_restoration_state::TileRestorationStateMut::new
             at /home/slomo/.cargo/git/checkouts/rav1e-14e16d25dd5cd9c0/aa4271e/src/tiling/tile_state.rs:0
   9: rav1e::tiling::tile_state::TileStateMut<T>::new
             at /home/slomo/.cargo/git/checkouts/rav1e-14e16d25dd5cd9c0/aa4271e/src/tiling/tile_state.rs:110
  10: <rav1e::tiling::tiler::TileContextIterMut<T> as core::iter::traits::iterator::Iterator>::next
             at /home/slomo/.cargo/git/checkouts/rav1e-14e16d25dd5cd9c0/aa4271e/src/tiling/tiler.rs:187
  11: alloc::vec::Vec<T>::extend_desugared
[...]

PR #1632, opened 2019-09-05
Add fuzzing support

### LINKED ISSUE #115: Add fuzz testing
Fuzzing raw input is generally easier, but fuzzing the decisions would be more useful (but probably more involved/intrusive).

Some potentially helpful links:

https://github.com/pingcap/fail-rs
https://rust-fuzz.github.io/book/introduction.html
https://github.com/BurntSushi/quickcheck
https://github.com/altsysrq/proptest

and might as well add this link from @lu-zero for posterity:
https://doc.rust-lang.org/std/num/struct.Wrapping.html

YaLTeR:
Find assertion fails in your own rav1e clone in one simple command.

Closes #115, although new issues could be opened for adding, for example, proptest, or the randomizing RDO decisions thing.

The design I chose here (putting actual fuzzable functions inside the `rav1e` crate) should make it easy to hook up other fuzzers such as AFL and honggfuzz (more fuzzers more better), although it doesn't make it easy to hook up proptest or quickcheck because those frameworks want you to generate each individual parameter through them, rather than feeding in a chunk of random bytes (this is so that they can meaningfully shrink and expand the inputs, which is the whole point).

Additionally included is the ability to inspect crashing state straight through the fuzz target executables via the `RUST_LOG=debug` environment variable, which is extremely handy for figuring out what's up and making test cases.

Thus far trophies include everything fixed in #1619 and #1630 and a more-interesting case for #1611.

### DIFF
--- a/src/fuzzing.rs
+++ b/src/fuzzing.rs
@@ -1,0 +1,216 @@
+use std::sync::Arc;
+
+use arbitrary::*;
+
+use crate::prelude::*;
+
+// Adding new fuzz targets
+//
+// 1. Add a function to this file which looks like this:
+//
+//    pub fn fuzz_something(data: &[u8]) {
+//      let mut g = create_generator!();
+//
+//      // Invoke everything you need.
+//      //
+//      // You should use g.g() to get an arbitrary value of any type that
+//      // implements Arbitrary [1]. This is how fuzzer affects the
+//      // execution—by feeding in different bytes, which result in different
+//      // arbitrary values being generated.
+//      // [1]: https://docs.rs/arbitrary/0.2.0/arbitrary/trait.Arbitrary.html
+//      //
+//      // Print out the structures you create with arbitrary data with
+//      // debug!().
+//    }
+//
+// 2. cargo fuzz add something
+// 3. Copy the contents of any other .rs file from fuzz/fuzz_targets/ into the
+//    newly created fuzz/fuzz_targets/something.rs and change the function
+//    being called to fuzz_something.
+//
+// Now you can fuzz the new target with cargo fuzz.
+
+// A helper for generating arbitrary data.
+struct Generator<'a> {
+  buffer: RingBuffer<'a>,
+}
+
+impl<'a> Generator<'a> {
+  fn new(data: &'a [u8]) -> Result<Self, BufferError> {
+    Ok(Self { buffer: RingBuffer::new(data, data.len())? })
+  }
+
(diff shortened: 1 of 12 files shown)


src/fuzzing.rs
```diff
@@ -0,0 +1,227 @@
...
+  config.enc.bitrate = g.g();
+  config.enc.tile_cols = g.g();
+  config.enc.tile_rows = g.g();
+  config.enc.tiles = g.g();
+  config.enc.rdo_lookahead_frames = g.g();
+  config.enc.speed_settings = SpeedSettings::from_preset(g.g());
+  config.enc.show_psnr = g.g();
+  config.enc.train_rdo = g.g();
+
+  if debug {
```
tmatth: Instea of having these `if debug` blocks, would it possible to wrap this in a macro or logging function that is a noop if the debug flag is disabled?

shssoichiro: ^ Now that we have the `log` crate, you can use the `debug!` macro instead of `println!`.

shssoichiro: ^ Now that we have the log crate in rav1e, you can use the `debug!` macro in place of `println!` to do this.

YaLTeR:
Changed to use `debug!()`.

src/test_encode_decode/mod.rs
```diff
@@ -38,7 +41,7 @@ fn fill_frame<T: Pixel>(ra: &mut ChaChaRng, frame: &mut Frame<T>) {
   }
 }
 
-pub(crate) fn read_frame_batch<T: Pixel>(
+fn read_frame_batch<T: Pixel>(
```
tmatth: Just out of curiosity, why did the visibility of this function change? Not seeing it called anywhere.

YaLTeR: Well, it is perhaps out of scope of this PR, but I noticed that it didn't need `pub(crate)` so I switched it to `pub`.

tmatth: I think the preference is for `pub(crate)` as we want to expose less API outside of the crate.

YaLTeR: Wait. I switched it to just `fn`, it's less visibility than `pub(crate)`.

tmatth: > Wait. I switched it to just `fn`, it's less visibility than `pub(crate)`.

Aha, I should have read your code and not your comment.

PR #1634, opened 2019-09-05
quantize: Bring back the very slightly smaller inter biasing

dwbuiten:
I had originally merged them since they were only 1/256 different, but it seems it actually does make a difference.

AWCY Link: https://beta.arewecompressedyet.com/?job=master-cc170c2a17675aae6fdcfd12ecc97eaeba90b974&job=deadzoning1%402019-09-02T14%3A31%3A02.785Z

Based on that AWCY link, do you guys think it's worth bringing them back / merging this PR?


YaLTeR:
I don't see why not.

PR #1638, opened 2019-09-06
Simplify deadzoning of trailing coefficients

barrbrain:
Total encode time reduction of ~3%. No metrics change.

### DIFF
--- a/src/quantize.rs
+++ b/src/quantize.rs
@@ -256,18 +256,15 @@ impl QuantizationContext {
   {
     // Find the last non-zero coefficient using our smaller biases and
     // zero everything else.
-    let mut is_zero = true;
-    let mut pos = coded_tx_size - 1;
-    while is_zero && pos != 1 {
-      let c = coeffs[pos] << (self.log_tx_scale as usize);
-      let qc = c + (c.signum() * T::cast_from(self.ac_offset_eob));
-      is_zero =
-        T::cast_from(divu_pair(qc.as_(), self.ac_mul_add)) == T::cast_from(0);
-      pos -= 1;
-    }
-
+    // This threshold is such that `abs(coeff) < deadzone` implies:
+    // (abs(coeff << log_tx_scale) + ac_offset_eob) / ac_quant == 0
+    let deadzone = (self.ac_quant as usize - self.ac_offset_eob as usize)
+      .align_power_of_two_and_shift(self.log_tx_scale)
+      as i32;
+    let pos =
+      coeffs[1..coded_tx_size].iter().rposition(|c| c.as_().abs() >= deadzone);
     // We skip the DC coefficient since it has its own quantizer index.
-    let last_pos = if is_zero { pos } else { pos + 1 };
+    let last_pos = pos.map(|pos| pos + 1).unwrap_or(1);
 
     qcoeffs[0] = coeffs[0] << (self.log_tx_scale as usize);
     qcoeffs[0] += qcoeffs[0].signum() * T::cast_from(self.dc_offset);


src/quantize.rs
```diff
@@ -256,18 +256,13 @@ impl QuantizationContext {
...
-    }
-
+    let deadzone = ((1 << self.log_tx_scale) - 1 + self.ac_quant as i32
+      - self.ac_offset_eob)
+      >> self.log_tx_scale;
+    let pos =
+      coeffs[1..coded_tx_size].iter().rposition(|c| c.as_().abs() >= deadzone);
     // We skip the DC coefficient since it has its own quantizer index.
-    let last_pos = if is_zero { pos } else { pos + 1 };
+    let last_pos = pos.map(|pos| pos + 1).unwrap_or(1);
```
dwbuiten: What type does `pos` end up that this weird rigamarole is needed?

src/quantize.rs
```diff
@@ -256,18 +256,13 @@ impl QuantizationContext {
...
-      let c = coeffs[pos] << (self.log_tx_scale as usize);
-      let qc = c + (c.signum() * T::cast_from(self.ac_offset_eob));
-      is_zero =
-        T::cast_from(divu_pair(qc.as_(), self.ac_mul_add)) == T::cast_from(0);
-      pos -= 1;
-    }
-
+    let deadzone = ((1 << self.log_tx_scale) - 1 + self.ac_quant as i32
+      - self.ac_offset_eob)
+      >> self.log_tx_scale;
```
dwbuiten: So my guess as to whats going on here is that youre converting a quantizer adjusted bias into a non-adjusted threshold?

I find the old code pretty straightforward, but the new one is pretty unclear to me... at the very least it needs a comment to explain this line (I got lost in the middle...).

Also AFAICT this is undoing the quant adjustment done in init, and is a constant? If so, may as well just do it once in init (ignoring the fact that encoder.rs is dumb and reinits a new Quantizer object on every single block... lol).

src/quantize.rs
shssoichiro: Option

src/quantize.rs
barrbrain: In short, the `is_zero` condition is logically equivalent between these two but it might help to spell out the relationship.

barrbrain: The following sequence of transformations is how I arrived at this expression:
```Rust
divu_pair(qc.as_(), self.ac_mul_add) == 0

qc.as_() / self.ac_quant as i32 == 0

qc.as_().abs() < self.ac_quant as i32

c.as_().abs() + self.ac_offset_eob < self.ac_quant as i32

c.as_().abs() < self.ac_quant as i32 - self.ac_offset_eob

(coeffs[pos] << self.log_tx_scale).as_().abs() < self.ac_quant as i32 - self.ac_offset_eob

(coeffs[pos].as_().abs() << self.log_tx_scale) < self.ac_quant as i32 - self.ac_offset_eob

coeffs[pos].as_().abs() <
  ((self.ac_quant as i32 - self.ac_offset_eob
  + (1 << self.log_tx_scale) - 1) >> self.log_tx_scale)
```

dwbuiten: It's this last step I don't understand:

```Rust
(coeffs[pos].as_().abs() << self.log_tx_scale) < self.ac_quant as i32 - self.ac_offset_eob

coeffs[pos].as_().abs() <
  ((self.ac_quant as i32 - self.ac_offset_eob
  + (1 << self.log_tx_scale) - 1) >> self.log_tx_scale)
```

And this may make sense to read, knowing what code came before, and knowing this derivation, but for a first time reader of this function, it's incredibly non-obvious, in my opinion. A detailed comment is necessary IMO.

barrbrain: This last step is equivalent to applying `ceil(x / 2^n)` to both sides of the inequality.
We continued this discussion on IRC to find a sufficiently clear description.

Issue #1617, opened 2019-08-30
Large reservoir_frame_delay results in massive slow-down even when flushed


YaLTeR:
Now sure how preventable this issue is because I'm not familiar with the complex logic in `guess_frame_subtypes`, but: setting `reservoir_frame_delay` to a very large value (for example, via `max_key_frame_interval`) results in this loop in `guess_frame_subtypes`
https://github.com/xiph/rav1e/blob/f8008ce9e43c885629456e63798b72f80db84560/src/api/mod.rs#L2227
spinning through it one by one which takes a really long while, even when the encoder has received a single frame and was flushed (which probably means the interval can be capped).

Now while I can theorize a simple fix to cap `reservoir_frame_delay` on flush, it's not clear to me if this issue can be prevented if some frames are being encoded before a flush.

YaLTeR:
After a discussion with derf, it's kind of supposed to do this, but it should probably have a max-cap which would prevent this kind of issue. Apparently it used to have a max-cap but then it was disabled because it wasn't working well?

PR #1639, opened 2019-09-06
Un-break reservoir constraints

dwbuiten:
The previous fix (#1637) was incorrect and removes part of the entire point of exposing it as an API (I should know... I added it). Capping to 240 makes no sense for user input... it make about as much sense as capping VBV buffer and maxrate in H.264. Maybe I want a 30 second frame reservoir. Maybe my frame rate isn't 24. This is also important for 2-pass, because with a capped reservori delay, 2nd pass could only hae the option of "240 frames or less" or "the whole file", which is not sufficient.

This PR reverts that change and makes the code match what the comment actually says (and what was actually asked for by @tdaede in #1324), as well as enforces a minimum of 12 for the default too.

#1637 was merged without any of the people who actually wrote this part of the code looking at it, which is not ideal. It was submitted in the middle of the night (for my timezone) and merged before I even got to a PC this morning.

### DIFF
--- a/src/rate.rs
+++ b/src/rate.rs
@@ -627,7 +627,7 @@ impl RCState {
     let reservoir_frame_delay = if maybe_reservoir_frame_delay.is_some() {
       maybe_reservoir_frame_delay.unwrap().max(12)
     } else {
-      ((max_key_frame_interval.checked_mul(3)?) >> 1).max(240)
+      ((max_key_frame_interval.checked_mul(3)?) >> 1).min(240)
     };
     // TODO: What are the limits on these?
     let npixels = (frame_width as i64) * (frame_height as i64);
--- a/src/rate.rs
+++ b/src/rate.rs
@@ -616,17 +616,19 @@ impl RCState {
     ac_qi_min: u8, max_key_frame_interval: i32,
     maybe_reservoir_frame_delay: Option<i32>,
   ) -> Option<RCState> {
-    // The default buffer size is set equal to 1.5x the keyframe interval.
-    // We enforce a minimum of 12 and a maximum of 240.
+    // The default buffer size is set equal to 1.5x the keyframe interval, or 240
+    //  frames; whichsever is smaller.
+    // For user set values, we enforce a minimum of 12.
     // The interval is short enough to allow reaction, but long enough to allow
     //  looking into the next GOP (avoiding the case where the last frames
     //  before an I-frame get starved), in most cases.
     // The 12 frame minimum gives us some chance to distribute bit estimation
     //  errors in the worst case.
-    let reservoir_frame_delay = maybe_reservoir_frame_delay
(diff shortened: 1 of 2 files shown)


src/rate.rs
```diff
@@ -627,7 +627,7 @@ impl RCState {
     let reservoir_frame_delay = if maybe_reservoir_frame_delay.is_some() {
       maybe_reservoir_frame_delay.unwrap().max(12)
     } else {
-      ((max_key_frame_interval.checked_mul(3)?) >> 1).min(240)
+      ((max_key_frame_interval.checked_mul(3)?) >> 1).min(240).max(12)
```
lu-zero:
``` rust
let reservoir_frame_delay = if let Some(frd) = maybe_reservoir_frame_delay {
  frd
} else {
  ((max_key_frame_interval.checked_mul(3)?) >> 1).min(240)
}.max(12);
```

maybe?

dwbuiten: Didn't know you could do a `{}.func()`. Will change.

src/rate.rs
```diff
@@ -616,17 +616,19 @@ impl RCState {
     ac_qi_min: u8, max_key_frame_interval: i32,
     maybe_reservoir_frame_delay: Option<i32>,
   ) -> Option<RCState> {
-    // The default buffer size is set equal to 1.5x the keyframe interval.
-    // We enforce a minimum of 12 and a maximum of 240.
+    // The default buffer size is set equal to 1.5x the keyframe interval, or 240
+    //  frames; whichsever is smaller, with a minimum of 12.
```
tmatth: nit: `whichever`

src/rate.rs
```diff
@@ -616,17 +616,21 @@ impl RCState {
...
-    let reservoir_frame_delay = maybe_reservoir_frame_delay
-      .unwrap_or(max_key_frame_interval.checked_mul(3)? >> 1)
-      .max(12)
-      .min(240);
+    let reservoir_frame_delay = if let Some(rfd) = maybe_reservoir_frame_delay
+    {
+      rfd
+    } else {
+      ((max_key_frame_interval.checked_mul(3)?) >> 1).min(240)
+    }
```
YaLTeR: This can be written in the same `unwrap_or` style if you want. I think it's nicer that way.

dwbuiten: I agree, that looks nicer. Updating.

YaLTeR:
As far as I can tell, reservoir should always be capped at `131072` so as to not overflow the `reservoir_max` computation.

dwbuiten:
@YaLTeR for #1617 I don't really have a good answer other than "users get what they set, welp"...

As for the `131072` cap, I would find that more confusing (since this a user-exposed value, users will see this cap of `131072`) than a checked_mul elsewhere, but I have no strong opinion.

PR #1640, opened 2019-09-06
Extract Config validation into a function

YaLTeR:
Please comment on the design.

- I have to store the maximal values to print them, unfortunately; I reported an issue into `err-derive` about this: https://gitlab.com/torkleyy/err-derive/issues/5
- Do we want to expose this in the C API somehow?
- Do we want to clearly expose all the min/max values as public constants or something?
- The hidden variant is added so that it's possible for us to add more variants without it being a breaking change (as more checks are added for example).
- Error is implemented for `EncoderStatus` so that it's possible to convert it to `Box<dyn Error>` (see doctests).

I don't _particularly_ like the magic number checks instead of checked operations, but on the other hand it results in clear range guarantees (rather than "we changed the order of operations in this update and suddenly this config value is no longer allowed").

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -514,6 +514,85 @@ impl fmt::Display for PredictionModesSetting {
   }
 }
 
+/// Enumeration of possible invalid configuration errors.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Error)]
+pub enum InvalidConfig {
+  /// The width is invalid.
+  #[error(display = "invalid width {} (expected > 0, <= 32767)", _0)]
+  InvalidWidth(usize),
+  /// The height is invalid.
+  #[error(display = "invalid height {} (expected > 0, <= 32767)", _0)]
+  InvalidHeight(usize),
+  /// RDO lookahead frame count is invalid.
+  #[error(
+    display = "invalid rdo lookahead frames {} (expected <= {})",
+    actual,
+    max
+  )]
+  InvalidRdoLookaheadFrames {
+    /// The actual value.
+    actual: usize,
+    /// The maximal supported value.
+    max: usize,
+  },
+  /// Maximal keyframe interval is invalid.
+  #[error(
+    display = "invalid max keyframe interval {} (expected > 0, <= {})",
+    actual,
+    max
+  )]
+  InvalidMaxKeyFrameInterval {
+    /// The actual value.
+    actual: u64,
+    /// The maximal supported value.
+    max: u64,
+  },
+  /// Tile columns is invalid.
+  #[error(display = "invalid tile cols {} (expected power of 2)", _0)]
+  InvalidTileCols(usize),
+  /// Tile rows is invalid.
+  #[error(display = "invalid tile rows {} (expected power of 2)", _0)]
+  InvalidTileRows(usize),
+  /// Framerate numerator is invalid.
+  #[error(
+    display = "invalid framerate numerator {} (expected > 0, <= {})",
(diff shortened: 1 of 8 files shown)


lu-zero:
> Please comment on the design.
> 
>     * I have to store the maximal values to print them, unfortunately; I reported an issue into `err-derive` about this: https://gitlab.com/torkleyy/err-derive/issues/5
> 
>     * Do we want to expose this in the C API somehow?

The C API does the validation in the setter right now.

> 
>     * Do we want to clearly expose all the min/max values as public constants or something?

The documentation should be enough for now.
 
>     * The hidden variant is added so that it's possible for us to add more variants without it being a breaking change (as more checks are added for example).
> 
>     * Error is implemented for `EncoderStatus` so that it's possible to convert it to `Box<dyn Error>` (see doctests).

This is good.

src/api/mod.rs
```diff
@@ -513,6 +514,81 @@ impl fmt::Display for PredictionModesSetting {
...
+    max: u64,
+  },
+  /// Reservoir frame delay is invalid.
+  #[error(
+    display = "invalid reservoir frame delay {} (expected >= 12, <= 131072)",
+    _0
+  )]
+  InvalidReservoirFrameDelay(i32),
+
+  #[doc(hidden)]
```
lu-zero: mention https://github.com/rust-lang/rust/issues/44109 in a comment.

src/api/mod.rs
```diff
@@ -513,6 +514,85 @@ impl fmt::Display for PredictionModesSetting {
...
+  /// Reservoir frame delay is invalid.
+  #[error(
+    display = "invalid reservoir frame delay {} (expected >= 12, <= 131072)",
+    _0
+  )]
+  InvalidReservoirFrameDelay(i32),
+
+  // This variant prevents people from exhaustively matching on this enum,
+  // which allows us to add more variants without it being a breaking change.
+  // This can be replaces with #[non_exhaustive] when it's stable:
```
lu-zero:
```suggestion
  // This can be replaced with #[non_exhaustive] when it's stable:
```

YaLTeR:
I just found this function:
https://github.com/xiph/rav1e/blob/a4bc23d5ee6462b32065989ce08639cc3f1a7113/src/api/mod.rs#L217-L226
I think it could be confusing that setting the max keyframe interval to zero manually vs. this function results in different behavior. Perhaps a better idea would be to manually call this function from within `new_context` before `validate`?

lu-zero:
What is using that function?

YaLTeR:
Apparently you're supposed to use that function to set the keyframe intervals because it handles the max keyint = 0 case.

lu-zero:
Might be a good idea or we could make 0 == std::u64::MAX both in the validation and in the context factory.

YaLTeR:
So this method sets it to `u64::MAX`, however that value is never actually specially checked for anywhere, and in fact leads to integer overflows. Should I just make it so `0` means "highest valid value"?

YaLTeR:
How does this look?

YaLTeR:
cc @rzumer

rzumer:
Fix seems good to me, I tested 0-keyint and didn't experience problems though. Did you test this alternative?

YaLTeR:
How should I test this?

YaLTeR:
I think the best way type-wise would be to make it an `Option` where `None` means infinite, but I'll leave this for a future refactor. Currently setting it to `0` seems to work as intended.

PR #1643, opened 2019-09-08
[WIP] Use inter and intra cost metrics from lookahead for scene change detection

### LINKED ISSUE #1528: Use Temporal RDO lookahead metric for scene detection instead of frame delta
The lookahead for temporal RDO computes the "complexity" of each block (inter vs intra cost) when it does the motion search, relative to all of its references. We should use those numbers instead of the current frame difference metric, both because it's "free" and because it's motion compensated, so it's far more accurate.

shssoichiro:
Closes #1528

[AWCY objective-1-fast](https://bit.ly/2lKsl1G) showing no difference on most clips (expected for a set with no scene changes), a couple of clips have a BD-Rate loss due to higher filesize from some frames being selected as keyframes that weren't being selected before. Probably some tweaks to intra/inter cost calculation would be helpful for this.

Speed cost at default settings is about 3% overall in the worst case (many keyframes), no speed difference in best case (few keyframes). Speed cost will also be less at slower speed settings.


src/scenechange/mod.rs
```diff
@@ -7,43 +7,25 @@
...
-use crate::util::CastFromPrimitive;
+use crate::api::{EncoderConfig, FrameType};
 use crate::util::Pixel;
 
+use crate::encoder::FrameInvariants;
 use std::cmp;
 use std::collections::BTreeSet;
-use std::sync::Arc;
+
+const SCENECUT_THRESHOLD: f32 = 40.;
```
dwbuiten: Where did this come from?

shssoichiro: x264. I did some testing as well to see if this is really the ideal threshold and I wasn't able to come to a solid conclusion. But the way it scales seems reasonable, because quality degradation gets worse the further you get from the last keyframe. So, this is the max threshold if you're near the max keyframe interval. Near the min interval, it's divided by 4, so it ends up starting at .1. Note also that the ratios can go negative (if inter cost is higher than intra cost), but we don't always want to wait for that to signal a new keyframe (on some sources, it may be very rare for inter cost to be greater than intra cost even on what a human would perceive as a "scene change").

src/api/mod.rs
```diff
@@ -1711,89 +1740,197 @@ impl<T: Pixel> ContextInner<T> {
...
 
-    while !self.needs_more_frame_q_lookahead(self.next_lookahead_frame) {
-      // Process the next unprocessed frame
-      // Start by getting that frame and all frames after it in the queue
-      let current_lookahead_frames = &lookahead_frames[lookahead_idx..];
+      // At this point we have `rdo_lookahead_frames` lookahead computed.
+      // Now we take frames until we no longer have `rdo_lookahead_frames`
+      // (or we are at the end of the video) and run frame type detection on them.
+      while self.lookahead_buffer_filled()
+        || self.is_flushing() && self.needs_more_frame_types_decided()
```
YaLTeR: Please put braces around here, || vs. && precedence is not obvious.

src/encoder.rs
```diff
@@ -830,6 +837,11 @@ impl<T: Pixel> FrameInvariants<T> {
     };
     fi.input_frameno = input_frameno;
     fi.me_range_scale = (inter_cfg.group_input_len >> fi.pyramid_level) as u8;
+    fi.lookahead_inter_costs = vec![
+      vec![0; fi.w_in_imp_b * fi.h_in_imp_b];
```
YaLTeR: Use boxed slice?

src/frame/mod.rs
```diff
@@ -108,18 +108,6 @@ impl<T: Pixel> Frame<T> {
...
-  /// Each point in the `PixelIter` is a triple consisting of a Y, U, and V component.
-  /// The `PixelIter` is laid out as contiguous rows, e.g. to get a given 0-indexed row
-  /// you could use `data.skip(width * row_idx).take(width)`.
-  ///
-  /// This data retains any padding, e.g. it uses the width and height specifed in
-  /// the Y-plane's `cfg` struct, and not the display width and height specied in
-  /// `FrameInvariants`.
-  pub(crate) fn iter(&self) -> PixelIter<'_, T> {
-    PixelIter::new(&self.planes)
-  }
```
YaLTeR: Should be in separate commit?

src/frame/mod.rs
```diff
@@ -130,10 +118,6 @@ pub(crate) struct PixelIter<'a, T: Pixel> {
 }
 
 impl<'a, T: Pixel> PixelIter<'a, T> {
-  pub fn new(planes: &'a [Plane<T>; 3]) -> Self {
-    PixelIter { planes, y: 0, x: 0 }
-  }
-
```
YaLTeR: Should be in separate commit?

src/frame/plane.rs
```diff
@@ -321,11 +321,6 @@ impl<T: Pixel> Plane<T> {
     PlaneMutSlice { plane: self, x: po.x, y: po.y }
   }
 
-  #[cfg(test)]
-  pub(crate) fn as_slice(&self) -> PlaneSlice<'_, T> {
-    self.slice(PlaneOffset { x: 0, y: 0 })
-  }
-
```
YaLTeR: Should be in separate commit?

src/api/mod.rs
```diff
@@ -218,7 +218,7 @@ impl EncoderConfig {
   pub fn set_key_frame_interval(
     &mut self, min_interval: u64, max_interval: u64,
   ) {
-    self.min_key_frame_interval = min_interval;
+    self.min_key_frame_interval = cmp::max(1, min_interval);
```
YaLTeR: This should probably be a config validate check once #1640 lands.

src/api/mod.rs
```diff
@@ -1232,12 +1228,14 @@ impl<T: Pixel> ContextInner<T> {
       frames_processed: 0,
       frame_q: BTreeMap::new(),
       frame_invariants: BTreeMap::new(),
-      keyframes: BTreeSet::new(),
+      // As an optimization for lookahead,
+      // Initialize this with the first frame set as a keyframe.
+      keyframes: [0].iter().copied().collect::<BTreeSet<_>>(),
```
YaLTeR: `iter::once(0).collect()`

src/encoder.rs
```diff
@@ -535,6 +538,8 @@ pub struct FrameInvariants<T: Pixel> {
   pub h_in_imp_b: usize,
   /// Intra prediction cost estimations for each importance block.
   pub lookahead_intra_costs: Box<[u32]>,
+  /// Inter prediction cost estimations for each importance block.
```
YaLTeR: This is confusing because it's computing inter-costs without taking the motion vectors into account (which is wrong?), and because block importance now uses different inter-costs (with motion vectors). Is there a mistake here?

src/frame/mod.rs
shssoichiro: The only code that was using this (and the other two removed sections below) was the old scenechange code, so I think it's fine to have it in this commit.

src/encoder.rs
shssoichiro: Good find, this comment is wrong, and more importantly failing to check the MVs is probably wrong as well. That may be behind some of the false positives showing in objective-1-fast.

shssoichiro:
This new windowed algorithm is the closest I've gotten to achieving metrics similar to master, but it still falls short. Every metric regresses with this branch, on every clip I've tested. The changes I've made also make this branch considerably slower than master--about 20% **overall** on speed 5 at 720p.

It goes against conventional wisdom, but the current SAD metric just produces better compression than using intra and inter costs. Which would lead me to believe the costs rav1e is calculating are inaccurate in some way. I did notice some oddities. For example, I was using a test clip of the first 30 frames of each objective-1-fast clip, in sequence. Without the windowed algorithm, frame 30 (first frame of clip 2) would be excluded as a scenechange because frame 28, despite being visually similar to frames 27 and 29, had considerably different motion vectors and therefore a considerably lower inter cost to frame 30.

Anyway, I will keep this PR open for the time being in a WIP state for discussion.

shssoichiro:
Closing this because it would just be too slow. I have a different approach that I'm going to experiment with.

PR #1646, opened 2019-09-09
Aligned array constructors

lu-zero:
Does not change the functionality, just makes it less strange.


lu-zero:
Amended to fix the bench usage as well.

YaLTeR:
Could you mark the function as `unsafe` if you're changing all the call sites anyway?

lu-zero:
I would not do that now, the changes were made using a simple sed pass, adding `unsafe {}` requires more time. Ideally once the const-generics land the whole thing could be reworked so we could potentially remove most if not all the uninit arrays.

PR #1650, opened 2019-09-11
Search all intra modes in lookahead

YaLTeR:
Instead of just DC, try all of them and use the minimal intra-cost.

[AWCY on `--speed 2`](https://beta.arewecompressedyet.com/?job=speed2-2019-09-06_163944-f39b1d7&job=all-intra-modes-importance-speed2-f0.9-a0.8-2019-09-09_230607-e01baca) shows good PSNR improvement.

Since this seems to significantly shift the importance, the segment ID heuristic needs to be updated. This is what I got so far: [AWCY on default speed with the updated heuristic](https://beta.arewecompressedyet.com/?job=master-f39b1d7529575a6540d348788074dd61771047f6&job=all-intra-modes-importance-2019-09-10_164115-7bdc32d), however, as can be seen by the [AWCY on default speed of the updated heuristic vs. segment ID RDO](https://beta.arewecompressedyet.com/?job=all-intra-modes-importance-2019-09-10_164115-7bdc32d&job=all-intra-modes-importance-2019-09-10_103124-1a93398) it can be considerably improved. I guess I'll try to do this throughout my remaining time.

cc #1643

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -32,7 +32,7 @@ use crate::encoder::*;
 use crate::frame::*;
 use crate::metrics::calculate_frame_psnr;
 use crate::partition::*;
-use crate::predict::PredictionMode;
+use crate::predict::RAV1E_INTRA_MODES;
 use crate::rate::RCState;
 use crate::rate::FRAME_NSUBTYPES;
 use crate::rate::FRAME_SUBTYPE_I;
@@ -1806,7 +1806,6 @@ impl<T: Pixel> ContextInner<T> {
           height: IMPORTANCE_BLOCK_SIZE,
         });
 
-        // TODO: other intra prediction modes.
         let edge_buf = get_intra_edges(
           &frame.planes[0].as_region(),
           TileBlockOffset(BlockOffset { x, y }),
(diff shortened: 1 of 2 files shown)


src/api/mod.rs
```diff
@@ -1819,47 +1818,55 @@ impl<T: Pixel> ContextInner<T> {
...
-          &[], // Not used by DC_PRED.
-          0,   // Not used by DC_PRED.
-          &edge_buf,
-        );
+        let bit_depth = self.config.bit_depth;
+        let intra_mode_set = RAV1E_INTRA_MODES;
+        let intra_cost = intra_mode_set
+          .iter()
+          .map(|&luma_mode| {
+            let mut plane_after_prediction_region = plane_after_prediction
```
lu-zero: The tile and the region seem to stay the same, is there a reason why they are in the iteration instead of out?

YaLTeR: Borrowck reasons. I need both mutable and immutable access in this closure, and if there's a mutable borrow outside of the closure I can't get the immutable one inside.

YaLTeR: The underlying issue is that `get_satd` accepts `PlaneRegion` and `PlaneRegionMut` is not a `PlaneRegion`.

src/api/mod.rs
```diff
@@ -1819,47 +1818,55 @@ impl<T: Pixel> ContextInner<T> {
...
-        );
+            let plane_after_prediction_region =
+              plane_after_prediction.region(Area::Rect {
+                x: (x * IMPORTANCE_BLOCK_SIZE) as isize,
+                y: (y * IMPORTANCE_BLOCK_SIZE) as isize,
+                width: IMPORTANCE_BLOCK_SIZE,
+                height: IMPORTANCE_BLOCK_SIZE,
+              });
+
+            get_satd(
```
tdaede: Our intra pruning during RDO actually uses get_sad, not get_satd. I think satd is actually better, but maybe that could be a cause of the PSNR vs PSNR-HVS difference?

tdaede:
Can you try rerunning this on master, which now uses satd for intra pruning?

ycho:
> Instead of just DC, try all of them and use the minimal intra-cost.
> 
> [AWCY on `--speed 2`](https://beta.arewecompressedyet.com/?job=speed2-2019-09-06_163944-f39b1d7&job=all-intra-modes-importance-speed2-f0.9-a0.8-2019-09-09_230607-e01baca) shows good PSNR improvement.
> 
Since its impact on enc time is very small but giving significant gain > 2%, we can consider enabling this PR at speed 2 (or <= 2)?

YaLTeR:
It's become worse, actually (unless I rebased something wrong): https://beta.arewecompressedyet.com/?job=speed2-2019-10-04_114925-45585ee&job=all-intra-modes-importance-speed2-2019-10-04_115057-fe648e4

PR #1654, opened 2019-09-12
Api refactor

lu-zero:
@shssoichiro it would probably collide with your ongoing work, let me know if I can fold it in before your set lands or you can afford to rebase on top of this.


shssoichiro:
Could you go ahead and rebase this off master? I'd like to review with the config portion excluded from the diff. I'm ready to move forward with this, though.

lu-zero:
Sure

PR #1656, opened 2019-09-13
Disable large LRU by default

rzumer:
It breaks encoding quite often (sample-dependent).


xiphmont:
As of right now, I'm not aware of any filed issues caused by large LRU.

rzumer:
As mentioned in IRC I was synced with the wrong branch, and the current master does fix all known tiling/large LRU errors as far as I'm aware.

PR #1661, opened 2019-09-16
Fix tile log2 validation in the C API

### LINKED ISSUE #1658: Valid number of tile columns/rows return an error in the C API
It seem the tile number validation in the C API is not working correctly. With 2x2 tiles specified, an error is returned, as it is wrongly detected as invalid. When the validation functions are skipped, everything works as expected.

rzumer:
The condition was inverted compared to its Rust API equivalent.

Should fix #1658.


rzumer:
Tested and seems to work as intended.

PR #1662, opened 2019-09-16
Remove input chroma sampling restriction in readme

rzumer:
All sampling formats are supported (though not fully for non-4:2:0 as mentioned above).


shssoichiro:
We should probably mention that monochrome (4:0:0) inputs are not yet supported.

PR #1664, opened 2019-09-16
Use const fn in more places

shssoichiro:
Related to #1464


shssoichiro:
I think it's unlikely to impact any other PRs. If it does, the merge resolution should be minimal.

PR #1670, opened 2019-09-17
Move the container_sequence_header logic in EncoderConfig

lu-zero:
It does not need any Context information.


lu-zero:
Let's postpone it for now.

PR #1674, opened 2019-09-18
Don't loop through all ref frames when multiref disabled

### LINKED ISSUE #1212: Multiple reference frames are prepared and tested even when multiref is false

shssoichiro:
[AWCY](https://beta.arewecompressedyet.com/?job=master-low-latency%402019-09-18T07%3A46%3A50.014Z&job=branch-low-latency%402019-09-18T07%3A54%3A57.812Z) no change

Closes #1212


ycho:
@tdaede, I will review but can you also review this patch please since you are more familiar with use of ref frames for inter mode?

ycho:
> [AWCY](https://beta.arewecompressedyet.com/?job=master-low-latency%402019-09-18T07%3A46%3A50.014Z&job=branch-low-latency%402019-09-18T07%3A54%3A57.812Z) no change
> 
> Closes #1212

There is a change in AWCY. Encoding time has reduced! Thanks!
(I checked AWCY, then found you already cited it in the beginning)

ycho:
@tdaede, btw, can you let me know what index [0] is used at:
https://github.com/xiph/rav1e/blob/57088e126978cfd47e6973bf0e649676353f1446/src/rdo.rs#L811 ?

Is it because, when ref_frames_set[] is created, NONE_FRAME is always added (line 770) so always the first entry, i.e. [0] is the current ref frame ?Would this be also true for multi-ref, i.e. compound reference mode?

@shssoichiro, basically I also wanted to check whether ref frame loop, https://github.com/xiph/rav1e/blob/57088e126978cfd47e6973bf0e649676353f1446/src/rdo.rs#L788, iterates only once!

shssoichiro:
> @shssoichiro, basically I also wanted to check whether ref frame loop,
> 
> https://github.com/xiph/rav1e/blob/57088e126978cfd47e6973bf0e649676353f1446/src/rdo.rs#L788
> , iterates only once!

Good find, I missed this one because I was only looking at places where we used INTER_REFS_PER_FRAME. We should be able to optimize this as well!

P.S. What I found, or at least this is my understanding of how our ref frames work and correct me if I'm wrong, when multiref is disabled, we actually still check _two_ ref frames: the last frame and the "golden frame", whatever that is. 🙂 But 2 is still faster than 7.

ycho:
I just checked in master, at speed 5 (multi-ref off for speed > 1)  that the frame ref loop in rdo_mode_decision iterate more than once.
First for LAST_FRAME, and 2nd LAST2_FRAME.
So, can possibly be another bug.

@tdaede  ^

ycho:
> > @shssoichiro, basically I also wanted to check whether ref frame loop,
> > https://github.com/xiph/rav1e/blob/57088e126978cfd47e6973bf0e649676353f1446/src/rdo.rs#L788
> > 
> > , iterates only once!
> 
> Good find, I missed this one because I was only looking at places where we used INTER_REFS_PER_FRAME. We should be able to optimize this as well!
> 
> P.S. What I found, or at least this is my understanding of how our ref frames work and correct me if I'm wrong, when multiref is disabled, we actually still check _two_ ref frames: the last frame and the "golden frame", whatever that is. 🙂 But 2 is still faster than 7.

My understanding is that if multi-ref is off, only one ref frame should be used? Bcz LAST + GOLDEN is apparently compound mode?
@tdaede ^

ycho:
All right, found evidence that our multi-ref = false does not work. It still uses multi ref, though not a compound mode.

Below is bitstream analyzer view for frame 4, awcy run by @shssoichiro (thanks!), and it shows both LAST_FRAM and LAST2_FRAME!
[AWCY of master at default speed 5 with --low-latency](https://beta.arewecompressedyet.com/analyzer/?maxFrames=4&decoder=https://people.xiph.org/~mbebenita/analyzer/inspect.js&decoderName=master-low-latency%402019-09-18T07%3A46%3A50.014Z&file=https://beta.arewecompressedyet.com/runs/master-low-latency@2019-09-18T07:46:50.014Z/objective-1-fast/red_kayak_360p_60f.y4m-172.ivf&decoder=https://people.xiph.org/~mbebenita/analyzer/inspect.js&decoderName=branch-low-latency%402019-09-18T07%3A54%3A57.812Z&file=https://beta.arewecompressedyet.com/runs/branch-low-latency@2019-09-18T07:54:57.812Z/objective-1-fast/red_kayak_360p_60f.y4m-172.ivf)

src/api/internal.rs
```diff
@@ -186,6 +186,14 @@ impl InterConfig {
   pub(crate) fn keyframe_lookahead_distance(&self) -> u64 {
     cmp::max(1, self.max_reordering_latency()) + 1
   }
+
+  pub(crate) fn inter_refs_per_frame(&self) -> usize {
```
tdaede: Please name this function something else so it's not confused with the INTER_REFS_PER_FRAME constant.

src/encoder.rs
```diff
@@ -832,8 +833,8 @@ impl<T: Pixel> FrameInvariants<T> {
     fi
   }
 
-  pub fn set_ref_frame_sign_bias(&mut self) {
-    for i in 0..INTER_REFS_PER_FRAME {
+  pub fn set_ref_frame_sign_bias(&mut self, inter_cfg: &InterConfig) {
+    for i in 0..inter_cfg.inter_refs_per_frame() {
```
tdaede: I don't think the optimization is really worth it here?

src/api/internal.rs
tdaede: Maybe a better way to do this would be to return a list of reference frames that are used?

tdaede:
> @tdaede, btw, can you let me know what index [0] is used at:
> 
> https://github.com/xiph/rav1e/blob/57088e126978cfd47e6973bf0e649676353f1446/src/rdo.rs#L811

ref_frames[0] is actually LAST_FRAME.

shssoichiro:
> My main concern here is that this limits multiref-off to only frame references 0 and 1. This is pretty non-intuitive (and are those the only references we actually want to use for multiref off?)

Those are effectively the only ones we use, but we loop through all 7 ref slots to accomplish it at the moment. The ref_frames array would be filled with something like `[1, 0, 0, 0, 0, 0, 0]`. So, the output is equivalent from only looking at the first two items, and this is purely a performance optimization.

tdaede:
Yeah I agree it's correct right now, but I want to make sure it stays correct if we change which reference frames we use in the future. Is there someplace we can hit an assert if we attempt to use a reference frame greater than 1 in this case?

Alternately, in new_inter_frame() we actually compute which reference frames we use, the assert could go there. Or the list of reference frames to loop through could be created there.

ycho:
In case you missed, I filed the related bug https://github.com/xiph/rav1e/issues/1675

src/encoder.rs
ycho: but it clearly corrected the looping thru unnecessary iterations.

tdaede: So the thing that bothers me is that this is spec designed behavior to propagate reference frame context. I think it's OK to skip in this case, but it's a different thing than not skipping searching the reference frames.

ycho:
Hmm, the definition of 'multiref' now confuses me.
https://aomediacodec.github.io/av1-spec/#ref-frames-semantics, it says:
SINGLE_REFERENCE indicates that the inter block uses only a single reference frame to generate motion compensated prediction.
COMPOUND_REFERENCE indicates that the inter block uses compound mode.
Then, I don't think we meant SINGLE_REFERENCE by our 'multiref', right?

Also, the def of SINGLE_REFERENCE does not seem that inter mode cannot use multiple ref frames from slots. I read it that just one ref_frame for one inter mode?

tdaede:
Yeah, our "multiref" setting is a bit underdefined. I think right now it controls whether you use compound modes... but even without compound (SINGLE) you can use multiple references in one frame, just not more than one per block. It probably should be renamed at minimum.

ycho:
> Yeah, our "multiref" setting is a bit underdefined. I think right now it controls whether you use compound modes... but even without compound (SINGLE) you can use multiple references in one frame, just not more than one per block. It probably should be renamed at minimum.

If that's the case (i.e. single vs compound reference), we probably need rename it. Otherwise, if multiref was part of our encoder options (not part of av1 spec), then need change code, so that only LAST_FRAME is used as ref frame!?

tdaede:
Aha, we did define it:
```
  /// Whether P-frames can use multiple references.
  pub(crate) multiref: bool,
```

So really we should change our code so we only use LAST_FRAME.

ycho:
> Aha, we did define it:
> 
> ```
>   /// Whether P-frames can use multiple references.
>   pub(crate) multiref: bool,
> ```
> 
> So really we should change our code so we only use LAST_FRAME.

Yes, indeed.

shssoichiro:
I went ahead and changed it to only look at LAST_FRAME, it passes tests but I'm going to run it through AWCY to verify as well.

ycho:
> My main concern here is that this limits multiref-off to only frame references 0 and 1. This is pretty non-intuitive (and are those the only references we actually want to use for multiref off?)

I think this issue is resolved now?

ycho:
> I went ahead and changed it to only look at LAST_FRAME, it passes tests but I'm going to run it through AWCY to verify as well.

It seems you haven't started awcy run for 2nd commit, so I started it here under your name.
[awcy, ](https://beta.arewecompressedyet.com/?job=branchc-low-latency%402019-09-19T22%3A57%3A20.488Z&job=master-low-latency%402019-09-18T07%3A46%3A50.014Z)

shssoichiro:
Ah, I did. Results [here](https://beta.arewecompressedyet.com/?job=master-low-latency%402019-09-18T07%3A46%3A50.014Z&job=ref-frames-low-latency%402019-09-19T19%3A25%3A27.807Z) and it looks like there's a pretty big quality difference from moving to single ref. But encoding speed also improves by 30%+.

ycho:
> Ah, I did. Results [here](https://beta.arewecompressedyet.com/?job=master-low-latency%402019-09-18T07%3A46%3A50.014Z&job=ref-frames-low-latency%402019-09-19T19%3A25%3A27.807Z) and it looks like there's a pretty big quality difference from moving to single ref. But encoding speed also improves by 30%+.

Wow, 35% loss! I briefly check the bitstream analys for bluesky then somehow keyframes are different, spending less bits. Maybe this is caused by temporal RDO (not sure temporal rdo is active with default speed 5 though)?
Also, did you see that the redkayak is least affected? Only 1% change... :) (Can't explain why....)

shssoichiro:
Lookahead RDO is active at all speed levels for determining MVs and block importances. So this would affect the intra blocks chosen. redkayak I can only guess the type of motion in that video didn't benefit much from multiple ref frames?

tdaede:
A change to use only 1 reference should update new_inter_frame() as well, so that all 7 slots are filled with the reference frame you use. That's why I'd really like some sort of check to make sure that matches what reference frames we are searching.

I guess the AWCY results show we actually want 3 modes - backward references only (what multiref disabled ends up doing right now), forward and backward references (default) and multiref disabled (one ref). It doesn't necessarily need to be an enum, we could replace the current code that checks for multiref to just check for reordering instead (to switch between backward references and forward/back), then don't actually disable multiref for low latency mode and make that a separate option.

shssoichiro:
I went ahead with the three-tier implementation, and added a check into new_inter_frame to make sure multiref=false is always filled with the same ref frame. I (somewhat arbitrarily) set multiref enabled at speed <= 7, since with the new change, there is both a sizeable speed and quality difference on most videos from having it turned on.

lu-zero:
Do you have AWCY runs to link to this change?

shssoichiro:
I will queue some up.

shssoichiro:
With `low-latency=false`: [No change](https://beta.arewecompressedyet.com/?job=master-baseline%402019-09-23T13%3A14%3A29.622Z&job=ref-frames-baseline%402019-09-23T13%3A16%3A31.752Z)

With `low-latency=true` and `multiref=true`: [Maybe will complete eventually, AWCY is being wonky](https://beta.arewecompressedyet.com/?job=master-ll-multiref%402019-09-23T13%3A17%3A50.225Z&job=ref-frames-ll-multiref%402019-09-23T13%3A18%3A34.408Z)

With `low-latency=true` and `multiref=false`: [~4% BD-Rate loss, ~40% speedup](https://beta.arewecompressedyet.com/?job=master-ll-no-multiref%402019-09-23T13%3A16%3A50.627Z&job=ref-frames-ll-no-multiref%402019-09-23T13%3A17%3A30.581Z)

lu-zero:
Looks like we have a target for a speed level update.

ycho:
>  multiref disabled (one ref)

Yes, I agree that checking and adding asserts are always help us our work easier.
Btw, I think I should have pointed out this earlier, I don't think the regression as much as 30% comes from the addition of forward ref to backward ref. Instead, I suspected other possible bugs, whether existing or there might be some missed places that this patch should address?

ycho:
Oh, now it is 4% loss if mutiref is disabled. "With low-latency=true and multiref=false: ~4% BD-Rate loss, ~40% speedup"
Then it seems to me that everything is correct.

ycho:
Also, I see that "With low-latency=true and multiref=true:" now gives no change in bdrate and almost no change in enc time, so looks correct as well.

PR #1683, opened 2019-09-21
Call sad/satd using new way

KyleSiefring:
Switches to indexing these functions using BlockSizes.
Removes 10 bit only sad functions and unused satd functions.

I didn't see any speedup or slowdown.

### DIFF
--- a/benches/dist.rs
+++ b/benches/dist.rs
@@ -10,6 +10,7 @@
 use criterion::*;
 use rand::{Rng, SeedableRng};
 use rand_chacha::ChaChaRng;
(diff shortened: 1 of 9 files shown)


KyleSiefring:
The benchmarks seem to be faster but I don't know how much this translates to real world performance.

KyleSiefring:
Do you want me to split the removal of the 10 bit only sad functions or just the satd functions?

lu-zero:
Split the removal of all the unused functions as a first patch and then make a patch with the actual refactor please :)

PR #1688, opened 2019-09-23
Prune intra modes with SATD instead of SAD.

tdaede:
Results on subset1:

master-2019-09-23_160214-b4d418a -> intra_prune_satd-2019-09-23_160204-1a821a8

-0.1007 | -0.0694 |     N/A |  -0.0845 | -0.1145 | -0.1157 |    -0.0402

https://beta.arewecompressedyet.com/?job=intra_prune_satd-2019-09-23_160204-1a821a8&job=master-2019-09-23_160214-b4d418a

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -968,7 +968,7 @@ pub fn rdo_mode_decision<T: Pixel>(
       };
 
       let intra_mode_set = RAV1E_INTRA_MODES;
-      let mut sads = {
+      let mut satds = {
         // FIXME: If tx partition is used, this whole sads block should be fixed
         debug_assert!(bsize == tx_size.block_size());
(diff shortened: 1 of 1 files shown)


src/rdo.rs
```diff
@@ -1011,7 +1011,7 @@ pub fn rdo_mode_decision<T: Pixel>(
 
             (
               luma_mode,
-              get_sad(
+              get_satd(
```
shssoichiro: This is going into a variable called `sads` which will need to be renamed to `satds`. Or some other, better variable name. Naming is hard.

tdaede: Fixed. Using the wrong name is a Theora tradition that I don't want to continue :)

PR #1694, opened 2019-09-24
Prevent segmentation from entering lossless mode.

### LINKED ISSUE #1649: Corrupted bitstream with speeds 2 to 5, quantizer <= 30
rav1e commit: 07ca598

System:
CPU: i7-4770 (with AVX2)
OS: Win7 64bit

Clip: https://mega.nz/#!5tp2nIgL!mFJnlgtY6r2O87_OHo3eItfMr9h9j9jDeoNtQ4lb3S0
Encoded file: [rav1e.cq30sp2.bugRepro.shorter.zip](https://github.com/xiph/rav1e/files/3597709/rav1e.cq30sp2.bugRepro.shorter.zip)

sha1sums:
```
781074334db079c9dd3c5d63ddc859b13d3a67be *rav1e.cq30sp2.bugRepro.shorter.y4m    (original clip)
823d6e103f2cbd24e62f140f4848d7320dd05c7d *rav1e.cq30sp2.bugRepro.shorter.mkv    (ffv1 compressed original clip)
576ebb4b3897aa95ae03e2e6b2b2eb18b69f3328 *rav1e.cq30sp2.bugRepro.shorter.ivf    (encoded clip)
359656d8c8e618867ead43e8f6d6e7cc40544dc9 *rav1e.cq30sp2.bugRepro.shorter.zip    (encoded clip archive)
```

Cmdlines used:
```
# rav1e -0
rav1e 0.1.0 (20190430-473-g07ca598)
# rav1e -o rav1e.cq30sp2.bugRepro.shorter.ivf --quantizer 30 -s 2 --tune Psnr rav1e.cq30sp2.bugRepro.shorter.y4m
...
# aomdec -o /dev/null rav1e.cq30sp2.bugRepro.shorter.ivf
Warning: Failed to decode frame 17: Corrupt frame detected
Warning: Additional information: Corrupted segment_ids
```

tdaede:
Fixes #1649

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -633,7 +633,7 @@ fn luma_chroma_mode_rdo<T: Pixel>(
       if fi.base_q_idx as i16
         + ts.segmentation.data[heuristic_sidx as usize]
           [SegLvl::SEG_LVL_ALT_Q as usize]
-        < 1
+        < 2
       {
         0..=0
       } else {
@@ -641,7 +641,7 @@ fn luma_chroma_mode_rdo<T: Pixel>(
       }
     } else if fi.base_q_idx as i16
       + ts.segmentation.data[2][SegLvl::SEG_LVL_ALT_Q as usize]
-      < 1
+      < 2
     {
       0..=1
     } else {
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -629,7 +629,21 @@ fn luma_chroma_mode_rdo<T: Pixel>(
         x if x >= 4. => 2,
         _ => unreachable!(),
       };
-      heuristic_sidx..=heuristic_sidx
+      // prevent the highest sidx from bringing us into lossless
+      if fi.base_q_idx as i16
+        + ts.segmentation.data[heuristic_sidx as usize]
+          [SegLvl::SEG_LVL_ALT_Q as usize]
+        < 1
+      {
+        0..=0
+      } else {
+        heuristic_sidx..=heuristic_sidx
+      }
+    } else if fi.base_q_idx as i16
+      + ts.segmentation.data[2][SegLvl::SEG_LVL_ALT_Q as usize]
+      < 1
+    {
+      0..=1
     } else {
       0..=2
     };
--- a/src/segmentation.rs
+++ b/src/segmentation.rs
@@ -28,6 +28,12 @@ pub fn segmentation_optimize<T: Pixel>(
   // showed this to be the optimal one.
   const TEMPORAL_RDO_QI_DELTA: i16 = 21;
 
+  // Avoid going into lossless mode by never bringing qidx below 1.
+  // Because base_q_idx changes more frequently than the segmentation
+  // data, it is still possible for a segment to enter lossless, so
+  // enforcement elsewhere is needed.
+  let offset_lower_limit = 1 - fi.base_q_idx as i16;
+
   // Fill in 3 slots with 0, delta, -delta. The slot IDs are also used in
   // luma_chroma_mode_rdo() so if you change things here make sure to check
   // that place too.
@@ -36,7 +42,7 @@ pub fn segmentation_optimize<T: Pixel>(
     fs.segmentation.data[i][SegLvl::SEG_LVL_ALT_Q as usize] = match i {
       0 => 0,
       1 => TEMPORAL_RDO_QI_DELTA,
-      2 => -TEMPORAL_RDO_QI_DELTA,
+      2 => (-TEMPORAL_RDO_QI_DELTA).max(offset_lower_limit),
       _ => unreachable!(),
     };
   }
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,7 +104,7 @@ jobs:
         - cargo doc --verbose --no-deps
         - rustup component add clippy
         - cargo clippy --version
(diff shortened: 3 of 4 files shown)


src/segmentation.rs
```diff
@@ -36,7 +39,7 @@ pub fn segmentation_optimize<T: Pixel>(
     fs.segmentation.data[i][SegLvl::SEG_LVL_ALT_Q as usize] = match i {
       0 => 0,
       1 => TEMPORAL_RDO_QI_DELTA,
-      2 => -TEMPORAL_RDO_QI_DELTA,
+      2 => -TEMPORAL_RDO_QI_DELTA.max(offset_lower_limit),
```
barrbrain: Operator precedence here is subtly misleading.
```suggestion
      2 => (-TEMPORAL_RDO_QI_DELTA).max(offset_lower_limit),
```

tdaede:
In fact this fix doesn't work, I must have missed an aomdec invocation in testing.

tdaede:
>     * the delta is added to `fi.base_q_idx` plus the ac or dc offset which differ on different pyramid levels, but I guess are always above zero so this is not an issue. At the very least it'd be nice to have a comment noting this.

It doesn't actually depend on the ac/dc offsets, the spec determines lossless mode just based on base_q_idx + offset. I can add a comment.

>     * I guess this depends on `fi.base_q_idx` being the same for every frame, because as per the condition above, the segmentation data is updated only on `fi.primary_ref_frame == PRIMARY_REF_NONE` frames.  If `fi.base_q_idx` is the same for every frame, perhaps it should be moved outside of `fi`?

Hmm, this is a problem. The reason it currently works is because base_q_idx is lowest on keyframes, so it chooses a conservative maximum offset. However, it's conceivable in ratecontrol mode that base_q_idx actually drops lower. So we need an additional fix. Either we can change the segmentation features when this happens, or have an additional condition to never select a segment that forces us into lossless mode. Thoughts?

ycho:
>  or have an additional condition to never select a segment that forces us into lossless mode

I think that in addition to the 2nd option you proposed as safe device (+ assert() as double safety), " or have an additional condition to never select a segment that forces us into lossless mode", the delta table, i.e. data[ ] should be computed based on rate-control adsjuted frame q_idx, instead of base_q_idx?

tdaede:
I added special case logic for sidx = 2. TBH it is sort of ugly but does seem to work. I should probably also come up with some sort of test that triggers the condition.

src/segmentation.rs
```diff
@@ -28,6 +28,12 @@ pub fn segmentation_optimize<T: Pixel>(
   // showed this to be the optimal one.
   const TEMPORAL_RDO_QI_DELTA: i16 = 21;
 
+  // Avoid going into lossless mode by never bringing qidx below 1.
```
ycho: I don't know where the rate-control adjusted frame qidx is stored, but at this point I believe that info is available (i.e. frame level rate control is finished). If so, can we check what block qidx will with each of three qi_delta from three seg id and disable them by setting segmentation.features[ ] inside the for loop below?
I simply hope, by doing so, we are able to block the use of qidx <= 1 from the origin?

tdaede: The segmentation adjusted qidx is fetched by the get_qidx() function. We can force it in the loop below instead. The only downside is that we'll end up searching sidx 1 twice if we're at the limit and doing segment rdo.

ycho: Hmm, I think I didn't get what you mean. I double checked that this function segmentation_optimize(), where this comment box is located, is called by encode_frame(), only once for a frame. Meanwhile, get_qidx() is called in the beginning of encode_tx_block() to fetch the qidx to use with quantizatioin and dequant.
One issue I now expect is, quant_rdo seems always assume that there are three qidx available (accessed by three seg id) then assign one to the block after quant_rdo is done!

ycho: > The only downside is that we'll end up searching sidx 1 twice if we're at the limit and doing segment rdo.

"The only downside is that we'll end up searching sidx 1 twice if we're at the limit and doing segment rdo.", this means seg id 1 and 2 get merged since seg id 2 introduce qidx <= 1, right?

tdaede: segmentation_optimize() only gets called for intra frames, so those values only change every keyframe. My limits in that function prevent it using lossless 99% of the time. But there's still a small chance that due to base_q_idx changing, segment 2 could still enter lossless, so we still need to check that one as well.

Yes, everything now assumes there are exactly 3 segment ids, and 3 qidx available. It's kind of intentionally hardcoded right now.

ycho: I see. I didn't acknowledge that the function is only called for keyframe. Then I guess that, w/o your PR here, the 1% chance is when certain inter frame to have qinx <= 21 (i.e. 22 - 21 = 0).
Please go ahead merging.

src/rdo.rs
```diff
@@ -629,7 +629,21 @@ fn luma_chroma_mode_rdo<T: Pixel>(
...
         _ => unreachable!(),
       };
-      heuristic_sidx..=heuristic_sidx
+      // prevent the highest sidx from bringing us into lossless
+      if fi.base_q_idx as i16
+        + ts.segmentation.data[heuristic_sidx as usize]
+          [SegLvl::SEG_LVL_ALT_Q as usize]
+        < 1
+      {
+        heuristic_sidx - 1..=heuristic_sidx - 1
```
barrbrain: As this is routing around the current segementation design, I note that:
- We expect to only encounter this condition when `heuristic_sidx == 2` and maps to `-TEMPORAL_RDO_QI_DELTA`
- In that case, `heuristic_sidx - 1 == 1` and maps to `+TEMPORAL_RDO_QI_DELTA`
- The remaining choice of `heuristic_sidx - 2 == 0` maps to zero and is a closer match.

```suggestion
        0..=0
```

tdaede:
Will merge as soon as travis passes.

PR #1696, opened 2019-09-25
Start adding cold to native implementations

KyleSiefring:


KyleSiefring:
Worth a couple % speedup from this old run.

https://beta.arewecompressedyet.com/?job=new_static_sad_satd%402019-09-21T17%3A07%3A31.862Z&job=cold_sad_satd%402019-09-21T21%3A48%3A12.186Z

KyleSiefring:
Better measurements.

https://beta.arewecompressedyet.com/?job=master-4faec275639df959a0c56dca09dbf1072975ba26&job=cold_as_ice%402019-09-25T15%3A50%3A16.337Z

shssoichiro:
WRT the native functions, these are controlled at some level by conditional compilation. We don't want to mark them as cold if ASM support isn't compiled in, so I think it would make sense to use `#[cfg_attr(all(target_arch = "x86_64", feature = "nasm"), cold)]`.

KyleSiefring:
> WRT the native functions, these are controlled at some level by conditional compilation. We don't want to mark them as cold if ASM support isn't compiled in, so I think it would make sense to use `#[cfg_attr(all(target_arch = "x86_64", feature = "nasm"), cold)]`.

I'm a little concerned about spewing this string everywhere. Is a way to construct a new cfg variable and use that instead?

KyleSiefring:
It's not going to be fast anyway if we don't have simd and I really don't want to have to change this for every platform added.

shssoichiro:
Someone on IRC already expressed concerns about rav1e getting slower on ARM platforms over the past month, so I'd like to avoid making those platforms slower. I know it's an ugly string. I wonder if there would be some other alternative to keep from having to put that cfg string everywhere?

lu-zero:
We can make custom attributes and I guess I can make a proc_macro for it while I'm waiting to have in stable the decoration of impl blocks.

lu-zero:
[here](https://crates.io/crates/simd-helpers) the first proc_macro, I'd add more as the needs arises.

ycho:
> cfg_attr

Never was able to guess what 'cold' really means!!
https://doc.rust-lang.org/reference/attributes/codegen.html#the-cold-attribute

PR #1697, opened 2019-09-25
Eliminate redundant integral image calculations in LRF RDO

xiphmont:
Right now, 32 functionally (though not practically) identical integral images are computed for each LRF filter optimization test.  The differences between the images can be reconciled to allow a single computation to be shared across the entire test process.


ycho:
I was seeing differences in those two functions where integral_image is computed, 'rows_iter' that is passed to sgrproj_solve() and sgrproj_stripe_filter(), i.e
https://github.com/xiphmont/rav1e/blob/b074992e54ab4067b3eabecd45110d451da95c2d/src/lrf.rs#L636
and 
https://github.com/xiphmont/rav1e/blob/b074992e54ab4067b3eabecd45110d451da95c2d/src/lrf.rs#L833

Then I though that might be the reason I was not able to not call setup_integral_image() for sgrproj_stripe_filter().

Btw, my PR (which was done before your max LRU landed) is here https://github.com/xiph/rav1e/pull/1600. It worked fine and gives ~5 % total encoding time reduction.

xiphmont:
AWCY results:
lrf-iiopt-baseline@2019-09-25T22:41:14.816Z -> lrf-iiopt-test@2019-09-25T22:41:44.061Z

-0.0006 | -0.0104 |  0.1664 |   0.0207 | 0.0170 |  0.0150 |    -0.0114

The small functional change is due to altering the boundary handling slightly in the RDO loop.  Previously each LRF filter was strictly bounded to its LRU with no addressing outside either the horizontal bounds or passed in vertical stripe.  Originally, this was to simplify corner cases in search processing.  With integral images it's longer necessary.  The filters are now allowed to peek outside the edges of the rectangle as they would in a normal filtering operation. This change allows easier unification of the II setup code.

As expected, the patches collectively get a roughly 5-8% overall encoding time speedup.

src/lrf.rs
```diff
@@ -1458,11 +1402,9 @@ impl RestorationState {
                 continue;
               }
 
-              sgrproj_stripe_filter(
-                set,
-                xqd,
-                fi,
+              setup_integral_image(
                 &mut stripe_filter_buffer,
+                STRIPE_IMAGE_STRIDE,
```
lu-zero: it seems constant over all the function calls, would make sense to embed it inside the function?

src/lrf.rs
```diff
@@ -1471,6 +1413,18 @@ impl RestorationState {
...
                 &pre_cdef.planes[pli]
                   .slice(PlaneOffset { x: x as isize, y: stripe_start_y }),
+              );
+
+              sgrproj_stripe_filter(
+                set,
+                xqd,
+                fi,
+                &mut stripe_filter_buffer,
+                STRIPE_IMAGE_STRIDE,
```
lu-zero: Same question here.

src/lrf.rs
xiphmont: It's not.  You missed the calls coming in from rdo.rs

src/lrf.rs
```diff
@@ -642,17 +641,6 @@ pub fn sgrproj_stripe_filter<T: Pixel>(
   let s_r2: u32 = SGRPROJ_PARAMS_S[set as usize][0];
   let s_r1: u32 = SGRPROJ_PARAMS_S[set as usize][1];
 
-  setup_integral_image(
```
ycho: Thanks for moving this out and having it called only once!

src/lrf.rs
```diff
@@ -812,17 +800,6 @@ pub fn sgrproj_solve<T: Pixel>(
   let mut h: [[f64; 2]; 2] = [[0., 0.], [0., 0.]];
   let mut c: [f64; 2] = [0., 0.];
 
-  setup_integral_image(
```
ycho: Thanks for moving this out and having it called only once!

PR #1699, opened 2019-09-26
Bump dependencies that rely on old syn version

shssoichiro:
Syn is our slowest-compiling dependency, so we would like
to avoid compiling two versions of it. We still have an old
version of bindgen used by aom-sys and avformat-sys which uses
old syn, but this moves us closer to compiling syn only once.

The new arrayvec also comes with some performance improvements
of its own.


shssoichiro:
Is someone in the org able to publish a new version of aom-sys? The dependency bump we need is in master of that repo, but hasn't been published to crates.io. If we can get that, I can update avformat-sys as well and we can get off of syn 0.x completely and shave a minute off of build times.

lu-zero:
On it, meanwhile we should land the cargo-c workaround.

shssoichiro:
Going to go all-in and clear out all the syn 0.x dependencies. Just waiting on https://github.com/rust-num/num-derive/issues/32 now.

shssoichiro:
Let it be done.

```
➜ rg 'syn 0' Cargo.lock | wc -l
0
```

shssoichiro:
This actually cut a full 28% off the release build on my machine (360s -> 260s)

PR #1717, opened 2019-10-03
Conform all ASM functions to a module structure

### LINKED ISSUE #1714: lrf_simd module doesn't follow conventions of other SIMD code
This module should be moved into the `asm` module, and refactored to choose the function based on the selected CpuFeatureLevel, and to use native code when the "nasm" feature is disabled,

shssoichiro:
This includes finding any ASM functions that are
scattered throughout the codebase and moving them
into the `asm` module, but also reorganizes
the `asm` module so that `x86` is a module
containing submodules for each native module.
This reduces the number of `cfg` statements needed,
and will also make it cleaner to implement assembly
for other architectures moving forward.

Partially addresses #1714


src/asm/x86/transform/inverse.rs
```diff
@@ -0,0 +1,155 @@
...
+    let stride = output.plane_cfg.stride as isize;
+
+    // perform the inverse transform
+    Self::match_tx_type(tx_type)(
+      output.data_ptr_mut() as *mut _,
+      stride,
+      coeff16.array.as_ptr(),
+      (coeff_w * coeff_h) as i32,
+    );
+    return;
```
barrbrain: Clippy disapproves of this stray return. 
```suggestion
```

shssoichiro:
Oh boy. Okay, I agree with you, although it's going to be _fun_ trying to split this into two PRs. I'll see what I can do.

shssoichiro:
@barrbrain This is now specific to moving the ASM code into a unified module structure. After this is merged, I'll open a separate PR for the CPU feature level changes.

shssoichiro:
Looking over this again, it looks like this only half-completes #1714. I will fix the subject. Thanks!

PR #1740, opened 2019-10-08
Use CpuFeatureLevel for all ASM functions

### LINKED ISSUE #1714: lrf_simd module doesn't follow conventions of other SIMD code

shssoichiro:
Closes #1714

### DIFF
--- a/src/api/config.rs
+++ b/src/api/config.rs
@@ -14,7 +14,6 @@ use num_derive::*;
 use crate::api::color::*;
 use crate::api::Rational;
 use crate::api::{Context, ContextInner};
-use crate::cpu_features::CpuFeatureLevel;
 use crate::encoder::Tune;
 use crate::partition::BlockSize;
 use crate::tiling::TilingInfo;
@@ -107,6 +106,8 @@ pub struct EncoderConfig {
   pub show_psnr: bool,
   /// Enables dumping of internal RDO training data.
   pub train_rdo: bool,
+  /// Target CPU feature level.
+  pub cpu_feature_level: crate::cpu_features::CpuFeatureLevel,
 }
 
 /// Default preset for EncoderConfig: it is a balance between quality and
@@ -159,6 +160,7 @@ impl EncoderConfig {
       speed_settings: SpeedSettings::from_preset(speed),
       show_psnr: false,
       train_rdo: false,
+      cpu_feature_level: Default::default(),
     }
   }
 
@@ -574,7 +576,7 @@ impl Config {
 
     self.validate()?;
 
-    info!("CPU Feature Level: {}", CpuFeatureLevel::default());
+    info!("CPU Feature Level: {}", self.enc.cpu_feature_level);
 
     let pool = rayon::ThreadPoolBuilder::new()
       .num_threads(self.threads)
--- a/src/api/config.rs
+++ b/src/api/config.rs
@@ -14,6 +14,7 @@ use num_derive::*;
 use crate::api::color::*;
 use crate::api::Rational;
 use crate::api::{Context, ContextInner};
+use crate::cpu_features::CpuFeatureLevel;
 use crate::encoder::Tune;
 use crate::partition::BlockSize;
 use crate::tiling::TilingInfo;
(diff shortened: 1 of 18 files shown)


shssoichiro:
Per IRC discussion, closing this in favor of a different approach.

shssoichiro:
Going to bring this up to date and move to a static instance of CpuFeatureLevel to remove the extra fn args all over the place. We need to decide between this and #1749.

src/api/config.rs
```diff
@@ -107,6 +106,8 @@ pub struct EncoderConfig {
   pub show_psnr: bool,
   /// Enables dumping of internal RDO training data.
   pub train_rdo: bool,
+  /// Target CPU feature level.
+  pub cpu_feature_level: crate::cpu_features::CpuFeatureLevel,
```
lu-zero: It should not live in a public field. `pub(crate)` please

src/asm/x86/ec.rs
```diff
@@ -1,17 +1,27 @@
...
 use crate::ec::native;
 #[cfg(target_arch = "x86")]
 use std::arch::x86::*;
 #[cfg(target_arch = "x86_64")]
 use std::arch::x86_64::*;
 
-pub fn update_cdf(cdf: &mut [u16], val: u32) {
-  if cdf.len() == 5 && cfg!(target_feature = "sse2") {
+pub fn update_cdf(cdf: &mut [u16], val: u32, cpu: CpuFeatureLevel) {
+  if cdf.len() == 5 && cpu >= CpuFeatureLevel::SSE2 {
```
lu-zero: currently  update_cdf_4_sse2 works on x86_64, so would make sense to not add a branch + a argument.

src/asm/x86/predict.rs
```diff
@@ -48,208 +58,207 @@ decl_cfl_pred_fn!(rav1e_ipred_cfl_128_avx2);
 decl_cfl_pred_fn!(rav1e_ipred_cfl_left_avx2);
 decl_cfl_pred_fn!(rav1e_ipred_cfl_top_avx2);
 
-pub trait Intra<T>: native::Intra<T>
+pub trait Intra<T: Pixel>: native::Intra<T>
```
lu-zero: Why?

src/asm/x86/predict.rs
```diff
@@ -48,208 +58,207 @@ decl_cfl_pred_fn!(rav1e_ipred_cfl_128_avx2);
...
-            output.data_ptr_mut() as *mut _,
-            output.plane_cfg.stride as libc::ptrdiff_t,
-            above.as_ptr().offset(-1) as *const _,
-            Self::W as libc::c_int,
-            Self::H as libc::c_int,
-            0,
-          )
-        };
-      }
+  fn pred_dc(
```
lu-zero: I'd rather have a separate impl and pick the correct one where this is used.

src/ec.rs
```diff
@@ -20,6 +20,7 @@ pub use self::native::*;
   any(target_arch = "x86", target_arch = "x86_64")
 ))]
 pub use crate::asm::x86::ec::*;
+use crate::cpu_features::CpuFeatureLevel;
```
lu-zero: this probably should be avoided at least for now (see comments above)

src/api/config.rs
shssoichiro: Our benchmark suite really does not like this... Should I conditionally make it `pub` if `bench` feature is enabled? It's kind of ugly...

src/asm/x86/predict.rs
shssoichiro: Merge conflict resolution fail.

src/api/config.rs
lu-zero: Do not put there the field alternatively.

src/asm/x86/predict.rs
shssoichiro: Could you clarify what you're looking for? I'm a bit confused.

shssoichiro: At what level should we be pickling the correct one? If we select it in `predict_intra_inner`, it makes the code messier and removes the isolation of the ASM functions, because of how we currently choose "asm" or "non-asm" at compile time by importing the correct crate. I don't believe we would be able to import the correct feature level at runtime, so we would need a different approach.

shssoichiro:
Rebased

src/asm/x86/predict.rs
shssoichiro: Per IRC discussion, barrbrain will work on this in a separate PR.

PR #1745, opened 2019-10-09
Define allowed clippy lints in code

### LINKED ISSUE #1744: Define clippy allowed lints in code
We can define e.g. `#[allow(clippy::annoying_lint)]` at the top of `lib.rs` and `main.rs` instead of specifying them inline in .travis.yml. This would make it easier for newcomers to the project to not need to know we have a special clippy command. This has been allowed for a few rust versions now (it used to break compiling if you weren't running clippy--now it's ignored if you don't have clippy and works fine). This would also remove the need for #1406.

shssoichiro:
Closes #1744

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -82,9 +82,7 @@ Clippy Lints:
   script:
     - touch src/lib.rs # Ensure cargo re-checks rav1e, since we ran `cargo check` earlier
     - >-
-      cargo clippy -- -D warnings -A clippy::cast_lossless -A clippy::cast_ptr_alignment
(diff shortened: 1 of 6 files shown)


tdaede:
I was worried that we'd need the big blocks of clippy allows in every file, but if it's only in lib.rs that's a lot more reasonable.

Does this bump the minimum rustc version?

shssoichiro:
@tdaede No, this works on 1.36 and up. Travis is happy. :slightly_smiling_face:

shssoichiro:
Split into two commits per derf's request on IRC

PR #1748, opened 2019-10-09
Unbreak x86

lu-zero:
@LigH-de could you please test this?


LigH-de:
Applying patch 1748 allows compilation of rav1e for Win32 (test for Win64 pending...); a lack of optimization will be a disadvantage, though.

lu-zero:
If you can help us set up a 32bit target for the CI it would be great, this is the quick bandaid since I do not have a x86 32bit cross-compiler ready here.

LigH-de:
I am mainly a user of the [media-autobuild suite](https://github.com/m-ab-s/media-autobuild_suite), my knowledge in UNIX like environments and in C family programming languages is pretty limited. I can recognize a few probable reasons of failures and tell educated guesses at best. If you have any Windows PC available, MABS is surely one of the most versatile projects to install and keep an MSYS2/MinGW environment up to date before building everything that may fit into ffmpeg. Apart from that, I barely remember that different Linux distros provide different cross-compiling packages, but I would not be able to keep them alive without support from e.g. Linux user forums...

lu-zero:
There is no lack of optimization since the nasm code is x86_64-only.

PR #1749, opened 2019-10-09
Remove CpuFeatureLevel in favor of std_detect

shssoichiro:
This uses the trait based approach discussed in IRC. In order to be able to set the env var for testing, https://github.com/rust-lang/stdarch/commit/b8aecbfb8ea2ac028426da48b0f9ca6a730b9730 will need to be pulled into the Rust tree and released in a nightly, but unfortunately it is not released yet. In lieu of that, I would like to make a follow-up PR to this one to create automated tests for each version of each ASM function that doesn't already have tests.

### DIFF
--- a/benches/dist.rs
+++ b/benches/dist.rs
@@ -10,7 +10,6 @@
 use criterion::*;
 use rand::{Rng, SeedableRng};
 use rand_chacha::ChaChaRng;
-use rav1e::bench::cpu_features::*;
 use rav1e::bench::dist;
 use rav1e::bench::frame::*;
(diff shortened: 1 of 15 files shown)


shssoichiro:
We need to decide between this or #1740.

shssoichiro:
From IRC, it sounds like the very compelling argument to go with the other approach is that a lookup table should be faster than if/else statements. Closing this in favor of that one.

ycho:
Thank you very much for spending your time & efforts trying this though!

PR #1760, opened 2019-10-12
fix spelling and grammar mistake

jagannathBhat:


README.md
```diff
@@ -12,11 +12,11 @@ rav1e is an AV1 video encoder. It is designed to eventually cover all use cases,
 * 64x64 superblocks
 * 4x4 to 64x64 RDO-selected square and 2:1/1:2 rectangular blocks
 * DC, H, V, Paeth, smooth, and a subset of directional prediction modes
-* DCT, ADST and identity transforms (up to 64x64, 16x16 and 32x32 respectively)
+* DCT, ADST, and identity transform (up to 64x64, 16x16 and 32x32 respectively)
 * 8-, 10- and 12-bit depth color
 * 4:2:0 (full support), 4:2:2 and 4:4:4 (limited) chroma sampling
```
vibhoothiiaanand: Anyway we are documenting I would propose we could have more nit fixing like having 
`4:2:2, and 4:4:4` like others, it would be good imo

vibhoothiiaanand: Also 
` 8-, 10- and 12-bit depth color`
to 
` 8-, 10-, and 12-bit depth color`

README.md
```diff
@@ -12,11 +12,11 @@ rav1e is an AV1 video encoder. It is designed to eventually cover all use cases,
 * 64x64 superblocks
 * 4x4 to 64x64 RDO-selected square and 2:1/1:2 rectangular blocks
 * DC, H, V, Paeth, smooth, and a subset of directional prediction modes
-* DCT, ADST and identity transforms (up to 64x64, 16x16 and 32x32 respectively)
+* DCT, ADST, and identity transform (up to 64x64, 16x16 and 32x32 respectively)
```
gnafuthegreat: I believe this was correct before.  It's referring to multiple transforms, so it should be plural.  I don't disagree with the added comma, though; I prefer an Oxford comma myself.

README.md
```diff
@@ -12,11 +12,11 @@ rav1e is an AV1 video encoder. It is designed to eventually cover all use cases,
...
 * 8-, 10- and 12-bit depth color
 * 4:2:0 (full support), 4:2:2 and 4:4:4 (limited) chroma sampling
 * Variable speed settings
-* Near real-time encoding at high speed levels
+* Near real-time encoding at high-speed levels
```
gnafuthegreat: This was more correct before.  Rather than referring to a single "high-speed" level (like a "high-speed train"), it is referring to high speed levels in general (the speed levels are high, rather than a level being high-speed).

README.md
```diff
@@ -92,7 +92,7 @@ rav1e has several optional features that can be enabled by passing --features to
 
 ### Online Analyzer
 
-If your `.ivf` file is hosted somewhere (and CORS is enabled on your web server) you can use:
+If your `.ivf` file is hosted somewhere (and CORS is enabled on your webserver) you can use:
```
gnafuthegreat: This may be more personal preference, but as an example the [Wikipedia article](https://en.wikipedia.org/wiki/Web_server) includes the space and it seems the more common usage is to spell it as two words.

jagannathBhat:
Thanks for the review @vibhoothiiaanand @gnafuthegreat

jagannathBhat:
Squashed

rzumer:
These are not grammar mistakes, almost all changes are valid one way or the other, and one (pluralizing "library") is wrong.

jagannathBhat:
agreed

PR #1764, opened 2019-10-14
Add tests for inverse transform ASM

shssoichiro:
These tests run the ASM code against the native code
and verify that the results are equal.

### DIFF
--- a/src/asm/x86/transform/inverse.rs
+++ b/src/asm/x86/transform/inverse.rs
@@ -249,123 +249,132 @@ decl_tx_fns! {
 }
 
 #[cfg(test)]
-#[rustfmt::skip]
 mod test {
   use super::*;
   use crate::frame::{AsRegion, Plane};
   use crate::transform::TxSize::*;
   use rand::random;
 
-  macro_rules! decl_tx_fn_test {
-    ($ENUM:pat, $TYPE1:ident, $TYPE2:ident, $W:expr, $H:expr, $OPT:ident, $OPTLIT:literal) => {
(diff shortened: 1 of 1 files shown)


shssoichiro:
I wanted to implement this for one set of ASM functions first, to make sure this is a sane approach, before adding similar tests for more ASM functions.

My primary concern is if the machine running the tests does not have the target feature available. I didn't want that to cause the tests to fail, but a passing test swallows any stderr output, so a developer on an SSSE3 CPU will have no feedback if they have broken AVX2 tests. I wasn't able to figure out a good solution to this, however, as there seems to be no way to "ignore" a test at runtime.

shssoichiro:
Reworking this since the ssse3 functions don't go through the same macro, so they don't get the tests.

src/asm/x86/transform/inverse.rs
```diff
@@ -247,3 +247,125 @@ decl_tx_fns! {
...
+#[cfg(test)]
+#[rustfmt::skip]
+mod test {
+  use super::*;
+  use crate::frame::{AsRegion, Plane};
+  use crate::transform::TxSize::*;
+  use rand::random;
+
+  macro_rules! decl_tx_fn_test {
+    ($ENUM:pat, $TYPE1:ident, $TYPE2:ident, $W:expr, $H:expr, $OPT:ident, $OPTLIT:literal) => {
```
barrbrain: The list of test cases would be a little nicer if this macro accepted a list. Then we might call it `test_itx_fns`.

PR #1766, opened 2019-10-15
Simd cleanup

lu-zero:
This is mainly to make easier to add aarch64 support.


lu-zero:
While I'm waiting for travis to unbreak arm rustup, here a further simplification.

PR #1767, opened 2019-10-15
Set speed 9 minimum partition size to 32x32

rzumer:
Otherwise speed 9 is too close to 10. This provides a better middle ground between the speed and quality tradeoff of 8x8 and 64x64 partitions.

[AWCY run](https://beta.arewecompressedyet.com/?job=master-s8%402019-10-10T21%3A23%3A14.371Z&job=sltest-9%402019-10-14T19%3A27%3A09.202Z&job=master-s9%402019-10-10T21%3A23%3A22.799Z&job=master-s10%402019-10-10T21%3A23%3A35.561Z) comparing speed 8, this PR, current speed 9, and speed 10. With this change the RD curve sits cleanly between 8 and 10, and the performance tradeoff is fair in my opinion.


ycho:
> Otherwise speed 9 is too close to 10. This provides a better middle ground between the speed and quality tradeoff of 8x8 and 64x64 partitions.
> 
> [AWCY run](https://beta.arewecompressedyet.com/?job=master-s8%402019-10-10T21%3A23%3A14.371Z&job=sltest-9%402019-10-14T19%3A27%3A09.202Z&job=master-s9%402019-10-10T21%3A23%3A22.799Z&job=master-s10%402019-10-10T21%3A23%3A35.561Z) comparing speed 8, this PR, current speed 9, and speed 10. With this change the RD curve sits cleanly between 8 and 10, and the performance tradeoff is fair in my opinion.
> 

Yes, looks nice change to me as well.
We used to vary the min block size for speed levels, the easiest control of speed vs quality, but ended up using 4x4, 8x8, and 64x64 only. So, I will even agree if we introduce 16x16 to regularize the gaps between speed levels.

> I have removed the claim that a minimum partition size of 32x32 without the reduced transform set is less efficient, since it does not seem accurate, but based on my tests it seems true for 16x16, which is why I have not added more levels for this setting.

Related to this, current 'the reduced transform set' is a bug for speed levels 5 and 6, since the feature does not work at those two levels because tx rdo  is  disabled for them. There is slight change in bitstream though, since reduced_tx_set flag changes the frame header.

PR #1763, opened 2019-10-14
Merge dav1d 0.1.0 x86 assembly

barrbrain:
Advances #1750.

Difference between upstream commit list and this PR:
```diff
+Integrate inv_txfm_add_..._4x4_ssse3 from itx_ssse3.asm
+Integrate ipred_h_ssse3 from ipred_ssse3.asm
+asm/x86: Use list syntax in macros
+Integrate avg_ssse3 from mc_ssse3.asm
+Add rav1e_dr_intra_derivative table for asm
-cdef: simplify sec_taps
+Add rav1e_mc_warp_filter table for asm
-Add a 4x4 cdef_filter AVX2 implementation
-Add 8x8 cdef_filter AVX2 implementation
-Add AVX2 SIMD implementation for cdef_dir
+Add rav1e_sgr_x_by_xplus1 table for asm
+Add rav1e_filter_intra_taps table for asm
-x86: Add chroma-from-luma intra prediction AVX2 asm
-x86: Add smooth intra prediction AVX2 asm
-x86: Add paeth intra prediction AVX2 asm
-x86: Add dc/h/v intra prediction AVX2 asm
-x86: Enable ITX AVX2 asm on 64-bit Windows
-x86: Fix out-of-bounds write in ITX asm
-x86-64: Inverse transforms AVX2
-x86: Enable MC AVX2 asm on 64-bit Windows
-x86: MC AVX2
-x86 asm infrastructure
+Move mc_subpel_filters table from x86 asm to shared Rust
```

The first 2 MC commits and the "x86 asm infrastructure" commit were already squashed in #639.
The ITX commits were squashed in #864, along with other commits from dav1d-0.2.0.
Our CDEF assembly is divergent from dav1d, and appears to be a subset of code in 0.2.0.

PR #1768, opened 2019-10-15
Nasm to rust

lu-zero:
Since @barrbrain wanted an example. It is less straightforward than it should indeed.


lu-zero:
There is something really fishy going on. Reproduce the dav1d failures locally.

src/asm/x86/mc.rs
```diff
@@ -363,3 +363,88 @@ pub(crate) static AVG_FNS: [Option<AvgFn>; CpuFeatureLevel::len()] = {
 
 pub(crate) static AVG_HBD_FNS: [Option<AvgHBDFn>; CpuFeatureLevel::len()] =
   [None; CpuFeatureLevel::len()];
+
+#[rustfmt::skip]
+static MC_SUBPEL_FILTERS: AlignedArray<[i8; 600]> = AlignedArray::new([
```
barrbrain: The following change passes tests for me with feature `decode_test_dav1d`:
```suggestion
#[no_mangle]
pub static rav1e_mc_subpel_filters: AlignedArray<[i8; 600]> = AlignedArray::new([
```

src/asm/x86/mc.rs
```diff
@@ -363,3 +363,88 @@ pub(crate) static AVG_FNS: [Option<AvgFn>; CpuFeatureLevel::len()] = {
...
+0,   0,   6,  26,  26,   6,   0,   0,
+0,   0,   6,  24,  27,   7,   0,   0,
+0,   0,   5,  23,  28,   8,   0,   0,
+0,   0,   4,  22,  29,   9,   0,   0,
+0,   0,   3,  21,  30,  10,   0,   0,
+0,   0,   2,  20,  31,  11,   0,   0,
+0,   0,   2,  18,  31,  13,   0,   0,
+0,   0,   1,  17,  31,  15,   0,   0 ]);
+
+#[no_mangle]
```
barrbrain:
```suggestion
```

src/asm/x86/mc.rs
```diff
@@ -363,3 +363,88 @@ pub(crate) static AVG_FNS: [Option<AvgFn>; CpuFeatureLevel::len()] = {
...
+pub static rav1e_mc_subpel_filters: &[i8] = &MC_SUBPEL_FILTERS.array;
```
barrbrain:
```suggestion
```

src/asm/x86/mc.rs
lu-zero: Right, that was my doubt, I still do not understand how it could work for the aom tests though.

barrbrain:
Thanks for the insight. This was handled in #1763.

PR #1769, opened 2019-10-15
Quantize in scan order

KyleSiefring:
Previous compression optimizations made the quantization process
dependent on the order that the coefficients are processed without
changing to using scan order. This patch corrects this.

Improves quality at the expense of encode speed.
```
-0.4854 | -0.2923 | -0.4541 |  -0.6873 | -0.3795 | -0.5577 |    -0.4489
```

### DIFF
--- a/src/quantize.rs
+++ b/src/quantize.rs
@@ -9,9 +9,10 @@
 
 #![allow(non_upper_case_globals)]
 
-use crate::transform::TxSize;
+use crate::transform::{TxSize, TxType};
 use crate::util::*;
 
+use crate::scan_order::av1_scan_orders;
 use num_traits::*;
 use std::convert::Into;
 use std::mem;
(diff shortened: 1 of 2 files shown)


src/quantize.rs
```diff
@@ -284,30 +289,25 @@ impl QuantizationContext {
     // To that end, we want to bias more toward rounding to zero for
     // that tail of zeroes and ones than we do for the larger coefficients.
     let mut level_mode = 1;
-    for (qc, c) in
-      qcoeffs[1..].iter_mut().zip(coeffs[1..].iter()).take(last_pos)
-    {
-      let coeff = *c << self.log_tx_scale;
+    for &pos in scan[1..].iter().take(last_pos) {
```
KyleSiefring: I just noticed I have some confusing naming conflicts I need to fix in this patch. Should change the earlier pos and last_pos.

KyleSiefring:
Ugh, breaking in the dequantize loop. It's accessing row/cols > 32 in there and it's not doing that now here.

KyleSiefring:
Will look into fixing tomorrow.

KyleSiefring:
#1770 should fix the problem I'm encountering here.

KyleSiefring:
Okay, it's passing now.

PR #1770, opened 2019-10-16
After quantizing, use only the codable area of coefficients

KyleSiefring:


KyleSiefring:
<derf> Okay, the patch itself looks fine. I guess the test can be addressed in a follow-up patch.
<derf> So r=me.

PR #1773, opened 2019-10-17
Implement FlipADST inverse transforms

barrbrain:
This is the most basic implementation of the 1D kernels: ADST followed by reversal of output.


shssoichiro:
I like the simplicity. Can you run an AWCY for this change?

rzumer:
We should consider whether or not it is worth enabling these for speed levels 2 to 6.

barrbrain:
[AWCY results at speed 4](https://beta.arewecompressedyet.com/?job=master-eb3be3a7-s4%402019-10-17T16%3A27%3A26.045Z&job=flipadst-s4%402019-10-17T16%3A27%3A43.836Z) are coming soon.

shssoichiro:
Actually, here's a concern: This seems to increase encoding time by about 20% at low QP, and by less as QP increases. Is this a significant enough concern to warrant changing speed levels (beyond the changes coming to fix limited tx set?)

rzumer:
I suggest enabling only for speed 2 and below. 2 is very close to 3 right now, so this will differentiate them a little better, and the BDrate improvement is too small to justify a large increase in encode time for most levels in my opinion.

barrbrain:
Squashing the latest change, this has no impact on metrics and encoding time. We can rework the speed settings for transform sets after more analysis.

PR #1777, opened 2019-10-18
Refactor iadst/flipadst

lu-zero:
We can do w/out the `reverse()`.

### DIFF
--- a/src/transform/inverse.rs
+++ b/src/transform/inverse.rs
@@ -61,6 +61,7 @@ pub fn av1_iflipadst4(input: &[i32], output: &mut [i32], range: usize) {
   output[..4].reverse();
 }
 
+#[inline(always)]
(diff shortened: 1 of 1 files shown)


shssoichiro:
From dumping the asm for the 8 and 16 functions, it looks like the improvement comes entirely from the inlining, and the compiler is able to optimize from there. The removal of `reverse` is unnecessary (which is good--I feel this change makes the code less maintainable). iflipadst4 is already inlined on master due to it being a small function.

rzumer:
I don't see much benefit in moving the logic in one place only to use an if-else where both branches are mostly duplicated logic. Why not use a variable index?

lu-zero:
if the compiler is smart enough to do the right thing with reverse +
inline(always) we can go that way.

>

PR #1781, opened 2019-10-18
Rename RDO(Partition)Output to Partition(Group)Parameters and reuse them where appropriate

rzumer:
Part of larger cleanup effort, but I'm opening this first in case there are comments on this change. I have thought for a while that the current names are a bit vague, and the parameters are reused without transformation in the encoder module, so they are not necessarily specific to RDO.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -101,6 +101,26 @@ pub struct PartitionParameters {
   pub sidx: u8,
 }
 
+impl Default for PartitionParameters {
+  fn default() -> Self {
+    PartitionParameters {
+      rd_cost: std::f64::MAX,
+      bo: TileBlockOffset::default(),
+      bsize: BlockSize::BLOCK_INVALID,
+      pred_mode_luma: PredictionMode::default(),
+      pred_mode_chroma: PredictionMode::default(),
+      pred_cfl_params: CFLParams::default(),
+      ref_frames: [RefType::INTRA_FRAME, RefType::NONE_FRAME],
+      mvs: [MotionVector::default(); 2],
+      skip: false,
+      has_coeff: true,
+      tx_size: TxSize::TX_4X4,
+      tx_type: TxType::DCT_DCT,
+      sidx: 0,
+    }
+  }
+}
+
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
 pub struct RDOTracker {
   rate_bins: Vec<Vec<Vec<u64>>>,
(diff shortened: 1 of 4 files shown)


src/rdo.rs
```diff
@@ -797,7 +797,7 @@ pub fn rdo_mode_decision<T: Pixel>(
   fi: &FrameInvariants<T>, ts: &mut TileStateMut<'_, T>,
   cw: &mut ContextWriter, bsize: BlockSize, tile_bo: TileBlockOffset,
   pmvs: &mut [Option<MotionVector>], inter_cfg: &InterConfig,
-) -> RDOPartitionOutput {
+) -> PartitionParameters {
```
ycho: Input to this function is a 'block', i.e, one partition, hence I expect the reader assumes that the output from this function is rdo results of one 'partition'. So, I doubt we need include 'partition' as part of the output name?

src/rdo.rs
```diff
@@ -78,14 +78,14 @@ impl RDOType {
...
+pub struct PartitionGroupParameters {
   pub rd_cost: f64,
   pub part_type: PartitionType,
-  pub part_modes: Vec<RDOPartitionOutput>,
+  pub part_modes: Vec<PartitionParameters>,
 }
 
 #[derive(Clone)]
-pub struct RDOPartitionOutput {
+pub struct PartitionParameters {
```
ycho: If you see the fields in this structure, they are all about the modes of one parittion, that are decided by rdo_mode_decision( ).
So, I don't think the new name 'Parameters' is more implying that meaning than current 'RDO'?
Current one has been not bad, or I need ever want to revise it again, I'd go simply swap two parts there, i,e. PartitionRDOResult, but that can easily mislead us to think it is about results of rdo-based partition search.
Thus, I'd end up with RDOOutput, i.e use already taken name for above struct, which can confuse us! :)

src/rdo.rs
rzumer: What is the problem? The struct contains the parameters for coding a single partition, so if it is not part of the name there is missing information. Specifying it makes the code self-documenting, and if it were called "Parameters" and used elsewhere in the program, it would be impossible to tell what kind of data is inside at a glance.

src/rdo.rs
```diff
@@ -78,14 +78,14 @@ impl RDOType {
 }
 
 #[derive(Clone)]
-pub struct RDOOutput {
+pub struct PartitionGroupParameters {
```
ycho: To me, most guidance for naming for this struct will be that this struct can have
the results from multiple partitions, i.e. split partitions. And I believe, your new name "PartitioGroup" wanted to care that point.
As non native English user,  I hope adding "ing" to below struct name, i.e. "RDOPartitioningOutput" would fit for this struct! Since partitioning implies to me that in the middle of splitting partition, thus info from multiple partitions.

src/rdo.rs
rzumer: It also contains parameters for transform size and type, motion vectors and other data that is not related to the prediction mode.

src/rdo.rs
rzumer: To me that would not be an intuitive name at all, but I'd like to hear other opinions on this.

rzumer:
I just noticed that there is also `EncodingSettings` which has almost the exact same layout as `RDOPartitionOutput`. Probably not worth keeping around but that is another name being used for roughly the same thing...

rzumer:
Pushed some additional changes, notably removing the redundant `EncodingSettings` struct. Next step is to reuse `PartitionParameters` when calling block encode functions, but my first attempt is desyncing, so currently debugging.

PR #1790, opened 2019-10-22
Fix tile rate restriction calculation in tiler

rzumer:
When calculating the minimum number of tiles needed
to code a valid bitstream, the rate restriction did
not return valid results. This uses the Rust log2()
function instead of an integer logarithm, and rounds
the result up rather than to the nearest integer.


ycho:
Thanks for fixing this!

PR #1774, opened 2019-10-17
Rebalance transform set speed presets

rzumer:
Closes #1707 via @ycho's suggested solution.

[Here](https://beta.arewecompressedyet.com/?job=master-s4%402019-10-17T18%3A18%3A25.883Z&job=master-s5%402019-10-17T18%3A19%3A08.579Z&job=s5-txset%402019-10-17T16%3A41%3A04.262Z&job=master-s6%402019-10-17T18%3A20%3A01.049Z) is a comparison between speed level 4, 5 as of now, 5 with this PR, and 6. It is clear that 5 and 6 are far too similar right now. This PR enables transform type and size RDO with a reduced transform set for speed level 5, which differentiates level 5 from 6 while remaining a reasonable speed/quality tradeoff compared to level 4.

PR #1794, opened 2019-10-23
Update speed settings documentation

shssoichiro:
These updates were missed as part of #1774


rzumer:
Thanks, #1767 and #1772 were also missing changes in the comments so if you can update those for speed levels 9 and 10 that would be great.

PR #1795, opened 2019-10-24
Optimizations to loops within write_coeffs_lv_map

shssoichiro:
Results in approximately 1% overall encoding time reduction
at default settings, mostly by eliding bounds checks,
allowing the compiler to auto-vectorize loops.

The speedup, although small, is consistently reproducible
and verifiable through reduced instruction counts and profiler
reporting 10% reduced time spent in write_coeffs_lv_map,
which is responsible for ~10% of the total encoding time,
adding up to 1% total reduction.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1789,18 +1789,19 @@ impl<'a> BlockContext<'a> {
     let txb_w_unit = tx_size.width_mi();
     let txb_h_unit = tx_size.height_mi();
 
+    let above_ctxs =
+      &self.above_coeff_context[plane][(bo.0.x >> xdec)..][..txb_w_unit];
+    let left_ctxs =
+      &self.left_coeff_context[plane][(bo.y_in_sb() >> ydec)..][..txb_h_unit];
+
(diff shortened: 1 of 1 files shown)


shssoichiro:
Regarding those asserts, Rust inserts them automatically as bounds checks on the slice accesses. The code generated is identical without explicitly asserting there.

ycho:
Had no chance to see the code, but thank you for looking and already showing the speed gains.
Coefficient coding in AV1 is new to most video codec engineers thus least discovered area.
I expect more speed gains in lower speed like 0, because more time are for coefficient encoding there. Also, I remember @KyleSiefring wanted to investigate the possibility of speedup of coefficient coding, maybe introducing or improving simd code.

ycho:
And, I hope you didn't blame too much who the h*** wrote all this code like C!

shssoichiro:
One other thing is that this should open up other future optimizations, per Intel VTune the bottleneck in this function has changed from front-end bound to branch mispredictions.

PR #1797, opened 2019-10-25
Fix an LRF crash when encoding tiny frames

rzumer:
Found this while fuzzing and took the easy way out. An alternative or complementary solution would be to assert that the frame size is valid and display a proper error message if it is not (I could not find the minimum frame size requirements in the spec, but I seem to recall they exist).

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -187,6 +187,10 @@ impl Sequence {
       tier[i] = 0;
     }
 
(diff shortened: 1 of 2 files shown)


lu-zero:
Another option would be to disable lrf for images smaller than something. I doubt it would have a beneficial impact if any.

rzumer:
@lu-zero I updated the PR with this solution and added an assertion when initializing the restoration context.

PR #1798, opened 2019-10-25
Ensure that frame sizes passed to the API are valid

rzumer:
The AV1 format specification, section A.3 specifies that
the frame width and height must be at minimum 16 px for
a valid AV1 stream. Since we are not upscaling or padding
input video below that size, ~panic with a friendly error message~ return an error in the API.

### DIFF
--- a/src/api/config.rs
+++ b/src/api/config.rs
@@ -465,10 +465,10 @@ impl fmt::Display for PredictionModesSetting {
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Error)]
 pub enum InvalidConfig {
   /// The width is invalid.
-  #[error(display = "invalid width {} (expected > 0, <= 32767)", _0)]
+  #[error(display = "invalid width {} (expected >= 16, <= 32767)", _0)]
   InvalidWidth(usize),
   /// The height is invalid.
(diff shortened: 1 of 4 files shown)


rzumer:
Discussion log:
```
<derf> Chagall: See what I just said above: I don't actually think we should panic.
<Chagall> it seems weird to allow encoding non-conformant streams
<derf> I mean, the "When the mapped level is contained in the tables above" part is important.
<derf> "Levels" are a thing for hardware. It seems perfectly reasonable for software to be able to process things outside of any level.
<Chagall> but it's not like there's a way to signal "no level", even "maximum parameters" is a level
<Chagall> maybe it's just poorly worded but I don't understand how there could not be a mapped level that corresponds to one that is in the tables
<Chagall> except an invalid one
<derf> I mean, I guess that depends on whether you think "contained in the tables above" means in _all_ of the tables or _any_ of the tables.
<derf> But "any" would be a vacuous statement, as you point out.
<Chagall> good thing the tables are not even labelled
<derf> I was noticing that extra-helpful feature also
```

tdaede:
The API should return an error rather than the library panicking, I think.

rzumer:
You're right, I will correct it later.

PR #1800, opened 2019-10-26
Eliminate loop RDO infinite limit cycle loop

xiphmont:
Fix a few bugs in LRF that weren't causing trouble (yet), and one that was (limit cycle infinite loop).

AWCY incoming.


xiphmont:
OK, AWCY is in and I frankly don't quite trust the one patch (intentionally) introducing a metric change.  I'm going to back it out and resubmit the fix.  At that point, AWCY is zeroes as the basic test set doesn't trip the bug this fixes.

PR #1806, opened 2019-10-29
Align end-of-encode stats to what AOMAnalyzer would show

shssoichiro:
AOMAnalyzer shows the percentage of pixels in a frame
that comprise each block size or transform type.
This commit updates rav1e to do the same.


shssoichiro:
Agreed, we had a brief discussion a couple of months ago because this mismatch was discovered. We never came to a hard conclusion, but I think it would be less confusing if we were consistent with other commonly used tooling.

PR #1807, opened 2019-10-30
Disable Travis ARM builds due to them not completing.

tdaede:


EwoutH:
Please describe such a PR a bit better next time. There are not a lot of good reasons to (partially) disable CI all together, but CI servers not booting up is definitely one of them (that is the problem right?), but definitely unlikely.

PR #1808, opened 2019-10-30
Split out inter and intra mode decision

shssoichiro:
Create easier-to-understand code by moving these into separate functions


ycho:
@shssoichiro, Would you like me to land?

shssoichiro:
@ycho I can merge. Thanks! Correct, no change in AWCY.

PR #1810, opened 2019-10-30
Add version APIs to capi.rs.

### LINKED ISSUE #1801: Ability to query version (string) from C API (FFI)
I'm hooking up rav1e support into libavif, and it'd be nice if there were capi calls such as:

```
const char * rav1e_version_short(void);
const char * rav1e_version_full(void);
```

Which returned lib.rs version::short, version::full string, etc, so I could expose those from my `avifCodecVersions()` call.

tdaede:
Fixes #1801

### DIFF
--- a/src/capi.rs
+++ b/src/capi.rs
@@ -226,6 +226,36 @@ pub struct Packet {
   pub frame_type: FrameType,
 }
 
+/// Version information as presented in `[package]` `version`.
+///
+/// e.g. `0.1.0``
+///
+/// Can be parsed by [semver](https://crates.io/crates/semver).
+/// This returns the version of the loaded library, regardless
+/// of which version the library user was built against.
+#[no_mangle]
+pub unsafe extern fn rav1e_version_short() -> *const c_char {
+  concat!(env!("CARGO_PKG_VERSION"), "\0").as_ptr() as *const c_char
+}
+
+/// Version information with the information
+/// provided by `git describe --tags`.
+///
+/// e.g. `0.1.0 (v0.1.0-1-g743d464)`
+///
+/// This returns the version of the loaded library, regardless
+/// of which version the library user was built against.
+#[no_mangle]
+pub unsafe extern fn rav1e_version_full() -> *const c_char {
+  concat!(
+    env!("CARGO_PKG_VERSION"),
+    " (",
+    env!("VERGEN_SEMVER_LIGHTWEIGHT"),
+    ")\0"
+  )
+  .as_ptr() as *const c_char
+}
+
 /// Simple Data
 ///
 ///


src/capi.rs
```diff
@@ -226,6 +226,45 @@ pub struct Packet {
...
+pub unsafe extern fn rav1e_version_short() -> *const c_char {
+    concat!(env!("CARGO_PKG_VERSION"), "\0").as_ptr() as *const c_char
+}
+
+/// Version information with the information
+/// provided by `git describe --tags`.
+///
+/// e.g. `0.1.0 (v0.1.0-1-g743d464)`
+///
+pub unsafe extern fn  rav1e_version_full() -> *const c_charg {
```
barrbrain:
```suggestion
pub unsafe extern fn  rav1e_version_full() -> *const c_char {
```

src/capi.rs
```diff
@@ -226,6 +226,45 @@ pub struct Packet {
...
+    let semver = env!("VERGEN_SEMVER_LIGHTWEIGHT");
+    concat!(format!("{} ({})", short(), semver), "\0").as_ptr() as *const c_char
```
barrbrain: Can merge `concat!` into `format!` here but perhaps it is leaky. 
```suggestion
    format!("{} ({})\0", short(), semver).as_ptr() as *const c_char
```

barrbrain: Alternatively, subject to formatting, this should evaluate to a constant:
```suggestion
    concat!(env!("CARGO_PKG_VERSION"), " (", env!("VERGEN_SEMVER_LIGHTWEIGHT"), ")\0").as_ptr() as *const c_char
```

jamrial:
Following what i mentioned yesterday on IRC, here's a suggestion of how to implement a version querying API, for both compile and runtime usage:

```
// Required. Used by applications that want to support more than one rav1e release
// using pre-processor checks.
#define RAV1E_VERSION_MAJOR 0
#define RAV1E_VERSION_MINOR 1
#define RAV1E_VERSION_PATCH 0

// Optional. convenience macro
#define RAV1E_VERSION \
    RAV1E_VERSION_MAJOR << 16 | \
    RAV1E_VERSION_MINOR << 8  | \
    RAV1E_VERSION_MAJOR

// Optional. Returns RAV1E_VERSION hardcoded in the compiled library, e.g. 0x00000100.
// Useful for applications to easily check if the runtime library version is the same
// as the one it was linked against.
unsigned int rav1e_version(void);

// returns compiled in version string, e.g. "0.1.0"
const char *rav1e_version_short(void);

// returns compiled in version string, e.g. `0.1.0 (v0.1.0-1-g743d464)`
const char *rav1e_version_full(void);
```

tdaede:
I should mention that I can't compile test this patch until the tables fix lands.

tdaede:
Updated patch to just include runtime version info. Version info inside the .h (@jamrial ) will come in the next PR with an update to cargo-c.

jamrial:
Alright. If cargo-c needs an update for that then it can wait. Thanks.

src/capi.rs
```diff
@@ -226,6 +226,34 @@ pub struct Packet {
...
 }
 
+/// Version information as presented in `[package]` `version`.
+///
+/// e.g. `0.1.0``
+///
+/// Can be parsed by [semver](https://crates.io/crates/semver).
+/// This returns the version of the loaded library, regardless
+/// of which version the library user was built against.
+pub unsafe extern fn rav1e_version_short() -> *const c_char {
```
jamrial: Both these functions appear to be missing a #[no_mangle] line. They don't show up in the generated rav1e.h without it here.

tdaede:
this half should be ready to land

PR #1811, opened 2019-10-31
Reenable get_sad 16x16 SSE2

### LINKED ISSUE #1715: Fix and reenable get_sad_16x16_sse2 (crashes when width not mod 64)
Hi
I was doing some tests with rav1e windows build 0.1.0 (20190926-19-ge11f06a) and found one video that I couldn't encode. The process started as usual with:
```
INFO  rav1e > Using y4m decoder: 550x872p @ 30000/1001 fps, 4:2:0, 8-bit
INFO  rav1e > Encoding settings: keyint_min=12 keyint_max=240 quantizer=150 bitrate=0 min_quantizer=0 low_latency=false tune=Psychovisual rdo_lookahead_frames=40 min_block_size=4x4 multiref=true fast_deblock=false reduced_tx_set=false tx_domain_distortion=false tx_domain_rate=false encode_bottomup=true rdo_tx_decision=true prediction_modes=Complex-All include_near_mvs=true no_scene_detection=false diamond_me=true cdef=true quantizer_rdo=true use_satd_subpel=true non_square_partition=false
INFO  rav1e::api::config > CPU Feature Level: SSSE3
INFO  rav1e::api::internal > Using 4 tiles (1x4)
```
and stopped here without saying why. Output was not readable by media player and had only 32KB. Is there some debug option I could turn on to get what's happening? The parameters I used are:
`-v -s 1 --threads 4 --tile-rows 4 --quantizer 150`
I tried changing -s and tiles.

Also worth mention is the fact I have the same video but in 1920x1080 and it works just fine. The one here is simply resized and cropped. Maybe it is somehow related to the resolution? Like SvtAv1Enc that works only with dimensions that can be divided by 8?

shssoichiro:
This has been fixed by one of the dav1d ASM merges.

Closes #1715

### DIFF
--- a/src/asm/x86/dist.rs
+++ b/src/asm/x86/dist.rs
@@ -53,7 +53,7 @@ declare_asm_dist_fn![
   (rav1e_sad8x8_sse2, u8),
   (rav1e_sad8x16_sse2, u8),
   (rav1e_sad8x32_sse2, u8),
-  //  (rav1e_sad16x16_sse2, u8),
+  (rav1e_sad16x16_sse2, u8),
   (rav1e_sad32x32_sse2, u8),
   (rav1e_sad64x64_sse2, u8),
   (rav1e_sad128x128_sse2, u8),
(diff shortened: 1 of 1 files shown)


shssoichiro:
I had failed to test this at the resolutions that were triggering the crash. Unfortunately this is still broken.

PR #1812, opened 2019-10-31
Allow "native" for setting the CPU feature level

shssoichiro:
It's confusing that the enum variant is called "NATIVE",
but the string to set it is called "rust".
This commit will match "rust" or "native" to disable ASM.


shssoichiro:
Going to take a different, cleaner approach using the arg_enum macro and then reopen this.

PR #1813, opened 2019-10-31
Integrate SATD ASM for SSSE3 and SSE4.1

shssoichiro:
This involves adding a new CpuFeatureLevel for SSE4.1.
Because of being unable to use a decimal in an enum variant name,
this required manually implementing FromStr and Display
instead of using the arg enum proc macro.

### DIFF
--- a/src/cpu_features/x86.rs
+++ b/src/cpu_features/x86.rs
@@ -7,15 +7,50 @@
 // Media Patent License 1.0 was not distributed with this source code in the
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
-use arg_enum_proc_macro::ArgEnum;
 use std::env;
+use std::fmt;
+use std::str::FromStr;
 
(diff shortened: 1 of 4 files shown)


lu-zero:
You may use the alias feature in the proc macro to map sse4.1 to SSE4_1

lu-zero:
[here the documentation about it](https://github.com/lu-zero/arg_enum_proc_macro#aliases)

src/cpu_features/x86.rs
```diff
@@ -9,12 +9,15 @@
 
 use arg_enum_proc_macro::ArgEnum;
 use std::env;
+use std::str::FromStr;
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, ArgEnum)]
 pub enum CpuFeatureLevel {
   NATIVE,
```
barrbrain: To keep the existing behaviour we also need an alias for `NATIVE`:
```rust
#[arg_enum(alias = "rust")]
NATIVE
```

PR #1817, opened 2019-11-01
Simplify aarch64 CpuFeatureLevel with FromStr

barrbrain:
This matches recent changes to the `x86_64` implementation.


lu-zero:
Looks good to me

PR #1826, opened 2019-11-04
Make rav1e_status_to_str return a static string

lu-zero:
As discussed on IRC.

### DIFF
--- a/src/capi.rs
+++ b/src/capi.rs
@@ -90,6 +90,21 @@ pub enum EncoderStatus {
   NotReady = -2,
 }
 
+impl EncoderStatus {
+  fn to_c(&self) -> *const u8 {
+    use self::EncoderStatus::*;
+    match self {
+      Success => "Normal operation\0".as_ptr(),
+      NeedMoreData => "The encoder needs more data to produce an output packet\0".as_ptr(),
+      EnoughData => "There are enough frames in the queue\0".as_ptr(),
+      LimitReached => "The encoder has already produced the number of frames requested\0".as_ptr(),
+      Encoded => "A Frame had been encoded but not emitted yet\0".as_ptr(),
+      Failure => "Generic fatal error\0".as_ptr(),
+      NotReady => "First-pass stats data not retrieved or not enough second-pass data provided\0".as_ptr(),
+    }
+  }
+}
+
 impl From<Option<rav1e::EncoderStatus>> for EncoderStatus {
   fn from(status: Option<rav1e::EncoderStatus>) -> Self {
     match status {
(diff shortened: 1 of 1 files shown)


src/capi.rs
```diff
@@ -90,6 +90,21 @@ pub enum EncoderStatus {
...
+  fn to_c(&self) -> *const u8 {
+    use self::EncoderStatus::*;
+    match self {
+      Success => "Normal operation\0".as_ptr(),
+      NeedMoreData => "The encoder needs more data to produce an output packet\0".as_ptr(),
+      EnoughData => "There are enough frames in the queue\0".as_ptr(),
+      LimitReached => "The encoder has already produced the number of frames requested\0".as_ptr(),
+      Encoded => "A Frame had been encoded but not emitted yet\0".as_ptr(),
+      Failure => "Generic fatal error\0".as_ptr(),
+      NotReady => "First-pass stats data not retrieved or not enough second-pass data provided\0".as_ptr(),
```
dwbuiten: Is there a way to add a test for these to make sure the Tust and C APIs do not diverge?

lu-zero: Currently it would not compile.

src/capi.rs
```diff
@@ -90,6 +90,21 @@ pub enum EncoderStatus {
   NotReady = -2,
 }
 
+impl EncoderStatus {
+  fn to_c(&self) -> *const u8 {
+    use self::EncoderStatus::*;
+    match self {
+      Success => "Normal operation\0".as_ptr(),
+      NeedMoreData => "The encoder needs more data to produce an output packet\0".as_ptr(),
+      EnoughData => "There are enough frames in the queue\0".as_ptr(),
```
jamrial: Seeing this is only returned when you send new data after signaling EOF, the message could be something more descriptive of the actual reason for the error code, like "No more data can be consumed after EOF" or similar instead.

lu-zero: It depends if we decide to cap the input queue or not, the message is future-proof but probably not as descriptive as the current state.

jamrial: Ok, in that case the string is fine. But if capping is a possibility in the future, then it would be better to state what the error will mean if its emitted before EOF is signaled in the doxy, since that's what an API user will look at to write their code. It's definitely not future-proof as is.

Right now, most API user would probably not bother to ever check for this error if they know they will never send new data after EOF, as that's the only time the doxy states this error is expected. But if you cap the input in a future release and reuse this error code for it, then their code will not be ready to handle it when it's emitted before EOF.

Making the doxy cover both cases will prevent the eventual cap to become an API change.

lu-zero: Agreed, I would do on another patch/PR

lu-zero:
I it ok to land it and then move on clarifying the API in another PR?

jamrial:
> 
> 
> I it ok to land it and then move on clarifying the API in another PR?

Yes, of course.

PR #1828, opened 2019-11-04
[WIP] disable cdef and lrf for parallelism testing 

tdaede:


lu-zero:
I'd land the speed feature, it is useful to craft specific bitstream in itself.

PR #1829, opened 2019-11-05
Replace rayon with WIP latch-target-thread branch.

tdaede:


jamrial:
Wouldn't it be safer to include a known good snapshot of this branch inside the crates folder, and use that instead of a remote private branch that may change in unpredictable ways?

Also, https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html mentions that [patch] should be used instead of [replace].

tdaede:
Yes, sorry I forgot to mark this WIP.

lu-zero:
With this patch on top of 7344daee25bc94620d7ac6577d1905089fab27f6, Bosphorus 4k 8bit.
```
 INFO  rav1e::api::internal > Using 16 tiles (4x4)
 INFO  rav1e::stats         > encoded 600 frames, 0.292 fps, 5481.45 Kb/s
```

PR #1832, opened 2019-11-05
hawktrace support

lu-zero:
it is pending a 0.5 release of rust-hawktracer, the code is complete.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,6 +63,16 @@ cfg_if::cfg_if! {
   }
 }
 
+mod hawktracer {
+  cfg_if::cfg_if! {
+    if #[cfg(feature="tracing")] {
+      pub use rust_hawktracer::*;
+    } else {
+      pub use noop_proc_macro::hawktracer;
+    }
+  }
+}
+
 #[cfg(any(cargo_c, feature = "capi"))]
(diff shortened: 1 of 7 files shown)


lu-zero:
It depends on https://github.com/AlexEne/rust_hawktracer/issues/19

lu-zero:
The previous refactoring broke it, now it works as intended on macOS and Linux.

AlexEne:
I've released a new (0.5.0) rust_hawktracer version.

src/lib.rs
```diff
@@ -63,6 +63,16 @@ cfg_if::cfg_if! {
   }
 }
 
+mod hawktracer {
+  cfg_if::cfg_if! {
+    if #[cfg(feature="tracing")] {
+      pub use rust_hawktracer::*;
+    } else {
+      pub use noop_proc_macro::hawktracer;
```
AlexEne: The macros don't touch the method if profiling is off so this might not be needed.
See: https://github.com/AlexEne/rust_hawktracer_proc_macro/blob/589a7070f9415497b946f6dee3d8d19c9b1fe0a3/src/lib.rs#L33

lu-zero: Thank you :)

PR #1833, opened 2019-11-05
Implement FusedIterator for all applicable iterators

KyleSiefring:
See https://doc.rust-lang.org/std/iter/trait.FusedIterator.html

"Calling next on a fused iterator that has returned None once is
guaranteed to return None again. This trait should be implemented by
all iterators that behave this way because it allows optimizing
Iterator::fuse."

### DIFF
--- a/src/frame/plane.rs
+++ b/src/frame/plane.rs
@@ -488,6 +488,8 @@ impl<'a, T: Pixel> Iterator for PlaneIter<'a, T> {
   }
 }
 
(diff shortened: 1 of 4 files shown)


rzumer:
Sounds good either way, but is there any measured performance improvement?

KyleSiefring:
@rzumer no measurable performance uplift.

lu-zero:
Can you please rebase and land it?

Issue #1720, opened 2019-10-04
Backport dav1d update_cdf() assembly


tdaede:
dav1d has a 2x as fast update_cdf() implementation, we should use it.

dodomorandi:
I noticed that the dav1d `update_cdf` function uses the following unstable intrinsics:
* [pavgw](https://doc.rust-lang.org/core/arch/x86_64/fn._m_pavgw.html)
* [psubw](https://doc.rust-lang.org/core/arch/x86_64/fn._m_psubw.html)
* [paddw](https://doc.rust-lang.org/core/arch/x86_64/fn._m_paddw.html)

I am also not sure about `psraw`, a _packed shift raw arithmetic_ from MMX set, which I am not able to find in the core docs. A possible alternative could be `_mm_sra_pi16`, but it does not exists as well. Maybe it could be possible to use the 128 bit version?

Last consideration: maybe the three functions could be replaced by [_mm_avg_pu16](https://doc.rust-lang.org/core/arch/x86_64/fn._mm_avg_pu16.html), [_mm_sub_pi16](https://doc.rust-lang.org/core/arch/x86_64/fn._mm_sub_pi16.html) and [_mm_add_pi16](https://doc.rust-lang.org/core/arch/x86_64/fn._mm_add_pi16.html), but I am a bit noob with SIMD intrinsics and I don't know if there are subtle differences between the two series of versions.

shssoichiro:
@dodomorandi What we have generally been preferring with dav1d's ASM is copying it as-is into rav1e (see the src/x86 directory) then compiling it with nasm. I don't know if you are more familiar with raw ASM syntax than with Rust's intrinsics.

tdaede:
In this particular case I wrote it as intrinsics first, then converted it to asm for dav1d. We could either switch this to calling the nasm version, or backport the changes back into intrinsics. In this particular case, the reduced calling overhead of intrinsics might make using them more feasable.

Unfortunately *none* of the m64 intrinsics are stable in Rust, so you'd need to actually convert those functions to the 128 bit version. That's totally OK to do. On modern Intel CPUs that's totally OK to do as the 128 bit versions are the same speed. You'll just want to test it to make sure it's functionally identical (the unit tests should do this already).

PR #1834, opened 2019-11-06
Rewrite of SSE2 `update_cdf`

dodomorandi:
A WIP rewrite of the SSE2 version of `update_cdf`, issue https://github.com/xiph/rav1e/issues/1720.

Current resulting assembly is pretty similar to the one written by hand in `dav1d`. I would like to see if I can do slightly better, and I would like to write a small documentation of what is happening.

### DIFF
--- a/src/asm/x86/ec.rs
+++ b/src/asm/x86/ec.rs
@@ -24,29 +24,66 @@ pub fn update_cdf(cdf: &mut [u16], val: u32) {
 #[target_feature(enable = "sse2")]
 unsafe fn update_cdf_4_sse2(cdf: &mut [u16], val: u32) {
   let nsymbs = 4;
-  let rate = 3 + (nsymbs >> 1).min(2) + (cdf[nsymbs] >> 4) as usize;
-  cdf[nsymbs] += 1 - (cdf[nsymbs] >> 5);
+  let rate = 5 + (cdf[nsymbs] >> 4) as usize;
+  cdf[nsymbs] += (cdf[nsymbs] < 32) as u16;
 
-  let val_idx = _mm_set_epi16(0, 0, 0, 0, 3, 2, 1, 0);
+  // A bit of explanation of what is happening down here. First of all, let's look at the simple
+  // implementation:
+  //
+  // ```
+  // if i as u32 >= val {
+  //   *v -= *v >> rate;
+  // } else {
+  //   *v += (32768 - *v) >> rate;
+  // }
+  // ```
+  //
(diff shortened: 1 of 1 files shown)


dodomorandi:
I think I have an idea about the problem. Unfortunately the `update_cdf` test is passing flawlessly, I probably need to write a specific regression test for this.

dodomorandi:
This is not the best solution: if the content of the auxiliary `k` variable is `-1` when `index >= val` instead of 0, it should be possible to replicate the behavior of `dav1d` and remove a SIMD instruction. However, I am not getting how to do that in a cheap way.

tdaede:
I'll take a look to see if that last instructions can be dropped, but the rest looks ok.

lu-zero:
Could you please rebase this set?

tdaede:
I'll also take a look in the next day to see if there's a way to drop the last insn.

dodomorandi:
I tried to find a cheaper solution, but without success. I am removing the `WIP` tag, but @tdaede if you find I way to optimize the code, suggestions are welcome.

tdaede:
Yeah, I think this is fine. If I come up with something better in the meantime I'll open that as a PR.

tdaede:
You'll need to squash the commits though. (or I can do it for you)

dodomorandi:
:+1: I was not sure if you would rather do a squash-merge.

src/asm/x86/ec.rs
```diff
@@ -24,29 +24,66 @@ pub fn update_cdf(cdf: &mut [u16], val: u32) {
 #[target_feature(enable = "sse2")]
 unsafe fn update_cdf_4_sse2(cdf: &mut [u16], val: u32) {
   let nsymbs = 4;
-  let rate = 3 + (nsymbs >> 1).min(2) + (cdf[nsymbs] >> 4) as usize;
+  let rate = 5 + (cdf[nsymbs] >> 4) as usize;
   cdf[nsymbs] += 1 - (cdf[nsymbs] >> 5);
```
barrbrain: To further align with dav1d assembly:
```suggestion
  cdf[nsymbs] += (cdf[nsymbs] < 32) as u16;
```

dodomorandi: Good catch, I overlooked that part. And honestly I overestimated the capabilities of the optimizer -- your solution produces a cleaner asm 👍

lu-zero:
Looks fine to me, I'd change the commit title to "Improve the sse2 update_cdf" though

lu-zero:
Let's land it then

PR #1835, opened 2019-11-09
Address the new lints introduced with the current clippy

lu-zero:
Make clippy happier

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,7 +107,7 @@ jobs:
         - cargo doc --verbose --no-deps
         - rustup component add clippy
         - cargo clippy --version
-        - cargo clippy -- -D warnings --verbose
+        - cargo clippy -- -D warnings --verbose -A clippy::wrong-self-convention
       env:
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -107,7 +107,7 @@ jobs:
         - cargo doc --verbose --no-deps
         - rustup component add clippy
         - cargo clippy --version
-        - cargo clippy -- -D warnings --verbose
+        - cargo clippy -- -D warnings --verbose -A clippy::wrong-self-convention
```
shssoichiro: This should be added to clippy.toml instead.

lu-zero: I'm not sure how to do it.

shssoichiro: Oh, that would be because I told you the wrong thing. Clippy.toml is the wrong place, it should actually be inline in lib.rs [like so](https://github.com/xiph/rav1e/blob/master/src/lib.rs#L39).

lu-zero: Ok, addressed, thank you :)

PR #1838, opened 2019-11-09
Make speed test start with all features enabled

shssoichiro:
As discussed a few times, because of how encoding tools interact,
we will get more accurate speed test results if we default to
having all encoding tools enabled, and disable one at a time
to see if that tool was actually useful or if another tool
compensates for it.


lu-zero:
I can bake a tiny proc-macro instead of using serde to avoid this kind of churn though.

lu-zero:
Please rebase on top of the current master.

PR #1839, opened 2019-11-09
Fix the doxy for EnoughData encoder status

jamrial:
Make it future proof in case a frame queue limit is introduced.


lu-zero:
Please rebase on top of the current master so the CI won't fail for unrelated code.

jamrial:
Done.

lu-zero:
Thank you!

PR #1841, opened 2019-11-10
Support setting timing info and overriding time base in CLI

### LINKED ISSUE #1840: Support writing Timing Info
Frame rate could be set using an encoder option for applications using the rav1e API. The CLI could also take it directly from the y4m input when available.

As far as i could see, it should be a matter or setting timing_info_present_flag to 1, time_base and num_units_in_display_tick in the timing_info section of sequence header to the framerate numerator and denominator respectively, and for simplicity equal_picture_interval to 1 and num_ticks_per_picture_minus_1 to 0.
Time stamps are not written in the bitstream with the above combination.

This will be useful for Annex B or raw Section 5 OBUs output, as some people have (unfortunately) started using and dumping samples in the wild.

rzumer:
Closes #1840, assuming I understood the request correctly. I have another branch with timing info support, but it does not seem to change anything. Time base is used to set the frame rate in the bitstream, so this allows overriding it from the CLI via `--frame-rate` and `--time-base` (default: 1) options.

There is already a function allowing setting the time base from the C API, so I'm not sure it is worth adding the equivalent string configuration options.


src/api/config.rs
```diff
@@ -28,7 +28,7 @@ const MAX_RDO_LOOKAHEAD_FRAMES: usize = usize::max_value() - 1;
 const MAX_MAX_KEY_FRAME_INTERVAL: u64 = i32::max_value() as u64 / 3;
 
 /// Encoder settings which impact the produced bitstream.
-#[derive(Clone, Debug)]
+#[derive(Clone, Copy, Debug)]
```
lu-zero: Why do you need to have it Copy? It is fairly large.

rzumer: Because fields get moved when writing the time scale from the config instead of the video details.

rzumer:
Actually added C API parameters as well to match, it is more user-friendly I think.

src/api/config.rs
```diff
@@ -58,6 +58,9 @@ pub struct EncoderConfig {
   // encoder configuration
   /// Video time base.
   pub time_base: Rational,
+  /// Flag specifying that the time base set in this configuration
+  /// should override any automatic time base detection.
+  pub time_base_override: bool,
```
lu-zero: This seems a binary-only field, it should live in `rav1e.rs`.

rzumer: Not anymore with the last commit.

lu-zero: It is still checked in the cli-only. Probably it could be done in a different way

rzumer: Ah okay

rzumer: Moved to `CliOptions`.

rzumer:
Note: currently timing info is not toggleable from the user's side, so hold off on merging until the interfaces are updated...

jamrial:
https://0x0.st/zE_q.y4m
If i force-toggle writing the timing info into the bistream and run an encode, the output of

`rav1e zE_q.y4m -o rav1e.ivf`

vs the output of

`rav1e zE_q.y4m --frame-rate 24 -o rav1e.ivf`

Is different. The numerator and denominator are swapped in the sequence header. The first command line appears to have the correct behavior.

jamrial:
Oh, it's also swapped in the ivf header.

src/bin/common.rs
```diff
@@ -541,6 +559,13 @@ fn parse_config(matches: &ArgMatches<'_>) -> Result<EncoderConfig, CliError> {
     panic!("Tile columns and rows may not be greater than 64");
   }
 
+  if let Some(frame_rate) = matches.value_of("FRAME_RATE") {
+    cfg.time_base = Rational::new(
+      frame_rate.parse().unwrap(),
+      matches.value_of("TIME_SCALE").unwrap().parse().unwrap(),
```
jamrial: These two lines should be swapped. Both the ivf timebase and the sequence header timing info fields will have invalid values otherwise.

rzumer:
> Oh, it's also swapped in the ivf header.

Yes, that's the issue, sorry for missing it.

src/bin/common.rs
rzumer: Done

jamrial:
It works now. Both the ivf header and the sequence header fields (if timing_info_present is toggled to true) are written correctly. Thanks.

That aside, if one wants to force a frame rate of 23.97fps, you need to pass --frame-rate 24000 --time-scale 1001, and it's not exactly intuitive. Could this instead be a single "frame-rate" option, which can accept values like 30, 30/1, 24000/1001, etc? It's less confusing, and in line with aomenc's --fps option.

rzumer:
I considered that initially but thought separating the options would actually be less confusing since there is no ambiguity on the format. I could also allow both formats, but then there is the issue of deciding which takes precedence over the other if both are set.

jamrial:
Alright, no problem then. Thanks for implementing this.

Next would be making writing or not the sequence header bits an user toggle-able option.

rzumer:
Indeed, I was just working on that. Please let me know if it works as expected.

jamrial:
> 
> 
> Indeed, I was just working on that. Please let me know if it works as expected.

At least when using the C API, if i do
`rav1e_config_parse_int(cfg, "enable_timing_info", 1);` it returns an error, whereas
`rav1e_config_parse(cfg, "enable_timing_info", "true");` works.

rzumer:
That's the same behavior as other settings like `still_picture` and `low_latency`, so seems to work as intended.

jamrial:
Ah, so boolean options wont work with rav1e_config_parse_int().

Looks good then, thanks again.

PR #1846, opened 2019-11-11
Convert Vecs in intra pruning to ArrayVecs.

tdaede:


shssoichiro:
I'm not sure if this is actually useful. The benefit of ArrayVec is that we never need to reallocate because it has a fixed size backing. In this case, we never push to the set of items after the `collect` operation, so there shouldn't be a reallocation anyway. Probably it doesn't matter which we use here.

tdaede:
It should still be useful as this is a stack allocation, not a heap allocation, and therefore is faster.

PR #1848, opened 2019-11-13
Speed up scenechange

lu-zero:
15-20x faster

On qcif:

before: 
```
encoded 600 frames, 39.575 fps, 52.29 Kb/s
```
after:
```
encoded 600 frames, 44.320 fps, 52.29 Kb/s
```

Because the code isn't implicitly upscaling chroma the results for high resolution might diverge a little, discussion regarding downscaling luma and/or decide the weight between luma and chroma are open.


shssoichiro:
Ah, you'll need to fix rustfmt however.

lu-zero:
Formatted, I'd land it today.

shssoichiro:
Good news! With the chroma scaling, [no change in AWCY](https://beta.arewecompressedyet.com/?job=master-scenecut%402019-11-14T03%3A15%3A23.198Z&job=sceneut-speedup-chroma-dec%402019-11-14T05%3A56%3A41.206Z).

Bad news, rustfmt and clippy are complaining again.

lu-zero:
I'll fix and merge once I'm back, the weighted version is still worse?

shssoichiro:
The earlier version was still worse, yes. Although I have a suspicion that things like scrolling credits interact in a strange way with this algorithm--they look a lot like constant scene flashes, especially if luma is prioritized more heavily.

If we want to work around this, it probably requires revisiting using cost estimates to determine where scene changes should happen, but I'm not sure if it's worth investing the time in that at the moment, given that it's inherently much slower than this approach. (We might be able to do it much more cheaply and accurately though if we can dedup some of the code that's shared between lookahead and encoding.)

PR #1850, opened 2019-11-14
Avoid marking the last frame in a clip as a scenecut

shssoichiro:
In real encoding, this will matter about 1/240 of the time. However, in our AWCY tests, this has slightly more of an impact. [AWCY vs master](https://beta.arewecompressedyet.com/?job=master-scenecut%402019-11-14T03%3A15%3A23.198Z&job=no-last-scenecut%402019-11-14T05%3A53%3A49.531Z); this will also help resolve some more noise in tests for tweaking of scenecuts in the future.

### DIFF
--- a/src/api/test.rs
+++ b/src/api/test.rs
@@ -633,15 +633,15 @@ fn output_frameno_reorder_scene_change_at(scene_change_at: u64) {
         ][..]
       }
       4 => {
+        // The algorithm cannot put a scenechange at the last frame of a video
         &[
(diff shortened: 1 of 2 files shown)


shssoichiro:
Yeah, this is probably just addressing a symptom. The root problem is certain types of material show as a long series of scene flashes, and then the last frame gets marked as the scenecut. The solution to this has to be choosing scenecuts using cost metrics.

PR #1853, opened 2019-11-18
Fix docs.rs generation

lu-zero:
Currently our documentation is not published correctly, this fixes it.


lu-zero:
It is not, probably, I guess I can undo that hunk and see what happens.

lu-zero:
Works as intended w/out this.

PR #1854, opened 2019-11-18
Add generate_bindings to the hawktracer features

lu-zero:
Should address the problem found on aarch64.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,7 @@ simd_helpers = "0.1"
 
 [dependencies.rust_hawktracer]
 version = "0.5.0"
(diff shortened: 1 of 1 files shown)


vaishnavsivadas:
Hey,

Updating hawktracer_sys actually solves the issue for aarch64 ubuntu devices as proposed by Alexandru Ene here: [
Build failing for rav1e enabling hawktracer on ubuntu aarch64 device](https://github.com/AlexEne/rust_hawktracer/issues/22) 

I tried building without adding `generate_bindings` in features and it builds fine too.

lu-zero:
Great so we can avoid this :)

PR #1862, opened 2019-11-20
Bump the avformat-sys version

lu-zero:
Alternatively I'd rather remove the feature.


lu-zero:
#1863 is a more radical solution.

PR #1863, opened 2019-11-20
Avformat removal

lu-zero:


src/ec.rs
```diff
@@ -326,7 +326,6 @@ impl<S> WriterBase<S> {
         debug: std::env::var_os("RAV1E_DEBUG").is_some(),
         fake_bits_frac: 0,
         s: storage,
-        cpu,
```
dwbuiten: ?

dwbuiten:
I never though we needed one in the first place ... so.

src/ec.rs
lu-zero: another problem fixed while testing the patch.

PR #1864, opened 2019-11-20
More scenechange refactor

lu-zero:


src/scenechange/mod.rs
```diff
@@ -137,7 +137,6 @@ impl SceneChangeDetector {
     // Where A and B are scenes: AAAAAABBBAAAAAA
     // If BBB is shorter than lookahead_distance, it is detected as a flash
     // and not considered a scenecut.
-    if lookahead_distance > 1 {
```
shssoichiro: I don't think we can remove this, unless you've verified it won't cause issues at the very end of the video (which AFAIK is the only time `lookahead_distance` will evaluate to 2).

PR #1866, opened 2019-11-22
Add is_filter_switchable false case to frame obu header

ycho:
Add missing SWITCHABLE to FilterMode enum type.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -632,7 +632,9 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
       }
 
       self.write_bit(fi.is_filter_switchable)?;
-      self.write(2, fi.default_filter as u8)?;
+      if !fi.is_filter_switchable {
+        self.write(2, fi.default_filter as u8)?;
+      }
       self.write_bit(fi.is_motion_mode_switchable)?;
 
       if (!fi.error_resilient && fi.sequence.enable_ref_frame_mvs) {
--- a/src/mc.rs
+++ b/src/mc.rs
@@ -64,6 +64,7 @@ pub enum FilterMode {
   SMOOTH = 1,
   SHARP = 2,
   BILINEAR = 3,
+  SWITCHABLE = 4,
 }
 
 pub const SUBPEL_FILTER_SIZE: usize = 8;


rzumer:
Typo on SWITCHABLE in the commit message, recommend amending.

ycho:
> Typo on SWITCHABLE in the commit message, recommend amending.

Ouch! Thanks-

src/mc.rs
```diff
@@ -64,6 +64,7 @@ pub enum FilterMode {
   SMOOTH = 1,
   SHARP = 2,
   BILINEAR = 3,
+  SWITCHABLE = 4,
```
barrbrain: This matches AV1 spec section [6.8.9 Interpolation filter semantics](https://aomediacodec.github.io/av1-spec/#interpolation-filter-semantics).

src/header.rs
```diff
@@ -632,7 +633,11 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
       }
 
       self.write_bit(fi.is_filter_switchable)?;
-      self.write(2, fi.default_filter as u8)?;
+      if fi.is_filter_switchable {
+        self.write(2, FilterMode::SWITCHABLE as u8)?;
```
barrbrain: This differs from AV1 spec section [5.9.10 Interpolation filter syntax](https://aomediacodec.github.io/av1-spec/#interpolation-filter-syntax) where `interpolation_filter` is implied rather than signaled when `is_filter_switchable == 1`.

ycho: Thanks for catching this. Yes. I did wrong. Don't need "self.write(2, FilterMode::SWITCHABLE as u8)?;", once "self.write_bit(fi.is_filter_switchable)?;" tells that the filter is switchable. Instead, each inter block encode which filter to use. The " interp_filter[ dir ]' in https://aomediacodec.github.io/av1-spec/#inter-block-mode-info-syntax

In fact, this commit is WIP and yet no chance to test itself, part of the lager work that actually is able to use best subpel interpolation filter based on RDO.

ycho:
@barrbrain, do you think now this patch is good to land? (Btw, I would not include other patches with this PR, so deleted the 2nd commit which I plan to submit as separate PR later)

tdaede:
Does the SWITCHABLE enum value ever get used?

ycho:
> Does the SWITCHABLE enum value ever get used?

It was meant to be used in my following patches. Yes, to check whether each inter block mode uses switchable filter.

tdaede:
Can't you just check fi.is_filter_switchable?

ycho:
> Can't you just check fi.is_filter_switchable?

I think we can do that way as well. When I added that type, I first followed the spec in fact.

ycho:
> > Can't you just check fi.is_filter_switchable?
> 
> I think we can do that way as well. When I added that type, I first followed the spec in fact.

Also, using fi.is_filter_switchable will make our work much easier and code neat. Since from my trial, It was quite tedious to add additional parameter 'interpolation_filter' from the top to function call path.

PR #1877, opened 2019-11-26
nasm-rs: Improve the error reporting

lu-zero:


lu-zero:
It can wait a full overhaul later

PR #1871, opened 2019-11-24
Reuse lookahead frame state and motion vectors

shssoichiro:
Results in 5-8% encoding time improvement with 1 tile. [AWCY](https://beta.arewecompressedyet.com/?job=master-edbc586999d6a1e49e1e35d84a0c8a2c9dcf7990&job=reuse-lookahead-mvs%402019-11-24T06%3A32%3A54.455Z)

Closes #1677

PR #1879, opened 2019-11-27
Avoid a very expensive clone in compute_block_importances

shssoichiro:
This clone, which was introduced in #1871,
was responsible for 40% of allocations
within rav1e. Although counterintuitive,
removing the frame data from the map
and reinserting it at the end of the iteration
is better because it avoids this clone.

Gives about a 4% speedup over master
with 1 tile on 720p test clip.

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -879,19 +879,23 @@ impl<T: Pixel> ContextInner<T> {
     );
 
     for &output_frameno in output_framenos.iter().skip(1).rev() {
-      // Copy fi from the map to avoid the borrow checker complaining
-      // when we mutably borrow another fi.
-      let fi = self
-        .frame_data
-        .get(&output_frameno)
-        .map(|data| data.fi.clone())
-        .unwrap();
-
       // TODO: see comment above about key frames not having references.
-      if fi.frame_type == FrameType::KEY {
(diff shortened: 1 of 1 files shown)


src/api/internal.rs
```diff
@@ -879,20 +879,24 @@ impl<T: Pixel> ContextInner<T> {
...
 
+      // Remove fi from the map temporarily and put it back in in the end of
+      // the iteration. This is required because we need to mutably borrow
+      // referenced fis from the map, and that wouldn't be possible if this was
+      // an active borrow.
+      //
+      // Performance note: Contrary to intuition,
+      // removing the data and re-inserting it at the end
+      // is more performant because it avoids a very expensive clone.
+      let output_frame_data = self.frame_data.remove(&output_frameno).unwrap();
```
lu-zero: Move this block at the start of the function please.

shssoichiro: derf thought it would be better to only perform the remove if it's not a keyframe. But I have no strong opinion, the performance impact of that portion should be minimal.

shssoichiro: Or wait, did you mean outside of the loop? That wouldn't work, because the output_frameno here is the loop index.

lu-zero: I see, Tim is right as I said anything :)

PR #1883, opened 2019-11-28
Automatically deploy rav1e

Luni-4:
`cargo` and `cargo-c` binaries are built on `Windows` and  then deployed on `Github` automatically through `Linux`. To create a new release, it is sufficient to add to the `CHANGELOG.md` file the necessary infos and push a new tag.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,103 @@
+name: deploy
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+  release:
+    types:
+      - prereleased
+
+jobs:
+  create-binaries:
+
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install nasm
+      run: |
+        $NASM_VERSION="nasm-2.14.02"
+        curl -LO "https://people.xiph.org/~tdaede/$NASM_VERSION-win64.zip"
+        7z e -y "$NASM_VERSION-win64.zip" -o"C:\nasm"
+    - name: Install Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable-x86_64-pc-windows-gnu
+        override: true
+    - name: Install cargo-c
+      run: |
+        cargo install cargo-c
+    - name: Set environment variables
+      run: |
+        echo "::add-path::C:\nasm"
+    - name: Build rav1e
+      run: |
+        cargo build --release
+    - name: Run cargo-c
+      run: |
+        cargo cinstall --release --destdir="C:\"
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-windows.yml
```diff
@@ -0,0 +1,83 @@
+name: deploy-windows
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
```
barrbrain: @lu-zero Are we planning to use the "v" prefix for future releases?

lu-zero: We should decide, probably it is a good idea.

lu-zero:
That reminds me we should keep a changelog.md up to date.

Luni-4:
The binaries can now be added to a pre-release automatically. Unfortunately, the entire building process took 20 minutes and I don't know if it's acceptable to have a pre-release without binaries for that amount of time

.github/workflows/deploy-windows.yml
lu-zero: I started using `p{date}` and `v{version}`.

.github/workflows/deploy-windows.yml
```diff
@@ -0,0 +1,110 @@
...
```
lu-zero: for prereleases I use `p{date}` as tag.

.github/workflows/deploy-windows.yml
```diff
@@ -0,0 +1,110 @@
...
+      with:
+        name: rav1e-binaries
+    - name: Unzip rav1e Windows binaries
+      run: |
+        unzip rav1e-binaries/rav1e.zip -d rav1e-binaries
+    - name: Handle release data and binaries
+      if: github.event_name == 'push'
+      id: data
+      run: |
+        VERSION=$(head -n 1 CHANGELOG.md | tr -d "## Version ")
```
lu-zero: Probably could be a good idea to take the version from the `git tag` or the `Cargo.toml`.

.github/workflows/deploy-windows.yml
Luni-4: The binaries will be added after the prerelease is created, so you can set the `p{date}' yourself. If we want to create a completely automatic prerelease, we need to add a `Prerelease.md` file containing all the changes.

Luni-4:
> Could you please remove the changelog from this set? I think it is good to land.

So do you want to add the release/prerelease's notes manually after the release/prerelease is created, right?

lu-zero:
At least for now yes, I prefer to keep the changes small here :)

Luni-4:
> At least for now yes, I prefer to keep the changes small here :)

Done :)

lu-zero:
Thank you a lot!

PR #1887, opened 2019-11-29
[WIP] Drop aom decode tests from grcov task

barrbrain:
Testing the significance of change in running time.


barrbrain:
This reduced the running time by about 4 minutes. However, the second-longest task is about 1 minute slower than grcov without this change. Closing because little benefit to wall time.

PR #1888, opened 2019-11-29
Add sccache locally

Luni-4:
Run `sscache` locally and then cache its output for the next runs. The cache size is 400M for this [reason](https://github.com/actions/cache#cache-limits).


Luni-4:
Any clue about the `unix` error @barrbrain?

Luni-4:
Thanks for the review! We can close this one in favor of #1892

PR #1892, opened 2019-11-30
Add sccache locally (with fixes)

barrbrain:
Original patch from @Luni-4 moved back to local sccache store, populated by an actions/cache rule. There was a small, blocking issue with the syntax for setting a path relative to `HOME`. As well as resolving that, @barrbrain optimised the cache size and partitioning rules with consideration for the statistics in earlier runs.

### DIFF
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -77,11 +77,29 @@ jobs:
     env:
       RUST_BACKTRACE: 1
       RUSTC_WRAPPER: sccache
+      SCCACHE_CACHE_SIZE: 300M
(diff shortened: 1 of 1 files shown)


barrbrain:
This is a continuation of #1888 with small fixes for implementation details of GitHub Actions.

.github/workflows/rav1e.yml
```diff
@@ -233,13 +238,28 @@ jobs:
 
     env:
       RUST_BACKTRACE: 1
+      RUSTC_WRAPPER: sccache
+      SCCACHE_CACHE_SIZE: 400M
```
Luni-4: I propose to change the cache size here too in order to maintain the coherence with the `unix` version

PR #1893, opened 2019-11-30
Relax the y4m allocation limit

lu-zero:


src/bin/rav1e.rs
```diff
@@ -265,8 +265,10 @@ fn init_logger() {
 
 fn run() -> Result<(), error::CliError> {
   let mut cli = parse_cli()?;
-  let mut y4m_dec =
-    y4m::decode(&mut cli.io.input).expect("input is not a y4m file");
+  // Maximum frame size by specification + maximum y4m header
+  let limit = y4m::Limits { bytes: 64 * 64 * 4096 * 2304 + 1024 };
```
ycho: So, dos this new size allow 8K image frame? Can you explain what is 64 * 64 for?

ycho: Hmm...  is it maybe 64 times of "1:19 PM <Chagall> 64 * 1024 * 1024 is the default memory limit apparently" ?

ycho: @barrbrain , Read further rational how the new limit is determined. 
Friday, Nov 29, "11:49 PM <barrbrain> max frame area is 38,654,705,664 = 64*64*4096*2304"

PR #1908, opened 2019-12-03
Remove allocations in get_sub_partitions_*.

tdaede:


src/rdo.rs
```diff
@@ -1392,33 +1392,39 @@ pub fn rdo_tx_type_decision<T: Pixel>(
 
 pub fn get_sub_partitions(
   four_partitions: &[TileBlockOffset; 4], partition: PartitionType,
-) -> Vec<TileBlockOffset> {
-  let mut partitions = vec![four_partitions[0]];
+) -> ArrayVec<[TileBlockOffset; EXT_PARTITION_TYPES]> {
+  let mut partition_offsets =
+    ArrayVec::<[TileBlockOffset; EXT_PARTITION_TYPES]>::new();
```
barrbrain: As far as I can tell, `EXT_PARTITION_TYPES` is 10 but the number of sub-partitions is in the range `1..=4`.

src/rdo.rs
```diff
@@ -1392,33 +1392,39 @@ pub fn rdo_tx_type_decision<T: Pixel>(
...
 }
 
 pub fn get_sub_partitions_with_border_check(
   four_partitions: &[TileBlockOffset; 4], partition: PartitionType,
   mi_width: usize, mi_height: usize, subsize: BlockSize,
-) -> Vec<TileBlockOffset> {
-  let mut partitions = vec![four_partitions[0]];
+) -> ArrayVec<[TileBlockOffset; EXT_PARTITION_TYPES]> {
+  let mut partition_offsets =
+    ArrayVec::<[TileBlockOffset; EXT_PARTITION_TYPES]>::new();
```
barrbrain: Likewise, `EXT_PARTITION_TYPES != 4` here.

src/rdo.rs
tdaede: I can switch it to plain ol' PARTITION_TYPES, but it will change when we actually support the other partitions. I wanted to do RAV1E_PARTITION_TYPES.len() but that's not allowed.

PR #1909, opened 2019-12-04
Document intentional dynamic allocations.

tdaede:


src/api/internal.rs
```diff
@@ -658,6 +658,7 @@ impl<T: Pixel> ContextInner<T> {
 
       let mvs = &fs.frame_mvs[index];
       use byteorder::{NativeEndian, WriteBytesExt};
+      // dynamic allocation: once per frame
```
shssoichiro: maybe worth noting this is from dump_lookahead_data?

tdaede: Ah, missed that.

PR #1910, opened 2019-12-04
Reduce number of allocations for planar data

barrbrain:
The LRF search and `ActivityMask` were allocating planar data as `Vec<Vec<_>>`.
Allocate as `Vec<_>` and use computed indices instead, e.g.:
`foo[x * w + y]` or `foo.chunks_exact(w)`.


barrbrain:
No significant impact to speed other than a 1.3% speed-up when running memory-profiler.

PR #1912, opened 2019-12-04
Reduce number of allocations for CDEF

barrbrain:
This continues #1910 - the allocating function `cdef_analyze_superblock`, was in another file.


lu-zero:
I would merge it sooner than later :)

Issue #1858, opened 2019-11-19
Desync with speed 0 and 1


ycho:
Tested with commit: acb7b69227ab29b42f8258b66d5c0dfee9a32470

./target/release/rav1e ~/Downloads/crowd_run_2160p50.y4m -o test.ivf -r test_rec.y4m --quantizer 128 --speed=1 --limit=3 --low_latency --tune=Psnr

../aom_build/aomdec test.ivf -o test_dec.y4m -v
../aom_build/aomdec test.ivf --rawvideo -o test_dec.yuv -v
ffmpeg -i test_rec.y4m test_rec.yuv -v 0
cmp test_rec.yuv test_dec.yuv

test_rec.yuv test_dec.yuv differ: char 8324073, line 1
==> difference in 2nd frame

Similar desync with speed 0. <s>1st frame is fine</s>.
Checked that it first happens at <s>the 2nd frame</s> both 1st and 2nd frame when two frames are encoded, and does. not happen if only one frame is encoded.
Speed 2, 6 and 10 were fine.

shssoichiro:
Could you give it a test at speed 2? I'd like to see if this is likely related to bottom-up encoding.

ycho:
Oh- Yeap, I missed stating that it was fine with speed 2!

ycho:
If I try narrowing down the problematic area, it is inter frame & bottom-up partition search.

ycho:
@shssoichiro 
It is possible "--tune Psnr" is causing desync for speed 0 and 1 (i.e. bottomup partition).
I tried:
 ./target/release/rav1e ~/Downloads/crowd_run_2160p50.y4m -o test.ivf -r test_rec.y4m --quantizer 128 --speed=1 --limit=2
, then it is fine.

And I tried again: 
./target/release/rav1e ~/Downloads/crowd_run_2160p50.y4m -o test.ivf -r test_rec.y4m --quantizer 128 --speed=1 --limit=2 --tune Psnr
, then desync happens.

test_rec.yuv test_dec.yuv differ: char 8418299, line 1

barrbrain:
As far as I can tell, the desync occurs where 4x4 blocks are chosen. If I increase `min_block_size` to `BlockSize::BLOCK_8X8` keeping all other settings, then it is fine.

ycho:
I am trying with tx_domain_distortion off now (its default is on, and used with --tune Psnr), which I suspect.

ycho:
[...]

PR #1564, opened 2019-08-19
Add heuristic quantizer adjustment

YaLTeR:
Introducing quantizer adjustment without speed loss.

[AWCY](https://beta.arewecompressedyet.com/?job=master-c8ca08c77c19514d251b497900d9a24e203caa0b&job=heuristic-sidx-2019-08-19_123457-9a6e4a7)

[AWCY vs. the RDO quantizers on --speed 2](https://beta.arewecompressedyet.com/?job=speed2-2019-08-19_131220-9a6e4a7&job=heuristic-sidx-speed2-2019-08-19_131918-69c0056): as expected, the RDO finds a solution that's slightly better.

PR #1520, opened 2019-08-01
Adjust quantizers via segmentation and RDO

YaLTeR:
[AWCY](https://beta.arewecompressedyet.com/?job=master-speed2-914fe8df0df9cdb87bd0f56fa8a57c8a3eab9489%402019-08-01T17%3A03%3A57.470Z&job=temporal-rdo-quantizer-27)

[AWCY but on --speed 5](https://beta.arewecompressedyet.com/?job=master-914fe8df0df9cdb87bd0f56fa8a57c8a3eab9489&job=temporal-rdo-quantizer-26)

With @tdaede we decided I should put this in `--speed 2` as it about doubles the encoding time.

This adds three quantizer index deltas: 0, -15 and 15, and adds searching them to the RDO. This way it's affected by any RDO biasing, be it block importance or activity. This showed better results than me trying to naively set the segment delta according to mean importance.

PR #1914, opened 2019-12-05
[WIP] Temporary fix for desync from quantizer rdo

ycho:
Temporary fIx for #1858.
It basically, revoke the commts from #1520 and #1564, since those commits are confirmed to cause the desync at speed 0,1,2.

Two commits that cause desync are:
https://github.com/xiph/rav1e/commit/3c79e025f7c13635558ce66370920fb5a8a38754 and
https://github.com/xiph/rav1e/commit/db0da3c97efb269cb23571f2c1f68afdf1ad9813
those needs different temporary fixes.

One set of test runs to see desync that I used can be:
./target/release/rav1e nyan.y4m -o test.ivf -r test_rec.y4m --quantizer 50 --speed=0 --limit=30
./target/release/rav1e nyan.y4m -o test.ivf -r test_rec.y4m --quantizer 50 --speed=1 --limit=30
./target/release/rav1e nyan.y4m -o test.ivf -r test_rec.y4m --quantizer 50 --speed=2 --limit=30

Note that the desync happens regardless of --tune.

### DIFF
--- a/src/api/config.rs
+++ b/src/api/config.rs
@@ -450,8 +450,9 @@ impl SpeedSettings {
     true
   }
 
-  const fn quantizer_rdo_preset(speed: usize) -> bool {
-    speed <= 1
+  const fn quantizer_rdo_preset(_speed: usize) -> bool {
+    // Until the desync caused by quantizer_rdo is fixed, disable this feature
+    false
   }
 
   const fn use_satd_subpel(speed: usize) -> bool {
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -578,6 +578,9 @@ fn luma_chroma_mode_rdo<T: Pixel>(
         0..=2
       }
     } else {
+      // FIXME:
+      // Until the desync caused by heuristic quantizer selection is fixed, disable this feature
+      /*
       let importance =
         compute_mean_importance(fi, ts.to_frame_block_offset(tile_bo), bsize);
       // Chosen based on the RDO segment ID statistics for speed 2 on the DOTA2
@@ -602,6 +605,8 @@ fn luma_chroma_mode_rdo<T: Pixel>(
       } else {
         heuristic_sidx..=heuristic_sidx
       }
+      */
+      0..=0
     };
 
     for sidx in sidx_range {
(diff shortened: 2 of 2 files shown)


tdaede:
If you want a quick fix, simply disabling segmentation seems like a more straightforward fix.

ycho:
> If you want a quick fix, simply disabling segmentation seems like a more straightforward fix.

Oh, that directs me that I forgot to not use segment.
So that, self.write_segment_data(fi, &fs.segmentation)?; signals that segment data is not written in frame header and on segment data in each partition.
Btw, fs.segmentation.enabled = true; is hard coded in segmentation_optimize() and the functions always runs even when segment is not used.

ycho:
> If you want a quick fix, simply disabling segmentation seems like a more straightforward fix.

If you prefer that way, please go ahead since I got knockdown thanks to chasing this mess! :)

tdaede:
OK. If segmentation = off doesn't work, we can do it this way. But you should at least merge patches 1 and 3 I think, they are two halves of the same thing.

ycho:
> OK. If segmentation = off doesn't work, we can do it this way. But you should at least merge patches 1 and 3 I think, they are two halves of the same thing.

Sure thing. I will also try add commits that avoids not writing any seg data when not in use.
Also, please away from rav1e, you are on PTO, folk! :)

lu-zero:
Sounds a good plan to me.

ycho:
> OK. If segmentation = off doesn't work, we can do it this way. But you should at least merge patches 1 and 3 I think, they are two halves of the same thing.

I think the reason why I split the changes into multiple commits is to be 'minimally invasive' so that the problematic but still more code are still being compiled by our CI. Otherwise, the whole quantizer_rdo else block (i.e. non heuristic part)  needed be commented out.
I might go landing this PR (with merging commits 1 and 3 as you requested), but I am considering that another possibly better way to keep all the code and not commenting out any code would be add the condition of segment enable, i.e. "if segmentation.enable { ... }".

barrbrain:
I agree with the recent comments, it would be preferable to have a path for disabling segmentation.

ycho:
@tdaede , @barrbrain , following your comments, i think I've devised better way to (temporary) fix desync. See, https://github.com/xiph/rav1e/pull/1923

ycho:
Since we have better fix which disables the use of segmentation, as in https://github.com/xiph/rav1e/pull/1923, closing this.

PR #1915, opened 2019-12-06
Add configuration for grcov action

barrbrain:
Remove some noise in results reported on Coveralls. Effectively limit results to the rav1e crate sources.

### DIFF
--- a/.github/actions-rs/grcov.yml
+++ b/.github/actions-rs/grcov.yml
@@ -1,0 +1,4 @@
+ignore-not-existing: true
+ignore:
+  - "/*"
+  - "../*"


lu-zero:
Sounds good

barrbrain:
Compare the result from [this PR](https://coveralls.io/builds/27453288) with [master](https://coveralls.io/builds/27436556):
* `../`
* `crates/`
* `ivf/`

shssoichiro:
We should probably also ignore src/test_encode_decode

barrbrain:
> We should probably also ignore src/test_encode_decode

Perhaps, I think that is part of a wider conversation about rationalising our test coverage.
The primary intent of this PR was reduce the noise from Coveralls.

PR #1918, opened 2019-12-06
Make sure we do not test invalid candidates to begin with

lu-zero:
The caller is in charge of assembling the candidates and should do so based on the same condition checked there.


lu-zero:
in `encoder.rs` we have also:
``` rust
  let must_split = (tile_bo.0.x + bsw as usize > ts.mi_width
    || tile_bo.0.y + bsh as usize > ts.mi_height
    || bsize.greater_than(BlockSize::BLOCK_64X64))
    && is_square;
```

while in `rdo.rs`
``` rust
 PARTITION_NONE if bsize <= BlockSize::BLOCK_64X64
```

probably we could short-circuit some more.

PR #1920, opened 2019-12-06
Add avx2 for forward txfms

KyleSiefring:
Putting this up here for other people to see. I haven't gotten the chance to read the entire diff in full. May need to split a few things into individual pull requests/commits.


lu-zero:
Yes please :) probably you can split in 3 :)

shssoichiro:
@barrbrain What you didn't point out is that this is a 10-15% speedup if we are to believe those AWCY numbers. :tada:  :rocket:

PR #1922, opened 2019-12-06
Fix wrong init for preskip

ycho:
The preskip should be initially false, then only become true if seg
feature >= SEG_LVL_REF_FRAME is enabled.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -468,7 +468,7 @@ impl Default for SegmentationState {
       enabled: false,
       update_data: false,
       update_map: false,
-      preskip: true,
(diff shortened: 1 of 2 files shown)


src/encoder.rs
```diff
@@ -3422,8 +3422,10 @@ pub fn encode_frame<T: Pixel>(
   debug_assert!(!fi.invalid);
   let mut packet = Vec::new();
 
+  if true {
```
lu-zero: why this block?

ycho:
I was too hungry and forgot that I left it there!

Thanks,
Yushin

PR #1925, opened 2019-12-08
Speed up me by removing unnecessary cost calculation

takehirokj:
The same best mv can be selected by only sad.
The selected mv and output file will be same as previous one which use rate and sad.
This removal speeds up the encoding up to 1.2-7.8% in fps in my test data.

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -971,6 +971,9 @@ fn full_search<T: Pixel>(
     height: (y_hi - y_lo) as usize + blk_h,
   });
 
+  let mut lowest_sad = std::u32::MAX;
+  let mut lowest_sad_mv = MotionVector::default();
+
   // Select rectangular regions within search region with vert+horz windows
   for vert_window in search_region.vert_windows(blk_h).step_by(step) {
     for ref_window in vert_window.horz_windows(blk_w).step_by(step) {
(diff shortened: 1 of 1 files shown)


lu-zero:
From what I can see the two computations aren't identical, how did you check that the results are identical?

rzumer:
[This](https://beta.arewecompressedyet.com/?job=master-2ec50e357aa60e7c69f5fec63d4c3761c6536e53&job=takehirokj-me%402019-12-08T15%3A25%3A36.904Z) confirms that it is equivalent.

barrbrain:
AWCY results for [this PR at default speed](https://beta.arewecompressedyet.com/?job=master-2ec50e357aa60e7c69f5fec63d4c3761c6536e53&job=takehirokj-me%402019-12-08T15%3A25%3A36.904Z) match observations by @takehirokj. I tried some quick `println!` debugging in `estimate_motion_ss4` and `lambda` was zero for the full quantizer range. Further digging showed that `fi.me_lambda` was also zero. I think the underlying bug is that `fi.me_lambda` has not been derived at the time that `estimate_motion_ss4` is called.

barrbrain:
> I think the underlying bug is that `fi.me_lambda` has not been derived at the time that `estimate_motion_ss4` is called.

#1926 addresses the issue of zero lambda; the impact on compression is slightly positive.

takehirokj:
I have confirm the same bug as @barrbrain.
Since this bug will be addressed in #1926, closing this.

PR #1927, opened 2019-12-09
[WIP] Avoid checking for scene flashes if scene detection disabled

shssoichiro:
~1% overall speedup when scene detection disabled.
(Tested on 1080p 60-frames, 8-tiles, speed 6.)

### DIFF
--- a/src/scenechange/mod.rs
+++ b/src/scenechange/mod.rs
@@ -75,7 +75,12 @@ impl SceneChangeDetector {
       }
     };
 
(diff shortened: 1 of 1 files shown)


lu-zero:
Why are we hitting that path if it is disabled? We shouldn't be there to begin with...

shssoichiro:
This struct still handles choosing keyframes using min/max keyframe interval, but you're right, it may be worth doing a larger refactor to simplify that path.

shssoichiro:
Superceded by #1940

PR #1928, opened 2019-12-09
Early exit exclude scene flashes loop

shssoichiro:
If we start with the furthest frame,
we can imply that all closer frames
are also scene flashes
and avoid testing them.

<1% overall speedup.
(Tested on 1080p 60-frames, 8-tiles, speed 6.)

### DIFF
--- a/src/scenechange/mod.rs
+++ b/src/scenechange/mod.rs
@@ -137,13 +137,16 @@ impl SceneChangeDetector {
     // Where A and B are scenes: AAAAAABBBAAAAAA
     // If BBB is shorter than lookahead_distance, it is detected as a flash
     // and not considered a scenecut.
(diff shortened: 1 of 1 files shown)


src/scenechange/mod.rs
```diff
@@ -137,13 +137,16 @@ impl SceneChangeDetector {
     // Where A and B are scenes: AAAAAABBBAAAAAA
     // If BBB is shorter than lookahead_distance, it is detected as a flash
     // and not considered a scenecut.
-    for j in 1..=lookahead_distance {
+    for j in (1..=lookahead_distance).rev() {
```
tdaede: A comment of why we do this in reverse would be nice.

lu-zero:
I guess it should land now?

PR #1931, opened 2019-12-09
Use intra and inter costs to determine frame type

### LINKED ISSUE #1528: Use Temporal RDO lookahead metric for scene detection instead of frame delta

shssoichiro:
Closes #1528

Provides .1-.2% BD Rate improvement on scenecut-720p set.
Because the threshold scales based on distance
since the last keyframe,
it is expected that the improvement will be more on real content
(i.e. clips with more than 240 frames).

Overall is about 1% slower (with 1 tile) compared to master.

[AWCY](https://beta.arewecompressedyet.com/?job=scenecut-master%402019-12-09T01%3A05%3A19.889Z&job=scenecut-cost-3%402019-12-09T12%3A36%3A04.496Z)

### DIFF
--- a/src/scenechange/mod.rs
+++ b/src/scenechange/mod.rs
@@ -7,31 +7,39 @@
 // Media Patent License 1.0 was not distributed with this source code in the
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
-use crate::api::EncoderConfig;
-use crate::api::InterConfig;
+use crate::api::lookahead::*;
+use crate::api::{EncoderConfig, InterConfig};
+use crate::cpu_features::CpuFeatureLevel;
+use crate::encoder::Sequence;
 use crate::frame::*;
+use crate::hawktracer::*;
 use crate::util::CastFromPrimitive;
 use crate::util::Pixel;
-
-use crate::hawktracer::*;
-
-use std::cmp;
 use std::collections::BTreeSet;
 use std::sync::Arc;
 
 /// Runs keyframe detection on frames from the lookahead queue.
 pub(crate) struct SceneChangeDetector {
   /// Minimum average difference between YUV deltas that will trigger a scene change.
-  threshold: u8,
-  /// Fast scene cut detection mode, ignoing chroma planes.
+  threshold: u64,
+  /// Fast scene cut detection mode, uses simple SAD instead of encoder cost estimates.
   fast_mode: bool,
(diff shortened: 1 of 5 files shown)


tdaede:
What actually makes this slower? Can we not reuse the estimated inter cost from temporal RDO?

shssoichiro:
The issue that prevents us from reusing the estimated costs from temporal RDO is that frame type selection runs before temporal RDO. My previous attempt tried to reverse that order, which did allow to reuse temporal RDO metrics, but meant that whenever a keyframe was chosen, the last 40 (number of lookahead) frames would all have to be recalculated. So it was quite slow and inefficient. This is actually faster than that approach, especially since it uses a faster inter cost estimation that only looks at motion vectors between two frames (no bi-directional, multiref, etc).

I would very much like to deduplicate the code, however. I'll probably take a pass at that along with figuring out why the build is failing.

src/scenechange/mod.rs
```diff
@@ -152,33 +224,80 @@ impl SceneChangeDetector {
...
+  /// in many cases.
   fn has_scenecut<T: Pixel>(
-    &self, frame1: &Frame<T>, frame2: &Frame<T>,
-  ) -> bool {
-    let mut len = frame2.planes[0].cfg.width * frame2.planes[0].cfg.height;
-    let mut delta = 0;
-
-    delta += self.delta_in_planes(&frame1.planes[0], &frame2.planes[0]);
+    &self, frame1: Arc<Frame<T>>, frame2: Arc<Frame<T>>, frameno: u64,
+    keyframes: &BTreeSet<u64>,
```
lu-zero: Would be better to not pass the set around, computing the last keyframe and passing it around usually works better (see my refactor)

lu-zero:
Why the tests change so much?

shssoichiro:
Primarily due to now counting frames within the last pyramid as a scene flash. In order to actually test the scene change detection, I modified the tests to read 5 extra frames, and then ignore those last 5 frames for the test assertions.

shssoichiro:
> <lu_zero> I wanted to do that after we can run that and the other send_frame tasks in a separate thread

PR #1933, opened 2019-12-09
Fix for issue #1903 (Krøyer_PS.y4m desync)

### LINKED ISSUE #1903: Decoding error on encoded Krøyer_PS.y4m from subset1 at high quantizers
The command line options are `--tiles 4 --quantizer [252|220]`, everything else is default. Lower quantizer values (among the default quantizers tested on AWCY) do not seem to generate corrupted bitstreams. I have not tested much besides that.

dav1d output:
```
dav1d 0.4.0-47-g556890b - by VideoLAN
Error decoding frame: Invalid argument
No data decoded
```

aomdec output:
```
Warning: Failed to decode frame 1: Corrupt frame detected
Warning: Additional information: Corrupted segment_ids
```

xiphmont:
Tiling had broken an edge-of-frame check in superblock queue flush triggering.  It was checking tile-local superblock offset against the edge of the frame instead of checking *global* superblock offset against the edge of frame.  This should fix the desync bug with no other functional change.

Fixes #1903

### DIFF
--- a/src/tiling/tile_restoration_state.rs
+++ b/src/tiling/tile_restoration_state.rs
@@ -225,18 +225,21 @@ macro_rules! tile_restoration_plane_common {
       pub fn restoration_unit_last_sb_for_rdo<T: Pixel>(
         &self,
         fi: &FrameInvariants<T>,
-        sbo: TileSuperBlockOffset,
+        global_sbo: PlaneSuperBlockOffset,
+        tile_sbo: TileSuperBlockOffset,
       ) -> bool {
         // there is 1 restoration unit for (1 << sb_shift) super-blocks
         let h_mask = (1 << self.rp_cfg.sb_h_shift) - 1;
         let v_mask = (1 << self.rp_cfg.sb_v_shift) - 1;
         // is this a stretch block?
-        let x_stretch = sbo.0.x >> self.rp_cfg.sb_h_shift >= self.units.cols;
-        let y_stretch = sbo.0.y >> self.rp_cfg.sb_v_shift >= self.units.rows;
-
-        let last_x = (sbo.0.x & h_mask == h_mask && !x_stretch) || sbo.0.x == fi.sb_width-1;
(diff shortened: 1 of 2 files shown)


src/tiling/tile_restoration_state.rs
```diff
@@ -225,18 +225,21 @@ macro_rules! tile_restoration_plane_common {
       pub fn restoration_unit_last_sb_for_rdo<T: Pixel>(
         &self,
         fi: &FrameInvariants<T>,
-        sbo: TileSuperBlockOffset,
+        global_sbo: PlaneSuperBlockOffset,
```
ycho: Does 'global' mean the sbo in the frame, instead of the tile, correct?

barrbrain:
I confirmed that this fixes #1903. Minor comment on commit subject: #1903 is an issue rather than a pull request, so the PR prefix is misleading. Also, including "Fixes #1903." in the body will enable GItHub to auto-close the issue.

src/tiling/tile_restoration_state.rs
xiphmont: yes.

xiphmont:
> Looks good, does awcy agree?

Yes, all zeroes (as expected, default AWCY didn't trip the bug).

src/tiling/tile_restoration_state.rs
tmatth: I might have called it `frame_sbo` but not really a dealbreaker

xiphmont:
We use 'Frames' in several places that are not the viewable frame....

Monty

xiphmont:
....but sure, it's certainly no worse, might be better.

src/tiling/tile_restoration_state.rs
ycho: > I might have called it `frame_sbo` but not really a dealbreaker

Yes, we will write a patch to rename it when monty is back to seep.... *_*

PR #1934, opened 2019-12-09
Fix temporal RDO biasing distortion

smarter:


src/rdo.rs
```diff
@@ -390,37 +390,69 @@ fn compute_tx_distortion<T: Pixel>(
...
+  //
+  //     scale = 2^(QP_delta / -3)
+  //           = (1 + (propagate_cost / intra_cost))^(strength / 3)
+  //
+  //  The original paper empirically chooses strength = 2.0, but strength = 1.0
+  //  seems to work best in rav1e currently, this may have something to do with
+  //  the fact that they use 16x16 blocks whereas our "importance blocks" are
+  //  8x8, but everything should be scale invariant here so that's weird.
+  //
+  // @article{mbtree,
```
lu-zero: [mbtree]: https://pdfs.semanticscholar.org/032f/1ab7d9db385780a02eb2d579af8303b266d2.pdf

If you want to have it rendered in a future.

tools/draw-importances.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
```
YaLTeR: Pretty sure this is not needed for Python 3.

tools/draw-importances.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/python3
```
lu-zero:
```suggestion
#!/usr/bin/env python
```

This should work even if you have python on a special path (quite common)

src/rdo.rs
```diff
@@ -390,37 +390,69 @@ fn compute_tx_distortion<T: Pixel>(
...
-) -> f64 {
-  let mean_importance = compute_mean_importance(fi, frame_bo, bsize);
 
-  // Chosen based on a series of AWCY runs.
-  const FACTOR: f32 = 3.;
-  const ADDEND: f64 = 0.65;
-
-  let bias = (mean_importance / FACTOR) as f64 + ADDEND;
-  debug_assert!(bias.is_finite());
+  // The mbtree paper \cite{mbtree} uses the following formula:
```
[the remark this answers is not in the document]
smarter: This is the style used in theora and daala, I'd like to keep it as is, at least until we settle on a better convention.

smarter:
Well, it's both of course :) but i'll update the commit message to say blocks since that's more precise.

lu-zero:
Shall we land it before or after 0.2.0 ?

barrbrain:
That segmentation is currently disabled by default is an important mitigation for loss of the segment heuristic. Without the heuristic, enabling segmentation incurs a ~2x slow-down. We could replace it with a trivial placeholder rather than deletion to minimize the API changes.

ycho:
> That segmentation is currently disabled by default is an important mitigation for loss of the segment heuristic. Without the heuristic, enabling segmentation incurs a ~2x slow-down. We could replace it with a trivial placeholder rather than deletion to minimize the API changes.

I think I am not following exactly. By "loss of the segment heuristic", did you mean that we have sacrificed the coding gain by "heuristic quantizer selection"? FYI, the workaround commit of segmentation switch and turned it off by default does disable both quantizer rdo and heuristic q selection. So, I read that, given that heuristic method gives significant coding gain with no change in enc time, while quantizer rdo cause x2.5~3 increase in enc time, we can consider fixing heuristic method in the future and keep the API path for now. Am I getting what you meant correctly?

PR #1935, opened 2019-12-10
Fix error_resilient mode and add tests.

tdaede:


tdaede:
It's actually a fix for FrameType::SWITCH.

Cargo.toml
```diff
@@ -62,46 +46,34 @@ num-traits = "0.2"
...
 paste = "0.1"
 noop_proc_macro = "0.2.0"
-serde = { version = "1.0", features = ["derive"], optional = true }
 dav1d-sys = { version = "0.2", optional = true }
 aom-sys = { version = "0.1.3", optional = true }
 scan_fmt = { version = "0.2.3", optional = true }
 ivf = { version = "0.1", path = "ivf/", optional = true }
 rayon = "1.0"
-toml = { version = "0.5", optional = true }
+bincode = { version = "1.1", optional = true }
```
shssoichiro: I'm thinking this was not intentional--it looks like this PR reverts a lot of changes in Cargo.toml.

src/encoder.rs
```diff
@@ -223,7 +223,7 @@ impl Sequence {
       enable_interintra_compound: false,
       enable_masked_compound: false,
       enable_dual_filter: false,
-      enable_order_hint: !config.still_picture,
```
lu-zero: is this part of the fix?

tdaede: Whoops no.

PR #1936, opened 2019-12-10
Build assembler files with dwarf debug symbols

cryptomilk:
This is needed for debuginfo package generation for distributions. E.g.
for openSUSE or Fedora

This should be set if '-C debuginfo=X' is passed to rust. However I'm a total Rust newbie especially when it comes to the build system.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -63,7 +63,7 @@ fn build_nasm_files() {
       "src/x86/cdef.asm",
       "src/x86/tables.asm",
     ],
-    &[&config_include_arg, "-Isrc/"],
+    &[&config_include_arg, "-Isrc/", "-Fdwarf"],
   );
   println!("cargo:rustc-link-lib=static=rav1easm");
   rerun_dir("src/x86");


lu-zero:
nasm-rs checks for the DEBUG env var and adds -g to its flags from what I could read.

cryptomilk:
So it would be enough to add `-Fdwarf` by default?

cryptomilk:
I guess the problem is that on Windows you can't pass `-Fdwarf`. I have no idea how to detect if you're building ELF objects.

lu-zero:
adding `-g` alone isn't enough?

cryptomilk:
No, it looks like the default is -Fstabs for nasm and you can't extract debuginfo packages out of that.

lu-zero:
Understood, adding the flag only on linux is just a matter of checking the target arch, do you want to try it or want me to? :)

cryptomilk:
I have no idea how to do that. I guess the best would be do address that in nasm-rs. But we could fix it here too.

cryptomilk:
Fixed by #1938

PR #1937, opened 2019-12-10
Refactor out intra/inter cost lookahead code

shssoichiro:
This code can be reusable in different parts
of the lookahead, for example from scenechange detection.

### DIFF
--- a/src/api/lookahead.rs
+++ b/src/api/lookahead.rs
@@ -1,0 +1,215 @@
+use crate::api::internal::InterConfig;
+use crate::config::EncoderConfig;
+use crate::context::{
+  BlockOffset, FrameBlocks, SuperBlockOffset, TileBlockOffset,
+  TileSuperBlockOffset,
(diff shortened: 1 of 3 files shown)


src/api/lookahead.rs
```diff
@@ -0,0 +1,215 @@
...
+
+pub(crate) fn compute_motion_vectors<T: Pixel>(
+  fi: &mut FrameInvariants<T>, fs: &mut FrameState<T>, inter_cfg: &InterConfig,
+) {
+  let mut blocks = FrameBlocks::new(fi.w_in_b, fi.h_in_b);
+  fs.half_res_pmvs = fi
+    .tiling
+    .tile_iter_mut(fs, &mut blocks)
+    .collect::<Vec<_>>()
+    .into_par_iter()
```
lu-zero: Note to self, make sure we install the right threadpool before this is called.

PR #1938, opened 2019-12-10
nasm_rs: Force the dwarf debug for elf targets

lu-zero:


lu-zero:
@cryptomilk This should address your need, please test it and report back :)

cryptomilk:
I've started a build with the patch. Thanks.

cryptomilk:
Works, thanks!

PR #1939, opened 2019-12-10
Avoid building the scene detection exclusion list when not needed

lu-zero:


lu-zero:
#1940 expands from here.

PR #1940, opened 2019-12-10
Scenechange simplification

lu-zero:
About a 1% faster for the speed 10 use-case.


src/scenechange/mod.rs
```diff
@@ -99,11 +73,20 @@ impl SceneChangeDetector {
       return true;
     }
 
-    // Skip smart scene detection if it's disabled
     if config.speed_settings.no_scene_detection {
       return false;
     }
 
+    self.exclude_scene_flashes(&frame_set, input_frameno, inter_cfg);
```
shssoichiro: I feel like I had a reason for running this on every frame, although logically it feels unnecessary. I'm going to throw this branch in AWCY just as a sanity test.

lu-zero: Thank you :)

PR #1943, opened 2019-12-11
Add switch frame support and switch-frame-interval CLI option.

tdaede:


src/api/config.rs
```diff
@@ -64,6 +64,9 @@ pub struct EncoderConfig {
   /// Flag to force all frames to be error resilient.
   pub error_resilient: bool,
 
+  /// Interval between switch frames (0 to disable)
+  pub switch_frame_interval: u64,
+
```
barrbrain: We might want to include this in the C-API. See `option_match` in `src/capi.rs`.
Validating the "multiple of `group_input_len`" constraint may not be straightforward.

tdaede: It's a pretty terrible constraint in general and I want to come up with a better one.

lu-zero: I guess adding a placement override could wait for that :)

tdaede: I do want to add a placement override to the API, but again it has to change the pyramid length.

lu-zero: Shall we land it as-is or wait for that?

tdaede: I made it reject non-low-latency case, so this is worked around for now.

tdaede:
I'm going to land this with it restricted to low latency only if that's OK.

lu-zero:
Two tests fail as in 4 not being a valid interval.

tdaede:
They actually failed because I forgot to delete the reordering test after disallowing reordering with switch frames.

PR #1947, opened 2019-12-12
CI: Update libdav1d to 0.5.2

barrbrain:
We are exchanging the time to fetch package lists on every CI task for effort to periodically update selected package metadata.


rzumer:
I'm not sure it's a great idea to break CI every time there is a patch update on a dependency.

shssoichiro:
Agreed, I'd be fine with the manual update if builds kept working on the old version, but having builds break every time Debian updates their package seems annoying.

barrbrain:
For what it's worth, this time around there was a 9 day window where both versions were distributed. But we did miss it.

barrbrain:
Since the package is small, we could use CI cache to keep it around between version bumps.

PR #1950, opened 2019-12-13
Optimize sgrproj_solve to avoid f64 arithmetic

KyleSiefring:
Results can always fit in i64 regardless of what bit depth is used. It
doesn't seem like much vectorization is happening but it is faster. Some
quick tests show that switching to a smaller type seems to enable
auto-vectorization.


KyleSiefring:
https://beta.arewecompressedyet.com/?job=master-664fa46b2c9bea79448473a3fe59ee6d3448d81c&job=sgr_solve_i64%402019-12-13T18%3A12%3A44.604Z

PR #1951, opened 2019-12-13
Move out the lookahead allocation

lu-zero:
And simplify it.


shssoichiro:
Is this toward any particular purpose? Makes it easier to move lookahead to its own thread?

Also, rustfmt. :slightly_smiling_face:

lu-zero:
> Is this toward any particular purpose? Makes it easier to move lookahead to its own thread?

That's the idea, we need probably to consider how many frames to process at time and if would make sense to take in an Iterator and not a slice.

> Also, rustfmt. 🙂

rustfmt and clippy are happy now :)

PR #1955, opened 2019-12-16
Limit transforms to single transpose

KyleSiefring:
Results in a small but measurable performance improvement.


KyleSiefring:
What is this touching that isn't getting tested? 1D, 64x32, 32x64, 64x16, 16x64 transforms? I want to make sure all this stuff works before committing.

ycho:
> What is this touching that isn't getting tested? 1D transform context writing? 64x32, 32x64, 64x16, 16x64 transforms? I want to make sure all this stuff works before committing.

You should test with speed 0 for it because non-square tx blocks are only used with non-square partition, which is enabled for speed 0 only.

tdaede:
We roundtrip test every transform AFAIK. But that means the coefficient writing could still fail.

PR #1957, opened 2019-12-16
Small fix to deblocking RDO

xiphmont:
chroma cropping was being handled slightly differently in RDO and actual filtering.  Actual filtering was correct (not a desync, just a very minor RDO effect).

AWCY reports zeroes-- that's due to our specific AWCY testset.


lu-zero:
Is it worth/feasible to factor it in a function and call it in both places?

xiphmont:
> Is it worth/feasible to factor it in a function and call it in both places?

Meh-- there's a couple examples of parallel code structure there that should be cleaned up, but might well introduce more hazards than they repair.

lu-zero:
Let's do that later, maybe open an issue about it.

PR #1958, opened 2019-12-17
Add a mini benchmark report to the cli

lu-zero:
Potentially we could add windows support using winapi-rs later.

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -33,6 +33,7 @@ pub struct CliOptions {
   pub override_time_base: bool,
   pub skip: usize,
   pub verbose: bool,
+  pub benchmark: bool,
   pub threads: usize,
   pub pass1file_name: Option<String>,
   pub pass2file_name: Option<String>,
@@ -303,6 +304,11 @@ pub fn parse_cli() -> Result<CliOptions, CliError> {
     )
     // DEBUGGING
     .arg(
+      Arg::with_name("BENCHMARK")
+        .help("Provide a benchmark report at the end of the encoding")
+        .long("benchmark")
+    )
+    .arg(
       Arg::with_name("VERBOSE")
         .help("Verbose logging; outputs info for every frame")
         .long("verbose")
@@ -429,6 +435,7 @@ pub fn parse_cli() -> Result<CliOptions, CliError> {
     color_range_specified: matches.occurrences_of("PIXEL_RANGE") > 0,
     override_time_base: matches.is_present("FRAME_RATE"),
     skip: matches.value_of("SKIP").unwrap().parse().unwrap(),
+    benchmark: matches.is_present("BENCHMARK"),
     verbose: matches.is_present("VERBOSE"),
     threads,
     pass1file_name: matches.value_of("FIRST_PASS").map(|s| s.to_owned()),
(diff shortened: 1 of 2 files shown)


src/bin/common.rs
```diff
@@ -302,6 +303,11 @@ pub fn parse_cli() -> Result<CliOptions, CliError> {
         .alias("still_picture")
     )
     // DEBUGGING
+    .arg(
+      Arg::with_name("BENCHMARK")
+        .help("Provide a benchmark report at the end of the encoding")
+        .long("benchmark")
```
shssoichiro: Would it make sense to make this option hidden from the user?

src/bin/rav1e.rs
```diff
@@ -265,6 +265,31 @@ fn init_logger() {
...
+    }
+  } else {
+    fn print_rusage() {
+      let (utime, stime, maxrss) = unsafe {
+        let mut usage = std::mem::zeroed();
+        let _ = libc::getrusage(libc::RUSAGE_SELF, &mut usage);
+        (usage.ru_utime, usage.ru_stime, usage.ru_maxrss)
+      };
+      eprintln!(
+        "utime: {} us",
```
shssoichiro: I assume this stands for "user time" and "system time"? It would be clearer to print out the full words.

src/bin/rav1e.rs
```diff
@@ -265,6 +265,31 @@ fn init_logger() {
   pretty_env_logger::init();
 }
 
+cfg_if::cfg_if! {
+  if #[cfg(target_os = "windows")] {
+    fn print_rusage() {
+      eprintln!("Windows is not supported.");
```
shssoichiro: Probably clearer to put "--benchmark is not supported on Windows". This message is a bit vague and might be misinterpreted as "rav1e is not supported on Windows".

src/bin/rav1e.rs
```diff
@@ -265,6 +265,31 @@ fn init_logger() {
...
+  } else {
+    fn print_rusage() {
+      let (utime, stime, maxrss) = unsafe {
+        let mut usage = std::mem::zeroed();
+        let _ = libc::getrusage(libc::RUSAGE_SELF, &mut usage);
+        (usage.ru_utime, usage.ru_stime, usage.ru_maxrss)
+      };
+      eprintln!(
+        "utime: {} us",
+        utime.tv_sec as i64 * 1_000_000 + utime.tv_usec as i64
```
shssoichiro: If this is benchmarking a whole encode, it probably makes more sense to show something like "2345.006 sec" than "23450067890 us".

src/bin/common.rs
lu-zero: It is a good idea indeed!

src/bin/rav1e.rs
lu-zero: Sure

src/bin/rav1e.rs
lu-zero: Formatted.

src/bin/rav1e.rs
lu-zero: Sure, I checked how to do that using winapi and it is feasible, maybe later some windows user can give it a try.

PR #1962, opened 2019-12-17
Unbreak the scenechange detection

lu-zero:
Pass all the frames in the buffer to the scenedetector only when in flushing state.


shssoichiro:
This seems more complicated than the prior version (before the breaking change). Could you explain why you chose this implementation rather than reverting?

lu-zero:
I prefer having the code explicit, alternatively I can revert and add a comment, but probably tomorrow, I'm falling asleep now ^^;

lu-zero:
Here a tidier version, I'm still sold in making explicit the flush condition.

lu-zero:
Appveyor got a networking issue but I have no mean to restart it =/

PR #1963, opened 2019-12-17
Refactor coefficient writing

KyleSiefring:
Improves performance slightly.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -3752,18 +3752,16 @@ impl<'a> ContextWriter<'a> {
   }
 
   fn txb_init_levels(
-    &self, coeffs: &[i32], width: usize, height: usize, levels_buf: &mut [u8],
+    &self, coeffs: &[i32], height: usize, levels: &mut [u8],
+    levels_stride: usize,
   ) {
     // Coefficients and levels are transposed from how they work in the spec
-    let mut offset = TX_PAD_TOP * (height + TX_PAD_HOR);
-
-    for x in 0..width {
-      let coeffs_col = &coeffs[(x * height)..][..height];
-      let levels_col = &mut levels_buf[offset..][..height];
+    for (coeffs_col, levels_col) in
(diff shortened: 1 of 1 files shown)


src/context.rs
```diff
@@ -4195,12 +4137,8 @@ impl<'a> ContextWriter<'a> {
       }
       // save extra golomb codes for separate loop
       if level > (COEFF_BASE_RANGE + NUM_BASE_LEVELS) as u32 {
```
KyleSiefring: same here

src/context.rs
```diff
@@ -4140,16 +4084,14 @@ impl<'a> ContextWriter<'a> {
       }
 
       if level > NUM_BASE_LEVELS as u32 {
```
KyleSiefring: level > ... as u32 then convert level to u16. This has code smell.

src/context.rs
```diff
@@ -3827,51 +3825,42 @@ impl<'a> ContextWriter<'a> {
...
-        //   if (row < 2) return 11 + ctx;
-        // } else if (width > height) {
-        //   if (col < 2) return 16 + ctx;
-        // }
-        // if (row + col < 2) return ctx + 1;
-        // if (row + col < 4) return 5 + ctx + 1;
-        // return 21 + ctx;
-        ctx
-          + av1_nz_map_ctx_offset[tx_size as usize][cmp::min(row, 4)]
+    let ctx = ((stats + 1) >> 1).min(4);
```
tdaede: any reason not to combine these into one statement

src/context.rs
tdaede: I think NUM_BASE_LEVELS was a #define in libaom so changing its type to u16 would also be a legit fix

tdaede: actually no we actually do the conversion at v.abs() as u16 right above. I think we should just drop all those conversions

KyleSiefring:
Closing till I fix the bug

PR #1964, opened 2019-12-17
Stop abs coeffs from being cast to u16 when writing

KyleSiefring:


tdaede:
Also is there any reason we can't leave level as just i32?

KyleSiefring:
I don't see why not.

PR #1967, opened 2019-12-18
CI: Use vswhere to query compiler path

barrbrain:
Follow up to #1965. The nasm-rs crate requires `link.exe` on the path.

### DIFF
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -280,7 +280,6 @@ jobs:
         7z e -y "$NASM_VERSION-win64.zip" -o"C:\nasm"
         echo "::add-path::C:\nasm"
     - name: Set MSVC x86_64 linker path
-      shell: pwsh
       run: |
         $LinkGlob = "VC\Tools\MSVC\*\bin\Hostx64\x64"
         $env:PATH = "$env:PATH;${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer"
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -282,9 +282,9 @@ jobs:
     - name: Set MSVC x86_64 linker path
       shell: pwsh
       run: |
-        $LinkGlob = "VC\Tools\MSVC\**\bin\Hostx64\x64\link.exe"
-        Set-Alias vswhere "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
-        $LinkPath = Split-Path (& vswhere -latest -products * -find "$LinkGlob" | Select-Object -Last 1) -Parent
+        $LinkGlob = "VC\Tools\MSVC\*\bin\Hostx64\x64"
+        $env:PATH = "$env:PATH;${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer"
+        $LinkPath = vswhere -latest -products * -find "$LinkGlob" | Select-Object -Last 1
         echo "::add-path::$LinkPath"
     - name: Install Rust
       uses: actions-rs/toolchain@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/rav1e.yml
```diff
@@ -280,9 +278,14 @@ jobs:
...
         curl -LO "https://people.xiph.org/~tdaede/$NASM_VERSION-win64.zip"
         7z e -y "$NASM_VERSION-win64.zip" -o"C:\nasm"
-    - name: Set environment variables
+        echo "::add-path::C:\nasm"
+    - name: Set MSVC x86_64 linker path
+      shell: pwsh
       run: |
-        echo "::add-path::$Env:VS_PATH\$Env:COMPILER_SUBPATH;C:\nasm"
+        $LinkGlob = "VC\Tools\MSVC\**\bin\Hostx64\x64\link.exe"
+        Set-Alias vswhere "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
```
1480c1: Why not just add this to PATH temporarily and then call vswhere? seems a bit unnecessary to set an alias

.github/workflows/rav1e.yml
```diff
@@ -280,9 +278,14 @@ jobs:
         $NASM_VERSION="nasm-2.14.02"
         curl -LO "https://people.xiph.org/~tdaede/$NASM_VERSION-win64.zip"
         7z e -y "$NASM_VERSION-win64.zip" -o"C:\nasm"
-    - name: Set environment variables
+        echo "::add-path::C:\nasm"
+    - name: Set MSVC x86_64 linker path
+      shell: pwsh
       run: |
-        echo "::add-path::$Env:VS_PATH\$Env:COMPILER_SUBPATH;C:\nasm"
+        $LinkGlob = "VC\Tools\MSVC\**\bin\Hostx64\x64\link.exe"
```
1480c1: Also, what's the reasoning behind separating out the glob string as a variable? Seems like an unnecessary line, not that it affects time very much

barrbrain: This is for readability. The query command was getting quite long with the glob embedded.

1480c1: I see, seemed a bit weird to have a variable just for that since a lot of powershell commands can easily go over 100 characters anyway

.github/workflows/rav1e.yml
barrbrain: The path suggestion is certainly more terse. Whether it is clearer than setting an alias is less certain.
I would rather not add another network dependency for a binary that is installed in the base image, just to find another binary also in the base image.

1480c1: Main reason why I suggested it was because usually, if github actions is up, github is probably also up, and since my personal preference is to not to rely on pre installed stuff on a machine unless the location is explicitly stated. Plus, that way the build would easily be reproducible on my local machine just by running the same exact commands

barrbrain:
@1480c1 Thanks for the comments. I am happy with the resulting code.

.github/workflows/rav1e.yml
```diff
@@ -280,9 +278,14 @@ jobs:
         $NASM_VERSION="nasm-2.14.02"
         curl -LO "https://people.xiph.org/~tdaede/$NASM_VERSION-win64.zip"
         7z e -y "$NASM_VERSION-win64.zip" -o"C:\nasm"
-    - name: Set environment variables
+        echo "::add-path::C:\nasm"
+    - name: Set MSVC x86_64 linker path
+      shell: pwsh
```
Luni-4: Is it possible to use the `Desktop Poweshell` (default shell used on Windows for GA) instead of the `PowerShell Core`?

barrbrain: Yes, from the logs we see that the exact same shell binary is run without explicitly setting `shell: pwsh`.

Luni-4: Great! :)

PR #1969, opened 2019-12-18
Prepare for release 0.2.0

lu-zero:


shssoichiro:
I'd personally rather see the dependency updates in a separate PR.

lu-zero:
Done, I guess we are set.

PR #1971, opened 2019-12-19
CI: Add cargo-c tests

### LINKED ISSUE #1383: Add C API build to CI
$subj

Luni-4:
This PR closes #1383

### DIFF
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -54,6 +54,7 @@ jobs:
          - grcov-coveralls
          - bench
          - doc
+         - cargo-c
          - check-no-default
          - check-extra-feats
         include:
@@ -68,6 +69,8 @@ jobs:
          - conf: bench
            toolchain: stable
          - conf: doc
+           toolchain: stable
+         - conf: cargo-c
            toolchain: stable
          - conf: check-no-default
            toolchain: stable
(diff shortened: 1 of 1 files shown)


lu-zero:
Something went wrong with the cache server though.

.github/workflows/rav1e.yml
```diff
@@ -283,15 +302,19 @@ jobs:
     - name: Set environment variables
       run: |
         echo "::add-path::$Env:VS_PATH\$Env:COMPILER_SUBPATH;C:\nasm"
-    - name: Install Rust
+    - name: Install ${{ matrix.toolchain }}
```
lu-zero: Is this related?

.github/workflows/rav1e.yml
```diff
@@ -283,15 +302,19 @@ jobs:
...
       uses: actions-rs/toolchain@v1
       with:
         profile: minimal
-        toolchain: stable
+        toolchain: ${{ matrix.toolchain }}
```
lu-zero: Same question here

Luni-4:
> I like it, later would be great having an action to package the binary package for the windows users I guess.

Does the binary package added in the `deploy action` not enough? Or are you talking about a `chocolatey` package?

.github/workflows/rav1e.yml
Luni-4: Hmm, I don't think so, but I can investigate on it

.github/workflows/rav1e.yml
Luni-4: Same reply as above

lu-zero:
> Does the binary package added in the `deploy action` not enough? Or are you talking about a `chocolatey` package?

As in having on one side the `.exe`, on the other the `.dll`, `.h` and `.pc`.

Luni-4:
> As in having on one side the `.exe`, on the other the `.dll`, `.h` and `.pc`.

For now they are in two different `.zip` files, but I could create a single `.zip` file for `Windows` containing all the binaries. Do you like this idea?

PR #1972, opened 2019-12-19
Refactor BlockWxH out of predict intra

KyleSiefring:
Reduces binary size from 42.5 MB to 37.8 MB.
Removes 40% of the LLVM-IR currently being created.
Reduces --release compile times by 15 seconds (Ryzen 5 3600).

### DIFF
--- a/src/asm/aarch64/transform/inverse.rs
+++ b/src/asm/aarch64/transform/inverse.rs
@@ -109,7 +109,7 @@ macro_rules! impl_itx_fns {
     )*
     paste::item! {
       // Implement InvTxfm2D for WxH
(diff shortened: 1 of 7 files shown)


KyleSiefring:
Change the commit message,
```
There is a performance penalty (up to 4x slower), but it shouldn't
  impact arm or x86_64 since they have assembly for this code.
```

shssoichiro:
One consideration that is often forgotten when dismissing native code speed is that our HBD paths have no assembly. Is the speed trade-off significant enough that it would be a serious concern for HBD encoding? I assume 4x slower is localized to the specific function, not to the whole encode, correct?

PR #1974, opened 2019-12-19
Fix for serious bug in deblocking RDO sse calculation

xiphmont:
At some point a code cleanup of the deblocking filter seriously broke the deblocking SSE calculation in optimization. The horizontal edge calculation was actually looking at random vertical edges through the frame.  Some potential for panic even if testing up to this point hadn't tripped it.

Better yet, fixing it actually improves metrics.

deblock-fix-baseline@2019-12-19T16:41:23.981Z -> deblock-fix-result@2019-12-19T16:41:55.514Z

-0.1511 | -0.1740 | -0.2082 |  -0.0815 | -0.1449 | -0.1509 |    -0.1157

### DIFF
--- a/src/deblock.rs
+++ b/src/deblock.rs
@@ -333,11 +333,10 @@ fn copy_vertical<T: Pixel>(
   }
 }
 
-fn stride_sse<T: Pixel>(a: &[T], b: &[i32], pitch: usize) -> i64 {
+fn stride_sse(a: &[i32], b: &[i32]) -> i64 {
   let mut acc: i32 = 0;
-  for (a, b) in a.iter().step_by(pitch).take(b.len()).zip(b) {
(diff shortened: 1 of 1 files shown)


src/deblock.rs
```diff
@@ -1145,23 +1235,39 @@ fn sse_h_edge<T: Pixel>(
     if filter_size > 0 {
       let po = {
         let mut po = bo.plane_offset(&rec_plane.cfg); // rec and src have identical subsampling
-        po.x -= (filter_size >> 1) as isize;
+        po.y -= (filter_size >> 1) as isize;
```
ycho: I believe this spot has been a bug!

xiphmont:
Yes, but not the only one-- an even more serious one lurks in the stride
addressing.

PR #1975, opened 2019-12-19
vergen: Make the new clippy happy

lu-zero:


lu-zero:
There is more to come, moving to a whole new set.

PR #1977, opened 2019-12-20
Fix desync when using multiple segments

smarter:


src/encoder.rs
```diff
@@ -1110,23 +1110,28 @@ pub fn encode_tx_block<T: Pixel>(
...
   bsize: BlockSize,
   po: PlaneOffset,
   skip: bool,
+  qidx: u8,
   ac: &[i16],
   alpha: i16,
   rdo_type: RDOType,
   need_recon_pixel: bool,
 ) -> (bool, ScaledDistortion) {
-  let qidx = get_qidx(fi, ts, cw, tx_bo);
```
ycho: I have tested your code to check whether the qidx obtained here (before your fix) matches to the qidx passed at line 1129 (after your fix).

Like using either of below:
  //debug_assert!(qidx == qidx2);  // ==> Does not work well in rust native code level debug in my VS Code, so // using below
  if qidx != qidx2 {
    let _test = true;
  }

Then, I am seeing that it consistently mismatches at plane 1 and 2.
The desync starts from the first 1,1 block position (i.e. MI grid) of luma 4x4 block.
Before your fix, instead of passing qidx directly, "cw.bc.blocks.set_segmentation_idx(tile_bo, bsize, sidx);" is used to set (or rememeber) the seg_id at MI grid, but someho chroma blocks fail to retrieve that info, possibly because of using incorrect offset in block context array.

smarter: The problem is that this didn't use tile_partition_bo, see the commit message for details.

src/encoder.rs
```diff
@@ -1110,27 +1110,32 @@ pub fn encode_tx_block<T: Pixel>(
...
-  by: usize, // tx block position within a partition, unit: tx block number
-  tile_bo: TileBlockOffset,
+  // tx block position within a partition, unit: tx block number
+  bx: usize, by: usize,
+  // Offset in the luma plane where this tx block is colocated. Note that for
+  // a chroma block, this offset might be outside of the current partition.
+  // For example in 4:2:0, four 4x4 luma partitions share one 4x4 chroma block,
+  // this block is part of the last 4x4 partition, but its `tx_bo` offset
+  // matches the offset of the first 4x4 partition.
+  tx_bo: TileBlockOffset,
```
ycho: Yes, this is really correct comment! The correct seg_id used by chroma 4x4 block is the seg_id for last luma 4x4.

src/encoder.rs
```diff
@@ -1110,27 +1110,32 @@ pub fn encode_tx_block<T: Pixel>(
...
   bsize: BlockSize,
   po: PlaneOffset,
   skip: bool,
+  qidx: u8,
   ac: &[i16],
   alpha: i16,
   rdo_type: RDOType,
   need_recon_pixel: bool,
 ) -> (bool, ScaledDistortion) {
-  let qidx = get_qidx(fi, ts, cw, tile_bo);
```
ycho: So, for chroma block with sub8x8 partition, instead of tile_bo it should be the last sub8x8 luma block.

src/encoder.rs
ycho: > The problem is that this didn't use tile_partition_bo, see the commit message for details.

As you exactly stated, I realized that using the existing parameter 'tile_partition_bo' to get qidx, i.e. doing: let qidx = get_qidx(fi, ts, cw, tile_partition_bo); fixes it as well.
However, I believe you didn't like fetching seg_id from block context and wanted passing qidx (since it is already available from caller functions, write_tx_blocks() and write_tx_tree()).

As in 
https://github.com/ycho/rav1e/commit/a8f10afdc9ab9221385906b8a8d205d11202a358

PR #1979, opened 2019-12-20
Use fern and console instead of pretty_env_logger

lu-zero:
It reduces significantly the build time and allows better output customization.

```
>  Using y4m decoder: 176x144p @ 30/1 fps, 4:2:0, 8-bit
>  Encoding settings: keyint_min=12 keyint_max=240 quantizer=100 bitrate=0 min_quantizer=0 low_latency=false tune=Psychovisual rdo_lookahead_frames=40 min_block_size=64x64 multiref=true fast_deblock=true reduced_tx_set=true tx_domain_distortion=false tx_domain_rate=false encode_bottomup=false rdo_tx_decision=false prediction_modes=Simple include_near_mvs=false no_scene_detection=false diamond_me=true cdef=true use_satd_subpel=false non_square_partition=false enable_timing_info=false
>  CPU Feature Level: AVX2
>  Using 6 tiles (2x3)
?  debug message
!  warning message 
!! error message
>  encoded 600 frames, 80.182 fps, 55.64 Kb/s
>  ----------
>  Key frame:             3 | avg QP:  79.00 | avg size:    2822 B
>  Inter frame:         597 | avg QP: 134.00 | avg size:     218 B
>  Intra only frame:      0 | avg QP:   0.00 | avg size:       0 B
>  Switching frame:       0 | avg QP:   0.00 | avg size:       0 B
```

TODO
- [ ] document the env var to override the logging level
- [ ] provide verboseness levels (from quiet to verbose)

### DIFF
--- a/src/bin/rav1e.rs
+++ b/src/bin/rav1e.rs
@@ -38,7 +38,6 @@ use rav1e::prelude::*;
 use crate::decoder::Decoder;
 use crate::decoder::VideoDetails;
 use crate::muxer::*;
-use std::env;
 use std::fs::File;
 use std::io::Read;
 use std::io::Seek;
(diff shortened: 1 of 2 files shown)


src/bin/rav1e.rs
```diff
@@ -259,10 +258,46 @@ fn main() {
...
+    // set the default log level. to filter out verbose log messages from dependencies, set
+    // this to Warn and overwrite the log level for your crate.
+    .level(log::LevelFilter::Warn)
+    // change log levels for individual modules. Note: This looks for the record's target
+    // field which defaults to the module path but can be overwritten with the `target`
+    // parameter:
+    // `info!(target="special_target", "This log message is about special_target");`
+    .level_for("rav1e", level)
+    // output to stdout
+    .chain(std::io::stdout())
```
shssoichiro: Probably we want to output to stderr instead.

lu-zero: Right!

lu-zero:
In numbers:
2b87fcff
```
real	2m3.306s
user	14m29.203s
sys	0m21.236s
```
This set
```
real	1m49.413s
user	13m18.152s
sys	0m19.093s
```

PR #1980, opened 2019-12-20
Use the same pool for send_frame

lu-zero:


shssoichiro:
I assume pool = thread pool. Does this mean that send_frame is now running parallel to receive_frame?

lu-zero:
No, it means that before this patch send_frame is using the default threadpool instead of the rav1e's one. This is just a bugfix =/

PR #1982, opened 2019-12-20
Refactor forward transforms to remove buffer weirdness

KyleSiefring:


tdaede:
"Split forward transform buffers into separate input and output buffers" is maybe more descriptive than buffer weirdness :)

PR #1984, opened 2019-12-21
Elide bounds checks in cdef_dist_wxh_8x8 inner loop

shssoichiro:
Although rustc still fails to autovectorize this function,
eliding the extra bounds check branches
gains 2.5% encode speed at speed 6
due to how hot this function is.


lu-zero:
Did you try to force inline it in the wxh variant?

lu-zero:
Looks like that's enough to make the autovectorizer do its job.

PR #1988, opened 2019-12-22
 Reduce float math in cdef_dist_wxh_8x8

shssoichiro:
Also creates unit benchmark for function.

Improves speed of function by ~1%

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -160,13 +160,13 @@ fn cdef_dist_wxh_8x8<T: Pixel>(
       sum_sd += (s * d) as i64;
     }
   }
-  let svar = (sum_s2 - ((sum_s as i64 * sum_s as i64 + 32) >> 6)) as f64;
-  let dvar = (sum_d2 - ((sum_d as i64 * sum_d as i64 + 32) >> 6)) as f64;
+  let svar = sum_s2 - ((sum_s as i64 * sum_s as i64 + 32) >> 6);
+  let dvar = sum_d2 - ((sum_d as i64 * sum_d as i64 + 32) >> 6);
   let sse = (sum_d2 + sum_s2 - 2 * sum_sd) as f64;
   //The two constants were tuned for CDEF, but can probably be better tuned for use in general RDO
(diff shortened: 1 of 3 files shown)


src/rdo.rs
```diff
@@ -160,13 +160,13 @@ fn cdef_dist_wxh_8x8<T: Pixel>(
...
-  let dvar = (sum_d2 - ((sum_d as i64 * sum_d as i64 + 32) >> 6)) as f64;
+  let svar = sum_s2 - ((sum_s as i64 * sum_s as i64 + 32) >> 6);
+  let dvar = sum_d2 - ((sum_d as i64 * sum_d as i64 + 32) >> 6);
   let sse = (sum_d2 + sum_s2 - 2 * sum_sd) as f64;
   //The two constants were tuned for CDEF, but can probably be better tuned for use in general RDO
   let ssim_boost = (4033_f64 / 16_384_f64)
-    * (svar + dvar + (16_384 << (2 * coeff_shift)) as f64)
-    / f64::sqrt((16_265_089u64 << (4 * coeff_shift)) as f64 + svar * dvar);
+    * (svar + dvar + (16_384 << (2 * coeff_shift))) as f64
+    / f64::sqrt(((16_265_089i64 << (4 * coeff_shift)) + svar * dvar) as f64);
```
tdaede: You're very close to no fp at all - if you just shift up ssim_boost and then shift it back down after the multiply below, this could be all u64.

lu-zero: While at it, maybe try to break this in some intermediates. I wonder if we have direct access to the reciprocal of square root or  if it is already doing `.sqrt().recip()` under the hood.

shssoichiro: No `rsqrt` found in the ASM. I also think my math skills are too stupid to figure out how to remove all float usage here. 😆

tdaede: We can land this incremental improvement then.

shssoichiro: I'd also be +1 for splitting out the incremental steps, but not sure what each step stands for. I was unable to find documentation on the formula used here.

PR #1997, opened 2019-12-27
Designate input coeff pointers as mutable on inverse transforms

KyleSiefring:
The inverse transform assembly implementation uses the input
coefficients as scratch to avoid stack allocations. Properly designate
the coefficent input pointers as mutable on the inverse transforms
assembly function definitions for the inverse transforms.


tdaede:
Do we have documentation anywhere that the itx functions trash the input?

PR #1998, opened 2019-12-27
Make it easier for cdef dist to auto-vectorize

KyleSiefring:
Cdef dist should autovectorize with default compiler flags now.

Since I don't see a performance benifit from it anymore, remove the forced
inlining.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -43,6 +43,7 @@ use crate::{encode_block_post_cdef, encode_block_pre_cdef};
 use crate::cpu_features::CpuFeatureLevel;
 use crate::partition::PartitionType::*;
 use arrayvec::*;
+use itertools::izip;
 use std;
 use std::vec::Vec;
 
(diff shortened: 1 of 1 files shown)


lu-zero:
If you enable avx2 what does it happen?

KyleSiefring:
Honestly, I couldn't seem to get any vectorized code on master with RUSTFLAGS="-C target-feature=+avx2,+fma". When I tested removing the inlining, I ran with those options. I was thinking of leaving it in if there were speedups with avx2.

shssoichiro:
Looks [optimized and unrolled](https://pastebin.com/v8UEVq4N) although only using SSE2 opts even with AVX2 enabled, but our microbenchmark on my machine shows that this version is 20% slower.

KyleSiefring:
I ran the microbenchmarks and did reproduce the a performance regression, but in the order of a couple percent instead.

Here are the awcy results. I do see performance improvements here.
https://beta.arewecompressedyet.com/?job=master-0a1335fb86686808df86654938fae420c77b4660&job=auto_vec_cdef_dist_2%402019-12-27T17%3A23%3A29.539Z

I guess more investigation is warranted.

shssoichiro:
I wonder if it depends on CPU architecture. My current machine is running an i7-6700.

KyleSiefring:
Try it now.

shssoichiro:
This version is 50% faster compared to master, strangely enough. Running a full encode, 1080p ducks take off, encode speed is about 3% faster at speed 6.

PR #2001, opened 2019-12-29
Tie coefficient type to pixel type

KyleSiefring:
I didn't see any speedup the last time I tested.


src/asm/aarch64/transform/mod.rs
```diff
@@ -7,4 +7,4 @@
 // Media Patent License 1.0 was not distributed with this source code in the
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
-pub mod inverse;
+//pub mod inverse;
```
lu-zero: Why this?

KyleSiefring: I missed it when reenabling the forward and inverse transforms.

PR #2007, opened 2019-12-30
Fix eob decision/computation in quantizer

KyleSiefring:
Fix a off by one error in the eob calculation. We were treating eob
as if it didn't include the dc value except for one place. When we
found no ac values below our deadzone, we would try to quantize one ac
value.

Rework quantize so that eob is related both dc and ac values. We can
now return the precalculated eob.

Add a debug_assert for the eob.

### DIFF
--- a/src/quantize.rs
+++ b/src/quantize.rs
@@ -246,21 +246,6 @@ impl QuantizationContext {
   {
     let scan = av1_scan_orders[tx_size as usize][tx_type as usize].scan;
 
-    // Find the last non-zero coefficient using our smaller biases and
-    // zero everything else.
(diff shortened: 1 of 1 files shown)


KyleSiefring:
https://beta.arewecompressedyet.com/?job=master-b6952b9a664606aa4d62088354c8a215f8ee3787&job=eob_precalculation1%402019-12-30T18%3A57%3A24.391Z

```
-0.1420 |  1.0591 |  0.9424 |  -0.0691 | -0.1227 | -0.0231 |     0.1238
```

KyleSiefring:
<derf> maseck: If no one else has looked at 2007, it LGTM.

PR #2010, opened 2019-12-31
Refactor inverse transform tests for asm

KyleSiefring:
Refactor so that the majority of the work happens outside of the
macros. Also, share code between the arm and x86 versions.


KyleSiefring:
Now to wait to see if I borked the arm version.

PR #2011, opened 2020-01-01
For speed 10, reduce self guided search space

KyleSiefring:
Search half as many sgr parameters for speed 10.


xiphmont:
There are smarter ways to reduce the search, but there's nothing actually wrong with this one.  LGTM.

PR #2013, opened 2020-01-02
Stop using Dimension traits for inverse transform

KyleSiefring:


lu-zero:
What is the impact overall ?

Also we might rediscuss how we dispatch the arch-specific functions sooner or later.

KyleSiefring:
It shaved a few seconds (1% off a clean build) off of compile times and doesn't seem to impact runtime.

lu-zero:
It should reduce the binary size as well.

KyleSiefring:
30.4MB to 29.4MB. not bad.

shssoichiro:
Unrelated to this PR: For release binaries we should run `strip` also. Since we build with debug symbols, that can reduce the binary size by 90%. (Even with default settings for Cargo, strip still gives 50% binary size reduction.)

lu-zero:
I assumed we split the debug information in the action already. I guess might be good to document it.

KyleSiefring:
I went and measured performance without the inverse assembly and there doesn't seem to be a performance impact. I changed the commit message to include this info.

lu-zero:
Thank you :)

PR #2014, opened 2020-01-02
SSE2 Optimizations for sse_wxh function

shssoichiro:
We can use SSE2 to optimize 8x8 block sse_wxh computations,
for both low-bit-depth and high-bit-depth.

This commit also includes minor improvements
to the native version of the function
to reduce redundant computations.

Microbenchmarks show a 60-65% improvement for 8x8 block sizes,
and a 3-25% improvement for smaller block sizes.

Image 2020-01-02 at 3 17 39 PM

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -234,7 +234,7 @@ pub fn cdef_dist_wxh<T: Pixel, F: Fn(Area, BlockSize) -> f64>(
 // Sum of Squared Error for a wxh block
 pub fn sse_wxh<T: Pixel, F: Fn(Area, BlockSize) -> f64>(
   src1: &PlaneRegion<'_, T>, src2: &PlaneRegion<'_, T>, w: usize, h: usize,
-  compute_bias: F,
+  compute_bias: F, _cpu: CpuFeatureLevel,
 ) -> Distortion {
   assert!(w & (MI_SIZE - 1) == 0);
   assert!(h & (MI_SIZE - 1) == 0);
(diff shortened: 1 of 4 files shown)


shssoichiro:
Running AWCY to verify no change in output and observe overall encode time impact.

shssoichiro:
I also broke ARM... I'll fix that.

shssoichiro:
[AWCY](https://beta.arewecompressedyet.com/?job=master-ecdc72996afa15bd94dd075ebb92eab384fda88e&job=sse_whx_optim%402020-01-02T20%3A26%3A01.077Z) shows no change in output but also no encoding speed change. I wonder if AWCY is not using SIMD optimizations? `sse_wxh` was one of the highest self time usages according to uftrace.

I'll dig a bit more later this evening to try and get a measure of overall encode time improvement.

shssoichiro:
Unlucky. Trashing this because it turns out that in practice, we never use 8x8 blocks in this function except for tune = Psnr.

tdaede:
Do we use it for tx domain dist + tune psnr? As that's still an important configuration.

shssoichiro:
I can poke at it some more and reopen when ready. Interestingly, I found a huge performance loss when running locally with default settings, probably branch misprediction kicking in badly.

shssoichiro:
I have absolutely 0 confidence in whole-encode benchmarks being accurate on my laptop...

src/rdo.rs
```diff
@@ -243,20 +243,36 @@ pub fn sse_wxh<T: Pixel, F: Fn(Area, BlockSize) -> f64>(
   // importance block size in a non-subsampled plane.
   let imp_block_w = IMPORTANCE_BLOCK_SIZE.min(w);
   let imp_block_h = IMPORTANCE_BLOCK_SIZE.min(h);
-  let imp_bsize = BlockSize::from_width_and_height(imp_block_w, imp_block_h);
   let block_w = imp_block_w >> src1.plane_cfg.xdec;
   let block_h = imp_block_h >> src1.plane_cfg.ydec;
 
+  #[cfg(nasm_x86_64)]
```
lu-zero:
``` rust
#[cfg(target_arch = "x86_64")]
```

PR #2015, opened 2020-01-02
x86: Fix inverse (flip)adst 8x4 clipping

KyleSiefring:


shssoichiro:
Do we need to make any Rust changes to accommodate this in the native code? The MR in dav1d had an associated C-code change: https://code.videolan.org/videolan/dav1d/merge_requests/879/diffs

KyleSiefring:
We're good.

dav1d had these negatives inside https://github.com/xiph/rav1e/blob/master/src/transform/inverse.rs#L527 of these clamps https://github.com/xiph/rav1e/blob/master/src/transform/inverse.rs#L487.

PR #2016, opened 2020-01-03
quantize: initially zero coefficients

barrbrain:
Effectively replaces a hot branchy loop with memset.
Reduces encoding time by 1.8% to 5% at default speed preset.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1197,8 +1197,8 @@ pub fn encode_tx_block<T: Pixel>(
     AlignedArray::uninitialized();
   let mut coeffs_storage: AlignedArray<[T::Coeff; 64 * 64]> =
     AlignedArray::uninitialized();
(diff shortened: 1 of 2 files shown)


tdaede:
We currently allocate qcoeffs with an Uninitialized AlignedArray. Maybe it would make sense to allocate it as initialized? That might zero more than this does though.

barrbrain:
> We currently allocate qcoeffs with an Uninitialized AlignedArray. Maybe it would make sense to allocate it as initialized? That might zero more than this does though.

After some thought, we could meet in the middle with a helper to go from uninitialized AllignedArray of constant size to initalized slice of runtime-determined length. Would also need to document that `quantize` assumes zeros in destination slice.

KyleSiefring:
There is a `AlignedArray::new()` that can be used like so `AlignedArray::new([0; length]);`

PR #2020, opened 2020-01-04
Add avx2 for dequantization

KyleSiefring:
Improves performance by ~1%

### DIFF
--- a/src/asm/x86/quantize.rs
+++ b/src/asm/x86/quantize.rs
@@ -1,0 +1,214 @@
+// Copyright (c) 2019, The rav1e contributors. All rights reserved
+//
+// This source code is subject to the terms of the BSD 2 Clause License and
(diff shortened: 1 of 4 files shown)


src/asm/x86/quantize.rs
```diff
@@ -0,0 +1,214 @@
...
+  }
+}
+
+#[cfg(test)]
+mod test {
+  use super::*;
+  use rand::{thread_rng, Rng};
+
+  #[test]
+  fn dequantize_test() {
```
lu-zero: This should have an effect only when check_asm is enabled, right?

KyleSiefring: No, I enabled to results checking code to be enabled either for check_asm or test.

lu-zero: Then I guess it is good to go :)

PR #2024, opened 2020-01-06
Start to safely partially initialize slices

KyleSiefring:
Currently we are using MaybeUninit::uninit().assume_init() is an
undefined way. This patch uses MaybeUninit::uninit().assume_init() to
create arrays of MaybeUninit<T> which is defined behavior. Slices of
these arrays have their elements set and the slices of initialized
elements are transmuted to a slice of T.

One instance of full array initialization is replaced and one instance
of undefined behaviour is replaced.

### DIFF
--- a/src/util/uninit.rs
+++ b/src/util/uninit.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2019, The rav1e contributors. All rights reserved
+// Copyright (c) 2019, The rav1e contributors. All rights reserved
 //
 // This source code is subject to the terms of the BSD 2 Clause License and
 // the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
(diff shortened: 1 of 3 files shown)


KyleSiefring:
yeah, clippy's suggestion works.

src/util/uninit.rs
```diff
@@ -0,0 +1,22 @@
...
+// This source code is subject to the terms of the BSD 2 Clause License and
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+use std::mem::MaybeUninit;
+
+pub fn init_slice_repeat_mut<T: Copy>(
```
lu-zero:
``` rust
trait InitMaybeSlice<T: Copy> {
   fn init_mut(&mut self, value: T) -> &mut [T];
}

impl<T: Copy> InitMaybeSlice<T> for [MaybeUninit<T>] {
    fn init_mut(&mut self, value: T) -> &mut [T] {
     // the code you wrote
    }
}
```
Is a bit more ergonomic IMHO.

I can make a PR on top of yours if you prefer.

KyleSiefring: I considered it, but it felt a little odd writing (&mut storage.array[..len]).init_mut(0).

KyleSiefring: I prefer this but, wouldn't you need const generics or some other nightly feature for that?

lu-zero: You are right, I guess we can go with your version and put a todo for once const_generics are available in stable.

PR #2032, opened 2020-01-07
Revert angle dir intrapred

ycho:
Because enabling this new coding modes, i.e. encode fine angle of directional intra prediction, causes desync at speed 0,1,2, revert the related commits.


ycho:
I want, but somehow I cannot include the author (takehirokj) of the commits that I try to revert as reviewer.

lu-zero:
He needs to be added in the team I guess, @tdaede ?

Issue #2025, opened 2020-01-06
Desync at speed 0,1,2 for 1 frame


ycho:
commit e89833ec45e8e412221f0e716afa79083588a64b

./target/release/rav1e nyan.y4m -o test.ivf -r test_rec.y4m --quantizer 50 --speed=0 --limit=1

test_rec.yuv test_dec.yuv differ: char 49469, line 4

Change --speed=0 to 1,2 and there is a desync at speed 1 and 2 as well.

With the above run, there is no desync for speeds 3~10.

lu-zero:
Why the tests didn't pick it? What introduced it?

ycho:
> Why the tests didn't pick it? What introduced it?

I think the desync depends on input, and the synthesized pixels luckily or easiiy pass our built-in tests.

From my quick probe the commit ecdc72996afa15bd94dd075ebb92eab384fda88e seems the origin of desync.

takehirokj:
I found there's case (alpha == 0) || (angle_delta == 0) is false in predict_intra().
This might cause desync as commented in #2023.
I will fix to ensure the condition above.

ycho:
> I found there's case (alpha == 0) || (angle_delta == 0) is false in predict_intra().
> This might cause desync as commented in #2023.
> I will fix to ensure the condition above.

@takehirokj  @barrbrain 
Hi, I wanted to say that I really appreciate your efforts and time for adding the missing coding tools of AV1!
As I stated above, it seems to me that ecdc72996afa15bd94dd075ebb92eab384fda88e is the origin of the desync.  My further test shows that changing min block size to 8x8 from 4x4 (via min_block_size_preset( ) ) stops desync for the test cases I used above, i.e. 1 frame of nyan.y4m at qidx 50 only for speed 0,1,2. Unfortunately, if the cause is related to partition size, esp. sub8x8, it signals me that the fix or workaround can take some more extra works.
So, I am sorry but I need to propose to disable this feature by reverting all related commits.

ycho:
@lu-zero @tdaede  ^

ycho:
Instead of revert, another option is #2033, if you prefer.

shssoichiro:
I prefer #2033 over reverting. Personal opinion only.

ycho:
> I prefer #2033 over reverting. Personal opinion only.

Hope you like #2034 better!

ycho:
[...]

PR #2036, opened 2020-01-07
Fix desync on sub8x8 partitions caused by fine dir intrapred

ycho:
Angle delta for directional intra prediction is only available for block
size >= 8x8 in AV1.


ycho:
Fix for #2025

PR #2040, opened 2020-01-08
Add additional hawktracer probes for filters.

tdaede:


src/lrf.rs
```diff
@@ -1434,6 +1435,7 @@ impl RestorationState {
     }
   }
 
+  #[hawktracer(lrf_filter_Frame)]
```
lu-zero:
```suggestion
  #[hawktracer(lrf_filter_frame)]
```

PR #2041, opened 2020-01-08
Use flat arrays in stats collection.

tdaede:
Reduces time taken for stats merging with 16 tiles from 575us
to 224us on my machine.

### DIFF
--- a/src/bin/stats.rs
+++ b/src/bin/stats.rs
@@ -179,7 +179,7 @@ impl ProgressInfo {
       .frame_info
       .iter()
       .filter(|frame| frame.frame_type == frame_type)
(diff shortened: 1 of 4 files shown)


tdaede:
Looking for tips to make this faster. It's still too slow, so for realtime I'll still need to add a flag to disable stats collection. But it would be nice if I didn't need the flag.

shssoichiro:
I should have realized these each had a known, finite number of options and done it like this in the first place. Nice.

As far as disabling, we only output the stats if the `verbose` flag is enabled, but maybe disabling stats collection using that flag is tying the encoder too closely to the CLI?

tdaede:
I mean, we can have an API for compute stats, and the CLI can use that API based on the verbose flag.

PR #2042, opened 2020-01-09
Disable activity mask computation until it is used.

tdaede:


src/activity.rs
```diff
@@ -72,6 +74,10 @@ impl ActivityMask {
     ActivityMask { variances, width, height, granularity }
   }
 
+  pub fn empty() -> ActivityMask {
+    ActivityMask { variances: [].to_vec(), width: 0, height: 0, granularity: 0}
+  }
+
```
barrbrain: `ActivityMask` derives `Default` which is equivalent to this.

src/api/internal.rs
```diff
@@ -1136,8 +1136,10 @@ impl<T: Pixel> ContextInner<T> {
         frame_data.fi.set_quantizers(&qps);
       }
 
+      // TODO: replace with ActivityMask::from_plane() when
+      // the activity mask is actually used.
       frame_data.fi.activity_mask =
-        ActivityMask::from_plane(&frame_data.fs.input.planes[0]);
+        ActivityMask::empty();
```
barrbrain:
```suggestion
        Default::default();
```

PR #2044, opened 2020-01-09
Use explicit block size checking

ycho:
Instead of enum type based comparison, which become clearly wrong once
1:4 ratio nonsquare partitions are introduced or any block sizes after
BLOCK_64X64 in enum type.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1649,7 +1649,7 @@ impl<'a> BlockContext<'a> {
     &mut self, bo: TileBlockOffset, bsize: BlockSize, xdec: usize, ydec: usize,
   ) {
     const num_planes: usize = 3;
-    let nplanes = if bsize >= BLOCK_8X8 {
+    let nplanes = if bsize.gte(BLOCK_8X8) {
       3
     } else {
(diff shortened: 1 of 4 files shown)


ycho:
> We should implement Ord correctly, then the operators will do the right thing.

The enum type for blocksizes and several others are based on libaom and 

> We should implement Ord correctly, then the operators will do the right thing.

Yes, if there is no dependency to av1 spec, that is a cure.
However, as stated in https://aomediacodec.github.io/av1-spec/#decode-partition-semantics, the order # assigned to each enum entry is defined in spec (that is still fine) and very likely the context index (which fetches the cdf) is based upon this. Similar cares are done by @barrbrain in https://github.com/xiph/rav1e/pull/2048#pullrequestreview-340514926,

tdaede:
Alternately, you should be able to remove Ord and PartialOrd traits from the BlockSize struct, now that we never use those comparisons anywhere.

lu-zero:
is this outdated now?

ycho:
> is this outdated now?

Yes, I think we can drop this PR, thanks to barrbrain!
And I see that similar change would be required for tx sizes.

PR #2046, opened 2020-01-09
Move avx2 fwd tx impl into non-duplicated fn

KyleSiefring:
Moves the avx2 forward transform code out of the Dim trait
implementation and into a seperate frunction that does not get
duplicated. We have already done this with the native code, so the next
step will be removing the Dim trait from the fwd transforms entirely.

In theory, there should be some performance penalty for doing this.
While there were no large differences in performance evident from
a quick look peek into perf, some minor changes were made to mitigate
any potential problems. Branches were hoisted outside of for loops and
changed into match statements.

Compile time and binary size:

Before
2m11.366s
28.9 MiB

After
2m8.374s
27.7 MiB

### DIFF
--- a/src/asm/x86/transform/forward.rs
+++ b/src/asm/x86/transform/forward.rs
@@ -296,81 +296,89 @@ unsafe fn round_shift_array_avx2(arr: &mut [I32X8], size: usize, bit: i8) {
   }
 }
 
-pub trait FwdTxfm2D: native::FwdTxfm2D {
-  fn fwd_txfm2d_daala<T: Coefficient>(
-    input: &[i16], output: &mut [T], stride: usize, tx_type: TxType,
-    bd: usize, cpu: CpuFeatureLevel,
-  ) {
-    if cpu >= CpuFeatureLevel::AVX2 {
-      return unsafe {
-        Self::fwd_txfm2d_daala_avx2(input, output, stride, tx_type, bd);
-      };
+/// For classifying the number of rows and columns in a transform. Used to
+/// select the operations to perform for different vector lengths.
+#[derive(Debug, Clone, Copy)]
+enum SizeClass1D {
+  X4,
+  X8UP,
+}
+
+impl SizeClass1D {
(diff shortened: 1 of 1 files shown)


src/asm/x86/transform/forward.rs
```diff
@@ -296,81 +296,89 @@ unsafe fn round_shift_array_avx2(arr: &mut [I32X8], size: usize, bit: i8) {
...
-      txfm_func_col(tx_in, tx_out);
+    txfm_func_col(tx_in, tx_out);
 
-      round_shift_array_avx2(tx_out, txfm_size_row, -cfg.shift[1]);
+    round_shift_array_avx2(tx_out, txfm_size_row, -cfg.shift[1]);
 
-      for rg in (0..txfm_size_row).step_by(8) {
-        if txfm_size_row >= 8 && txfm_size_col >= 8 {
+    // Transpose the array. Select the appropriate method to do so.
+    match (row_class, col_class) {
```
lu-zero: You may move this further up, you may use hawktrace to compare before and after.

src/asm/x86/transform/forward.rs
```diff
@@ -481,4 +501,28 @@ pub trait FwdTxfm2D: native::FwdTxfm2D {
...
+        fwd_txfm2d_avx2(
+          input,
+          output,
+          stride,
+          TxSize::by_dims(Self::W, Self::H),
+          tx_type,
+          bd,
+        );
+      };
+    }
```
lu-zero:
``` rust
else {
...
```
please

lu-zero:
Also please remove the typos from the commit message while at it ^^

src/asm/x86/transform/forward.rs
KyleSiefring: tx_out is overwritten in each loop iteration.

lu-zero: I wanted move the transposes out and set the right transpose outside the loop. I can do myself later :)

PR #2049, opened 2020-01-09
Use eob in inverse transforms

KyleSiefring:
The inverse transforms from dav1d have eob optimizations. It is as
simple as providing those functions with the needed parameters.

Very small performance improvement. Measures in the scale of tenths of a
percent.


shssoichiro:
Looks good aside from rustfmt

PR #2050, opened 2020-01-10
Don't initialize planes used for downsampling.

tdaede:
Negligible speed impact.

### DIFF
--- a/benches/plane.rs
+++ b/benches/plane.rs
@@ -1,7 +1,7 @@
 use criterion::*;
 use rand::{Rng, SeedableRng};
 use rand_chacha::ChaChaRng;
-use rav1e::prelude::Plane;
+use rav1e::bench::frame::*;
 
 fn init_plane_u8(width: usize, height: usize) -> Plane<u8> {
   let mut ra = ChaChaRng::from_seed([0; 32]);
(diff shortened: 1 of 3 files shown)


shssoichiro:
I forgot to PR the benchmarks I wrote for this function last night, but it's showing a 60% regression for downsampling on 8-bit input. Interestingly, 10-bit showed no such regression.

```
downsample_8bit         time:   [1.4270 ms 1.4326 ms 1.4391 ms]                             
                        change: [+57.070% +59.069% +61.065%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 14 outliers among 100 measurements (14.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  11 (11.00%) high severe

downsample_10bit        time:   [922.60 us 929.11 us 935.82 us]                             
                        change: [+1.5419% +2.8097% +3.9266%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe
```

tdaede:
So this moves the pad into downsample_from, because I wanted it to always return a fully initialized plane. You'll need to move the pad back out to get a fair comparison.

tdaede:
Oh, it also moves the allocation into the function, so you'll have to undo that too. Or just change new_uninitialized to new.

shssoichiro:
Okay, matching up the internals, `new_uninitialized` seems to be about 2% slower for 8-bit and 10% faster for 10-bit. For some reason.

shssoichiro:
Feel free to merge whenever you're ready

PR #2052, opened 2020-01-10
Add benchmarks for plane downsample

shssoichiro:


tdaede:
(drop the merge commit btw)

PR #2053, opened 2020-01-10
WIP: Do both downsamples in the same loop.

tdaede:


src/frame/plane.rs
```diff
@@ -437,6 +456,62 @@ impl<T: Pixel> Plane<T> {
...
+          }
+        }
+        let q_avg = (h_avg[0] + h_avg[1] + h_avg[2] + h_avg[3] + 2) >> 2;
+        let base = (yorigin + q_row) * stride + xorigin;
+        let dst = &mut qres.data[base..base + width];
+        dst[q_col] = T::cast_from(q_avg);
+      }
+    }
+    hres.pad(frame_width, frame_height);
+    hres.pad(frame_width, frame_height);
```
KyleSiefring: qres

PR #2061, opened 2020-01-10
Implement Ord and PartialOrd for BlockSize

lu-zero:
Here is what I had in mind.


shssoichiro:
Thinking this can be closed because #2076 has been merged instead?

lu-zero:
yes :)

ycho:
> Thinking this can be closed because #2076 has been merged instead?

Sure, thanks for headup and let me close it now.

PR #2063, opened 2020-01-10
Add plane downsampling test.

tdaede:


src/frame/plane.rs
```diff
@@ -707,9 +707,9 @@ pub mod test {
   }
 
   #[test]
-  fn test_plane_pad() {
```
KyleSiefring: It's your intention to get rid of the test_plane_pad?

tdaede: whoops no, I'll restore it

PR #2065, opened 2020-01-11
Use correct dimension to get x and y coeff in test

KyleSiefring:
I thought I had to reverse what dav1d did. I didn't have to reverse it.

### DIFF
--- a/src/asm/shared/transform/inverse.rs
+++ b/src/asm/shared/transform/inverse.rs
@@ -54,7 +54,7 @@ pub mod test {
   use rand::{random, thread_rng, Rng};
 
   pub fn pick_eob<T: Coefficient>(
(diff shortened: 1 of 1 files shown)


src/asm/shared/transform/inverse.rs
```diff
@@ -54,7 +54,7 @@ pub mod test {
   use rand::{random, thread_rng, Rng};
 
   pub fn pick_eob<T: Coefficient>(
-    coeffs: &mut [T], tx_size: TxSize, tx_type: TxType, sub_w: usize,
+    coeffs: &mut [T], tx_size: TxSize, tx_type: TxType, sub_h: usize,
```
takehirokj: Do we also need to change the sub_w in L136 which calls this function?

KyleSiefring: I should change its name, so yes.

PR #2066, opened 2020-01-11
Unify angle_delta and alpha to enum

takehirokj:
Only one of angle_delta or alpha is used for predict_intra().
This ensures that (alpha == 0) || (angle_delta == 0) which avoids invalid calculation of angle.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -22,7 +22,7 @@ use crate::me::*;
 use crate::partition::PartitionType::*;
 use crate::partition::RefType::*;
 use crate::partition::*;
-use crate::predict::{AngleDelta, PredictionMode};
+use crate::predict::{AngleDelta, IntraParam, PredictionMode};
 use crate::quantize::*;
 use crate::rate::bexp64;
 use crate::rate::q57;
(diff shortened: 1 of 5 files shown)


takehirokj:
AWCY shows no difference in the output and slight speed up in encoding time(~1%).
[master -> unify_alpha_angle_delta](https://beta.arewecompressedyet.com/?job=unify_alpha_angle_delta%402020-01-11T06%3A17%3A54.132Z&job=master-5929763a0d90097fcf3077c0cf1826d95f9f266b)

src/encoder.rs
```diff
@@ -1971,8 +1968,11 @@ pub fn write_tx_blocks<T: Pixel>(
             skip,
             qidx,
             &ac.array,
-            angle_delta.uv,
-            alpha,
+            if angle_delta.uv != 0 {
```
barrbrain: The condition `chroma_mode.is_cfl()` matches the choice here, with `IntraParam::Alpha(alpha)` as the positive case.

src/predict.rs
```diff
@@ -146,6 +145,15 @@ impl PredictionMode {
...
 
+    let alpha = match intra_param {
+      IntraParam::Alpha(val) => val,
+      _ => 0,
+    };
+    let angle_delta = match intra_param {
+      IntraParam::Angle_delta(val) => val,
+      _ => 0,
+    };
+
```
barrbrain: These matches could be merged with the match at line 168. It may not be obvious how to cover all cases correctly, so we can take care of this in a future pull request if necessary.

src/encoder.rs
takehirokj: Updated by using chroma_mode.is_cfl().

src/predict.rs
takehirokj: Yes. It will be improved in future PR.

PR #2067, opened 2020-01-11
Switch mutable slice with mutable region in sgr

KyleSiefring:
Mutable regions should be preferred over mutable slices.


lu-zero:
You might also write down why :)

PR #2076, opened 2020-01-13
 Implement PartialOrd for BlockSize

barrbrain:
Drop derivation of Ord, there is not a total order. Remove greater_than() and gte() from BlockSize; replace with the equivalent operators. Alternative to #2061.

### DIFF
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -338,16 +338,6 @@ impl BlockSize {
     (offset_x, offset_y)
   }
 
-  pub fn greater_than(self, other: BlockSize) -> bool {
-    (self.width() > other.width() && self.height() >= other.height())
-      || (self.width() >= other.width() && self.height() > other.height())
-  }
-
-  pub fn gte(self, other: BlockSize) -> bool {
-    self.greater_than(other)
-      || (self.width() == other.width() && self.height() == other.height())
-  }
-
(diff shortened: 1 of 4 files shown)


src/partition.rs
```diff
@@ -143,6 +141,21 @@ pub enum BlockSize {
...
+  fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+    use std::cmp::Ordering::{Equal, Greater, Less};
+    match (
+      self.width().cmp(&other.width()),
+      self.height().cmp(&other.height()),
+    ) {
+      (Greater, Less) | (Less, Greater) => None,
+      (Equal, Equal) => Some(Equal),
+      (Greater, _) | (_, Greater) => Some(Greater),
+      (Less, _) | (_, Less) => Some(Less),
```
shssoichiro: I'm thinking the `or` conditions in the last two match arms are unreachable, because they're handled by the first match arm.

barrbrain: With this change, the match is incomplete:
```
patterns `(Equal, Less)` and `(Equal, Greater)` not covered
```
I distilled [an example to explore](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a37ce7e35a15fe9e4a7bfaaf4a13ed5e).

shssoichiro: Ohh you're right, good call.

PR #2079, opened 2020-01-14
Cache the distortion scale computation

lu-zero:
About a 4% faster across all the speed levels.

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -1032,10 +1032,21 @@ impl<T: Pixel> ContextInner<T> {
       self.frame_data.insert(output_frameno, output_frame_data);
     }
 
-    #[cfg(feature = "dump_lookahead_data")]
-    {
-      if !output_framenos.is_empty() {
-        let fi = &mut self.frame_data.get_mut(&output_framenos[0]).unwrap().fi;
+    if !output_framenos.is_empty() {
+      let fi = &mut self.frame_data.get_mut(&output_framenos[0]).unwrap().fi;
+      let block_importances = fi.block_importances.iter();
+      let lookahead_intra_costs = fi.lookahead_intra_costs.iter();
+      let distortion_scales = fi.distortion_scales.iter_mut();
+      for ((&propagate_cost, &intra_cost), distortion_scale) in
+        block_importances.zip(lookahead_intra_costs).zip(distortion_scales)
+      {
+        *distortion_scale = crate::rdo::distortion_scale_for(
+          propagate_cost as f64,
+          intra_cost as f64,
+        );
+      }
+      #[cfg(feature = "dump_lookahead_data")]
+      {
         use byteorder::{NativeEndian, WriteBytesExt};
         let mut buf = vec![];
         buf.write_u64::<NativeEndian>(fi.h_in_imp_b as u64).unwrap();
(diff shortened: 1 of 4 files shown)


src/api/internal.rs
```diff
@@ -1032,10 +1032,21 @@ impl<T: Pixel> ContextInner<T> {
...
+        *distortion_scale = crate::rdo::distortion_scale_for(
+          propagate_cost as f64,
+          intra_cost as f64,
+        );
+      }
+      #[cfg(feature = "dump_lookahead_data")]
+      {
         use byteorder::{NativeEndian, WriteBytesExt};
         let mut buf = vec![];
         buf.write_u64::<NativeEndian>(fi.h_in_imp_b as u64).unwrap();
```
smarter: In the code below, we can get rid of `let res = ...; buf.write_f32::<NativeEndian>(res)` and just write the  `distortion_scales` array in the buffer, it's the value we want to visualize.

src/rdo.rs
```diff
@@ -439,9 +439,14 @@ pub fn compute_distortion_scale<T: Pixel>(
   let x = frame_bo.0.x >> IMPORTANCE_BLOCK_TO_BLOCK_SHIFT;
   let y = frame_bo.0.y >> IMPORTANCE_BLOCK_TO_BLOCK_SHIFT;
 
-  let propagate_cost = fi.block_importances[y * fi.w_in_imp_b + x] as f64;
-  let intra_cost = fi.lookahead_intra_costs[y * fi.w_in_imp_b + x] as f64;
+  //  let propagate_cost = fi.block_importances[y * fi.w_in_imp_b + x] as f64;
```
smarter: Don't leave commented out code.

src/rdo.rs
```diff
@@ -439,9 +439,14 @@ pub fn compute_distortion_scale<T: Pixel>(
...
+  //  let intra_cost = fi.lookahead_intra_costs[y * fi.w_in_imp_b + x] as f64;
 
+  //  distortion_scale_for(propagate_cost, intra_cost)
+  fi.distortion_scales[y * fi.w_in_imp_b + x]
+}
+
+pub fn distortion_scale_for(propagate_cost: f64, intra_cost: f64) -> f64 {
```
smarter: This method is only used in api/internal.rs, so it should be moved there (later we could also move all the block importance related code in its own file)

src/rdo.rs
```diff
@@ -439,9 +439,14 @@ pub fn compute_distortion_scale<T: Pixel>(
   let x = frame_bo.0.x >> IMPORTANCE_BLOCK_TO_BLOCK_SHIFT;
   let y = frame_bo.0.y >> IMPORTANCE_BLOCK_TO_BLOCK_SHIFT;
```
smarter: The `compute_distortion_scale` method should be renamed to just `distortion_scale` since it doesn't really compute anything anymore.

src/api/internal.rs
lu-zero: I wanted to do that on a second patch.

src/rdo.rs
lu-zero: I'd do that once we move the block importance code.

src/rdo.rs
lu-zero: I'll bake anothe patch on top of it. I wanted to keep the changes to the bare minimum.

src/api/internal.rs
lu-zero: Done.

lu-zero:
Sure

PR #2082, opened 2020-01-14
Ease auto vectorization of tx distortion computation

KyleSiefring:
The compiler was not autovectorizing this section of code. Switches the
code to doing two seperate iteration on the input coefficients from using
repeat and chain. The first iteration processes the stored reconstructed
coefficients and the second processes the reconstructed coefficients that
can be assumed to be zero. This isn't used by default, but Thomas is
working on a project that uses it.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1279,18 +1279,26 @@ pub fn encode_tx_block<T: Pixel>(
 
   let tx_dist = if rdo_type.needs_tx_dist() {
     // Store tx-domain distortion of this block
+    // rcoeffs above 32 rows/cols aren't held in the array, because they are
+    // always 0. The first 32x32 is stored first in coeffs so we can iterate
+    // over coeffs and rcoeffs for the first 32 rows/cols. For the
+    // coefficients above 32 rows/cols, we iterate over the rest of coeffs
+    // with the assumption that rcoeff coefficients are zero.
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -1279,18 +1279,26 @@ pub fn encode_tx_block<T: Pixel>(
 
   let tx_dist = if rdo_type.needs_tx_dist() {
     // Store tx-domain distortion of this block
+    // rcoeffs above 32 rows/cols aren't held in the array, because they are
+    // always 0. The first 32x32 is stored first in coeffs so we can iterate
+    // over coeffs and rcoeffs for the first 32 rows/cols. For the
+    // coefficients above 32 rows/cols, we iterate over the rest of coeffs
+    // with the assumption that rcoeff coefficients are zero.
```
ycho: As far as coeffs are not touched, it is correct that you assume outside of top-left 32x32 coefficients is all zeros.
I believe you want to test your change with 64x64 partitions with tx-domain distortion enabled.

KyleSiefring:
They are the same. https://beta.arewecompressedyet.com/?job=master_e081cbc_ref_s0%402020-01-14T18%3A55%3A40.987Z&job=tx_dist_auto_vec_s0%402020-01-14T18%3A56%3A58.170Z

PR #2083, opened 2020-01-14
Support intra edge filtering

### LINKED ISSUE #1065: Add intra edge filtering
We currently turn this off in the sequence header. This would be a pretty cheap compression win.

rzumer:
This adds support for and enables the intra edge filter for directional prediction and closes #1065.

[This AWCY run](https://beta.arewecompressedyet.com/?job=master-s2-e9db8ff6c255491f44e321505d1d272ff1bcc855%402020-01-13T18%3A34%3A47.902Z&job=ief-rdo-s2-e9db8ff6c255491f44e321505d1d272ff1bcc855%402020-01-14T17%3A30%3A16.615Z) has the following results (at speed level 2):
```
-0.4245 | -0.7415 | -0.5551 |  -0.4450 | -0.4508 | -0.5157 |    -0.4762
```

Run time when called natively is around 4% worse than master on this run. The native version could probably be better optimized.

I was considering adding a speed preset for it, but I don't think it would be too useful since it is only applied when directional prediction modes are enabled, which is only at lower levels where we would want to turn this on anyway.

The AVX2 version does not work (it desyncs). My guess is that it is due to pixels being initialized when they should not, or vice versa, in the edge buffer, as changing the `needs_foo` variables in `get_intra_edges()` changes the location of the desync. The current combination may be invalid for IEF.

### DIFF
--- a/src/asm/x86/predict.rs
+++ b/src/asm/x86/predict.rs
@@ -134,13 +134,29 @@ pub fn dispatch_predict_intra<T: Pixel>(
         | PredictionMode::D157_PRED
         | PredictionMode::D203_PRED
         | PredictionMode::D67_PRED => {
-          (if angle <= 90 {
-            rav1e_ipred_z1_avx2
-          } else if angle < 180 {
-            rav1e_ipred_z2_avx2
+          let enable_ief = ief_params.is_some() as libc::c_int;
+
+          if enable_ief > 0 {
+            // FIXME: segfaults with smooth filter, desyncs without
+            call_native(dst);
           } else {
-            rav1e_ipred_z3_avx2
-          })(dst_ptr, stride, edge_ptr, w, h, angle);
+            let ief_smooth_filter = if let Some(params) = ief_params {
+              params.use_smooth_filter()
+            } else {
+              false
+            } as libc::c_int;
+            // dav1d assembly uses the unused integer bits to hold IEF parameters
+            let angle_with_flags =
+              angle | (enable_ief << 10) | (ief_smooth_filter << 9);
+
+            (if angle <= 90 {
+              rav1e_ipred_z1_avx2
+            } else if angle < 180 {
+              rav1e_ipred_z2_avx2
+            } else {
+              rav1e_ipred_z3_avx2
+            })(dst_ptr, stride, edge_ptr, w, h, angle_with_flags);
+          }
         }
         PredictionMode::SMOOTH_PRED => {
           rav1e_ipred_smooth_avx2(dst_ptr, stride, edge_ptr, w, h, angle);
--- a/src/asm/x86/predict.rs
+++ b/src/asm/x86/predict.rs
@@ -9,7 +9,9 @@
 
 use crate::context::MAX_TX_SIZE;
 use crate::cpu_features::CpuFeatureLevel;
-use crate::predict::{native, PredictionMode, PredictionVariant};
+use crate::predict::{
+  native, IntraEdgeFilterParameters, PredictionMode, PredictionVariant,
+};
 use crate::tiling::PlaneRegionMut;
 use crate::transform::TxSize;
 use crate::util::AlignedArray;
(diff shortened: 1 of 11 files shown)


src/encoder.rs
```diff
@@ -1162,6 +1164,82 @@ pub fn encode_tx_block<T: Pixel>(
     "mode.is_intra()={:#?}, plane={:#?}, tx_size.block_size()={:#?}, plane_bsize={:#?}, need_recon_pixel={:#?}",
     mode.is_intra(), p, tx_size.block_size(), plane_bsize, need_recon_pixel);
 
+  let ief_params =
+    if mode.is_directional() && fi.sequence.enable_intra_edge_filter {
```
lu-zero: This should be a separate function I guess.

src/tiling/tile_state.rs
```diff
@@ -62,10 +63,31 @@ pub struct TileStateMut<'a, T: Pixel> {
   pub restoration: TileRestorationStateMut<'a>,
   pub half_res_pmvs: &'a mut Vec<BlockPmv>,
   pub mvs: Vec<TileMotionVectorsMut<'a>>,
+  pub coded_block_info: Vec<Vec<CodedBlockInfo>>, // indexed in MI units
```
lu-zero: It could be a `Box<[CodedBlockInfo]>` maybe?

src/encoder.rs
```diff
@@ -1745,6 +1824,19 @@ pub fn encode_block_post_cdef<T: Pixel>(
     }
   }
 
+  if fi.sequence.enable_intra_edge_filter {
+    for y in 0..bsize.height_mi() {
```
lu-zero: It could be make simpler if coded_block_info is a `Box<[]>` instead of a `Vec<Vec<>>`.

tdaede:
I would also agree no speed preset. The gain relative to the speed loss is quite good.

src/encoder.rs
rzumer: Done in three functions, two for getting the surrounding block information, and one for `IntraEdgeFilterParameters` initialization.

lu-zero: Neat :)

src/tiling/tile_state.rs
rzumer: Changed to a new structure with a single vector and `Index`/`IndexMut` implementations instead.

src/encoder.rs
rzumer: Changed to a new structure with a single vector and `Index`/`IndexMut` implementations instead.

rzumer:
Since IEF is enabled for encode-decode tests, and they catch most issues, I'm not sure having unit tests specifically for it would bring much value.

src/asm/x86/predict.rs
```diff
@@ -131,13 +134,27 @@ pub fn dispatch_predict_intra<T: Pixel>(
...
+            call_native(dst);
           } else {
-            rav1e_ipred_z3_avx2
-          })(dst_ptr, stride, edge_ptr, w, h, angle);
+            let ief_smooth_filter = if let Some(params) = ief_params {
+              params.use_smooth_filter()
+            } else {
+              false
+            } as libc::c_int;
+            let angle = angle | (enable_ief << 10) | (ief_smooth_filter << 9);
```
takehirokj:
```suggestion
            let angle_with_flg = angle | (enable_ief << 10) | (ief_smooth_filter << 9);
```
This and update in L156 fixes segfaults with smooth filter but still with desync.

A comment explaining enable_ief and ief_smooth_filter is encoded in bit of angle will help.

src/asm/x86/predict.rs
```diff
@@ -131,13 +134,27 @@ pub fn dispatch_predict_intra<T: Pixel>(
...
+
+            (if angle <= 90 {
```
takehirokj: This angle should be original value as above.

src/asm/x86/predict.rs
```diff
@@ -131,13 +134,27 @@ pub fn dispatch_predict_intra<T: Pixel>(
...
+              rav1e_ipred_z1_avx2
+            } else if angle < 180 {
+              rav1e_ipred_z2_avx2
+            } else {
+              rav1e_ipred_z3_avx2
+            })(dst_ptr, stride, edge_ptr, w, h, angle);
```
takehirokj:
```suggestion
            })(dst_ptr, stride, edge_ptr, w, h, angle_with_flg);
```

takehirokj:
> Since IEF is enabled for encode-decode tests, and they catch most issues, I'm not sure having unit tests specifically for it would bring much value.

I agreed that encode-decode tests is enough if IEF is enabled.

src/asm/x86/predict.rs
takehirokj: I mean the angle in from L150 and L152.

rzumer: Thanks, that was a silly mistake. Will fix later via force push.

src/asm/x86/predict.rs
rzumer: Done.

src/asm/x86/predict.rs
rzumer: Done.

src/asm/x86/predict.rs
rzumer: And done.

PR #2084, opened 2020-01-14
Avoid invoking memset in avx2 forward transforms

KyleSiefring:
First commit is single line patch that needs to be resolved for the next patch to work.

Use uninitialized memory to avoid invoking memset. Avoid assuming
uninitialized are initialized in one case and in the other, brazenly
use assume_init on an uninitialized array.

### DIFF
--- a/src/asm/x86/transform/forward.rs
+++ b/src/asm/x86/transform/forward.rs
@@ -12,6 +12,7 @@ use crate::transform::forward::native;
 use crate::transform::forward_shared::*;
 use crate::transform::*;
 use crate::util::*;
+use std::mem::MaybeUninit;
 
 #[cfg(target_arch = "x86")]
 use std::arch::x86::*;
@@ -336,7 +337,6 @@ unsafe fn fwd_txfm2d_avx2<T: Coefficient>(
   let mut tmp: AlignedArray<[I32X8; 64 * 64 / 8]> =
     AlignedArray::uninitialized();
   let buf = &mut tmp.array[..txfm_size_col * (txfm_size_row / 8).max(1)];
-  let tx_in = &mut [I32X8::zero(); 64];
   let cfg = Txfm2DFlipCfg::fwd(tx_type, tx_size, bd);
 
   let txfm_func_col = get_func_i32x8(cfg.txfm_type_col);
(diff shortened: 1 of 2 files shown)


src/asm/x86/transform/forward.rs
```diff
@@ -357,21 +357,30 @@ unsafe fn fwd_txfm2d_avx2<T: Coefficient>(
...
+        *out_reg = MaybeUninit::new(load_columns(in_slice.as_ptr(), shift));
       }
     }
 
-    txfm_func_col(tx_in, tx_out);
+    // TODO: Don't assume uninitialized values are initialized
+    let tx_out =
+      &mut MaybeUninit::<[I32X8; 64]>::uninit().assume_init()[..txfm_size_row];
+
+    txfm_func_col(assume_slice_init(tx_in), tx_out);
```
lu-zero: txfm_func_col could take in the MaybeUninit directly maybe?

KyleSiefring: You are talking about tx_out, right? I thought about it, but didn't want to do it, because I was being lazy. I would have to edit the native implementation and I want to leave that version using a slice.

I'm willing to do it. I just wasn't as willing to when I wrote this originally.

tdaede: txfm_func_col could also just return tx_out, couldn't it? Or is that obnoxious with the simd?

KyleSiefring: I decided to mutate the input and now the patch is a lot bigger than I originally planned.

src/asm/x86/transform/forward.rs
```diff
@@ -356,21 +357,27 @@ unsafe fn fwd_txfm2d_avx2<T: Coefficient>(
         shift,
       )
     }
+
+    // Avoid zero initialization
+    let tx_in = &mut MaybeUninit::<[MaybeUninit<I32X8>; 64]>::uninit()
+      .assume_init()[..txfm_size_row];
```
lu-zero: [MaybeUninit<I32X8>; 64] alone isn't enough?

KyleSiefring: Fixed

PR #2086, opened 2020-01-15
[WIP] Don't require writing coeffs for txfm reconstruction

KyleSiefring:


KyleSiefring:
I think this is only triggered when tx_domain_rate is true. It never is. If I turn it on, I get a segfault with my patch.

Looking at the commit that added this if statement, I feel that I should remove the branch. Going to have to ask Thomas.

https://github.com/xiph/rav1e/commit/85e882932ed16fd77885bd17b19ab7192ef5e069

src/encoder.rs
```diff
@@ -1232,8 +1232,9 @@ pub fn encode_tx_block<T: Pixel>(
   );
 
   let eob = ts.qc.quantize(coeffs, qcoeffs, tx_size, tx_type);
+  let has_coeff: bool = eob != 0;
 
-  let has_coeff = if need_recon_pixel || rdo_type.needs_coeff_rate() {
+  if rdo_type.needs_coeff_rate() {
```
ycho: I think you have missed the condition "need_recon_pixel ||" here? That become true, for ex, when there are multiple tx blocks in a partition and the prediction mode of the partition is intra prediction. Since ref pixels for there should be pixel domain.

src/context.rs
```diff
@@ -3987,7 +3987,7 @@ impl<'a> ContextWriter<'a> {
 
     if eob == 0 {
       self.bc.set_coeff_context(plane, bo, tx_size, xdec, ydec, 0);
-      return false;
```
tdaede: With this change you aren't setting the coefficient context anymore. This *might* be OK, *if* you verify that the initial value of the coefficient context in bc is correct, *and* it's always restored to that after an RDO rollback.

PR #2091, opened 2020-01-16
Add intra_mode_to_angle

takehirokj:


takehirokj:
Updated by using match.

PR #2092, opened 2020-01-16
Perform 1-d forward transform inplace

KyleSiefring:
Instead of taking both an input and output, overwrite over the input
slice. Avoids using an extra array to perform forward transforms. This
will help us remove uninitialized arrays from the forward transforms.


lu-zero:
Looks good, the benchmarks aren't happy though :)

PR #2095, opened 2020-01-17
frame: plane: Fix Downsampling of odd sized frames

### LINKED ISSUE #2055: Fix downsampling on odd sized frames
Currently, hres and qres are 1/2 and 1/4 the width of the full res frame, rounded down. They should really be rounded up.

vibhoothiiaanand:
Hi,

Currently, the hres and qres are talking half(1/2) and one-fourth(1/4) of the
width and height which is rounded down, when we have an odd-sized frame this becomes
a problem so to solve that we are rounding one up for both width and height.

Earlier: src.cfg.width / 2
Now    : (src.cfg.width + 1) / 2,

Fixes #2055

I am not 100% if doing this causes any other harm, but it would be great, to hear the comments with this

### DIFF
--- a/benches/plane.rs
+++ b/benches/plane.rs
@@ -24,6 +24,15 @@ pub fn downsample_8bit(c: &mut Criterion) {
   });
 }
 
+pub fn downsample_odd(c: &mut Criterion) {
+  let input = init_plane_u8(1919, 1079);
+  c.bench_function("downsample_odd", move |b| {
+    b.iter(|| {
+      let output = input.downsampled(960, 540);
+    })
+  });
+}
+
 pub fn downsample_10bit(c: &mut Criterion) {
   let input = init_plane_u16(1920, 1080);
   c.bench_function("downsample_10bit", move |b| {
@@ -33,4 +42,4 @@ pub fn downsample_10bit(c: &mut Criterion) {
   });
 }
 
-criterion_group!(plane, downsample_8bit, downsample_10bit);
+criterion_group!(plane, downsample_8bit, downsample_odd, downsample_10bit);
--- a/src/frame/plane.rs
+++ b/src/frame/plane.rs
@@ -437,8 +437,8 @@ impl<T: Pixel> Plane<T> {
     // unsafe: all pixels initialized in this function
     let mut new = unsafe {
       Plane::new_uninitialized(
-        src.cfg.width / 2,
-        src.cfg.height / 2,
+        (src.cfg.width + 1) / 2,
+        (src.cfg.height + 1) / 2,
         src.cfg.xdec + 1,
         src.cfg.ydec + 1,
         src.cfg.xpad / 2,
--- a/src/frame/plane.rs
+++ b/src/frame/plane.rs
@@ -452,8 +452,8 @@ impl<T: Pixel> Plane<T> {
     let yorigin = new.cfg.yorigin;
     let stride = new.cfg.stride;
 
-    assert!(width * 2 == src.cfg.width);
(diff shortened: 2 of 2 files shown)


vibhoothiiaanand:
https://beta.arewecompressedyet.com/?job=%402020-01-17T15%3A01%3A19.232Z&job=master-caef43f3fcadcde053ddd6c49caef188724209ea

Did AWCY and noticed there is a very small encoding time improving by less than a second or second which is neglible,

Good news is nothing is broke.

tdaede:
Could you make a downsampling test that shows this? See the existing downsample test - make an odd sized ones.

src/frame/plane.rs
```diff
@@ -787,7 +793,55 @@ pub mod test {
...
+        0, 0, 0, 0, 0, 0, 0, 0,  
+        0, 0, 0, 0, 0, 0, 0, 0,  
+        6, 6, 0, 0, 0, 0, 0, 0,  
+        0, 0, 0, 0, 0, 0, 0, 0,  
+        0, 0, 0, 0, 0, 0, 0, 0,  
+        0, 0, 0, 0, 0, 0, 0, 0,  
+        0, 0, 0, 0, 0, 0, 0, 0,  
+        0, 0, 0, 0, 0, 0, 0, 0,  
+        0, 0, 0, 0, 0, 0, 0, 0,  
+        0, 0, 0, 0, 0, 0, 0, 0,
```
takehirokj: How about changing the width and height to 3 from 5 and expected result as same as
test_plane_downsample() as L784-791?

src/frame/plane.rs
```diff
@@ -451,9 +451,15 @@ impl<T: Pixel> Plane<T> {
...
-    assert!(height * 2 == src.cfg.height);
+  
+    //if width % 2 == 0 && height % 2 == 0 {
+     // assert!(width * 2 == src.cfg.width);
+     // assert!(height * 2 == src.cfg.height);
+    //} else {
+      //assert!(width * 2  == src.cfg.width + 1);
+     // assert!(height * 2 == src.cfg.height + 1); 
+   // }
+
```
takehirokj: This part can be deleted.

src/frame/plane.rs
vibhoothiiaanand: Yes, very much we can have that,

I thought of having higher values for both width and height, the only issue was when we have large values for Width/Height like 9,11,13 it would crash. 
IIRC we have a hard limit of 16x16 for width and height
Ref: https://github.com/xiph/rav1e/blob/master/src/api/config.rs#L774
I am not sure what we are missing.

I did update the PR too, please do check it

benches/plane.rs
```diff
@@ -24,6 +24,14 @@ pub fn downsample_8bit(c: &mut Criterion) {
...
 }
 
+pub fn downsample_odd(c: &mut Criterion) {
+  let input = init_plane_u8(1919, 1079);
+  c.bench_function("downsample_odd", move |b| {
+    b.iter(|| {
+      let output = input.downsampled(960, 540);
+    })
+  });
+}
```
takehirokj:
```suggestion
}

```
It's better to add new line as in other functions.

src/frame/plane.rs
takehirokj: > I thought of having higher values for both width and height, the only issue was when we have large values for Width/Height like 9,11,13 it would crash.
For this, we need to change the stride and alloc_height too.
I have confirmed Width/Height=9 works without crash by changing stride and alloc_height.

For testing, I believe Width/Height=3 is enough.

PR #2096, opened 2020-01-17
Replace 1d fwd tx benchmarks with 2d benchmarks

KyleSiefring:
The 1d benchmarks profile the native code and aren't very representitive
of the performance of the transforms as a whole. Switching to 2d
transforms also lowers the exposure of the benchmark code to the more
unstable internals of our forward transforms.

### DIFF
--- a/benches/transform.rs
+++ b/benches/transform.rs
@@ -10,7 +10,9 @@
 use criterion::*;
 use rand::{Rng, SeedableRng};
 use rand_chacha::ChaChaRng;
+use rav1e::bench::cpu_features::*;
 use rav1e::bench::transform;
+use rav1e::bench::transform::{forward_transform, TxSize, TxType};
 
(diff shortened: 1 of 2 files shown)


benches/transform.rs
```diff
@@ -72,54 +74,78 @@ pub fn av1_iadst8(c: &mut Criterion) {
...
-  let (coeffs, _) = init_buffers(4);
-
-  c.bench_function("daala_fdct4", move |b| {
-    b.iter_batched(
-      || coeffs.clone(),
-      |mut coeffs| transform::forward::native::daala_fdct4(&mut coeffs[..]),
-      BatchSize::SmallInput,
-    )
-  });
+fn get_valid_txfm_types(tx_size: TxSize) -> Vec<TxType> {
```
lu-zero: it is for benchmarks so we don't care much, you can return references to 'static arrays in case you need that in other code.

PR #2099, opened 2020-01-19
Add missing Copy impl on two structs that can use it

shssoichiro:


rzumer:
Does that actually make any different performance-wise?

shssoichiro:
Not significantly, it might have a minor impact because I've heard the compiler is more likely to allocate `Copy` structs on the stack as opposed to the heap, although it's possible LLVM optimizes that anyway. However, the general recommendation is to always implement Copy when possible.

src/context.rs
```diff
@@ -1509,7 +1510,7 @@ const PARTITION_CONTEXT_MAX_WIDTH: usize =
 
 const COEFF_CONTEXT_MAX_WIDTH: usize = MAX_TILE_WIDTH / MI_SIZE;
 
-#[derive(Clone)]
+#[derive(Clone, Copy)]
 pub struct BlockContextCheckpoint {
```
lu-zero: Looks fairly large, I assume we never pass it around functions.
Is there a clippy lint to prevent it?

shssoichiro:
To clarify my previous underinformed comment, a long time ago I heard this impacts stack/heap usage, but I can no longer find any source to support that. What I did find was a [comment by Steve Klabnik](https://www.reddit.com/r/rust/comments/2xxjda/when_should_my_type_be_copy/) on Reddit talking about this:

> If you don't think that you'll be adding a non-Copy element to your struct later, than you should implement Copy.
>
> Remember, a move is just a copy where the compiler disallows using it afterward. That's the only thing this modifies, it's not more or less expensive.

So apparently, this likely has no direct impact on performance, just on semantics. It's strictly not necessary unless we're using an explicit `clone` somewhere on this struct that doesn't need to be a `clone` (clones are at best equivalent performance to a copy, and can perform worse). Clippy does have a lint to [warn if we use clone on a copy type](https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy) and to [warn if a small copy type is passed by ref](https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy) to help ensure Copy types are used optimally.

Based on this, the recommendation "If you don't think that you'll be adding a non-Copy element to your struct later, than you should implement Copy." seems reasonable. (The only reason it's not recommended to add Copy to a struct that may become non-Copy later is for API stability of public types.) There's no downside to implementing Copy, and although there's no immediate performance gain, it does enable other performance optimizations.

shssoichiro:
There is also an allow-by-default lint built into the compiler which can warn if Copy is missing from a type that can implement Copy. It's worth asking if it makes sense to enable this lint for rav1e.

rzumer:
Personally I prefer avoiding boilerplate that effectively does not do anything.

PR #2102, opened 2020-01-19
Eliminate buffer and copy for sifting codable coefficients

KyleSiefring:
For large transforms we reorder coefficients from our forward transform
so that the codable coefficients are first. Before this patch, we would
do this by buffering the output forward transform then writing it out
in the correct order.

I'm a little suprised by the lack of impact in the microbenchmark
results. I expected the copy to be inefficient enough to provide a
performance boost when removed. It seems that a copy in the L1 cache
just doesn't have a noticable impact.

In the real world, performance does seem to improve on awcy by about
half a percent.

https://beta.arewecompressedyet.com/?job=master-redo-70dc401df%402020-01-19T14%3A02%3A47.242Z&job=embed_fwd_reorder%402020-01-19T14%3A25%3A07.898Z

### DIFF
--- a/src/asm/x86/transform/forward.rs
+++ b/src/asm/x86/transform/forward.rs
@@ -462,26 +462,44 @@ unsafe fn fwd_txfm2d_avx2<T: Coefficient>(
     // this size.
     match row_class {
       SizeClass1D::X8UP => {
-        for c in 0..txfm_size_col {
-          match T::Pixel::type_enum() {
-            PixelType::U8 => {
-              let lo = _mm256_castsi256_si128(row_coeffs[c].vec());
-              let hi = _mm256_extracti128_si256(row_coeffs[c].vec(), 1);
-              _mm_storeu_si128(
-                output[c * txfm_size_row + rg..].as_mut_ptr() as *mut _,
-                _mm_packs_epi32(lo, hi),
-              );
-            }
-            PixelType::U16 => {
-              _mm256_storeu_si256(
-                output[c * txfm_size_row + rg..].as_mut_ptr() as *mut _,
-                row_coeffs[c].vec(),
-              );
+        // Store output in at most 32x32 chunks. See native code for details.
+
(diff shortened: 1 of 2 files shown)


KyleSiefring:
Test at speed 0 with min block size is 32x32.

https://beta.arewecompressedyet.com/?job=64fwd_reorder_impl_ref_2%402020-01-18T19%3A51%3A50.344Z&job=64fwd_reorder_impl_test_2%402020-01-18T19%3A55%3A01.984Z

src/transform/forward.rs
```diff
@@ -153,8 +153,31 @@ pub mod native {
...
+      //  - First 32 rows and last 32 cols
+      //  - Last 32 rows and last 32 cols
+
+      // Output is grouped into 32x32 chunks so a stride of at most 32 is
+      // used for each chunk.
+      let output_stride = txfm_size_row.min(32);
+
+      // Split the first 32 rows from the last 32 rows
+      let output = &mut output
+        [(r >= 32) as usize * output_stride * txfm_size_col.min(32)..];
```
lu-zero: Might be better to split the loop in two one for `< 32` and the other for `>= 32` instead of having the branch in the inner loop.

src/asm/x86/transform/forward.rs
```diff
@@ -462,26 +462,44 @@ unsafe fn fwd_txfm2d_avx2<T: Coefficient>(
...
-              );
+        // Store output in at most 32x32 chunks. See native code for details.
+
+        // Output is grouped into 32x32 chunks so a stride of at most 32 is
+        // used for each chunk
+        let output_stride = txfm_size_row.min(32);
+
+        // Split the first 32 rows from the last 32 rows and offset by rg % 32
+        let output = &mut output[(rg & 31)
+          + (rg >= 32) as usize * output_stride * txfm_size_col.min(32)..];
```
lu-zero: Might be ideal to split the loop for `>=32` and `< 32`, if you factorize the inner loop below in a local function it should be even easier to read.

PR #2108, opened 2020-01-21
Fix desyncs and bitstream corruption in still picture mode

### LINKED ISSUE #1000: Use reduced_still_picture_header when still_picture flag is set

rzumer:
`force_screen_content_tools` was not set correctly. It must be 2 only when the reduced still picture header is enabled, which it cannot yet be, since we do not write palette mode info.

Also adds a test for it.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -216,10 +216,10 @@ impl Sequence {
       delta_frame_id_length: DELTA_FRAME_ID_LENGTH,
       use_128x128_superblock: false,
       order_hint_bits_minus_1: 5,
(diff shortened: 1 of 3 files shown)


rzumer:
Some test failures to take care of, if there are desyncs it should show up in the metrics [here](https://beta.arewecompressedyet.com/?job=master-stillpicture%402020-01-21T21%3A30%3A40.784Z&job=master-stillpicture%402020-01-21T21%3A50%3A38.806Z) as well once the second run has completed.

rzumer:
Unfortunately still getting desyncs in my local tests, investigation ongoing...

PR #2109, opened 2020-01-22
CI: move grcov-coveralls conf to beta channel

barrbrain:
The required compiler flags are in rust version: `1.41.0-beta.1 (eb3f7c2d3 2019-12-17)`


barrbrain:
Nope, `-Zprofile` is still nightly-only.

PR #2114, opened 2020-01-22
Enable tx partition for inter mode at speed 10, always split once

ycho:
This commit is prepared for the urgent demo purposes,
which requires tx partitions for inter mode blocks and one level split only at speed 10.

- DCT_DCT only

Known bug or limitation.
- Only works for reduced_tx_set enabled, with unknown reason.

TODO: Fix the bug!

### DIFF
--- a/src/api/config.rs
+++ b/src/api/config.rs
@@ -326,6 +326,9 @@ pub struct SpeedSettings {
 
   /// Use segmentation.
   pub enable_segmentation: bool,
+
+  /// Enable tx split for inter mode block.
+  pub enable_inter_tx_split: bool,
 }
 
 impl Default for SpeedSettings {
@@ -355,6 +358,7 @@ impl Default for SpeedSettings {
       use_satd_subpel: true,
       non_square_partition: true,
       enable_segmentation: true,
+      enable_inter_tx_split: false,
     }
   }
 }
@@ -395,6 +399,7 @@ impl SpeedSettings {
       use_satd_subpel: Self::use_satd_subpel(speed),
       non_square_partition: Self::non_square_partition_preset(speed),
       enable_segmentation: Self::enable_segmentation_preset(speed),
+      enable_inter_tx_split: Self::enable_inter_tx_split_preset(speed),
     }
   }
 
@@ -505,6 +510,11 @@ impl SpeedSettings {
   // solution we should be able to enable segmentation at all speeds.
   const fn enable_segmentation_preset(speed: usize) -> bool {
     speed == 0
+  }
+
+  // FIXME: With unknown reasons, inter_tx_split does not work if reduced_tx_set is false
+  const fn enable_inter_tx_split_preset(speed: usize) -> bool {
+    speed == 10
   }
 }
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -87,6 +87,9 @@ pub const TX_SETS_INTER: usize = 4;
 
 pub const TXFM_PARTITION_CONTEXTS: usize =
   ((TxSize::TX_SIZES - TxSize::TX_8X8 as usize) * 6 - 3);
+
+const MAX_VARTX_DEPTH: usize = 2;
+
 const MAX_REF_MV_STACK_SIZE: usize = 8;
 pub const REF_CAT_LEVEL: u32 = 640;
 
@@ -110,7 +113,7 @@ pub static av1_tx_used: [[usize; TX_TYPES]; TX_SETS] = [
 ];
 
 // Maps set types above to the indices used for intra
-static tx_set_index_intra: [i8; TX_SETS] = [0, -1, 2, 1, -1, -16];
+static tx_set_index_intra: [i8; TX_SETS] = [0, -1, 2, 1, -1, -1];
 // Maps set types above to the indices used for inter
 static tx_set_index_inter: [i8; TX_SETS] = [0, 3, -1, -1, 2, 1];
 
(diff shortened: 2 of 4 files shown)


ycho:
[AWCY result for speed 10 on the way](https://beta.arewecompressedyet.com/?job=master-a496fd9ec-s10%402020-01-22T19%3A42%3A11.910Z&job=test_tx_split_inter-s10%402020-01-22T19%3A49%3A33.593Z)

ycho:
No tx-split at 1st frame, i.e. intra frame.
image

And 16x16 tx used for 32x32 partitions.
image

ycho:
Checked that there is no change at default speed 6, [awcy](https://beta.arewecompressedyet.com/?job=master-a496fd9ec69ae5c0264ac1d51065fa98f5bc694f&job=test_tx_split_inter-s6%402020-01-22T19%3A52%3A21.615Z)

ycho:
Purple colored blocks are skipped.
Blocks with no tx split are coded intra mode.

image

ycho:
11:58 AM 
<•derf> codeview: in write_tx_size_inter(), you call write_tx_size_inter() recursively, but it seems to be misisng the check that the tx_size is not already 4x4?
11:58 AM 
<•derf> Since you cannot split a 4x4 transform, there is no need to write a txfm_split == false symbol.
11:59 AM 
<codeview> yeah, I am sure I leave lots of bugs there... comment welcome
11:59 AM 
<codeview> I saw somewhere it checked but probably and possibly not in the place you just seen
11:59 AM 
<•derf> Yes, this is my comment ^
11:59 AM 
<•derf> I cannot log into Github.
11:59 AM 
<codeview> i will see and fix it anyway, atm I am more busy to soothe cargo fmt....
12:01 PM 
<codeview> derf: yes, what happens is before the function is call, there is debug_assert!(bsize > BlockSize::BLOCK_4X4);
12:01 PM 
<codeview> https://github.com/xiph/rav1e/pull/2114/files#diff-26006a0497afb96d876927e114a7b29eR1746
12:02 PM 
<codeview> but better we double secure
12:02 PM 
<codeview> since someone else in the future can call it else where
12:03 PM 
<•derf> codeview: No, I mean, if the block size is 8x8, then inside that function, it calls itself, without the check.
12:03 PM 
<•derf> Which would explain why it does not work for partition sizes smaller than 16x16.
12:03 PM 
<codeview> oh, yes- it calls inside it....
12:03 PM 
<codeview> oh!
12:04 PM 
<•derf> No idea why it needs the reduced transform set yet, though.
12:04 PM 
<codeview> was it recursive? the last parameter is txfm_split == false
12:04 PM 
<•derf> codeview: Yes, but you are still writing a symbol when you don't need to.
12:05 PM 
<codeview> let me look at it....
12:05 PM 
<•derf> The first thing write_tx_size_inter does is call symbol_with_update() to code the value of txfm_split.
12:05 PM 
<codeview> oh i always call symbol_with_update!( ).... dumb
12:06 PM 
<•derf> Exactly.
12:08 PM 
<•derf> Shouldn't that have tripped the debug_assert!() in txfm_partition_context()?
12:09 PM 
<codeview> No, it was not hit. btw, I will add a condition :    if txfm_split && tx_size != TX_4X4 {.   }
12:10 PM 
<codeview> you mean debug_assert!(category != TXFM_PARTITION_CONTEXTS); ?
12:10 PM 
<•derf> Yes.
12:11 PM 
<•derf> Oh, but it only checks max_tx_size against bsize... but you did not make bsize smaller when recursing.
12:11 PM 
<codeview> Because, the max_tx_size for "if max_tx_size >= TxSize::TX_8X8 {"  comes from et max_tx_size: TxSize = bsize.tx_size().sqr_up();   and 
12:11 PM 
<•derf> So something seems fishy there.
12:11 PM 
<codeview> bszie is partition size.
12:11 PM 
<codeview> yes
12:11 PM 
<codeview> partition size is required for context decision
12:12 PM 
<codeview> for fetching above and left tx context
12:12 PM 
<codeview> sure, pls help us find more bugs
12:17 PM 
<•derf> Honestly, I don't know what the if max_tx_size >= TxSize::TX_8X8 check in txfm_partition_context() is all about.
12:17 PM 
<•derf> It's not in the spec.
12:18 PM 
<•derf> The debug_assert!() is just checking whether or not that if was hit, but it is not the proper condition to test if you have a valid context.
12:21 PM 
<•derf> The proper check is txSz != TX_4X4 && depth < MAX_VARTX_DEPTH (see read_var_tx_size in the spec).
12:21 PM 
<•derf> But it does not make any sense to have an if() followed by a debug_assert!() that the if was taken.
12:23 PM 
<•derf> codeview: Looking at read_var_tx_size()... there is another potential issue with your patch.
12:23 PM 
<•derf>   if (row >= MiRows || col >= MiCols)
12:23 PM 
<•derf>     return
12:23 PM 
<•derf> ^ You have no equivalent of that check, for partitions which straddle the frame border.
12:24 PM 
<codeview> Well, it has.
12:24 PM 
<•derf> Where?
12:25 PM 
<codeview> tby == 0. or  tbx == 0. in let above = (self.get_above_tx_width(bo, bsize, tx_size, tby == 0)
12:26 PM 
<codeview> in txfm_partition_context()
12:26 PM 
<codeview> for each of above and left, it passes whether the tx blocks is the first one in row in col in a partition, for above and left, resp.
12:27 PM 
<•derf> Okay.
12:27 PM 
<codeview> that is tested.
12:27 PM 
<•derf> I don't see how that relates to the issue I'm raising?
12:27 PM 
<codeview> btw, thanks for catching the bug, it seem fixing  my 16x16 min tx bug.
12:28 PM 
<codeview> if (row >= MiRows || col >= MiCols) ?
12:28 PM 
<•derf> Yes.
12:28 PM 
<codeview> in read_var_tx_size() ?
12:28 PM 
<•derf> Yes, in the spec.
12:29 PM 
<•derf> write_tx_size_inter should only write txfm_split in the same cases that read_var_tx_size reads that symbol in the spec.
12:29 PM 
<•derf> But you are not checking all of the conditions that the spec does.
12:30 PM 
<•derf> So there will be a mismatch.
12:32 PM 
<codeview> :) btw, do you know, what are MiRows and MiCols?
12:32 PM 
<codeview> Yes, I missed frame border check.
12:32 PM 
<•derf> The width and height of the frame in 4x4 blocks.
12:33 PM 
<•derf> It is saying, if a transform block is completely outside the frame, then it can never be split.
12:33 PM 
<codeview> Right. So, i mostly tested with nyan, and might had lucks so far.
12:33 PM 
<•derf> Yes.
12:33 PM 
<codeview> Yeap-
12:33 PM 
<•derf> Maybe we should crop nyan a bit to trigger more edge cases.
12:34 PM 
<codeview> The one I responded wrong was in fact more challenging,  if wrong it cannot decode from the first SB. I thought you asked me that one.
12:34 PM 
<•derf> Oh, no, that was not what I was asking.
12:35 PM 
<codeview> Well, I haven't invited you as reviewer but you are most helping!
12:36 PM 
<codeview> let me fix after I fix the clippy ones...
12:36 PM 
<codeview> thx again---
12:57 PM 
<codeview> TD-Linux, maseck, unlord, derf, So, there is super slight increase of enc time but we have coding gain of -3.70, https://beta.arewecompressedyet.com/?job=master-a496fd9ec-s10%402020-01-22T19%3A42%3A11.910Z&job=test_tx_split_inter-s10%402020-01-22T19%3A49%3A33.593Z
12:59 PM 
<•derf> Encoding time went up by 2.6%? Seems like a good trade...
1:03 PM 
<codeview> TD-Linux, maseck, unlord, derf,  Pls go ahead plan demo with this PR. For other possible fixes like derf gave me, I will address them as separate PRs and commits.

2:35 PM 
<codeview> derf: I've addressed on your comments. 1) condition of non splitable min tx size 4x4 and max depth (2),  https://github.com/xiph/rav1e/pull/2114/files#diff-e0934bb6f70a4f8c3baf22a773c14ec7R2258
2:36 PM 
<codeview> 2) mysterious condition in txfm_partition_context() https://github.com/xiph/rav1e/pull/2114/files#diff-e0934bb6f70a4f8c3baf22a773c14ec7L2233,
2:37 PM 
<codeview> I removed this. At first I copied whole this function from my reference, libaom, and I think the condition is equivalent to bsize > 4x4, not tx_size. So 8x4, 4x8 is fine, but only bszie 4x4 will fail it. But txfm_partition_context( ) is not called for bsize 4x4 (and there is assert before it), the condition always pass.
2:38 PM 
<codeview> as we know, the depth check in rav1e is currenlly not used since we always do 1 split only. But  added for future extension.
2:39 PM 
<codeview> oh, frame border check is not there, probably go in separate PR.
2:48 PM 
<codeview> One  simple idea to have 64x64 partition skip back at speed 10 would be: if all four of 32x32 are skipped then we can change it as 64x64 skip.
2:48 PM 
<•derf> codeview: After the if tx_size != TX_4X4 && depth < MAX_VARTX_DEPTH, maybe an else debug_assert!(!txfm_split)?
2:48 PM 
<•derf> Then you can drop the debug_assert!() on line 2273.
2:48 PM 
<codeview> oh sure-
2:49 PM 
<•derf> (i.e., it makes sure the conditions match)
2:49 PM 
<•derf> (because right now they don't)
2:49 PM 
<codeview> you mean the if condition fails 
2:49 PM 
<•derf> I mean, it makes sure if we don't code the symbol, then we are not splitting.
2:50 PM 
<codeview> "derf> Then you can drop the debug_assert!() on line 2273." meas 'remove' it....
2:51 PM 
<•derf> Yes.
2:53 PM 
<codeview> yeap- make sense. btw, noticed this https://github.com/xiph/rav1e/pull/2114/files#diff-e0934bb6f70a4f8c3baf22a773c14ec7R2259 ?
2:53 PM 
<•derf> codeview: The && ... part on line 2242 can also go away, I think?
2:53 PM 
<codeview> i.e. moved        let ctx = self.txfm_partition_context(bo, bsize, tx_size, tbx, tby); inside the condition, bcz ctx only needed when encoding a symbol.
2:54 PM 
<•derf> codeview: Yes, I noticed and approved of that.
2:54 PM 
<codeview> This? "&& max_tx_size > TxSize::TX_8X8" 
2:56 PM 
<•derf> Yes.
2:56 PM 
<•derf> I don't see anything that corresponds to that in the spec.
2:57 PM 
<codeview> Hmm not sure, it is clearly not is spec but bcz it will be false if tx size is 4x8 or 8x4, which are splittable
2:57 PM 
<codeview> it was in libaom. And possibly, legacy code from vp9
2:57 PM 
<•derf> Hmm.
2:57 PM 
<•derf> Well, that's concerning.
2:57 PM 
<codeview> when sub8x8 is not a partition
2:57 PM 
<codeview> :) IT IS possible spec misses it...
2:58 PM 
<codeview> first, i need clearly understand the context of tx_split to confirm this. We know there are 21.
2:59 PM 
<codeview> and for each depth, there are 3 (above + left)
3:01 PM 
<codeview> In the spec,  "(txSzSqrUp != maxTxSz) * 3 " read as first 3 ctx, i.e. 0,1,2 are for when tx size == block size (both up_square)
3:02 PM 
<codeview> i.e. if current tx_size to split is the bsize, then first three cdfs are used.
3:03 PM 
<•derf> I looked at the libaom code and I think it gave me cancer.
3:04 PM 
<•smarter> rip
3:04 PM 
<codeview> that is one category, we have 21 - 3  = 18 left.  18/3 = 6 categories/
3:05 PM 
<codeview> derf: need deep breathe before read it
3:08 PM 
<•derf> codeview: So, there are 4 values of maxTxSz possible, (8x8, 16x16, 32x32, 64x64).
3:09 PM 
<BitBot> (rav1e) [PR] Luni-4 requested #2115 merge into master: Fix AppVeyor build - https://git.io/JvLbp
3:09 PM 
<•derf> And 6 contexts for each size: 3 if txSzSqrUp == maxTxSz and 3 if txSzSqrUp != maxTxSz.
3:09 PM 
<•derf> So ostensibly there are 24 possible contexts... but in the last case, where maxTxSz is 8x8, then txSzSqrUp is always equal to maxTxSz.
3:09 PM 
<•derf> So really there are only 21.
3:11 PM 
<•derf> So, in the case where max_tx_size > TxSize::TX_8X8, then tx_size.sqr_up() != max_tx_size is never true.
3:12 PM 
<•derf> So the condition is in fact redundant, and the spec is correct.
3:12 PM 
<•derf> And libaom is just dumb.
3:12 PM 
<codeview> I think your interpretation is based on libacom function, including && max_tx_size > TX_8X8)
3:12 PM 
<codeview> "&& max_tx_size > TX_8X8)"
3:12 PM 
<codeview> ?
3:12 PM 
<•derf> No.
3:13 PM 
<•derf> My interpretation is just based on the spec, and counting.
3:13 PM 
<•derf> But you can compare the two, and the libaom code is equivalent to the spec, it just includes an extra check for no reason.
3:14 PM 
<•derf> The extra check would cover contexts 21...23, but those cannot occur.
3:16 PM 
<codeview> I am not sure how you got "•derf> And 6 contexts for each size: 3 if txSzSqrUp == maxTxSz and 3 if txSzSqrUp != maxTxSz."
3:16 PM 
<codeview> In spec, "(txSzSqrUp != maxTxSz) * 3" term just shift 3 entries  if true
3:16 PM 
<•derf> Yes, exactly.
3:16 PM 
<codeview> not multiplied by 6
3:16 PM 
<•derf> It is (TX_SIZES - 1 - maxTxSz) that is multiplied by 6.
3:17 PM 
<•derf> I.e., 6 for each size.
3:17 PM 
<codeview> so whether txSzSqrUp == maxTxSz or not only affect offset of 3
3:17 PM 
<codeview> 6 for each size is correct.
3:17 PM 
<codeview> sqr tx size.
3:17 PM 
<•derf> More correctly, 6 for each maxTxSz.
3:18 PM 
<•derf> And then within maxTxSz, 3 for the first split, and then 3 for any additional splits.
3:18 PM 
<codeview> hmm.. so first 3 is 6-3!
3:18 PM 
<•derf> But for the smallest maxTcSz, 8x8, you can only have one split.
3:18 PM 
<•derf> *maxTxSz
3:43 PM 
<codeview> reading dav1d code, "(txSzSqrUp != maxTxSz) " is depth info, i.e, 0 or 1
3:44 PM 
<codeview> https://code.videolan.org/videolan/dav1d/blob/master/src/decode.c#L153
3:53 PM 
<codeview> so, 4  bsize_up_sqr or maxTxSz (64x64 ... 8x8) x 2 depths (0, 1) x 3 neighbor types ( above + left) - 3  (there only depth 0 for 8x8) = 21
3:56 PM 
<codeview> in libaom, if max_tx_size == TX_8X8,  always txsize_sqr_up_map[tx_size] == max_tx_size, so the term (txsize_sqr_up_map[tx_size] != max_tx_size && max_tx_size > TX_8X8) is always 0. So, don't need "max_tx_size > TX_8X8" term.
3:57 PM 
<•derf> codeview: Now we're on the same page.

4:48 PM 
<codeview> TD-Linux: "4:07 PM <•TD-Linux> codeview, btw master runs demo at 50fps, but we only need 30fps. so any trade to slow it down for a big quality increase is ok", if that is the case, we can enable tx-split inter mode for speed 9 as well, whose partition can choose either 64x64 or 32x32?
4:48 PM 
<codeview> Before that, pls review and approve https://github.com/xiph/rav1e/pull/2114, I think derf has approved it already?
4:51 PM 
<•derf> I did not carefully review the whole patch, but the parts I did review look good now.
4:53 PM 
<•derf> codeview: Should I interpret from your analyzer pictures that the blocks that do not have transform split are skip blocks?
5:02 PM 
<codeview> derf: I could never understand what 6 means until Jan 2021, if I haven't read dav1d code.
5:02 PM 
<•derf> Oh. It seemed pretty clear to me.
5:06 PM 
<codeview> derf: https://github.com/xiph/rav1e/pull/2114#issuecomment-577459058
5:06 PM 
<codeview> Yes, if no tx-split, those are either skip or intra mode block.
5:07 PM 
<•derf> Makes sense.
5:07 PM 
<codeview> I did run for nyan again for you...

PR #2117, opened 2020-01-23
Use ArrayVec instead of array of MaybeUninit in write_coeffs_lv_map

shssoichiro:
This enables us to know the length of the values
in coeffs_storage at all times,
and allows the compiler to auto-vectorize the
calculation of cul_level.

Overall speed difference is <1%;
the two loops at the end of this function
are responsible for most of the time consumption here.
This commit will hopefully unblock further
optimizations on those loops.

The aligned crate is being vendored here
because the upstream crate does not
have a type for 32-byte alignment,
which we need.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -34,7 +34,6 @@ use crate::util::*;
 
 use arrayvec::*;
 use std::default::Default;
-use std::mem::MaybeUninit;
 use std::ops::{Index, IndexMut};
 use std::*;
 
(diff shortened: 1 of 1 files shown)


shssoichiro:
I would also like to, in a followup PR, use the Aligned crate in place of our custom AlignedArray. The performance difference from that should be zero, but the Aligned struct provided is more generic and will allow us to do similar optimizations such as using ArrayVecs in places where it makes sense.

shssoichiro:
It probably makes more sense to PR a 32-byte aligned type into the aligned crate, rather than vendoring. So I will also open a PR to do that upstream.

shssoichiro:
Removed the third-party library in favor of using our existing AlignedArray, which I did not realize could take items other than arrays.

PR #2118, opened 2020-01-23
Channel api

lu-zero:
This set provides a channel-based API that should make much simpler the normal usage.

TODO:
- [x] Basic encoding API
  - [x] Works
  - [x] Produces the same output for the same input
- [ ] Multipass encoding API
   - [x] Single pass
   - [ ] Second pass
     - [ ] Produces the same output for the same input
   - [ ] Multipass

Open questions:
- Fit a `last-error` inside the ReceiverPacket abstraction and a mean to retrieve it
- For multipass the api can be a single function and use the Config rc information to decide if it has to be multipass/firstpass/secondpass/normal

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,6 +73,7 @@ ivf = { version = "0.1", path = "ivf/", optional = true }
 v_frame = { version = "0.1", path = "v_frame/" }
 av-metrics = { version = "0.5.1", optional = true }
 rayon = "1.0"
+crossbeam = "0.7"
 toml = { version = "0.5", optional = true }
 arrayvec = { version = "0.5", features = ["array-sizes-33-128"] }
 thiserror = "1.0"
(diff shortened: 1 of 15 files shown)


shssoichiro:
@lu-zero is this still WIP or is it ready for review?

lu-zero:
You can start reviewing it, the multipass variant is using exactly the same approach (and I had been otherwise busy to complete it)

shssoichiro:
I don't have any major concerns about the API as-is. I assume this is still a WIP, as there's definitely cleanup to be done.

Was the rav1e-ch binary just for demonstration purposes? I'd prefer to only have to maintain one binary, so ideally the channel-based CLI would replace the current CLI.

lu-zero:
It is to compare the two API and impl side by side. the rate control integration is still incomplete (I need to fix the data feeding system so it works as I intended)

PR #2125, opened 2020-01-24
Simplify logic in Counter/Recorder store method

shssoichiro:
This function is particularly hot in WriterCounter,
taking about 5% of total encode time at speed 10.
This commit reduces the generated assembly
by 3 instructions and 1 branch.

Overall encode time reduced by 1%.

### DIFF
--- a/src/ec.rs
+++ b/src/ec.rs
@@ -188,17 +188,15 @@ impl StorageBackend for WriterBase<WriterCounter> {
   fn store(&mut self, fl: u16, fh: u16, nms: u16) {
     let (_l, r) = self.lr_compute(fl, fh, nms);
     let d = 16 - r.ilog();
-    let mut c = self.cnt;
-    let mut s = c + (d as i16);
+    let mut s = self.cnt + (d as i16);
 
-    if s >= 0 {
-      c += 16;
-      if s >= 8 {
-        self.s.bytes += 1;
-        c -= 8;
-      }
+    if s >= 8 {
       self.s.bytes += 1;
-      s = c + (d as i16) - 24;
+      s -= 8;
+    }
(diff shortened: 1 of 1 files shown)


tdaede:
Looking at the tests in ec.rs, it looks like there's no coverage of WriterRecorder. Could you either add a unit test, or run on AWCY to ensure there's 0 change in encoded output?

shssoichiro:
Coveralls shows it's covered (probably by encode/decode tests), but yeah looks like there are no unit tests. I'll at least throw it in AWCY. I'm not sure what the best way would be to test the Counter and Recorder, although I'm willing to write some unit tests if I could get pointed in the right direction on those.

tdaede:
The main problem is that Counter is only used for RDO, so if it produces wrong results the encode/decode tests will still work.

shssoichiro:
[AWCY checks out](https://beta.arewecompressedyet.com/?job=master-5b94130e5701b61bb81f7c2aa54763043242bcfd&job=counter-recorder-optim%402020-01-24T02%3A10%3A46.232Z)

src/ec.rs
```diff
@@ -188,17 +188,15 @@ impl StorageBackend for WriterBase<WriterCounter> {
   fn store(&mut self, fl: u16, fh: u16, nms: u16) {
     let (_l, r) = self.lr_compute(fl, fh, nms);
     let d = 16 - r.ilog();
-    let mut c = self.cnt;
-    let mut s = c + (d as i16);
+    let mut s = self.cnt + (d as i16);
 
+    if s >= 8 {
```
tdaede: might be interesting to make this a if/elif chain for s>=16 and s>=8

Issue #2107, opened 2020-01-21
Desyncs when using AVX2 dav1d assembly with the intra edge filter


rzumer:
Assembly for directional prediction is currently disabled, because it causes desyncs when enabling the intra edge filter. My guess is that it is related to the edge buffer. Depending on the prediction mode, different edges and corners are initialized or not, and the assembly may be sensitive to that. However, simply reverting the change made to the edge buffer initialization for native intra edge filter support (setting the top left pixel for all directional modes) does not solve the desync, so more research is needed to determine whether it is the actual cause, and if so, what is the correct way to fill the buffer.

rzumer:
Seems like [this](https://code.videolan.org/videolan/dav1d/blob/master/src/ipred_prepare_tmpl.c#L198) needs to be implemented in partition::get_intra_edges() and handled in the native version.

takehirokj:
I found that rav1e_ipred_z2_avx2 causes desync.
Replacing call_native to each rav1e_ipred_z1_avx2 and  rav1e_ipred_z3_avx2 doesn't
cause desync.

rzumer:
The top left pixel filter is applied only for z2 prediction, so that makes sense. I am planning on fixing it next week.

rzumer:
#2135 makes some progress towards fixing this, but not there yet.

rzumer:
It seems the missing piece is the dx and dy parameters that must be passed to the z2 prediction function. It is used only for z2 and, I guess, only for the intra edge filter. I'll take a shot at it.

Corresponding dav1d code for filling those arguments [here](https://code.videolan.org/videolan/dav1d/blob/master/src/recon_tmpl.c#L1219).

PR #2126, opened 2020-01-25
Exclude unrequired mode for needs_top decision in get_intra_edges

takehirokj:
The output is same and encode time is almost same as [AWCY](https://beta.arewecompressedyet.com/?job=remove_unrequired_mode_in_selecting_part%402020-01-25T06%3A24%3A27.450Z&job=master-8ebe9233f6fbea4b3c83dd8f76129d196082e3d7).

### DIFF
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -572,7 +572,7 @@ pub fn get_intra_edges<T: Pixel>(
         || mode == PredictionMode::D203_PRED
         || mode == PredictionMode::D67_PRED
         || mode == PredictionMode::PAETH_PRED;
-      needs_top = !dc_or_cfl || y != 0;
+      needs_top = (!dc_or_cfl || y != 0) && mode != PredictionMode::D203_PRED;
       needs_topright = mode == PredictionMode::V_PRED
         || mode == PredictionMode::D45_PRED
         || mode == PredictionMode::D67_PRED;


src/partition.rs
```diff
@@ -563,16 +563,9 @@ pub fn get_intra_edges<T: Pixel>(
...
-        || mode == PredictionMode::D135_PRED
-        || mode == PredictionMode::D113_PRED
-        || mode == PredictionMode::D157_PRED
-        || mode == PredictionMode::D203_PRED
-        || mode == PredictionMode::D67_PRED
-        || mode == PredictionMode::PAETH_PRED;
-      needs_top = !dc_or_cfl || y != 0;
+      needs_topleft =
+        mode.is_directional() || mode == PredictionMode::PAETH_PRED;
+      needs_top = (!dc_or_cfl || y != 0) && mode != PredictionMode::D203_PRED;
```
rzumer: > && mode != PredictionMode::D203_PRED
Why is this added?

takehirokj: This angle mode is not required for "needs_top" and this does not affect to the output.

rzumer: Can you specify where this is mentioned, e.g. spec section, page, or reference to libaom?

src/partition.rs
```diff
@@ -563,16 +563,9 @@ pub fn get_intra_edges<T: Pixel>(
...
-        || mode == PredictionMode::H_PRED
-        || mode == PredictionMode::D45_PRED
-        || mode == PredictionMode::D135_PRED
-        || mode == PredictionMode::D113_PRED
-        || mode == PredictionMode::D157_PRED
-        || mode == PredictionMode::D203_PRED
-        || mode == PredictionMode::D67_PRED
-        || mode == PredictionMode::PAETH_PRED;
-      needs_top = !dc_or_cfl || y != 0;
+      needs_topleft =
```
rzumer: This is good to simplify now, but not all directional modes need the top-left pixel if IEF is disabled. Since I am planning to fix the top-left pixel pre-processing for IEF and put it in this function it will probably end up being reverted.

takehirokj: OK. I understand this simplification need to be changed after fixing #2107.
It is better to keep it as now.

src/partition.rs
takehirokj: "AboveRow" is not required "if pAngle is greater than 180" as
in, 7.11.2.4.Directional intra prediction process (9.)
[AV1 spec](https://aomediacodec.github.io/av1-spec/av1-spec.pdf)

src/partition.rs
takehirokj: This part is changed to original and change for "needs_top" is kept.

PR #2127, opened 2020-01-25
avx512 support

lu-zero:
Now copy dav1d in checking for pre-Icelake and Icelake avx512.

### DIFF
--- a/src/cpu_features/x86.rs
+++ b/src/cpu_features/x86.rs
@@ -20,11 +20,13 @@ pub enum CpuFeatureLevel {
   #[arg_enum(alias = "sse4.1")]
   SSE4_1,
   AVX2,
+  AVX512,
+  AVX512ICL,
 }
 
 impl CpuFeatureLevel {
   pub const fn len() -> usize {
-    CpuFeatureLevel::AVX2 as usize + 1
+    CpuFeatureLevel::AVX512ICL as usize + 1
   }
 
   #[inline(always)]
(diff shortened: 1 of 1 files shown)


src/cpu_features/x86.rs
```diff
@@ -35,7 +36,9 @@ impl CpuFeatureLevel {
 
 impl Default for CpuFeatureLevel {
   fn default() -> CpuFeatureLevel {
-    let detected: CpuFeatureLevel = if is_x86_feature_detected!("avx2") {
+    let detected: CpuFeatureLevel = if is_x86_feature_detected!("avx512f") {
```
KyleSiefring: With AVX512, Intel decided to divide features among many cpuflags. You are going to have to check more cpu flags if you want something useful.

You can see which flags dav1d uses in the comments here.
https://code.videolan.org/videolan/dav1d/commit/40891aab9bc2f60bdbef16d1a499f36944e1738d#0dc601e13d41edea90a29ed0c5a00c738474626d_854_848

Wikipedia can also give you the info you need. https://en.wikipedia.org/wiki/AVX-512

KyleSiefring:
Now that I think about it, you probably have to make sure avx512 functions get all of the avx2 functions. Ugh.

Maybe I should try to fix this problem again. I tried a few things but rust didn't support them yet. It might be a good idea to mutate those tables on startup for now and fix it when a solution becomes possible.

lu-zero:
We can discuss it and see how to solve it in rust if needed.

src/cpu_features/x86.rs
lu-zero: dav1d basically checks for icelake in the detection routine, I updated our check accordingly, I might need your help to add the missing feature bits in stdarch if we want to be thorough.

shssoichiro:
> Now that I think about it, you probably have to make sure avx512 functions get all of the avx2 functions.

Could you clarify? There shouldn't be any CPUs with AVX-512 support but no AVX2 support, AFAIK.

lu-zero:
Meanwhile I prepared https://github.com/rust-lang/stdarch/pull/838

src/cpu_features/x86.rs
lu-zero: I added the checks, even if even a partial match should be fine for now at least :)

barrbrain:
> Now that I think about it, you probably have to make sure avx512 functions get all of the avx2 functions. Ugh.

This was resolved in #2136.

src/cpu_features/x86.rs
```diff
@@ -35,7 +37,29 @@ impl CpuFeatureLevel {
...
+      is_x86_feature_detected!("avx512bw")
+        && is_x86_feature_detected!("avx512cd")
+        && is_x86_feature_detected!("avx512dq")
+        && is_x86_feature_detected!("avx512f")
+        && is_x86_feature_detected!("avx512vl")
+    }
+    fn avx512icl_detected() -> bool {
+      avx512_detected()
+        // && is_x86_feature_detected!("avx512bitalg")
+        // && is_x86_feature_detected!("avx512clmulqdq")
```
shssoichiro: Are these commented lines important?

lu-zero: Not until Intel releases a cpu that has a set of avx512 extensions that do not match the pre-Icelake and Icelake cpu.

We just need to know that we are on an Icelake (so probe for at least some of the Icelake-introduced extensions) or not.

lu-zero:
@barrbrain are you fine with the non-exhaustive check?

barrbrain:
Given the rustc version constraint, testing just the widely available features is fine.

lu-zero:
@KyleSiefring you have to approve or I won't be able to merge it.

KyleSiefring:
You need to change this line https://github.com/xiph/rav1e/blob/master/src/cpu_features/x86.rs#L80 .

Issue #1981, opened 2019-12-20
Upstream nasm-rs uses stabs debug info for x86_64 elf


eclipseo:
I built the rav1e binary but when I try to extract debuginfo and strip the binary, I'm encountering an issue:

```
explicitly decompress any DWARF compressed ELF sections in /builddir/build/BUILDROOT/rust-rav1e-0.2.0-1.fc32.x86_64/usr/bin/rav1e
extracting debug info from /builddir/build/BUILDROOT/rust-rav1e-0.2.0-1.fc32.x86_64/usr/bin/rav1e
Stabs debuginfo not supported: /builddir/build/BUILDROOT/rust-rav1e-0.2.0-1.fc32.x86_64/usr/bin/rav1e
Failed to update file: invalid section entry size
```
It seems there are Stabs debuginfo in the binary.

Is this normal? It prevents me from extracting the debuginfo. How can I turn that off?

eclipseo:
(This is Rust 1.40 on x86_64)

tdaede:
I think this should have fixed that: https://github.com/xiph/rav1e/pull/1938
Is your rav1e newer than that commit?

eclipseo:
Yes I'm onto this commit, my nasm-rs is from upstream, that's why it doesn't include this commit. I'm in the process of patching it and see how this goes.
Can this be upstreamed otherwise?

lu-zero:
Let's try to sync up with upstream then :) 

https://github.com/medek/nasm-rs/pulls

lu-zero:
https://github.com/medek/nasm-rs/pull/26 is still ope with no feedback.

lu-zero:
The upstream version 0.1.6 should work fine, could you please test it?

lu-zero:
It does not since I managed to land a typo when I resolved a collision after a rebase :(

barrbrain:
The upstream version 0.1.7 includes the typo fix and we now use that.

PR #2128, opened 2020-01-25
Use the upstream nasm-rs

lu-zero:
All our changes got upstreamed now.

See #1981


lu-zero:
This should be updated once https://github.com/medek/nasm-rs/pull/27 lands.

lu-zero:
Now upstream `nasm-rs 0.1.7` supports everything we need :)

PR #2129, opened 2020-01-26
Remove y4m header in desync finding step of contribution guide

takehirokj:
Original step showed difference in header even if there's no difference in frame data.
This is confusing.

This additional step excludes the difference in Y4M header as desync detection in
rav1e/build.sh (L56-59).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,7 +108,13 @@ Install `cargo-fuzz` with `cargo install cargo-fuzz`. Running fuzz targets requi
 /path/to/dav1d -i out.ivf -o dec.y4m
 ```
 
-3. Compare if the reconstruction and decoded video match.
+3. Remove the y4m sequence header to see the difference in frame header or data
 ```
-cmp rec.y4m dec.y4m
+tail -n+2 rec.y4m > rec
+tail -n+2 dec.y4m > dec
+```
+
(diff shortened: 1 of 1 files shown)


ycho:
I think this approach removes the sequence header thus works for y4m file having 1 frame only, bcz each frame in y4m has a frame header as well?

In https://wiki.multimedia.cx/index.php/YUV4MPEG2,
"Each frame begins with the 5 bytes 'FRAME' followed by zero or more parameters each preceded by 0x20, ending with 0x0A. "

takehirokj:
This works for y4m with multiple frame headers.

This removes only sequence header which has difference.
For each frame, there's no difference.

takehirokj:
@ycho 
Y4m frame header contains parameters below:
  - Detail of "'Im' = Mixed modes" for interlace.
  - Original metadata
All my rec/dec sample has zero parameter in frame header and I think it's 
usually zero parameters for most cases.
Therefore, this update works for most cases.

takehirokj:
Updated the sentence to show that the removed header is sequence header.

lu-zero:
I think it is already an improvement, we can consider adding an option to output `.yuv` raw files later.

takehirokj:
> I think it is already an improvement, we can consider adding an option to output .yuv raw files later.

Thanks. Further improvement will be done later if required.

PR #2135, opened 2020-01-27
Pre-process the top-left intra edge pixel for dav1d IEF assembly

rzumer:
This works towards supporting the assembly versions of directional predictors with the intra edge filter. However it seems like there is extra work to be done, as I am still getting desyncs locally when enabling the Z2 prediction.

As pointed out by @takehirokj, assembly for Z1 and Z3 prediction does not cause desyncs, so is enabled as well.


rzumer:
I force-pushed a change to remove a stale comment, will merge once tests pass again.

PR #2136, opened 2020-01-28
Use macros to define static lookup tables

KyleSiefring:


lu-zero:
The idea is really pretty. clippy is a little sad though...

PR #2137, opened 2020-01-28
Provide dx and dy arguments to the Z2 prediction dav1d assembly

### LINKED ISSUE #2107: Desyncs when using AVX2 dav1d assembly with the intra edge filter

rzumer:
This allows Z2 prediction dav1d assembly to
be enabled along with the intra edge filter.

Context for why this data is needed:
https://bugs.chromium.org/p/aomedia/issues/detail?id=2239

Tested only on nyan.y4m locally. AWCY run started [here](https://beta.arewecompressedyet.com/?job=master-131b2916ff290147e843d6e0ae704c8d6acf5ed1&job=ief-fix-131b2916ff290147e843d6e0ae704c8d6acf5ed1%402020-01-28T20%3A11%3A43.730Z). I expect a small performance improvement, but it might be within the margin of error, since Z1 and Z3 assembly is already enabled.

Fixes #2107.

### DIFF
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -568,37 +568,23 @@ pub fn get_intra_edges<T: Pixel>(
         _ => mode,
       };
 
+      let p_angle = intra_mode_to_angle(mode)
+        + match intra_param {
+          IntraParam::AngleDelta(val) => (val * ANGLE_STEP) as isize,
+          _ => 0,
+        };
+
       let dc_or_cfl =
         mode == PredictionMode::DC_PRED || mode == PredictionMode::UV_CFL_PRED;
 
-      needs_left = (!dc_or_cfl || x != 0)
-        && !(mode == PredictionMode::D45_PRED
-          || mode == PredictionMode::D67_PRED);
+      needs_left = (!dc_or_cfl || x != 0) || (p_angle > 90 && p_angle != 180);
       needs_topleft = mode == PredictionMode::PAETH_PRED
-        || if enable_intra_edge_filter {
-          mode.is_directional()
-        } else {
-          mode == PredictionMode::V_PRED
-            || mode == PredictionMode::H_PRED
-            || mode == PredictionMode::D135_PRED
-            || mode == PredictionMode::D113_PRED
-            || mode == PredictionMode::D157_PRED
-        };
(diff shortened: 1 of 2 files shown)


src/partition.rs
```diff
@@ -568,37 +568,23 @@ pub fn get_intra_edges<T: Pixel>(
...
-          + match intra_param {
-            IntraParam::AngleDelta(val) => (val * ANGLE_STEP) as isize,
-            _ => 0,
-          };
-
-        enable_intra_edge_filter && p_angle > 90 && p_angle < 180
-      };
+        || (mode.is_directional() && p_angle != 90 && p_angle != 180);
+      needs_top = (!dc_or_cfl || y != 0) || (p_angle != 90 && p_angle < 180);
+      needs_topright = p_angle <= 90;
```
barrbrain: For non-directional `mode`, we have `p_angle == 0` which also matches this condition.
```suggestion
      needs_topright = mode.is_directional() && p_angle <= 90;
```

src/partition.rs
```diff
@@ -568,37 +568,23 @@ pub fn get_intra_edges<T: Pixel>(
         _ => mode,
       };
 
+      let p_angle = intra_mode_to_angle(mode)
+        + match intra_param {
+          IntraParam::AngleDelta(val) => (val * ANGLE_STEP) as isize,
+          _ => 0,
+        };
```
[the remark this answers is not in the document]
rzumer: Yeah, I was thinking the same thing while implementing that.

src/partition.rs
rzumer: Sorry for merging too early, if you want to submit a separate PR for this I'll approve it, otherwise I will open one tomorrow.

PR #2142, opened 2020-01-29
build.sh: Clean up and add some checks

1480c1:
Use /bin/sh
Check if cargo exists before running it and use :+ for IS_RELEASE
Remove bashism
Add curl alternative if wget doesn't exist
Remove unnecessary  subshell on echos
Test if aomdec and mpv exist before testing them
Convert comments at the bottom to block comments

Right now testing to make sure it isn't broken

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -15,15 +15,14 @@ for arg in "$@"; do
   esac
 done
 
-if [ -z "$SEQ" ]; then
-  SEQ=nyan.y4m
+if [ -z "$SEQ" ] && SEQ=nyan.y4m && [ ! -f nyan.y4m ]; then
   #SEQ10=nyan10.y4m
   #SEQ12=nyan12.y4m
 
-  if type wget > /dev/null 2>&1; then
-    wget -nc https://mf4.xiph.org/~ltrudeau/videos/nyan.y4m
-  elif type curl > /dev/null 2>&1; then
-    curl -fkLs https://mf4.xiph.org/~ltrudeau/videos/nyan.y4m -o nyan.y4m
+  if ! wget -nc https://mf4.xiph.org/~ltrudeau/videos/nyan.y4m &&
+    ! curl -fkLs https://mf4.xiph.org/~ltrudeau/videos/nyan.y4m -o nyan.y4m; then
+    echo "wget and curl not found or failed!" >&2
+    exit 1
   fi
   #wget -nc https://people.xiph.org/~tdaede/nyan10.y4m
   #wget -nc https://people.xiph.org/~tdaede/nyan12.y4m
--- a/build.sh
+++ b/build.sh
@@ -1,85 +1,105 @@
-#!/bin/bash
+#!/bin/sh
 
 # Stop on first error
 set -e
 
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -1,85 +1,105 @@
-#!/bin/bash
+#!/bin/sh
 
 # Stop on first error
 set -e
```
barrbrain: Note that the existing behavior is to stop on error.

build.sh
```diff
@@ -1,85 +1,105 @@
...
-
-  wget -nc https://mf4.xiph.org/~ltrudeau/videos/nyan.y4m
+  #SEQ10=nyan10.y4m
+  #SEQ12=nyan12.y4m
+
+  if type wget > /dev/null 2>&1; then
+    wget -nc https://mf4.xiph.org/~ltrudeau/videos/nyan.y4m
+  elif type curl > /dev/null 2>&1; then
+    curl -fkLs https://mf4.xiph.org/~ltrudeau/videos/nyan.y4m -o nyan.y4m
+  fi
```
barrbrain: We could error here or at least warn the user that we expected wget or curl.

build.sh
```diff
@@ -1,85 +1,105 @@
...
 
 # Print the backtrace on error
 export RUST_BACKTRACE=1
 
 # Build and run encoder
-BUILD_TYPE=""
-if [ $IS_RELEASE == 1 ]; then
-  BUILD_TYPE="--release"
+
+if ! type cargo > /dev/null 2>&1; then
```
barrbrain: This check is superfluous, because of  `set -e` we would stop at the following call with `sh: 50: cargo: not found`.

build.sh
```diff
@@ -1,85 +1,105 @@
...
+  echo "cargo not found" >&2
+  exit 1
 fi
 
-cargo run --bin rav1e $BUILD_TYPE -- $SEQ -o $ENC_FILE -s 3 -r $REC_FILE
+if ! cargo run --bin rav1e ${IS_RELEASE:+--release} -- "$SEQ" -o $ENC_FILE -s 3 -r $REC_FILE; then
```
barrbrain: Likewise, rav1e should provide an error message on failure and the error code will cause us to stop here.

build.sh
1480c1: I mainly did this to not rely on the shell's error reporting and that way if we want, we can print out a more helpful message like `cargo not found, please install rust either from your package manager or from https://rustup.rs/`, but I first wanted some feedback on this before adding links or changing error messages

build.sh
1480c1: Yeah, that's why I added a lot more ifs than I would prefer

build.sh
1480c1: Sure

build.sh
1480c1: Would it be better to split it into a cargo build then run rav1e to know at what point it failed?

barrbrain: This script is targeted to new contributors, so it may be helpful to distinguish between build errors and run-time errors.

PR #2147, opened 2020-02-03
Remove unnecessary allocations from rdo_loop_optimization

shssoichiro:
Currently this function never looks at more than one
superblock at a time, even if we were to enable
128x128 superblocks. This means we can avoid allocating
Vecs within this function.

This change reduces the total allocation count by ~25%,
by removing these 1-4 byte allocations.
This also results in an encoding time reduction of ~1%.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -45,7 +45,6 @@ use arrayvec::*;
 use itertools::izip;
 use std;
 use std::fmt;
-use std::vec::Vec;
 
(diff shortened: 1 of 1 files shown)


tdaede:
Maybe give a bit for @xiphmont to review though

shssoichiro:
You are correct, I was able to trigger this with a quantizer of 200. I'll have to attempt to find the theoretical maximum and use an ArrayVec to avoid the allocations.

Will open a clean PR once that is available.

PR #2148, opened 2020-02-03
Reduce the buffer of coeff encoder from 64x64 downto 32x32

ycho:
The largest size of tx block that can be encodeded by AV1 coefficient
encoder is 32x32, thus reduce it down to around~1/4 of previous size.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -59,7 +59,9 @@ pub const SB_SIZE: usize = (1 << SB_SIZE_LOG2);
 const SB_SQUARE: usize = (SB_SIZE * SB_SIZE);
 
 pub const MAX_TX_SIZE: usize = 64;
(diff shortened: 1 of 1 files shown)


ycho:
> The logic is sound. There appears to be another over-allocation at line 4200, `coeff_contexts`. We may add `MAX_CODED_TX_SIZE` and `MAX_CODED_TX_SQUARE` or similar for clarity.

Indeed!

ycho:
> The logic is sound. There appears to be another over-allocation at line 4200, `coeff_contexts`. We may add `MAX_CODED_TX_SIZE` and `MAX_CODED_TX_SQUARE` or similar for clarity.

Forgot to say: Great thanks for review! :)

PR #2149, opened 2020-02-04
Optimizations to rdo_loop_decision

shssoichiro:
Part 1:
There is a known upper bound to the number of superblocks
and LRUs that this function can analyze at once,
so we can allocate a fixed sized array for the results.

Reduces the total allocation count by 25%.
No measurable impact on encoding speed.

Part 2:
Because looping through the planes is the inner-most loop,
reversing the layout of the nested arrays helps with linear
cache access.

Improves encoding time by 1%.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -1935,7 +1935,7 @@ pub fn rdo_loop_decision<T: Pixel>(
   // need to make sure we don't fall into a limit cycle.  Track our
   // best cost at LRF so that we can break if we get a solution that doesn't
   // improve at the reconstruction stage.
-  let mut best_lrf_cost = [[-1.0; MAX_LRU_SIZE * MAX_LRU_SIZE]; PLANES];
+  let mut best_lrf_cost = [[-1.0; PLANES]; MAX_LRU_SIZE * MAX_LRU_SIZE];
 
   // Loop filter RDO is an iterative process and we need temporary
   // scratch data to hold the results of deblocking, cdef, and the
@@ -2129,7 +2129,7 @@ pub fn rdo_loop_decision<T: Pixel>(
                 )
               } {
                 // We have a valid LRU, apply LRF, compute error
-                match best_lrf[pli][lru_y * lru_w[pli] + lru_x] {
+                match best_lrf[lru_y * lru_w[pli] + lru_x][pli] {
                   RestorationFilter::None {} => {
                     err += rdo_loop_plane_error(
                       base_sbo,
@@ -2147,7 +2147,7 @@ pub fn rdo_loop_decision<T: Pixel>(
                       cw.count_lrf_switchable(
                         w,
                         &ts.restoration.as_const(),
-                        best_lrf[pli][lru_y * lru_w[pli] + lru_x],
+                        best_lrf[lru_y * lru_w[pli] + lru_x][pli],
                         pli,
                       )
                     } else {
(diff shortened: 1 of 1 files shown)


src/rdo.rs
```diff
@@ -1897,18 +1896,19 @@ pub fn rdo_loop_decision<T: Pixel>(
     lru_h[pli] = sb_h / (1 << sb_v_shift);
   }
 
-  let mut best_index = vec![-1; sb_w * sb_h];
-  let mut best_lrf = ArrayVec::<[Vec<RestorationFilter>; 3]>::new();
+  // Based on `RestorationState::new`
+  const MAX_SB_SHIFT: usize = 4;
```
lu-zero: Can we keep those in a single place?

shssoichiro: You mean make these available more generally? I'm not sure if these variables as they're defined here are useful outside the context of this function. It doesn't seem so, looking at usages of `ts.restoration.planes[pli].rp_cfg.sb_h_shift`.

lu-zero:
the comment made me wonder :)

shssoichiro:
I think ideally we could use a const fn to compute this based on the logic in `RestorationState::new`, but that would require if blocks in const fns to be stable first.

lu-zero:
I'll wait for @xiphmont opinion then, no objection from me.

xiphmont:
Part 1:  This is true now, but not necessarily going to be true forever.  Also, I routinely run tests where I use the LF RDO loop on larger segments.  That requires changing the area determination in the code, but still useful for testing.

Part 2: Array setup was arbitrary.  Have at.

shssoichiro:
Hmm, all right, sounds like it's reasonable to cherry pick part 2 and commit that, and leave out part 1 for future compatibility. Thanks!

xiphmont:
As per further discussion in channel:
<barrbrain> xiphmont_: primary motivation was speeding up memory tracing
<xiphmont_> so although the LF RDO loop *can* run on bigger sections, if we have reason to hard-code the current invariant, I'm not actually fundamanetally opposed.
<soichiro> It looks like part 2 isn't really feasible without part 1
<xiphmont_> ah, so there is a practical performance reason
<xiphmont_> even if it's testing, I'm still in favor of it
<xiphmont_> OK
<xiphmont_> In that case, sure.
<xiphmont_> if I need bigger chink handling later, I can put dynamic alloc back temporarily, or enlarge the static allocation.
<xiphmont_> chunk
<xiphmont_> darnit
<xiphmont_> anyway, add a comment that the new static allocation relies on the 'minimal LRU area for all 3 planes' invariant and go for it.
<xiphmont_> I'll add a comment as such
<barrbrain> I think the other reason is to do with analysis, less noise - more signal
<xiphmont_> all fine
<xiphmont_> Improving testability is good

So yes, in favor of both. Add the comment and I'll get started on looking more carefully at the code changes.

src/rdo.rs
```diff
@@ -1923,18 +1922,21 @@ pub fn rdo_loop_decision<T: Pixel>(
...
   // due to imprecision in the reconstruction parameter solver, we
   // need to make sure we don't fall into a limit cycle.  Track our
   // best cost at LRF so that we can break if we get a solution that doesn't
   // improve at the reconstruction stage.
-  let mut best_lrf_cost = ArrayVec::<[Vec<f64>; 3]>::new();
-  for pli in 0..PLANES {
-    best_lrf.push(vec![RestorationFilter::None; lru_h[pli] * lru_w[pli]]);
-    best_lrf_cost.push(vec![-1.0; lru_h[pli] * lru_w[pli]]);
-  }
+  let mut best_lrf_cost = [[-1.0; PLANES]; MAX_LRU_SIZE * MAX_LRU_SIZE];
```
xiphmont, line 1940: Hm, we lose some implicit bounds checking here.  Before, if we accidentally overflowed the expected number of LRUs in a plane, we'd get a panic.  Now it's potentially possible to overflow and continue on none the wiser.

Issue #2155, opened 2020-02-05
Boolean values don't work with rav1e_config_parse_int()


tdaede:
e.g. 
```
rav1e_config_parse_int(configuration, "low_latency", 1);
```

lu-zero:
What mapping would you propose? 0 vs non zero ?

tdaede:
C boolean semantics, so yes.

PR #2157, opened 2020-02-06
Add rav1e_config_parse_bool

lu-zero:
Act as helper to set boolean values.

Fixes: https://github.com/xiph/rav1e/issues/2155


dwbuiten:
I feel like using `parse_int` with 0 and 1 feels more like idiomatic C, but also I don't have a stong opinion on it.

@tdaede any opinion from you?

lu-zero:
If you tested it I'd go and merge it

lu-zero:
Poke.

lu-zero:
I'd land it for 0.5.

PR #2161, opened 2020-02-07
Refactor compute_motion_vectors_per_tile

lu-zero:


lu-zero:
I'd keep the prefixes matching between the two functions

PR #2162, opened 2020-02-08
Add a quiet option to the cli

lu-zero:


src/bin/common.rs
```diff
@@ -33,6 +33,7 @@ pub struct CliOptions {
   pub override_time_base: bool,
   pub skip: usize,
   pub verbose: bool,
+  pub quiet: bool,
```
shssoichiro: IMO it would be cleaner to have one field for `verbosity` with a tri-state enum with states `Quiet`, `Normal`, `Verbose`.

src/bin/common.rs
```diff
@@ -322,6 +323,12 @@ pub fn parse_cli() -> Result<CliOptions, CliError> {
         .long("verbose")
         .short("v")
     )
+    .arg(
+      Arg::with_name("QUIET")
+        .help("Do not output any status message")
+        .long("quiet")
+        .short("q")
```
shssoichiro: What is the desired behavior if both quiet and verbose are specified? Should we assume one, or display an error?

lu-zero: Quiet beats verbose in my implementation.

src/bin/common.rs
lu-zero: You are right.

PR #2164, opened 2020-02-10
Split the Frame-related data structures

lu-zero:
This should make easier sharing code with the stand-alone av-scenechange and av-metrics.

- [x] Move Plane
   - [x] Move the math and pixel dependencies
   - [ ] Add a copy_plane function
- [x] Move Frame
   - [x] Make so the exported constructors do not depend on rav1e internals, make the rav1es use them with the internal constant.
- [ ] Move tiling/*


v_frame/Cargo.toml
```diff
@@ -6,7 +6,13 @@ license = "BSD-2-Clause"
...
 edition = "2018"
 
-
 [dependencies]
 cfg-if = "0.1"
 num-traits = "0.2"
+num-derive = "0.3"
+noop_proc_macro = "0.2.0"
+
+[dependencies.rav1e]
```
lu-zero: It should not depend on rav1e at all.

src/frame/mod.rs
```diff
@@ -41,70 +41,69 @@ pub struct FrameParameters {
...
   #[cfg(test)]
-  pub(crate) fn as_tile(&self) -> Tile<'_, T> {
+  fn as_tile(&self) -> Tile<'_, T> {
     let PlaneConfig { width, height, .. } = self.planes[0].cfg;
     Tile::new(self, TileRect { x: 0, y: 0, width, height })
   }
+}
+
+/// Public Trait for new Mutable Tile of a frame
+pub trait AsTileMut<T: Pixel> {
```
lu-zero: You may use a single trait for both methods.

src/frame/mod.rs
```diff
@@ -41,70 +41,69 @@ pub struct FrameParameters {
...
+  #[cfg(test)]
+  fn as_tile_mut(&mut self) -> TileMut<'_, T>;
+}
 
+impl<T: Pixel> AsTileMut<T> for Frame<T> {
   #[inline(always)]
   #[cfg(test)]
```
lu-zero: This way `#[cfg(test)]` can be put before the impl block

src/frame/mod.rs
```diff
@@ -100,9 +100,9 @@ pub trait AsTileMut<T: Pixel> {
   fn as_tile_mut(&mut self) -> TileMut<'_, T>;
 }
 
+#[cfg(test)]
 impl<T: Pixel> AsTileMut<T> for Frame<T> {
```
lu-zero: AsTileMut should not exist. Have both methods in an AsTile trait.

v_frame/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+mod serialize {
+  cfg_if::cfg_if! {
+     if #[cfg(feature="serialize")] {
+       pub use serde::*;
+      } else {
+        pub use noop_proc_macro::{Deserialize, Serialize};
+     }
+  }
+}
+pub mod preamble {
```
shssoichiro: Typically in Rust this is called `prelude`. For consistency, I think it makes sense to use the community-agreed term here.

lu-zero: Agreed, I'm not sure what I thought when I typoed it :)

lu-zero:
The tests need to be fixed.

vibhoothiiaanand:
Let us see now if tests are passing

PR #2165, opened 2020-02-11
Thread CDEF with tiles

### LINKED ISSUE #2056: Thread CDEF application with tiles
We can split the frame into arbitrary tiles during the application, and just run the CDEF application on those. Should be pretty easy.

takehirokj:
Closes #2056.
~1.2% performance improvements with 2x2 tiles in [AWCY](https://beta.arewecompressedyet.com/?job=master_tile_2_2%402020-02-09T04%3A02%3A27.514Z&job=thread_cdef_04%402020-02-11T12%3A18%3A23.482Z).

There's difference in the encoding output between with/wo threading when tile size is more than 2.
The output is same when tile is 1x1.
I'm not sure this is OK and need review.

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -401,6 +401,71 @@ pub fn cdef_padded_tile_copy<T: Pixel>(
   out
 }
 
+// Allocates and returns a new Frame with its own memory that is
+// padded with the input frame
+pub fn cdef_padded_frame_copy<T: Pixel>(in_frame: &Frame<T>) -> Frame<u16> {
+  let mut out: Frame<u16> = Frame {
+    planes: {
+      let new_plane = |pli: usize| {
+        Plane::new(
+          in_frame.planes[pli].cfg.width,
+          in_frame.planes[pli].cfg.height,
+          in_frame.planes[pli].cfg.xdec,
+          in_frame.planes[pli].cfg.ydec,
+          2,
+          2,
+        )
+      };
+      [new_plane(0), new_plane(1), new_plane(2)]
+    },
+  };
+
+  for p in 0..3 {
+    let rec_w = in_frame.planes[p].cfg.width;
+    let rec_h = in_frame.planes[p].cfg.height;
+    let mut out_region = out.planes[p].region_mut(Area::Rect {
+      x: -2,
+      y: -2,
+      width: rec_w + 4,
+      height: rec_h + 4,
+    });
+    for row in 0..out_region.rect().height {
+      // pad first two elements of current row
+      {
+        let out_row = &mut out_region[row][..2];
+        out_row[0] = CDEF_VERY_LARGE;
+        out_row[1] = CDEF_VERY_LARGE;
+      }
+      // pad out end of current row
+      {
+        let out_row = &mut out_region[row][rec_w + 2..];
+        for x in out_row {
+          *x = CDEF_VERY_LARGE;
+        }
+      }
+      // copy current row from input frame if we're in data, or pad if we're in first two rows/last N rows
+      {
+        let out_row = &mut out_region[row][2..rec_w + 2];
+        if row < 2 || row >= rec_h + 2 {
+          for x in out_row {
+            *x = CDEF_VERY_LARGE;
+          }
+        } else {
+          let in_stride = in_frame.planes[p].cfg.stride;
+          for (x, y) in out_row.iter_mut().zip(
+            in_frame.planes[p].data_origin()
+              [(row - 2) * in_stride..(row - 1) * in_stride]
+              .iter(),
+          ) {
+            *x = u16::cast_from(*y);
+          }
+        }
+      }
+    }
+  }
+  out
+}
+
 // We assume in is padded, and the area we'll write out is at least as
 // large as the unpadded area of in
 // cdef_index is taken from the block context
@@ -542,6 +607,7 @@ pub fn cdef_filter_frame<T: Pixel>(
   let fb_height = (rec.planes[0].cfg.height + 63) / 64;
 
   // Construct a padded copy of the reconstructed frame.
+  let in_padded_frame = cdef_padded_frame_copy(rec);
   let mut cdef_frame: Frame<u16> = Frame {
     planes: {
       let new_plane = |pli: usize| {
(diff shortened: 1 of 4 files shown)


lu-zero:
Looks like there is a little mistake in how a corner case is managed.

src/encoder.rs
```diff
@@ -2948,17 +2948,29 @@ fn encode_tile_group<T: Pixel>(
     let pre_cdef_frame = fs.rec.clone();
 
     /* TODO: Don't apply if lossless */
-    let rec = Arc::make_mut(&mut fs.rec);
     if fi.sequence.enable_cdef {
-      cdef_filter_frame(fi, rec, &blocks);
+      ti.tile_iter_mut(fs, &mut blocks)
```
tdaede: For purposes of CDEF application, instead of using the existing tileinfo, it might make sense to create a new one. There is no reason that we need to use the same number of tiles for CDEF as we do for the rest of the encode - in fact you might often want to use more tiles for CDEF because it has no compression loss (unlike bitstream tiles).

src/cdef.rs
```diff
@@ -558,22 +559,23 @@ pub fn cdef_filter_superblock<T: Pixel>(
 // Reference: http://av1-spec.argondesign.com/av1-spec/av1-spec.html#cdef-process
 #[hawktracer(cdef_filter_frame)]
 pub fn cdef_filter_frame<T: Pixel>(
```
tdaede: Renaming this to cdef_filter_tile() might be clearer.

src/cdef.rs
```diff
@@ -613,34 +615,24 @@ pub fn cdef_filter_frame<T: Pixel>(
             *x = CDEF_VERY_LARGE;
           }
         } else {
-          let rec_stride = rec.planes[p].cfg.stride;
-          for (x, y) in cdef_row.iter_mut().zip(
-            rec.planes[p].data_origin()
-              [(row - 2) * rec_stride..(row - 1) * rec_stride]
-              .iter(),
-          ) {
+          for (x, y) in cdef_row.iter_mut().zip(rec.planes[p][row - 2].iter())
```
tdaede: The old version looks like it iterated over only the top two lines, but it looks like this iterates over all the lines?

src/encoder.rs
takehirokj: Yes. I will create new tileinfo and decide number of tiles independent of the number for encode.

src/cdef.rs
takehirokj: I will update the name to cdef_filter_tile().

src/cdef.rs
takehirokj: The old version iterates over each one line as "range bounded inclusively below and exclusively above (start..end)." as in [doc](https://doc.rust-lang.org/std/ops/struct.Range.html).
New version also iterates over one line because the index of PlaneRegionMut returns rows.

takehirokj:
ts plane sboでアクセス
Sbo global

takehirokj:
I found there's problem in old commit.

cdef_frame is padded in each tile.
However some of padded region is referred in cdef filter process.

This is explained with 4x4 simple example as below:
(p: pad)

1 2 3 4
5 6 7 8
9 0 1 2
3 4 5 6

This plane is separated to 2x1 tiles as

1 2 3 4
5 6 7 8

and

9 0 1 2
3 4 5 6

These were padded as
p p p p p p p p
p p p p p p p p
p p 1 2 3 4 p p
p p 5 6 7 8 p p
p p P P P P p p
p p P P P P p p

and

p p P P P P p p
p p P P P P p p
p p 9 0 1 2 p p
p p 3 4 5 6 p p
p p p p p p p p
p p p p p p p p

Some part of padded region (shown in "P") had problem in cdef filter process.
These are normal pixel value when tile is not used. This caused the difference.

Therefore, I have updated as below:
1 2 3 4
5 6 7 8
9 0 1 2
3 4 5 6

By padding:
p p p p p p p p
p p p p p p p p
p p 1 2 3 4 p p
p p 5 6 7 8 p p
p p 9 0 1 2 p p
p p 3 4 5 6 p p
p p p p p p p p
p p p p p p p p

This is separated to 2x1 tiles as:
p p p p p p p p
p p p p p p p p
p p 1 2 3 4 p p
p p 5 6 7 8 p p
p p 9 0 1 2 p p
p p 3 4 5 6 p p
 
and

p p 1 2 3 4 p p
p p 5 6 7 8 p p
p p 9 0 1 2 p p
p p 3 4 5 6 p p
p p p p p p p p
p p p p p p p p

As a result, the cdef output becomes same as original one.

xiphmont:
Oops.  This conflict pretty heavily with another changeset  I've been working on since December.  There's no conceptual problem, but the patchet rearranges an API that gets rewritten to allow better buffer management in the loop filter RDO loop.

We'll need to coordinate how to land this; not only is my pending patchset in conflict, but I need further buffer management work after it lands that the multithreading will likely complicate.

Aside from that, great patch :-)

xiphmont:
[to be clear, the problem here is only that we're both working heavily in exactly the same chunk of code, and the changes will need to be serialized.  I'd have to undo some amount of your work above to achieve what I'm after, so I might have to claim precedence.  That said, I also want to see the multithreading land, so I'm at your disposal to help make that happen]

takehirokj:
@xiphmont, OK.
I will update and continue this patch after your preceding patchess land.
It will help if you notify after your preceding patches land.

xiphmont:
OK!  PR #2207 has landed, we can rebase this now.

src/cdef.rs
```diff
@@ -401,11 +402,76 @@ pub fn cdef_padded_tile_copy<T: Pixel>(
   out
 }
 
+// Allocates and returns a new Frame with its own memory that is
+// padded with the input frame
+pub fn cdef_padded_frame_copy<T: Pixel>(in_frame: &Frame<T>) -> Frame<u16> {
```
xiphmont: Is there a reason not to write this in terms of or just use cdef_padded_tile_copy()?  The frame can always be wrapped on the fly using Frame::as_tile()

src/cdef.rs
```diff
@@ -526,30 +593,61 @@ pub fn cdef_filter_superblock<T: Pixel, U: Pixel>(
...
 ) {
   // Each filter block is 64x64, except right and/or bottom for non-multiple-of-64 sizes.
   // FIXME: 128x128 SB support will break this, we need FilterBlockOffset etc.
-  let fb_width = (rec.planes[0].cfg.width + 63) / 64;
-  let fb_height = (rec.planes[0].cfg.height + 63) / 64;
+  let fb_width = (rec.planes[0].rect().width + 63) / 64;
+  let fb_height = (rec.planes[0].rect().height + 63) / 64;
 
-  // Construct a padded copy of the reconstructed frame.
+  // Construct a part of copy from the padded input frame.
```
xiphmont: ?  Do you mean:
// Construct a padded copy of part of the input tile?

src/cdef.rs
```diff
@@ -559,67 +657,41 @@ pub fn cdef_filter_frame<T: Pixel>(
...
-          let rec_stride = rec.planes[p].cfg.stride;
-          for (x, y) in cdef_row.iter_mut().zip(
-            rec.planes[p].data_origin()
-              [(row - 2) * rec_stride..(row - 1) * rec_stride]
-              .iter(),
-          ) {
-            *x = u16::cast_from(*y);
-          }
-        }
-      }
```
xiphmont: Oh wow, don't know how I missed this extra padding step as having become redundant.  Nice catch.  Maaaaybe... break this out as a separate change that happens first?  Just in case?

src/cdef.rs
takehirokj: This is for performance reason. I can replace it to cdef_padded_tile_copy() and the output is same. But the speed is little bit slow. Therefore, I make a function that do padded process as in original cdef_filter_frame().

src/cdef.rs
takehirokj: Yes. I will update the comment.

src/cdef.rs
takehirokj: Do you mean separating this change for padding as different PR?

src/cdef.rs
xiphmont: Ah.  Mid-term, padding goes away, so both will get removed.  I'm surprised as_tile() has that much of a performance impact (and I am using as_tile() extensively elsewhere, so now I'm intrigued as to why).

In any case, as this is temporary and there is a reason (maybe not a good one) I have no serious objection.

src/cdef.rs
xiphmont: It need not be a separate PR, just a separate commit.  

Looking at it carefully, I think you're right to remove the padding code.  However, my intuition tells me there is a possibility I could be wrong.  To make future maintenance easier in that case (even if I'm the one who has to debug it), being able to git bisect this change separately would be very useful.

takehirokj: I make it to separate commit.

src/cdef.rs
takehirokj: I have not checked which one of as_tile() or cdef_padded_tile_copy() has performance impact. For now, current approach solves this.

PR #2171, opened 2020-02-14
Disable fine directional intra prediction between speed levels 2 and 5

### LINKED ISSUE #2167: Fine directional prediction seems slower than it's worth at some speed levels
Based on local tests and AWCY runs, I get about 10-20% higher frame rate with fine directional prediction disabled on speed level 2, while the BD-rate benefit is below 1%. I think it should be disabled by default for speed levels where it does not provide a worthwhile gain.

https://beta.arewecompressedyet.com/?job=no-fine-directional-modes-s2%402020-02-07T17%3A57%3A23.932Z&job=v0.3.0-s2%402020-02-07T12%3A33%3A40.400Z

takehirokj:
Closes #2167.

~30% speed up in speed 5 in [AWCY](https://beta.arewecompressedyet.com/?job=master_s_5%402020-02-14T14%3A11%3A21.320Z&job=disable_fine_dir_s_5%402020-02-14T14%3A07%3A13.758Z).
| PSNR | PSNR Cb | PSNR Cr | PSNR HVS | SSIM | MS SSIM | CIEDE 2000 |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 0.6833 | 1.4316 | 1.3890 | 0.6723 | 0.7174 | 0.6667 | 0.8463 |

### DIFF
--- a/src/api/config.rs
+++ b/src/api/config.rs
@@ -251,6 +251,10 @@ impl fmt::Display for EncoderConfig {
         self.speed_settings.non_square_partition.to_string(),
       ),
       ("enable_timing_info", self.enable_timing_info.to_string()),
+      (
+        "fine_directional_intra",
+        self.speed_settings.fine_directional_intra.to_string(),
+      ),
     ];
     write!(
       f,
@@ -329,6 +333,9 @@ pub struct SpeedSettings {
 
   /// Enable tx split for inter mode block.
   pub enable_inter_tx_split: bool,
+
+  /// Use fine directional intra prediction
+  pub fine_directional_intra: bool,
 }
 
 impl Default for SpeedSettings {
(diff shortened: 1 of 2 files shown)


src/api/config.rs
```diff
@@ -516,6 +525,10 @@ impl SpeedSettings {
   const fn enable_inter_tx_split_preset(speed: usize) -> bool {
     speed >= 9
   }
+
+  const fn fine_directional_intra_preset(speed: usize) -> bool {
+    speed <= 1
```
rzumer:
```suggestion
    speed <= 1 || speed >= 6
```

Based on our tests, the performance penalty is not bad for speed 6 and up, so we can leave it enabled for higher speed levels too.

takehirokj: Fixed it and update the title as above.

rzumer: Great, thanks!

PR #2173, opened 2020-02-15
Add tests and benchmarks for put_8tap and prep_8tap

shssoichiro:
- Add put_8tap and prep_8tap x86_64 asm tests
- Add benchmarks for put_8tap and prep_8tap

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -8,6 +8,7 @@
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
 mod dist;
+mod mc;
 mod plane;
 mod predict;
 mod rdo;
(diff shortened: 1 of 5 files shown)


shssoichiro:
Trying to figure out why tests are failing in CI; they pass locally. :man_facepalming:

shssoichiro:
Oh, it is now consistently segfaulting locally--happens on the same test every time if I run single threaded. So, now I'm wondering if there's a bug in the 8-bit ASM...

shssoichiro:
TIL that prep_8tap AVX2 seems to fail intermittently if the plane is less than 64x64. So... hopefully that's a known thing that we handle.

PR #2177, opened 2020-02-17
Unbreak the configuration serialization

lu-zero:


shssoichiro:
Might make sense to add a test around this.

lu-zero:
I will do that later.

PR #2186, opened 2020-02-27
api: Introduce reference frame for Packets

vibhoothiiaanand:
This commit adds Reference Frame from the FrameData(Input) to the packets, so we could take this for calculation, for instance, Metrics calculations can be independent of API changes and will be cleaner to implement with fewer API breakages in favour of 1.0 Future Release.

### DIFF
--- a/src/api/util.rs
+++ b/src/api/util.rs
@@ -160,6 +160,8 @@ pub struct Packet<T: Pixel> {
   pub data: Vec<u8>,
   /// The reconstruction of the shown frame.
   pub rec: Option<Arc<Frame<T>>>,
+  /// The Reference Frame
+  pub source: Option<Arc<Frame<T>>>,
   /// The number of the input frame corresponding to the one shown frame in the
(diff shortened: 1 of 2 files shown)


src/api/util.rs
```diff
@@ -160,6 +160,8 @@ pub struct Packet<T: Pixel> {
   pub data: Vec<u8>,
   /// The reconstruction of the shown frame.
   pub rec: Option<Arc<Frame<T>>>,
+  /// The Reference Frame
+  pub ref_f: Option<Arc<Frame<T>>>,
```
lu-zero: `input` or `source` might be a better name in the end.

To be discussed if `Option<(Arc<Frame<T>>, Arc<Frame<T>>)` wouldn't be more future proof even if it would have the problem of documenting what is what.

shssoichiro: Are you referring to putting rec and source in one element? I'm not sure that that would be cleaner/easier to reason about.

PR #2198, opened 2020-03-09
Add mimalloc allocator support

lu-zero:
It is yet to be tested


shssoichiro:
I'd like some tests/benchmarks/etc. I tested a similar change for jemalloc a while ago, and found that it provides no significant change (at least on Linux), so opted to keep the default allocator.

tdaede:
TBH I think a better course of action is removing expensive mallocs if there are any.

PR #2203, opened 2020-03-18
Bump sccache to 0.2.13

vibhoothiiaanand:


barrbrain:
We should also try bumping it in `.travis/install-sccache.sh`.

PR #2209, opened 2020-03-23
Fix two-pass encoding and create basic CLI tests

### LINKED ISSUE #2208: process_frame() refactor broke 2-pass encoding
https://github.com/xiph/rav1e/commit/45ef0af03575180a589351dd88615c863a6bed62 causes only one frame to be encoded in my 2-pass encoding tests (happens on the first pass, so could also occur for single-pass). After reverting this commit only, master works as expected.

```
<derf> I mean, by the time you're at Result<Option<Vec<...>>, ...>, and all of different possibilities (Ok(None/Some(empty/full))/Err) have different meanings for the caller, maybe it is time to start thinking about an enum or something to tell people what is actually going on.
<derf> That patch collapsed two of the meanings (None and an empty Vec), but didn't actually preserve the existing code's behavior in those cases.
```

shssoichiro:
Closes #2208


tests/.gitignore
```diff
@@ -0,0 +1 @@
+!small_input.y4m
```
rzumer: Suggest adding a trailing newline for consistency.

tests/binary.rs
```diff
@@ -0,0 +1,93 @@
...
+    .unwrap();
+    let mut data = Vec::new();
+    input.read_to_end(&mut data).unwrap();
+    data
+  }
+
+  fn get_tempfile_path(extension: &str) -> PathBuf {
+    let mut path = temp_dir();
+    let filename =
+      thread_rng().sample_iter(&Alphanumeric).take(12).collect::<String>();
```
rzumer: Might be personal preference, but I think tests should avoid using random values. A second opinion would be nice.

tests/small_input.y4m
```diff
@@ -0,0 +1,7 @@
+YUV4MPEG2 W64 H64 F25:1 Ip A16:9 C420jpeg XYSCSS=420JPEG
```
rzumer: What is the source of this file?

shssoichiro: objective-1-fast blue_sky scaled down to 64x64

shssoichiro: (To clarify: I tried generating a test file in Rust using random values, but kept failing to get the headers right, so I opted to let ffmpeg do the work)

tests/binary.rs
```diff
@@ -0,0 +1,93 @@
...
+      "{}/tests/small_input.y4m",
+      env!("CARGO_MANIFEST_DIR")
+    ))
+    .unwrap();
+    let mut data = Vec::new();
+    input.read_to_end(&mut data).unwrap();
+    data
+  }
+
+  fn get_tempfile_path(extension: &str) -> PathBuf {
```
lu-zero: You may use [tempfile](https://docs.rs/tempfile/3.1.0/tempfile/index.html) maybe?

tests/small_input.y4m
rzumer: Sounds good.

tests/binary.rs
shssoichiro: I looked into that, but tempfile doesn't have an option to just generate a filename--it also creates the file. For this, we just want a filename in the temp directory that rav1e can write to.

.github/workflows/rav1e.yml
```diff
@@ -273,8 +273,8 @@ jobs:
            toolchain: stable
          - conf: cargo-test
            toolchain: stable
-         - conf: cargo-c
-           toolchain: stable-x86_64-pc-windows-gnu
+#         - conf: cargo-c
```
lu-zero: Please mention https://github.com/rust-lang/rust/issues/68518 and https://github.com/rust-lang/rust/pull/67429

Or we could set the `LIBRARY_PATH` to sidetrack the problem.

lu-zero: Setting `LIBRARY_PATH` does not work. I would build cargo-c with the msvc toolchain and run it for the gnu toolchain.

tests/binary.rs
lu-zero: We could use a directory and use non-random filenames.

PR #2211, opened 2020-03-25
Update README.md

rzumer:
* Revise the list of features
* Remove rav1erepl.rs from the Design section

More suggestions/corrections welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,17 @@ rav1e is an AV1 video encoder. It is designed to eventually cover all use cases,
 
 ## Features
 
-* Intra and inter frames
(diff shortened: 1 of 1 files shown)


rzumer:
@vibhoothiiaanand 
Agreed with most of these.

Fine directional prediction I think is covered in the prediction modes line. 

The intra edge filter might fit into a new line about the different supported filters, but it doesn't bring a large improvement so it's probably not worth mentioning on its own.

I'll update this with the rest in a bit.

lu-zero:
Can we land it now? (and then @vibhoothiiaanand patch)

PR #2215, opened 2020-04-01
Version bump

lu-zero:


negge:
Not right to 0.4.1 ?

rzumer:
I commented on this in a call last Friday, but forgot to write it down.

I don't think we're ready for a 0.4 release. There are some bugfixes introduced recently that could justify a v0.3.2, but actual improvements to performance and quality are negligible.

I also don't see why we would jump from 0.3 straight to 0.4.1.

shssoichiro:
I commented on this in IRC but may as well post here for posterity.

The reasoning for 0.4 is in regards to semver--there was a change to the Rust API, which constitutes a breaking change and a bump to 0.4. We could do 0.3.2 as well, but it would mean cherry-picking out the commit with API changes. Agreed there's no reason to jump to 0.4.1.

rzumer:
Ah right, I forgot about the API change. Sounds good then.

rzumer:
@ycho found an apparent issue with speed level 10 that was introduced recently, which could be a blocker. Since regressions are not uncommon, once we decide to cut a release, we should test the master (or release) branch under multiple different configurations (e.g. speed levels, keyframe interval, modes), and on different test sets to include high bit depth, 4:2:2 and 4:4:4.

ycho:
I see. However, the issue I am seeing seems more related to other part not specific to s 10. Because awcy runs for ra1ve has halted since March 5, as I reported in #daala irc.
https://beta.arewecompressedyet.com/?job=master-de515e0875424836c28462fff85851c610157944&job=master-e43d84490302c3dcc3fe2247dbd9c5324ea475ca
seems the last successful rav1e rune.

EwoutH:
Is there an expected or targeted release date of rav1e 0.4? I'm looking to update [libavif](https://github.com/AOMediaCodec/libavif) to enable monochrome support with rav1e. If 0.4 is coming soon I will wait, otherwise I will use the last/next pre-release.

PR #2217, opened 2020-04-02
Split the rate control twopass_in() in smaller functions

lu-zero:
`twopass_in()` is a function that does too much at the same time:
-  It initializes the second pass rc.
-  It provides the user the amount of bytes the same function would like to consume next
-  It consumes and parses the initial global header, called in the code summary.
-  It consumes and parses the per-frame information, depending on the configuration either a single per-frame data or a variable number of them in a single pass.

This PR addresses this problem by splitting the function in smaller ones with, I hoped, self-describing names:

- [x] rate control initialization (`twopass_init`)
- [x] summary parsing (`twopass_in_summary`)
- [x] frame data parsing (`twopass_in_frame_data`)

### DIFF
--- a/src/rate.rs
+++ b/src/rate.rs
@@ -742,6 +742,20 @@ impl RCState {
     }
   }
 
+  pub(crate) fn select_first_pass_qi(
+    &self, bit_depth: usize, fti: usize,
+  ) -> QuantizerParameters {
+    // Adjust the quantizer for the frame type, result is Q57:
+    let log_q = ((self.pass1_log_base_q + (1i64 << 11)) >> 12)
+      * (MQP_Q12[fti] as i64)
+      + DQP_Q57[fti];
+    QuantizerParameters::new_from_log_q(
+      self.pass1_log_base_q,
+      log_q,
+      bit_depth,
+    )
+  }
+
   // TODO: Separate quantizers for Cb and Cr.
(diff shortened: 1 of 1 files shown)


rzumer:
Can you please clarify the goal of this PR, what it achieves, how it improves the code? If anything it is useful for documentation purposes. Considering it adds more lines than it removes, at first glance it looks to add complexity more than remove it.

I also find the term "summary" unclear. In my opinion, if one cannot understand a function's purpose just by reading the name, it needs to be documented. If you could specify what this "summary" is in the code that would be helpful. Maybe there is a better term that could be used.

lu-zero:
The term `summary` is already used in the codebase. If you have opinions on how to name it differently feel free to suggest a different name, but it is unrelated to this PR.

`twopass_in()` is a function that does too much:
- It initializes the second pass rc.
- It provides the user the amount of bytes the same function would like to consume next
- It consumes and parses the initial global header, called in the code `summary`.
- It consumes and parses the per-frame information, depending on the configuration either a single per-frame data or a variable number of them in a single pass.

This PR addresses this problem by splitting the function in smaller ones with, I hoped, self-describing names.

I started working on this to debug why my implementation of multipass channel API was not working: I wrongly assumed the function is able consume partial buffers, that's not currently always the case.

PR #2218, opened 2020-04-03
Add a test for reservoir-frame-delay

lu-zero:
The default (unset) and the option set exercise two different codepaths in the
rate control.

### DIFF
--- a/tests/binary.rs
+++ b/tests/binary.rs
@@ -90,4 +90,39 @@ mod binary {
       .assert()
       .success();
   }
+  #[test]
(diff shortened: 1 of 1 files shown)


shssoichiro:
Not specifically related to this PR, but I hope to eventually test that the files produced by these integration tests decode correctly. But I avoided that to start off with, to avoid adding libaom/dav1d dependencies outside of the encode/decode test suite.

lu-zero:
That would be interesting as well indeed.

PR #2223, opened 2020-04-06
Add NASM minimum version check

shssoichiro:


shssoichiro:
Fixed up the logic to parse the `nasm -v` output correctly, then finished fighting clippy ^^;

PR #2228, opened 2020-04-10
Unbreak the serialize feature

lu-zero:


barrbrain:
When I run `cargo check --features=serialize` with this change, I see an error:
```
error[E0463]: can't find crate for `num_derive`
  --> v_frame/src/pixel.rs:12:5
   |
12 | use num_derive::FromPrimitive;
   |     ^^^^^^^^^^ can't find crate
```

PR #2230, opened 2020-04-11
Do not try to rebuild the assembly files if not needed

lu-zero:
The build is triggered only if the assembly files or the configuration
changes.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -14,7 +14,7 @@ use rustc_version::{version, Version};
 #[allow(unused_imports)]
 use std::env;
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::exit;
 
 #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -30,44 +30,84 @@ fn rerun_dir<P: AsRef<Path>>(dir: P) {
...
+
+  let hash_path = Path::new(&out_dir).join("asm.hash");
+
+  if let Ok(old_hash) = std::fs::read(&hash_path) {
+    if old_hash == hash {
+      return true;
+    }
+  }
+
+  std::fs::write(hash_path, &hash[..]).unwrap();
```
barrbrain, line 64: There is a race condition here since the hash file is written before calling nasm.
If nasm is interrupted, future builds will not call nasm again as the hash already matches.
Conversely, if the hash is written after and there is an interruption then nasm will be called again on the next build.

tdaede:
What does cc-rs do?

build.rs
lu-zero: I guess the latter condition even if sub-optimal is better than the current status. Thank you for spotting the problem. :)

lu-zero:
> What does cc-rs do?

Nothing different from `nasm-rs`.

build.rs
lu-zero: Done :)

tdaede:
I mean, does it do this same file hash thing? What about dep generation?

lu-zero:
Doesn't do anything in that regard that I could see.

PR #2232, opened 2020-04-13
Print simpler error messages from the CLI.

tdaede:


tdaede:
@lu-zero @shssoichiro updated to cover all errors

PR #2236, opened 2020-04-14
Add compound prediction mode variants for drl=2 and drl=3.

tdaede:


src/context.rs
```diff
@@ -3529,7 +3518,23 @@ impl<'a> ContextWriter<'a> {
     };
 
     assert!(mode >= PredictionMode::NEAREST_NEARESTMV);
-    let val = mode as u32 - PredictionMode::NEAREST_NEARESTMV as u32;
+    let val = match mode {
```
lu-zero, line 3521: what about making a small method `to_u8` or such?

src/stats.rs
```diff
@@ -21,9 +21,9 @@ pub struct EncoderStats {
   /// Stores count of pixels belonging to each transform type in this frame
   pub tx_type_counts: [usize; TX_TYPES],
   /// Stores count of pixels belonging to each luma prediction mode in this frame
-  pub luma_pred_mode_counts: [usize; PREDICTION_MODES],
+  pub luma_pred_mode_counts: Vec<usize>,
   /// Stores count of pixels belonging to each chroma prediction mode in this frame
-  pub chroma_pred_mode_counts: [usize; PREDICTION_MODES],
+  pub chroma_pred_mode_counts: Vec<usize>,
```
lu-zero, line 26: why this?

src/context.rs
tdaede: can't derive #[Debug] for arrays larger than 32

src/stats.rs
tdaede: can't derive #[Debug] for arrays longer than 32

src/context.rs
tdaede: could but this is the only case where this numeric conversion is used

src/stats.rs
tdaede: I don't know why this derives debug anyway, maybe it shouldn't

lu-zero: Probably Debug can be dropped.

src/context.rs
lu-zero: Ok.

shssoichiro:
Does this impact metrics? If so, AWCY incoming?

shssoichiro:
Also, this seems to depend on #2237? There's a lot of overlap, at least. Is that correct?

tdaede:
No, #2237 is the one that impacts metrics.

src/stats.rs
tdaede: Unfortunately can't do that as Packet also derives Debug.

src/predict.rs
```diff
@@ -61,9 +61,11 @@ pub static RAV1E_INTER_COMPOUND_MODES: &[PredictionMode] = &[
   PredictionMode::NEW_NEWMV,
   PredictionMode::NEAREST_NEWMV,
   PredictionMode::NEW_NEARESTMV,
-  PredictionMode::NEAR_NEARMV,
+  PredictionMode::NEAR_NEAR0MV,
```
shssoichiro, line 64: Should this be updated to add the new modes?

src/rdo.rs
```diff
@@ -1114,7 +1114,7 @@ fn inter_frame_rdo_mode_decision<T: Pixel>(
           [MotionVector::default(); 2]
         }
       }
-      PredictionMode::NEAR0MV | PredictionMode::NEAR_NEARMV => {
+      PredictionMode::NEAR0MV | PredictionMode::NEAR_NEAR0MV => {
```
shssoichiro, line 1117: Should this be updated to add the new modes?

barrbrain: When the new modes are added to RAV1E_INTER_COMPOUND_MODES in #2237, the logic for those modes will be added here.

src/predict.rs
barrbrain: The new modes will be added in #2237, which extends the search. This PR is for preparing the accounting.

Cargo.toml
```diff
@@ -73,7 +73,7 @@ v_frame = { version = "0.1", path = "v_frame/" }
 av-metrics = { version = "0.4", optional = true }
 rayon = "1.0"
 toml = { version = "0.5", optional = true }
-arrayvec = "0.5"
+arrayvec = { version = "0.5", features = ["serde", "array-sizes-33-128"] }
```
barrbrain, line 76: Nice, this resolves my concern about introducing new allocations.

PR #2237, opened 2020-04-14
Enable NEAR_NEAR1MV compound mode. 

tdaede:


shssoichiro:
Does this impact metrics? If so, AWCY incoming?

tdaede:
There is AWCY info in the commit message.

barrbrain:
Now that this includes the speed level check, we should measure the impact at speed level 3, the fastest level for which near modes are currently enabled. I expect we still see a small improvement in BD-rate for a proportionate increase in encoding time.

tdaede:
Seems the most recent rebase actually broke it:
https://beta.arewecompressedyet.com/?job=near0b-2020-06-30_162405-39390f84
I'll see if I can fix.

barrbrain:
> Seems the most recent rebase actually broke it:

Embarrassingly, when analysing this PR locally... I missed checking out the commit. :frowning_face:

src/rdo.rs
```diff
@@ -1124,7 +1124,16 @@ fn inter_frame_rdo_mode_decision<T: Pixel>(
           true,
         ));
         for &x in RAV1E_INTER_COMPOUND_MODES {
-          inter_mode_set.push((x, ref_frames_set.len() - 1));
+          // exclude any NEAR mode based on speed setting
+          if fi.config.speed_settings.include_near_mvs || !x.has_nearmv() {
+            let mv_stack_idx = ref_frames_set.len() - 1;
+            // exclude NEAR modes if the mv_stack is too short
+            if !(x.has_nearmv()
+              && x.ref_mv_idx() >= mv_stacks[mv_stack_idx].len())
```
barrbrain, line 1132: The push of `mv_stack` into `mv_stacks` occurs after this loop. It is not clear that `mv_stack_idx`is the right index here anyhow.
```suggestion
              && x.ref_mv_idx() >= mv_stack.len())
```

src/predict.rs
```diff
@@ -183,6 +184,21 @@ impl PredictionMode {
       || self == PredictionMode::NEW_NEAR1MV
       || self == PredictionMode::NEW_NEAR2MV
   }
+  pub fn ref_mv_idx(self) -> usize {
```
barrbrain, line 187: For consistency with LTO-enabled builds and the neighbouring methods of similar complexity, we may add `#[inline]` here.

src/rdo.rs
tdaede: Indeed. I think I broke this when changing it to exclude NEAR.

PR #2238, opened 2020-04-15
Add the unstable feature and document it

lu-zero:


shssoichiro:
Is this used by anything at the moment?

lu-zero:
It will be used by the channel-api and the new rc api

PR #2243, opened 2020-04-17
CI: Upgrade actions/checkout to v2

barrbrain:
Performance improved, authentication persisted, local branch created.

See [What's New](https://github.com/marketplace/actions/checkout?version=v2.0.0#whats-new) on GitHub Marketplace.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: windows-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


vibhoothiiaanand:
Can we use [v2.1](https://github.com/marketplace/actions/checkout?version=v2.1.0) which is the latest version which is being tagged?

shssoichiro:
It should bump to 2.1 automatically, it looks to be semver-major based.

PR #2245, opened 2020-04-18
Distinguish Arc clones from expensive clones

shssoichiro:
This should make future performance refactoring easier,
by allowing us to more easily identify which clones
we would want to potentially eliminate.

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -322,7 +322,7 @@ impl<T: Pixel> ContextInner<T> {
       let lookahead_frames = self
         .frame_q
         .range(self.next_lookahead_frame - 1..)
-        .filter_map(|(&_input_frameno, frame)| frame.clone())
+        .filter_map(|(&_input_frameno, frame)| frame.as_ref().map(Arc::clone))
         .collect::<Vec<_>>();
 
       if is_flushing {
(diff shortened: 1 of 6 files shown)


src/api/internal.rs
```diff
@@ -322,7 +322,9 @@ impl<T: Pixel> ContextInner<T> {
       let lookahead_frames = self
         .frame_q
         .range(self.next_lookahead_frame - 1..)
-        .filter_map(|(&_input_frameno, frame)| frame.clone())
+        .filter_map(|(&_input_frameno, frame)| {
+          frame.as_ref().map(|f| Arc::clone(f))
```
linkmauve, line 326: Small nit, you can remove the closure and replace it with `Arc::clone` itself:
```suggestion
          frame.as_ref().map(Arc::clone)
```

rslife:
(*just an opinion from an outsider*)

If I wanted to do this, I would use an extension trait for it (and maybe `#![no_implicit_prelude]` as any solution without it is unenforceable).

So, something like this:

``` rust
use std::sync::Arc;
use std::rc::Rc;

// maybe a better name is needed
trait IncrClone: Clone {
    fn incr_clone(&self) -> Self {
        self.clone()
    }
}

impl<T> IncrClone for Arc<T> {}
impl<T> IncrClone for Rc<T> {}
```

Yes it's less explicit, but it's more generic, the diff becomes simpler, chain-abilitiy is not lost, and future diffs would be smaller if anything other than `std::sync::Arc` starts to get used.

---

And maybe another trait for more distinction (if inner clones are used at all):

``` rust
trait InnerClone<T: Clone>: Deref<Target=T> {
    fn inner_clone(&self) -> T {
        self.deref().clone()
    }
}

impl<T: Clone> InnerClone<T> for Arc<T> {}
impl<T: Clone> InnerClone<T> for Rc<T> {}
```

PR #2246, opened 2020-04-22
Further cleanup the rate control code

lu-zero:
Separate the rate control logic from the serialization/deserialization logic.

- [x] Have a RCSummary struct as we have a RCFrameMetrics already
- [x] Move all the deserialization code in a separate struct
  - [x] Return a RCSummary instead of directly touching the state
- [ ] Move all the serialization code in a separate struct

### DIFF
--- a/src/rate.rs
+++ b/src/rate.rs
@@ -1444,7 +1444,7 @@ impl RCState {
       if ntus_total < 1 {
         return Err(());
       }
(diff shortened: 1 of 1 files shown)


lu-zero:
I added some documentation as you suggested, I'll land this once the CI pass completes.

PR #2247, opened 2020-04-23
[WIP][review this thing already]Spatial AQ

cyanreg:
The people who need to look at it know about it, so I'm not going to explain.
I've forgotten most of what it did anyway because its been gathering rust for so long.

### DIFF
--- a/src/activity.rs
+++ b/src/activity.rs
@@ -136,7 +136,7 @@ impl ActivityMask {
             let element = variances.get_mut(y * (width >> granularity) + x);
             match element {
                 Some(x) => {
-                    *x = (*x * temporal_act) * 0.6472725012 - 24.32294559;
+                    *x = ((*x * temporal_act) * 0.6472725012 - 24.32294559).max(0f64);
                     avg_var += *x;
                     max = max.max(*x);
                 }
--- a/src/activity.rs
+++ b/src/activity.rs
@@ -135,7 +135,7 @@ impl ActivityMask {
             let element = variances.get_mut(y * (width >> granularity) + x);
             match element {
                 Some(x) => {
-                    *x = *x;
+                    *x = *x + temporal_act;
                     avg_var += *x;
                     max = max.max(*x);
                 }
--- a/src/activity.rs
+++ b/src/activity.rs
@@ -63,6 +63,10 @@ impl ActivityMask {
             _ => unreachable!(),
         };
 
+    let tot_bins = variances.len();
+    let mut avg_var = 0f64;
+    let mut max = 0f64;
+
     for y in 0..height >> act_granularity {
       for x in 0..width >> act_granularity {
         let block_rect = Area::Rect {
@@ -98,6 +102,8 @@ impl ActivityMask {
         }
 
         sum_f /= (tot_pix - 4*4) as f64;
+        avg_var += sum_f;
+        max = max.max(sum_f);
 
         /* Copy down to granularity */
         for i in 0..(1 << (act_granularity - granularity)) {
@@ -113,8 +119,11 @@ impl ActivityMask {
       }
     }
 
-    let mut avg_var = 0f64;
-    let mut max = 0f64;
+    let old_avg_var = avg_var / (tot_bins as f64);
+    let old_max = max;
+
+    avg_var = 0f64;
+    max = 0f64;
 
     /* Merge temporal activity */
     for y in 0..fi.h_in_imp_b {
@@ -129,13 +138,13 @@ impl ActivityMask {
                 } else {
                     let strength = 1.0; // empirical, see comment above
                     let frac = (intra_cost + propagate_cost) / intra_cost;
-                    frac.powf(strength / 3.0) * 1.0f64
+                    frac.powf(strength / 3.0)
                 };
 
             let element = variances.get_mut(y * (width >> granularity) + x);
             match element {
                 Some(x) => {
-                    *x = *x;
+                    *x = (*x + temporal_act * old_avg_var) * 0.940799829805f64 - 19.7703361851f64;
                     avg_var += *x;
                     max = max.max(*x);
                 }
@@ -147,7 +156,7 @@ impl ActivityMask {
 //    println!("Avg var = {}", avg_var);
 
     let mut seg_bins = [0usize; 8];
-    for i in 0..variances.len() {
+    for i in 0..tot_bins {
         let element = variances.get_mut(i);
         match element {
             Some(x) => {
@@ -157,8 +166,6 @@ impl ActivityMask {
             None => unreachable!(),
         }
     }
-
-    let tot_bins = variances.len();
 
     avg_var /= tot_bins as f64;
     avg_var /= max;
--- a/src/activity.rs
+++ b/src/activity.rs
@@ -185,7 +185,7 @@ impl ActivityMask {
     let res = self.variances.get((x >> self.granularity) + dec_width * (y >> self.granularity));
     match res {
         /* avg_var is inversely proportional to the scaling necessary */
-        Some(val) => return (*val * self.var_scale) * (1500f64/((self.avg_var * self.var_scale).log(2f64))),
+        Some(val) => return (*val * self.var_scale) * (1300f64/((self.avg_var * self.var_scale).log(2f64))),
         None => unreachable!(),
     }
   }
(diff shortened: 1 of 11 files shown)


src/activity.rs
```diff
@@ -11,107 +11,182 @@ use crate::frame::*;
...
-        let variance: f64 = block
-          .rows_iter()
-          .flatten()
-          .map(|&pix| {
-            let pix: i16 = CastFromPrimitive::cast_from(pix);
-            (pix as f64 - mean).powi(2)
-          })
-          .sum::<f64>();
-        variances.push(variance);
+        let src = &mut src_storage.data[..tx_size.area()];
```
tdaede, line 79: I know you mentioned on the call you were converting this to some other sort of "hybrid"?

With this current metric, you're grabbing only the high frequency bins of a DCT. It is an interesting approximation of the contrast sensitivity metric used in PSNR-HVS, however I think it has two problems:

- The most expensive part is the forward transform. Instead of dropping the lowest coefficients, you could actually weight most of the coefficients instead.
- It is hard to evaluate this metric standalone as the patch also adds the very first usage of it.

I think it would be best to convert this to a plain pixel-domain variance and test/land the patch with that, and then upgrade it to e.g. a frequency selective DCT variance next, showing the improvements on AWCY. That would also help directly evaluate alternatives against each other, e.g. you could also implement this faster but less accurately with a hadamard transform. I think a plain variance is the best baseline to compare these options to.

cyanreg:
> Review of activity.rs and/or #2247
> Sorry for taking so long on this review. The actual bulk of the comments are on the original tree you referred me to on Feb. 18. The older pull request looked like a WIP with lots of partial checkpoint commits and reverts and so on, and not something cleaned up for review. Obviously when just looking at the files on the tip of a branch, it's hard to know which parts are things you were actually working on and which issues were pre-existing in the code, so apologies if I comment on stuff that is not your fault.
> 
> Big picture:
> 
> There is a lot going on here, and it is headed in a different direction than David Barr, Guillaume, and I have been pushing things. Our goal is to move the psychovisual tuning into a place where each component is able to be understood separately, and makes sense in a consistent overall framework. Your work mixes a lot of things (which in and of itself is not bad), but does not show clear justifications, measurements, or at least document the testing that went into each of those things (which is the part that I am less excited about). That makes it really hard for others (specifically, me) to evaluate. It would also make it hard for other people to pick up and work on your stuff.
> 
> There are three specific things that we know work well:
> 
> 1) Weighting distortion by pixel variance (actual variance!) raised to the 1/3rd power, scaled by some overall tuning constant (see x264 and also Jean-Marc's Journal of Dubious Results). There's literally years of work showing that this produces good results.
> 
> 2) A measure of temporal RDO that also scales distortion in a manner consistent with the original x264 work. There are grounds to believe that what x264 did could be done better, but so far we have (a) not actually implemented the original correctly and (b) achieved worse results. Guillaume had work to move us in this direction, picking a scale factor that matches the original x264 implementation and achieving superior results to the current ad-hoc method, in a framework that makes consistent sense with all of the other pieces of the puzzle.
> 
> 3) Quantizer (actual quantizer, not quantizer index) adjustments are directly proportional to the square root of distortion scaling.
> 
> You have combined (1) and (2) in a measure that sums them instead of multiplying them, which doesn't match my first inclination (to multiply independent contributing factors instead of summing them).
> 
> To help think about which approach is more likely to be a useful model, one experiment to consider is the "completely still scene" test. Imagine a constant image shown for N frames. Obviously the best way to code this is to spend all your bits on the initial keyframe, and then SKIP everything in all subsequent frames. If you hold the bitrate constant and increase the number of frames showing the still image, the quantizers used for the initial frame should decrease. Specifically, one would expect the quantizer to decrease by a uniform factor in all blocks. That is true under the multiplicative combination scheme, but it is not true under additive one.
> 
> You are also introducing a new, non-variance based measure to replace (1). Maybe your proposal is actually better, but I don't have any way to tell. What I would want to see to judge that are some experiments or test results on how that compares to the variance-based measure we know gives good results. Specifically, how _just_ that piece compares.
> 
> With respect to (3), I tried to take a peek into segmentation.rs (even though that wasn't the file you asked me to review) to try to figure out how you were coming up with quantizer offsets from this, but you seem to be operating on quantizer indices directly, and doing so in a linear relation relative to what you scale distortion by. The quantizer index to quantizer mapping tables in AV1 are not consistent, in the sense that the relationship is linear for some sections of the table and exponential for others. So trying to make adjustments to those indices directly confounds analysis, and will be suboptimal for some part of the curve.
> 
> There's still some work to do to unravel the quantizer table mapping in AV1. My original hope was that we could just do whatever we wanted to pick quantizers and then figure out how to map that to quantizer indices, but because segmentation only gives us a single offset to quantizer index for all six ({AC,DC}x{color plane}) quantizers, we can only approximate that. That means we need a combined measure for the mismatch between the quantizers we want and the quantizers we can actually achieve with AV1. The "average quantizer theory" I had been working on in av1-metric.pdf is one approach to doing that, but there are still emperical measurements that need to be done to support that (specifically, getting average AC and DC variance of the prediction residual in each plane with our chosen chroma scalings and the efficacy of AV1's predictors, and seeing how that varies with quantizer). If I (or anyone else who wants to help) can make progress on that, though, it also directly extends to supporting quantization matrices in the future.
> 
> I think the optimization you are doing in segmentation.rs to pick segment quantizers is over-complicated, too. A much simplier approach is to pick one segment for the base quantizer and two segments for a somewhat larger and somewhat smaller quantizer. If you know your distortion scaling, then you know what quantizers to use, based on (3) above. That is something that I have used in the past and know works. You can expand that to adjust the number of segments at higher or lower bitrates (it's a vector quantization problem), but the above is enough to get good results. I would start there and then worry about making something more complicated once things are working and everyone is happy with them (and again, this is something you can test and show improvements for in isolation).
> 
> Small picture:
> 
> activity.rs:39
> 
> Why are you always adding (1 << act_granularity) instead of rounding up?
> 
> activity.rs:68
> 
> Feels like we ought to have an iterator method on PlaneRegion for this, but that's a larger refactor and mostly irrelevant to what you're trying to accomplish.
> 
> activity.rs:102
> 
> You're skipping (x < 2) && (y < 2) in the loop above (a total of 2*2 == 4 pixels), but subtracting 4*4 == 16 from the number of pixels here. That does not seem correct.
> 
> activity.rs:103
> 
> old_avg_var is not a variance, since you are summing the L1 norm of transform coefficients. Please do not call it one.
> 
> activity.rs:136:
> 
> There is no comment above that refers to this.
> 
> activity.rs:138
> 
> What is going on here?
> 
> activity.rs:141
> 
> Again, the way you are using this, these are not variances at all (I honestly have no idea what these are). Please do not call them variances.
> 
> activity.rs:144
> 
> I am guessing that the choice of granularity == importance block size is intentional, but this code looks like it would be hopelessly broken if they ever differed. You are doing an accumulation in *x (also, please do not shadow your loop counter like this... I know, I know, Rust, but I still think it's bad form in this instance), but then adding *x repeatedly to avg_var. The only way this can hope to make sense is if *x is used exactly once, which would not be true if importance block size was larger than granularity. If importance block size is smaller than granularity, then you are only going to add temporal activity to some small fraction of the variance entries.
> 
> So if the two *have* to match, why not make that explicit, instead of giving people the false hope that they could be different? It would also let you zip up an iterator over the variances array, instead of using this complicated get_mut/match construct.
> 
> activity.rs:154
> 
> What is this trying to accomplish?
> It looks like maybe you are trying to group things into 8 bins, but the way you have done this, the bins are not equal sizes. Also, the "max" statistic is incredibly sensitive to outliers, and I don't think this is going to be the best way to accomplish what you are trying to do.
> 
> activity.rs:175
> 
> It would be good to have a name for "dec_width" that is common in both this and the routine above, since it is the same value serving the same purpose. I'm not sure "dec_width" is a good name ("dec" is easily confused with "decoder", although I guess you meant "decimate", but it's still a pretty generic name).
> 
> activity.rs:158
> 
> Please use an iterator for this... the whole get_mut/match construct here is totally unnecessary.
> 
> acitvity.rs:188
> 
> Where does all of the magic here come from? (and why are we repeating complex calculations that are static for the whole frame on every lookup into this array?)
> 
> activity.rs:183
> 
> This also isn't returning a variance. Please don't call it one.

BlueSwordM:
I see that this has not progressed much since its inception :P

shssoichiro:
Given that there have been 2 years worth of commits, some of which have majorly rewritten the variance and distortion code, rebasing this ended up being too difficult. Based on the review suggestions, I ended up cherry-picking portions of this changeset against master in #2921.

PR #2248, opened 2020-04-23
Add monochrome support

### LINKED ISSUE #2190: Monochrome Support (RA_CHROMA_SAMPLING_CS400)
It appears that CS400 (YUV400) isn't actually fully supported yet. rav1e asserts if I try to use it at all. 

(When this gets addressed, I'll add monochrome support to libavif using it!)

cyanreg:
Making a PR so I can spend 10 days fixing what clippy doesn't like.
Need to check that LRF and deblock still work and aren't broken for 420.

Fixes #2190

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -7,6 +7,7 @@
 // Media Patent License 1.0 was not distributed with this source code in the
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
+use crate::color::ChromaSampling::Cs400;
 use crate::context::*;
 use crate::encoder::{FrameInvariants, FrameState};
 use crate::frame::*;
@@ -349,7 +350,7 @@ pub fn cdef_block8_frame<T: Pixel>(
 // the CDEF_VERY_LARGE flag.
 pub fn cdef_padded_tile_copy<T: Pixel>(
   tile: &Tile<'_, T>, sbo: TileSuperBlockOffset, w_8: usize, h_8: usize,
-  pad: usize,
+  pad: usize, planes: usize,
 ) -> Frame<u16> {
   let ipad = pad as isize;
   let mut out = {
@@ -364,7 +365,7 @@ pub fn cdef_padded_tile_copy<T: Pixel>(
     }
   };
   // Copy data into padded frame
-  for pli in 0..3 {
+  for pli in 0..planes {
     let PlaneOffset { x, y } = sbo.plane_offset(tile.planes[pli].plane_cfg);
     let in_width = tile.planes[pli].rect().width as isize;
     let in_height = tile.planes[pli].rect().height as isize;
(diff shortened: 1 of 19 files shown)


src/cdef.rs
```diff
@@ -420,9 +421,13 @@ pub fn cdef_padded_frame_copy<T: Pixel>(in_frame: &Frame<T>) -> Frame<u16> {
     },
   };
 
-  for p in 0..3 {
+  for p in 0..MAX_PLANES {
     let rec_w = in_frame.planes[p].cfg.width;
     let rec_h = in_frame.planes[p].cfg.height;
+
+    /* Its a monochrome frame but we have no way to signal that */
+    if rec_w == 0 || rec_h == 0 { break; };
```
cyanreg, line 429: We can leave this out for some small overhead when encoding monochrome, but if possible I'd rather have a better solution.

takehirokj:
Do we need a encode and decode test for 400 [here](https://github.com/xiph/rav1e/blob/53f068fa003b674b46a2a0ab15f8e41f2fbb282e/src/test_encode_decode/mod.rs#L718)?

src/cdef.rs
tdaede: This optimization is definitely worth keeping. Passing in # of planes or putting it in the Frame struct are other valid solutions. In Frame might be a bit awkward because it's really an encoder wide setting.

cyanreg:
Tests added.
Tried to test but the libaom wrapper seems broken (cfg is a missing field) and when fixed, all subsampling tests get skipped.

tdaede:
I think the problem with the libaom test is that in CI it uses the 1.0.0 branch, which is API incompatible. Maybe there is a way to work around this.

All of the encode-decode tests also work with dav1d, that's what I normally do. Also I think these tests are ignored by default because of their high runtime. This should run the tests:

```
cargo test --features=decode_test_dav1d --release -- --ignored
```

barrbrain:
At [src/test_encode_decode/dav1d.rs:177](https://github.com/xiph/rav1e/blob/master/src/test_encode_decode/dav1d.rs#L177) and [src/test_encode_decode/aom.rs:134](https://github.com/xiph/rav1e/blob/master/src/test_encode_decode/aom.rs#L134), the encode-decode tests iterate over the planes. These should also be modified for monochrome.

src/quantize.rs
```diff
@@ -83,7 +83,7 @@ pub fn select_dc_qi(quantizer: i64, bit_depth: usize) -> u8 {
     8 => &dc_qlookup_Q3,
     10 => &dc_qlookup_10_Q3,
     12 => &dc_qlookup_12_Q3,
-    _ => unimplemented!(),
+    _ => return 255,
```
barrbrain, line 86: This could be handled in rate.rs, selecting a quantizer index for a non-existent bitdepth is strange.

src/frame/mod.rs
```diff
@@ -72,8 +72,8 @@ pub(crate) trait FramePad {
 
 impl<T: Pixel> FramePad for Frame<T> {
   fn pad(&mut self, w: usize, h: usize) {
-    for p in self.planes.iter_mut() {
-      p.pad(w, h);
+    for pli in 0..1 {
```
barrbrain, line 75: Padding has normative effects, we shouldn't skip padding the chroma planes for non-monochrome encodes.

src/quantize.rs
cyanreg: That's just for testing.
The real fix would be modifying the rate control system. Rust makes this so not nice.

cyanreg:
Tests pass, and I'm somewhat contempt with the code itself, so this is ready for a review.

PR #2250, opened 2020-04-23
Disable icelake detection

lu-zero:
Due some miscommunication upstream disabled all the icelake flags in the
current stable release.

The issue is tracked in https://github.com/rust-lang/rust/issues/71473


lu-zero:
We aren't ready yet to leverage avx512 anyway.

PR #2253, opened 2020-04-24
Add beta to CI matrix

est31:
In order to find regressions like
https://github.com/rust-lang/rust/issues/71473
before stable releases.

### DIFF
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -48,6 +48,7 @@ jobs:
     strategy:
       matrix:
         conf:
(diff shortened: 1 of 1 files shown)


.github/workflows/rav1e.yml
```diff
@@ -48,6 +48,7 @@ jobs:
     strategy:
       matrix:
         conf:
+         - build
```
shssoichiro, line 51: I think this is a great idea. However, could we have the run named something indicating this is for Rust beta (e.g. `beta-build`)?

PR #2259, opened 2020-04-27
Add profile name to version string 

### LINKED ISSUE #2257: Add release vs debug profile to version string

shssoichiro:
This is to easily identify if we have a
debug or release binary in use.

Closes #2257


lu-zero:
It needs a rebase (and the rework mentioned above)

PR #2260, opened 2020-04-27
CI: Install cargo-c package on macOS

barrbrain:
We may prefer to have the latest release even if it is not yet in homebrew.
This change tests the alternative, installing a pre-built release.


Luni-4:
@lu-zero 

Yep, feel free to assign that task to me

PR #2264, opened 2020-04-29
Fix assert failure in bench of downsample_odd

### LINKED ISSUE #2263: Assertion failed on bench of downsample_odd
```
$ cargo bench --features=bench downsample_odd

Benchmarking downsample_odd: Warming up for 3.0000 sthread 'main' panicked at
 'assertion failed: width * 2 <= src.cfg.stride - src.cfg.xorigin',
```

Failed in [here](https://github.com/xiph/rav1e/blob/41464e701e6ea2052120e1a14c7b9e68de3b563c/v_frame/src/plane.rs#L418).
Padding is required for downsampling of odd sized frames.

takehirokj:
Closes #2263.
Padding is added for downsampling of odd sized input.

### DIFF
--- a/benches/plane.rs
+++ b/benches/plane.rs
@@ -33,7 +33,7 @@ pub fn downsample_8bit(c: &mut Criterion) {
   let input = init_plane_u8(1920, 1080);
   c.bench_function("downsample_8bit", move |b| {
     b.iter(|| {
-      let _ = input.downsampled(960, 540);
+      let _ = input.downsampled(input.cfg.width, input.cfg.height);
     })
   });
 }
@@ -42,7 +42,7 @@ pub fn downsample_odd(c: &mut Criterion) {
   let input = init_plane_u8(1919, 1079);
   c.bench_function("downsample_odd", move |b| {
     b.iter(|| {
-      let _ = input.downsampled(960, 540);
+      let _ = input.downsampled(input.cfg.width, input.cfg.height);
     })
   });
 }
@@ -51,7 +51,7 @@ pub fn downsample_10bit(c: &mut Criterion) {
   let input = init_plane_u16(1920, 1080);
   c.bench_function("downsample_10bit", move |b| {
     b.iter(|| {
-      let _ = input.downsampled(960, 540);
+      let _ = input.downsampled(input.cfg.width, input.cfg.height);
     })
   });
 }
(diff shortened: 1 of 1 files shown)


takehirokj:
@tdaede I've fixed the args for the function to the original size, not the downsampled size.

But the assertion fail is due to the padding of odd sized input, not the wrong args.
The other bench with wrong args didn't have assertion failure.

tdaede:
It's a bug in the downsampled() implementation then. Is it the same assert? downsampled() is supposed to round up with odd sized frames and do the padding automatically.

takehirokj:
Yes. Both fails in the same assert [here](https://github.com/xiph/rav1e/blob/41464e701e6ea2052120e1a14c7b9e68de3b563c/v_frame/src/plane.rs#L418).

This assert is to check the padding of odd sized input plane as in #2095.

tdaede:
Hmm OK. So I guess downsampled() requires padded to even inputs. All of our current uses of Plane use padded to even (in fact padded to multiple of 8) so this didn't happen. I think there are two possible fixes:

1. Change downsampled() to support images that are not padded to even.

2. Document that planes have a minimum padding requirement, and enforce it e.g. in the plane constructors. Note that even downsampled() does not enforce the minimum padding requirement on its output, because it divides the padding by 2, rounded down.

takehirokj:
@tdaede, I consider and try 1. and 2. and prefer other possibility below.
	

3. Use assert to stop using odd plane without padded to even in required place.
This is already done in current downsampled() .

If  
> All of our current uses of Plane use padded to even

and only this bench doesn't, we don't need to add support for this special case now.
Because this support may slow down in the normal case and makes the code more complex.
If we choose 3, we only need to fix the bench as this PR.

tdaede:
Sorry for the slow reply!

I think your option 3) is fine too. It is kind of what I meant by "enforcing the minimum padding requirement" in 2.

benches/plane.rs
```diff
@@ -19,16 +19,23 @@ pub fn downsample_8bit(c: &mut Criterion) {
...
     b.iter(|| {
-      let _ = input.downsampled(960, 540);
+      let _ = input.downsampled(input.cfg.width, input.cfg.height);
     })
   });
 }
 
 pub fn downsample_odd(c: &mut Criterion) {
   let input = init_plane_u8(1919, 1079);
+  let mut padded = Plane::new(input.cfg.width, input.cfg.height, 0, 0, 1, 1);
```
tdaede, line 29: I think it would make more sense for this padding code to be in init_plane_u8(), if we decide that all planes must be padded like this.

takehirokj: Thanks. I have moved this part to init_plane_u8().

takehirokj: Fixed.

PR #2271, opened 2020-05-01
api: Properly validate minimum rdo_lookahead_frames value

dwbuiten:
Fixes crash / panic when rdo_lookahead_frames was set to 0.


dwbuiten:
@lu-zero Good idea. Done.

PR #2272, opened 2020-05-01
Optimize chroma quantizer offsets for subset3 4:4:4

barrbrain:
While rationalising the speed presets in preparation for MSU Annual Video Codecs Comparison 2020, we identified that color performance has not yet been optimised for non-4:2:0 subsamplings. This is relevant because half of the test sequences are 4:2:2.

Numeric analysis indicated 401 +/- 2 in Q12.
Gradient for 4:2:2 is estimated by the midpoint with 4:2:0, rounded down.

Before this change, it is likely that too many bits are given to chroma for 4:2:2 and 4:4:4.

### DIFF
--- a/src/rate.rs
+++ b/src/rate.rs
@@ -7,6 +7,7 @@
 // Media Patent License 1.0 was not distributed with this source code in the
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
+use crate::api::color::ChromaSampling;
 use crate::api::ContextInner;
 use crate::encoder::TEMPORAL_DELIMITER;
 use crate::quantize::{ac_q, dc_q, select_ac_qi, select_dc_qi};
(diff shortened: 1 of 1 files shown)


src/rate.rs
```diff
@@ -676,21 +677,29 @@ const Q57_SQUARE_EXP_SCALE: f64 =
...
+) -> (i64, i64) {
   let x = log_target_q.max(0);
-  // Gradient 0.266 optimized for CIEDE2000+PSNR on subset3
-  let y = (x >> 2) + (x >> 6);
+  // Gradient optimized for CIEDE2000+PSNR on subset3
+  let y = match chroma_sampling {
+    ChromaSampling::Cs400 => unimplemented!(),
+    ChromaSampling::Cs420 => (x >> 2) + (x >> 6), // 0.266
+    ChromaSampling::Cs422 => (x >> 3) + (x >> 4) - (x >> 7), // 0.180
+    ChromaSampling::Cs444 => (x >> 4) + (x >> 5) + (x >> 8), // 0.098
```
cyanreg, line 689: I think you should have a 1.0 entry for when the colorspace is RGB and the subsampling is 444.

barrbrain: That makes sense and we should plumb `ColorDescription` next. In this change, we are still presuming BT.709 but no longer assuming 4:2:0 chroma quantization behavior. Ultimately, we should map all color configurations to optimal chroma quantizer offsets.

cyanreg:
Could we have a specialcase for RGB before committing this?

barrbrain:
We don't yet have any data on the optimal offsets for RGB but when we do they may be applied independent of this change.

cyanreg:
Wouldn't a 1-1-1 offset be more-or-less right?

barrbrain:
A cursory review of literature indicates a ratio similar to 13-10-17 for RGB quantizers is optimal for video. Given that there are differences in sensitivity to the RGB planes, our experience with BT.709 YCbCr suggests that the optimal ratio likely varies with the quantization level and has a different curve per subsampling.

cyanreg:
It doesn't have to be perfect or close, just somewhat better in theory than using yuv444's offsets which should decimate G and B.

barrbrain:
Measuring the impact for 4:2:2 is confounded by the present disabling CDEF/LRF, so only 4:4:4 results will be presented. [AWCY results on derf-720p-444 at the default speed](https://beta.arewecompressedyet.com/?job=444-master-b5b5b603%402020-05-09T17%3A10%3A55.639Z&job=444-chroma-subsampling-quantizers%402020-05-09T17%3A14%3A36.891Z) show a significant impact on metrics.  The runs were incomplete due a storage issue on one of the nodes, so the aggregate result is misleading. It is also a small set of sequences. The median BD-rate of luma metrics was -31.165% and for CIEDE2000 the median BD-rate was -14.355%. For the sequence `in_to_tree_444_720p50.y4m` which had complete results, median luma BD-rate was -36.645% and the CIEDE2000 BD-rate was -15.45%.

PR #2274, opened 2020-05-02
Cleanup PlaneData implementation and enable Serialize/Deserialize

shssoichiro:
The existing PlaneData implementation essentially
replicated the Vec implementation,
so this commit switches the internal implementation
to use a boxed slice.
This allows the removal of some unsafe code.

This also enables implementing serialize/deserialize
in a reasonable way, which is needed for distributed encoding.

This commit has no breaking changes to the public API.

### DIFF
--- a/v_frame/src/frame.rs
+++ b/v_frame/src/frame.rs
@@ -10,9 +10,10 @@
 use crate::math::*;
 use crate::pixel::*;
 use crate::plane::*;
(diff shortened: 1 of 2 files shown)


shssoichiro:
All right, I'll have to do the manual serialize implementation. Have to figure out how to do that without an intermediate allocation.

PR #2275, opened 2020-05-02
Implement Serialize and Deserialize for Frame and Plane types

shssoichiro:
When serialize feature is enabled, otherwise uses the
dummy noop trait.

### DIFF
--- a/v_frame/src/plane.rs
+++ b/v_frame/src/plane.rs
@@ -16,9 +16,10 @@ use std::ops::{Index, IndexMut, Range};
 
 use crate::math::*;
 use crate::pixel::*;
+use crate::serialize::{Deserialize, Serialize};
 
 /// Plane-specific configuration.
-#[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


v_frame/src/plane.rs
```diff
@@ -138,6 +140,37 @@ impl<T: Pixel> std::ops::Drop for PlaneData<T> {
...
+impl<T: Pixel + Serialize> Serialize for PlaneData<T> {
+  fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+  where
+    S: serde::Serializer,
+  {
+    use serde::ser::SerializeSeq;
+
+    let mut data = serializer.serialize_seq(Some(self.len))?;
+    for byte in
+      unsafe { std::slice::from_raw_parts(self.ptr.as_ptr(), self.len) }
```
lu-zero, line 153: plane data derefs to &[] you can avoid unsafe here

shssoichiro: I tried to do a deref but couldn't figure out the magic syntax to make the compiler happy. I guess I could always do `.deref()`

PR #2286, opened 2020-05-09
Disable RDO of transform type for inter blocks

barrbrain:
While rationalising the speed presets in preparation for MSU Annual Video Codecs Comparison 2020, we identified that speed 5 is a good candidate for 1080p at 1fps encoding speed with a few changes. To reduce the large jump in encoding time relative to speed 6 while preserving the majority of the benefit from transform type and size RDO, disable the type RDO for inter. After this change, transform type and size RDO is only performed for intra blocks. In future, we should add speed settings to enable a wider search at slower speed levels.

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -622,6 +622,7 @@ pub fn rdo_tx_size_type<T: Pixel>(
 
     let do_rdo_tx_type = tx_set > TxSet::TX_SET_DCTONLY
       && fi.config.speed_settings.rdo_tx_decision
+      && !is_inter
       && !skip;
 
     if !do_rdo_tx_size && !do_rdo_tx_type {


shssoichiro:
In IRC I think you said you had an AWCY for this?

I'm wondering if it would be prudent to go ahead and add this as a speed setting now. I think that probably depends on AWCY results.

barrbrain:
[AWCY results at speed 5](https://beta.arewecompressedyet.com/?job=speed-5-fine-dir%402020-05-09T12%3A44%3A38.812Z&job=disable-inter-tx-type-rdo%402020-05-09T14%3A03%3A09.263Z) show a 23% increase in encoding speed with mixed results in metrics:

| 0.6422 | -0.8994 | -0.6326 |  -0.0105 | 0.1596 | -0.1235 |    -0.1157 |

shssoichiro:
Interesting, PSNR seems to take a decent hit on screen content, but MSSSIM is mostly unaffected, and that's the metric I consider to be most representative of visual quality.

I'm good with this change as is. I wonder if it might make sense to enable tx-rdo for inter frames if we're using Tune = Psnr? But I also care approximately not at all about Tune = Psnr because it's not real-world useful.

PR #2292, opened 2020-05-11
Simplify deploy script

Luni-4:
- Remove a useless job
- Clean up the script a bit

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,7 +7,7 @@ on:
       - 'p*'
 
 jobs:
-  create-binaries:
+  deploy-windows-binaries:
 
     runs-on: windows-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -37,60 +37,36 @@ jobs:
...
 
-    - name: Handle binaries
+    - name: Get the version
+      if: startsWith(github.ref, 'refs/tags/v')
+      shell: bash
+      id: tagName
       run: |
-        mv rav1e-binaries/local rav1e-binaries/$FILENAME
-        strip rav1e-binaries/rav1e.exe
+        VERSION=$(cargo pkgid | cut -d# -f2 | cut -d: -f2)
```
shssoichiro, line 57: I think we need to use https://github.com/marketplace/actions/get-tag-name#get-tag-name instead, `cut` doesn't seem to work on Windows, as I learned from oxipng.

Luni-4: Yep, you're right, it shouldn't work on Windows, but if you set `bash: shell`, the bash shell included with `Git` for `Windows` is used. I tested that [here](https://github.com/Luni-4/rav1e/runs/665110574?check_suite_focus=true#step:10:2) and it works.

shssoichiro:
Thanks!

PR #2293, opened 2020-05-12
v_frame: Add timestamp member

dwbuiten:
This requires a bump to v0.2.0 since it is an API break.

# Notes

**Motivation**: We want input and output timestamps in the future. This is step 1 to allowing that, since our API returns a `Frame` directly. I have more work, but this PR a prerequisite for all of it.

**Help needed**: This patch only modifies the `v_frame` crate. When I tried to modify rav1e as well, av-metrics cannot build, as it requires `0.1.0`, which can't be found in the subdirectory pointed to in `Cargo.toml`. As far as I can tell it's actually impossible to have rav1e build at every commit, because `av-metrics` requires its `v_frame` usage to be updated at the same time as... `v_frame` is updated. It's a circular dependency. Please help.

### DIFF
--- a/v_frame/Cargo.toml
+++ b/v_frame/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "v_frame"
-version = "0.1.0"
+version = "0.2.0"
 description = "Video Frame data structures, part of rav1e"
 license = "BSD-2-Clause"
 authors = ["Luca Barbato <lu_zero@gentoo.org>"]
--- a/v_frame/src/frame.rs
+++ b/v_frame/src/frame.rs
@@ -17,6 +17,7 @@ use crate::serialize::{Deserialize, Serialize};
 pub struct Frame<T: Pixel> {
   /// Planes constituting the frame.
   pub planes: [Plane<T>; 3],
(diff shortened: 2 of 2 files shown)


dwbuiten:
```
[15:44] < mindfreeze> https://www.irccloud.com/pastebin/6UJIeXNY/temp.patch
[15:44] < mindfreeze> Daemon404: The above patch can atleast trigger your changes for testing and updating rav1e
[15:45] < mindfreeze> This is the output of build:  https://www.irccloud.com/pastebin/1aZlqMys/errors
[15:45] < Daemon404> when i tried, av-metrics failed to build, as it also relieas on v_frame
[15:45] < Daemon404> unless i missed something
[15:46] < mindfreeze> I think you missed to update the v_frame in the `[replace]` part since that takes the local version instead of the cargo one IIRC when we do for rav1e
[15:46] < Daemon404> lemme try
[15:47] < Daemon404> yep that was it
[15:47] < Daemon404> thanks mindfreeze!
[15:47] < mindfreeze> \o/
```

Marking this PR as [WIP] for now, but letting people comment on the idea of adding this member at all in the meantime.

Issue #2282, opened 2020-05-07
Split the context.rs into smaller blocks


vibhoothiiaanand:
Currently, [context.rs](https://github.com/xiph/rav1e/blob/master/src/context.rs) is getting larger and larger, now it is having ~4471 Lines of Code, it would be nice if we start breaking functions into smaller functions to increase readability.

~Currently, it has functions for the following:~
- [ ] ~Partitions~
- [ ] ~Blocks~
- [ ] ~CFL~
- [ ] ~Transformations~
- [ ] ~CDF Context~
- [ ] ~Block, Planes offsets~
- [ ] ~Motion Vectors~
and more...

Splitting based on Block hierarchy is a better choice :
- [ ] Frame
- [ ] Tile
- [ ] Block
- [ ] Superblock
- [ ] Partition
- [ ] Transform

tdaede:
I think if you want to split context.rs, the better dividing line is hierarchy in the bitstream. e.g. superblock, partition, block.

vibhoothiiaanand:
@tdaede Yeah that makes more sense, I will try to do it like that.

vibhoothiiaanand:
Fixed with #2303

PR #2303, opened 2020-05-14
Split context.rs 

vibhoothiiaanand:
Hi,

This patchset introduces a new module context which breaks context.rs into smaller chunks based on bitstream hierarchy 

Ref: #2282


src/context/frame.rs
```diff
@@ -0,0 +1,322 @@
+// Copyright (c) 2017-2020, The rav1e contributors. All rights reserved
+//
+// This source code is subject to the terms of the BSD 2 Clause License and
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+use super::*;
```
lu-zero, line 10: This module is related to filters, not frames.

vibhoothiiaanand:
okay, did discuss with @xiphmont regarding this,
So here are the fine pointers/summary of the discussion
+ The patchset is splitting without compartmentalising things For eg, „none of the shared cdf context is currently split up". Thus leading to a state where there are small files but also have the working bits for any one thing in the bitstream spread out over multiple files.“. Which is the reason for having all in one big file as it is all shared.
+ Here we are only just moving it around which does not hurt also does not help anything either.
+ We need to compartmentalize the data flow, not just the code flow. The data flow is the important part!  Code is relatively unimportant to the data it manages.
+ Compartmentalizing the global data state by bitstream function might be a way to start. Then the code will naturally separate into pieces.
+ By compartmentalizing the global data by bitstream we will have underlying context abstraction, and then the functional code will just sit on top and should break into pieces more easily.

Having such an approach over this will make things better in the right way.

tdaede:
Unfortunately the big block of CDF definitions isn't really splittable, in the data structure we currently have defined. There is probably still some value to splitting, and I do prefer the current style over earlier proposals. I think the goal here is *NOT* just to shrink a large file.

A great start would just be rename the files. E.g. frame_header, partition_unit, transform_unit (or coding_unit). You can see that @lu-zero was confused by a filter function being in a file called frame.rs.

tdaede:
Secondly, I'd like to double check with a couple of reviewers @xiphmont and @lu-zero that we think this sort of split is something that will work, even without splitting the CDF context. I don't want you to spend a whole lot of time on this without a clear direction.

lu-zero:
The new names look better. @vibhoothiiaanand you might add the speed difference (your patch should make building faster assuming it runs on a many-threads machine)

vibhoothiiaanand:
This gives an approx speedup on build time between ~4% to 10%

on macOS
Master Branch(02018bf) 
```
    Finished release [optimized + debuginfo] target(s) in 8m 05s
      485.51 real       812.85 user        31.64 sys
```

split-context branch
```
    Finished release [optimized + debuginfo] target(s) in 7m 13s
      433.08 real       767.43 user        28.46 sys
```

src/context/cdf_context.rs
```diff
@@ -0,0 +1,869 @@
...
+    [CFL_SIGN_POS, CFL_SIGN_ZERO],
+    [CFL_SIGN_POS, CFL_SIGN_NEG],
+    [CFL_SIGN_POS, CFL_SIGN_POS],
+  ];
+
+  static cfl_context: [[usize; 8]; 2] =
+    [[0, 0, 0, 1, 2, 3, 4, 5], [0, 3, 0, 1, 4, 0, 2, 5]];
+
+  #[test]
+  fn cfl_joint_sign() {
```
tdaede, line 494: cfl functions should be in partition_unit

src/context/cdf_context.rs
```diff
@@ -0,0 +1,869 @@
...
+    }
+
+    cw
+  }
+
+  pub fn cdf_element_prob(cdf: &[u16], element: usize) -> u16 {
+    (if element > 0 { cdf[element - 1] } else { 32768 }) - cdf[element]
+  }
+
+  pub fn write_skip(
```
tdaede, line 561: skip should be in partition_unit

src/context/cdf_context.rs
```diff
@@ -0,0 +1,869 @@
...
+    &mut self, w: &mut dyn Writer, bo: TileBlockOffset, skip: bool,
+  ) {
+    let ctx = self.bc.skip_context(bo);
+    symbol_with_update!(self, w, skip as u32, &mut self.fc.skip_cdfs[ctx]);
+  }
+
+  fn get_segment_pred(&self, bo: TileBlockOffset) -> (u8, u8) {
```
tdaede, line 568: segmentation functions should be in partition_unit

src/context/cdf_context.rs
```diff
@@ -0,0 +1,869 @@
...
+    );
+    symbol_with_update!(
+      self,
+      w,
+      coded_id as u32,
+      &mut self.fc.spatial_segmentation_cdfs[cdf_index as usize]
+    );
+  }
+
+  pub fn write_coeffs_lv_map<T: Coefficient>(
```
tdaede, line 662: coeff functions should be in block_unit

barrbrain:
Once we settle on a new structure, it would be good to also update [doc/STRUCTURE.md](https://github.com/xiph/rav1e/blob/master/doc/STRUCTURE.md).

shssoichiro:
Needs a rebase due to CI updates

PR #2305, opened 2020-05-14
Reduce the number of checkpoints used in tx rdo

KyleSiefring:
I want awcy working again so I can test.


src/rdo.rs
```diff
@@ -1510,7 +1516,11 @@ pub fn rdo_tx_type_decision<T: Pixel>(
 
   let is_inter = !mode.is_intra();
 
-  let cw_checkpoint = cw.checkpoint();
+  if cw_checkpoint.is_none() {
+    // Only run the first call
+    // Prevents creating multiple checkpoints for onw version of cw
```
lu-zero, line 1521:
```suggestion
    // Prevents creating multiple checkpoints for its own version of cw
```

PR #2307, opened 2020-05-15
Make Config opaque

lu-zero:
In preparation of the rate control API update.

Add Config::new(), Config::with_encoder_config() and Config::with_threads().


lu-zero:
Added a commit to make it fully opaque.

lu-zero:
If it is ok, I'd like to have it landed so I can rebase my pending sets just once :)

src/test_encode_decode/mod.rs
```diff
@@ -192,7 +192,7 @@ fn setup_encoder<T: Pixel>(
   enc.tile_rows = 1 << tile_rows_log2;
   enc.still_picture = still_picture;
 
-  let cfg = Config { enc, threads: 0 };
+  let cfg = Config::new().with_encoder_config(enc);
```
tdaede, line 195: review the warning in this file

PR #2312, opened 2020-05-18
Fix intermittent dequantize test failure

### LINKED ISSUE #2276: Intermittent asm::x86::dequantize_test failure
asm::x86::dequantize_test is intermittently failing due to an off-by-one error on one byte of the output. This has been replicated on Windows and Linux.

```
failures:

---- asm::x86::quantize::test::dequantize_test stdout ----
thread 'asm::x86::quantize::test::dequantize_test' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
 [
     -2100,
     -28672,
     -2688,
     -768,
     14720,
     -7808,
     -3968,
     15232,
     3584,
     1152,
     0,
     -4480,
     5120,
     7936,
     -29568,
     30336,
     -32640,
     4608,
<    -32767,
>    -32768,
     -12928,
     -10240,
     -23168,
     1152,
     12672,
     2304,
     12928,
     17152,
     -18432,
     -4224,
     17280,
     0,
     0,
 ]

', src\asm\x86\quantize.rs:83:5
```

KyleSiefring:
The test was generating values one less than -i16::MAX.
Fixes #2276

### DIFF
--- a/src/asm/x86/quantize.rs
+++ b/src/asm/x86/quantize.rs
@@ -154,6 +154,7 @@ unsafe fn dequantize_avx2(
 #[cfg(test)]
 mod test {
   use super::*;
+  use rand::distributions::{Distribution, Uniform};
   use rand::{thread_rng, Rng};
 
   #[test]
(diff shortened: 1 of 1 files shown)


KyleSiefring:
I'm going to make sure this is the case before landing this patch.

lu-zero:
Maybe this patch can be landed as-is and refined later?

barrbrain:
This change is fine as-is. Regarding the range of output from forward transforms, we can add more tests specifically for that.

PR #2317, opened 2020-05-21
Change the minimum value of rdo_lookahead_frames to 0.

tdaede:
Has no effect on minimum latency, effectively changes the
meaning of this value by 1.


src/api/internal.rs
```diff
@@ -605,30 +605,6 @@ impl<T: Pixel> ContextInner<T> {
       .unwrap();
     }
 
-    // Do not modify the next output frame's FrameInvariants.
```
lu-zero, line 608: Why this?

tdaede: Basically this prevents us from ever modifying the "current" fi. But if we want to run with zero latency, we *must* modify the current fi, to put in its lookahead mvs etc.

tdaede: FWIW I don't understand why this condition was here in the first place.

lu-zero: the tests are currently failing though.

YaLTeR: So, as far as I remember, this condition was here because the current fi contains the reconstruction buffer filled with reconstructed frames, but for lookahead we set the reconstruction buffer to original frames (since we need it to compute MVs). Overwriting the current frame's reconstruction buffer might break everything that needs the correct reconstructed frame.

tdaede: Right, so what I want to change this to is to these two steps happening sequentially in one receive_packet. As per the comment on rec_buffer we end up rewriting this when starting the final encode, so I don't think we actually need to avoid writing it here.

YaLTeR: Should be good then (especially if breaking the reconstruction buffer is signalled by some tests).

src/api/test.rs
```diff
@@ -1492,156 +1492,155 @@ struct LookaheadTestExpectations {
 }
 
 #[test]
-fn lookahead_size_properly_bounded_8() {
-  const LOOKAHEAD_SIZE: usize = 8;
+fn lookahead_size_properly_bounded_7() {
```
YaLTeR, line 1495: Why the test changes?

tdaede: oh whoops I think I mixed up two versions of this patch, one which allowed setting to 0 and one which changed the meaning so that 1 was the new minimum. if I allow 0 then the tests shouldn't change, you're right

src/api/internal.rs
tdaede: I don't think we have any tests for this. Maybe I can add a flag in fi just for signaling this condition for assert / debugging purposes.

YaLTeR: Shouldn't breaking the reconstruction buffer break sync tests?

tdaede: Yes, maybe that's enough.

src/api/test.rs
tdaede: I might squash these 2 commits into 1 because I think it's a bit confusing to follow them separately.

lu-zero:
Can we rebase and land it?

lu-zero:
It needs another rebase, good for 0.5?

tdaede:
I rebased this and it passes tests, but it really does need some closer look before we merge I think. I would almost suggest skipping 0.5.0 because of the high chance for regressions.

shssoichiro:
I happened upon this PR in my review queue. Is this still something we want to do? I'm assuming given the time since this PR was created, there's going to be some nasty merge conflicts.

PR #2318, opened 2020-05-21
Re-enable the avx512 runtime check

lu-zero:
Rust 1.43.1 fixed the avx512 runtime check support.

It contains https://github.com/xiph/rav1e/pull/2329


lu-zero:
We need to bump the nightly version to a current one.

PR #2320, opened 2020-05-22
Update CI

Luni-4:
- Install `cargo-c` using its binaries to speed up CI
- Shorten up long lines


Luni-4:
@barrbrain 

Now this PR updates `cargo-c` only

PR #2321, opened 2020-05-22
Update the y4m dependency

### LINKED ISSUE #1983: Improve error message when feeding a non-y4m input
Currently rav1e gives this:
````
The application panicked (crashed).
  cannot decode the input: ParseError
in src/libcore/result.rs, line 1165
thread: main
````

rzumer:
This closes #1983.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ bitstream-io = "0.8"
 cfg-if = "0.1"
 clap = { version = "2", optional = true, default-features = false }
 libc = "0.2"
-y4m = { version = "0.5", optional = true }
+y4m = { version = "0.6", optional = true }
 backtrace = { version = "0.3", optional = true }
 num-traits = "0.2"
 num-derive = "0.3"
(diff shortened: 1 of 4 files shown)


lu-zero:
Looks like the error type changed.

rzumer:
Fixed the error type mapping, but now there are many errors because of https://github.com/image-rs/y4m/commit/9f25fd03a43ed4ba086f3a7cc793b24fb796b354 that I don't want to deal with right now.

lu-zero:
Should be is just removing `'_,` in 3 places.

rzumer:
I'd pushed my last change to my fork instead of here, but as you can see that's nowhere near enough.

lu-zero:
Almost, you need to rebase to the current master and add 
https://github.com/rust-av/rav1e/commit/8c5d6091a9bae2589f256820b4607900e9a87322

Since you the y4m let you store the Write/Read directly in the struct, you have to remove some &/&mut around the constructors.

rzumer:
Thanks, often seems like Rust error messages are an unreadable mess even when the fix is simple.

lu-zero:
They were swamped in the errors uncovered by the paste update. After the rebase they were much simpler to follow :)

PR #2323, opened 2020-05-24
Decoders version bump

lu-zero:
- [x] needs a valid aom 2.0.0 package, so far I couldn't find a .deb that works with ubuntu.

### DIFF
--- a/.travis/install-aom.sh
+++ b/.travis/install-aom.sh
@@ -1,27 +1,21 @@
 #!/bin/bash
 set -ex
 
-AOM_VERSION="1.0.0.errata1-3~18.04.york0"
-PKG_URL="http://ppa.launchpad.net/jonathonf/ffmpeg-4/ubuntu/pool/main/a/aom"
-
-case "$ARCH" in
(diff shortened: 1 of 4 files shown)


lu-zero:
This should be squashed on merge.

Luni-4:
We need to update `aom` for `travis` too.

lu-zero:
@Luni-4 will you take care of it?

.travis/install-aom.sh
```diff
@@ -1,27 +1,21 @@
 #!/bin/bash
 set -ex
 
-AOM_VERSION="1.0.0.errata1-3~18.04.york0"
-PKG_URL="http://ppa.launchpad.net/jonathonf/ffmpeg-4/ubuntu/pool/main/a/aom"
-
-case "$ARCH" in
-  x86_64) ARCH=amd64 ;;
```
vibhoothiiaanand, line 8: Why are we removing x86_64 case, as it is in GitHub Actions?

Luni-4: Exactly, we are testing it with `GitHub Actions`

PR #2325, opened 2020-05-25
Backport to 0.3 branch

lu-zero:
This would lead to `0.3.2`.


lu-zero:
Let's land this first and think about more backporting later I guess.

PR #2326, opened 2020-05-25
Improve the deploy script

Luni-4:
The script deploys:

- linux-musl binary
- aarch64-musl binary
- wasm32 binaries
- macos binary
- windows binary
- windows-msvc binaries produced by `cargo-c`
- windows-gnu binaries produced by `cargo-c`
- Cargo.lock

You can find a pre-release [here](https://github.com/Luni-4/rav1e/releases/tag/p20200229) and a release [here](https://github.com/Luni-4/rav1e/releases/tag/v0.4.9)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,7 +7,19 @@ on:
       - 'p*'
 
 jobs:
-  deploy-windows-binaries:
+
+  windows-binaries:
+
+    strategy:
+      matrix:
+        conf:
+         - msvc
+         - gnu
+        include:
+         - conf: msvc
+           toolchain: stable
+         - conf: gnu
+           toolchain: stable-x86_64-pc-windows-gnu
 
     runs-on: windows-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -60,13 +90,226 @@ jobs:
...
+        path: rav1e-${{ steps.tagName.outputs.version }}-windows-${{ matrix.conf }}.zip
+
+
+  linux-binaries:
+
+    strategy:
+      matrix:
+        target:
+         - x86_64-unknown-linux-musl
+         - aarch64-unknown-linux-gnu
```
barrbrain, line 128: Perhaps `aarch64-unknown-linux-musl` is preferable as then the binary will run on both Android and arbitrary Linux distributions. However, my experience is that getting a working aarch64 musl cross-compilation environment may require more effort.

.github/workflows/deploy.yml
```diff
@@ -60,13 +90,226 @@ jobs:
...
+        UPCASE=$(echo ${{ matrix.target }} |
+                 awk '{ print toupper($0) }' | sed 's/-/_/g')
+        echo ::set-env name=CARGO_TARGET_${UPCASE}_LINKER::aarch64-linux-gnu-gcc
+
+    - name: Build rav1e
+      run: |
+        cargo build --target ${{ matrix.target }} --release
+
+    - name: Strip musl rav1e
+      if: matrix.target == 'x86_64-unknown-linux-musl'
```
barrbrain, line 187: Perhaps we should also strip `aarch64-unknown-linux-gnu`.

.github/workflows/deploy.yml
Luni-4: I've seen an `aarch64-unknown-linux-musl` binary in the last pre-release. Perhaps who had built that could explain us how to do that.

Luni-4:
@barrbrain 

I will address those problems in another PR, thanks!

.github/workflows/deploy.yml
```diff
@@ -60,13 +90,237 @@ jobs:
...
+         - target: wasm32-unknown-unknown
+           name: wasm32-linux
+           binaries: librav1e.a rav1e.wasm
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install nasm
```
barrbrain, line 148: Nasm is only needed by the `x86_64-unknown-linux-musl` target.

PR #1209, opened 2019-04-12
WIP: Fix Arm64 Windows builds on AppVeyor

EwoutH:
Now that the winapi crashes have been fixed upstream, I tried to get the Arm64 Windows builds working again. See https://ci.appveyor.com/project/EwoutH/rav1e.

Currently, AppVeyor seems to fail on compiling another dependency, [paste](https://github.com/dtolnay/paste).

```
   Compiling paste v0.1.5
error: linking with `link.exe` failed: exit code: 1112
  |
[...]

PR #2327, opened 2020-05-25
Remove appveyor

Luni-4:
From now on, the `msvc` binaries will be produced by the deploy script and the respective tests are run by the `GitHub Actions`, so the `appveyor.yml` is no more necessary.

This PR closes #1209 and #2115


shssoichiro:
I also deleted the webhook in repo settings so we'll no longer get that red X.

Urhengulas:
This is just a small thing, but probably it's best to also remove the appveyor-shield in the headline of the `README.md`.

Luni-4:
Thanks @shssoichiro! :)

I agree with you @Urhengulas

PR #2330, opened 2020-05-26
Backport CI updates to 0.3 branch

barrbrain:


lu-zero:
Looks like before this we'll need to merge in the fixes for the unused parens.

lu-zero:
the commit that removes betterpanic is needed to support 1.36.0.

PR #2332, opened 2020-05-26
Add details of valid key value pairs accepted by rav1e_config_parse.

amCap1712:


src/capi.rs
```diff
@@ -587,9 +587,25 @@ unsafe fn option_match(
 /// Set a configuration parameter using its key and value as string.
 ///
 /// Available keys and values
-/// - "quantizer": 0-255, default 100
+/// - "width": width of the frame, default 640
+/// - "height": height of the frame, default 480
 /// - "speed": 0-10, default 6
+/// - "threads": maximum number of threads to be used
 /// - "tune": "psnr"-"psychovisual", default "psychovisual"
+/// - "quantizer": 0-255 (0 denotes auto), default 100
```
shssoichiro, line 595: Sorry, I probably was confusing about this. 0=auto applies to the tiles setting, not to the quantizer setting.

PR #2334, opened 2020-05-26
Backport bug fixes to 0.3

barrbrain:
Initially, this includes almost everything in the master branch. We can whittle this down.


rzumer:
I would omit new features, dependency updates and refactors at least, readme and CI changes possibly.

lu-zero:
The dependency updates are needed for the release.

Must not land:
- v_frame split
- Rename "native" CPU feature level to "rust"
- avx512 support

I couldn't spot any other changes that are either API breaking or incomplete.

lu-zero:
I do not see any change that would cause API issue. It is good to go in my opinion.

rzumer:
With so many commits I feel like it's almost 0.4 in disguise. There are still new features included in here that modify BD-rate, personally I would expect backports to only include bugfixes, but maybe it's just me.

lu-zero:
We could omit
- monochrome
- compound prediction mode

From a quick look I cannot find other commits that are feature-only, if you have other to remove in mind please add them to the list :)

rzumer:
> 
> 
> We could omit
> 
>     * monochrome
> 
>     * compound prediction mode
> 
> 
> From a quick look I cannot find other commits that are feature-only, if you have other to remove in mind please add them to the list :)

I agree with those two and personally I would remove anything that changes BD-rate for a maintenance release, except maybe fixes that allow enabling previously broken features (like loop filters for 4:2:2).

barrbrain:
I made a slow pass over the set of patches to exclude any new bitstream features, changes that affect BD-rate other than bugfixes (allowing fixes for RDO bugs) and changes that expose a known bug to API/CLI users.

PR #2344, opened 2020-05-28
Refactor the binary tests

### LINKED ISSUE #2337: Refactor the binary tests
`tests/binary.rs` has quite a bit of code duplication.
- [ ] factor out the common part of the `Command` in a function returning it
- [ ] use it in all the tests.

Urhengulas:
This PR applies various refactoring to [`tests/binary.rs`](tests/binary.rs):
* add 
   ```rust
   fn get_common_cmd(outfile: &str) -> Command
   ```
* combine related arguments in `Command.args(&[arg1, arg2])`
* refactor `fn get_tempfile_path(extension: &str)` to return `OsString`
* don't store `Command` in intermediate variable

Fixes #2337

### DIFF
--- a/tests/binary.rs
+++ b/tests/binary.rs
@@ -42,7 +42,7 @@ mod binary {
 
   fn get_common_cmd(outfile: &PathBuf) -> Command {
     let mut cmd = get_rav1e_command();
-    cmd.arg("--bitrate").arg("1000").arg("-o").arg(outfile);
+    cmd.args(&["--bitrate", "1000"]).arg("-o").arg(outfile);
     cmd
   }
 
@@ -52,8 +52,7 @@ mod binary {
     let outfile = get_tempfile_path("ivf");
 
     cmd
-      .arg("--quantizer")
-      .arg("100")
+      .args(&["--quantizer", "100"])
       .arg("-o")
       .arg(&outfile)
       .arg("-")
@@ -100,8 +99,7 @@ mod binary {
 
     let mut cmd1 = get_common_cmd(&outfile);
     cmd1
-      .arg("--reservoir-frame-delay")
-      .arg("14")
+      .args(&["--reservoir-frame-delay", "14"])
       .arg("--first-pass")
       .arg(&passfile)
       .arg("-")
@@ -111,8 +109,7 @@ mod binary {
 
     let mut cmd2 = get_common_cmd(&outfile);
     cmd2
-      .arg("--reservoir-frame-delay")
-      .arg("14")
+      .args(&["--reservoir-frame-delay", "14"])
       .arg("--second-pass")
       .arg(&passfile)
       .arg("-")
--- a/tests/binary.rs
+++ b/tests/binary.rs
@@ -40,6 +40,12 @@ mod binary {
     Command::cargo_bin("rav1e").unwrap()
   }
 
+  fn get_common_cmd(outfile: &PathBuf) -> Command {
+    let mut cmd = get_rav1e_command();
+    cmd.arg("--bitrate").arg("1000").arg("-o").arg(outfile);
+    cmd
+  }
+
   #[test]
   fn one_pass_qp_based() {
     let mut cmd = get_rav1e_command();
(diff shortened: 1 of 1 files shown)


tests/binary.rs
```diff
@@ -19,12 +18,12 @@ mod binary {
     data
   }
 
-  fn get_tempfile_path(extension: &str) -> PathBuf {
+  fn get_tempfile_path(extension: &str) -> String {
```
lu-zero, line 21: This is potentially wrong, why do you need to do this?

tests/binary.rs
```diff
@@ -40,16 +39,19 @@ mod binary {
     Command::cargo_bin("rav1e").unwrap()
   }
 
+  fn get_common_cmd(outfile: &str) -> Command {
+    let mut cmd = get_rav1e_command();
+    cmd.args(&["--bitrate", "1000"]).args(&["-o", outfile]);
```
lu-zero, line 44: Why not factoring in also 
``` rust
      .arg("-")
      .write_stdin(get_y4m_input())
      .assert()
```

tests/binary.rs
Urhengulas: To use the filepath in `Command.args(...)` (e.g. [here](https://github.com/xiph/rav1e/pull/2344/files#diff-b1aba0aa09c8f43383202cbf5085a064R54) as `&*outfile`), it needs to be `&str`. So I though doing the conversion in one place is easier than multiple times just before the `.args(...)`.

What error scenarios are there?

tests/binary.rs
Urhengulas: Some uses of the fn need to add another arg before the `-` (e.g. [here](https://github.com/xiph/rav1e/pull/2344/files#diff-b1aba0aa09c8f43383202cbf5085a064R78), so I think I can't add it already in the function.

Also I understood, that after the call to `.assert()` the `Command` get's executed, so it should only be called after all the args are added.

Urhengulas:
Thanks for the review :+1:

### functional
I added my reasoning to your comments.

### non-functional
I normally add the `Fixes #...` in the commit message, if the fix is only 1 commit. If the fix consists of multiple commits I am always not sure in which commit to put it. :see_no_evil: 

But, since I have the `Fixes #...` in the body of the PR it should be fine, right?

tests/binary.rs
Urhengulas: **lu-zero** on irc: String cannot represent paths in the operating systems having non-utf8 locale, Command arg and args take any type that can be an OsStr for that reason

Urhengulas: Addressed this in https://github.com/xiph/rav1e/pull/2344/commits/e984ddbc7c0c41d76ca9128f876bc8735dbb2992

lu-zero:
> Thanks for the review 👍
> ### functional
> 
> I added my reasoning to your comments.
> ### non-functional
> 
> I normally add the `Fixes #...` in the commit message, if the fix is only 1 commit. If the fix consists of multiple commits I am always not sure in which commit to put it. 🙈
> 
> But, since I have the `Fixes #...` in the body of the PR it should be fine, right?

It is fine.

tests/binary.rs
lu-zero: Do not use `args()` and keep the pathbuf please.

tests/binary.rs
```diff
@@ -74,65 +72,39 @@ mod binary {
...
-      .arg("--reservoir-frame-delay")
-      .arg("14")
-      .arg("--bitrate")
-      .arg("1000")
-      .arg("-o")
-      .arg(&outfile)
-      .arg("--first-pass")
-      .arg(&passfile)
+    get_common_cmd(outfile)
+      .args(&["--reservoir-frame-delay", "14"])
```
lu-zero, line 98: This is fine but the line below is not.

tests/binary.rs
lu-zero: You are right

tests/binary.rs
Urhengulas: What is the problem?

tests/binary.rs
Urhengulas: Okay. Also fine.

tests/binary.rs
```diff
@@ -74,65 +72,39 @@ mod binary {
...
+      .args(&[OsStr::new("--first-pass"), passfile])
```
lu-zero, line 99: It is more readable to use `arg().arg()` instead.

Urhengulas: That's true.

Urhengulas:
@lu-zero I dropped the two commits. Should be fine now.

PR #2345, opened 2020-05-28
Version bump

lu-zero:
Version bump and known working `Cargo.lock`.


lu-zero:
the Cargo.lock file is probably too old, time to update the CI once more. (or reconsider committing the lockfile)

PR #2348, opened 2020-05-29
Update structure documentation

vibhoothiiaanand:
After the splitting of context(#2303), the file structure is
changed.


doc/STRUCTURE.md
```diff
@@ -6,6 +6,7 @@
 
 ![Image](structure.png)
 
+Source: [draw.io](https://drive.google.com/file/d/1ZoVobQRdWoE8qrE-Ma1hhybDXxwNMAJN/view?usp=sharing)
```
lu-zero, line 9: might be a better idea to store it directly in git.

vibhoothiiaanand: Indeed, 
I did update the PR ^^;

PR #2349, opened 2020-05-29
Prompt overwrite file message

### LINKED ISSUE #2295: Overwrite prompt missing.
Please add something to the effect of

`Are you sure you want to override "preciousbits.ivf"? [y/N]`

I just got burned by the lack of this little tidbit.
Thanks.

Dayof:
PR to fix #2295 :
- Add function ``IvfMuxer::check_file(path)``
- Calling from ``bin/muxer/mod.rs``, ``create_muxer(path)``
- Default value "N" 
- Add arg ``-y`` to overwrite uninteractive

### DIFF
--- a/src/bin/muxer/ivf.rs
+++ b/src/bin/muxer/ivf.rs
@@ -11,6 +11,7 @@
 use super::Muxer;
 use ivf::*;
 use rav1e::prelude::*;
+use std::fs;
 use std::fs::File;
 use std::io;
 use std::io::Write;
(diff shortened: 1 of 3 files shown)


rzumer:
The implementation looks good to me, but it needs a new CLI option that will bypass the check and always overwrite, for applications that don't want to rely on user input.

Dayof:
@rzumer do you think is good to merge the cli addition in the PR ?

rzumer:
Yes, you might want to make it a separate commit but it's definitely better to have it merged at the same time.

src/bin/muxer/ivf.rs
```diff
@@ -45,6 +46,27 @@ impl Muxer for IvfMuxer {
 }
 
 impl IvfMuxer {
+  pub fn check_file(path: &str) -> Result<(), CliError> {
+    if fs::metadata(path).is_ok() {
+      print!("File '{}' already exists. Overwrite ? [y/N] ", path);
```
lu-zero, line 51: `eprint` and `eprintln` below might be better.

src/bin/muxer/ivf.rs
```diff
@@ -45,6 +46,27 @@ impl Muxer for IvfMuxer {
...
+      let mut option_input = String::new();
+      io::stdin()
+        .read_line(&mut option_input)
+        .expect("Failed to read option.");
+
+      match option_input.as_str().trim() {
+        "y" | "Y" => return Ok(()),
+        _ => {
+          println!("Not overwriting, exiting.");
+          std::process::exit(0);
```
lu-zero, line 63: You should return a CliError and forward it up.

Dayof: Sorry for the silly question, but instead of what ? ``std::process::exit(0);`` ?

lu-zero:
``` rust
return Err(CliError::NotOverwriting)
```

PR #2350, opened 2020-05-30
Rate Control new C-API

lu-zero:
It needs some testing and opinions.

### DIFF
--- a/src/capi.rs
+++ b/src/capi.rs
@@ -256,6 +256,36 @@ impl EncContext {
       EncContext::U16(ctx) => ctx.twopass_out(),
     }
   }
+
+  fn rc_summary_size(&self) -> usize {
+    match self {
+      EncContext::U8(ctx) => ctx.rc_summary_size(),
+      EncContext::U16(ctx) => ctx.rc_summary_size(),
+    }
+  }
+
+  fn rc_receive_pass_data(&mut self) -> rav1e::RcData {
+    match self {
+      EncContext::U8(ctx) => ctx.rc_receive_pass_data(),
+      EncContext::U16(ctx) => ctx.rc_receive_pass_data(),
+    }
+  }
+
+  fn rc_second_pass_data_required(&self) -> usize {
+    match self {
+      EncContext::U8(ctx) => ctx.rc_second_pass_data_required(),
+      EncContext::U16(ctx) => ctx.rc_second_pass_data_required(),
+    }
+  }
+
+  fn rc_send_pass_data(
+    &mut self, data: &[u8],
+  ) -> Result<(), rav1e::EncoderStatus> {
+    match self {
+      EncContext::U8(ctx) => ctx.rc_send_pass_data(data),
+      EncContext::U16(ctx) => ctx.rc_send_pass_data(data),
+    }
+  }
 }
 
 /// Encoder context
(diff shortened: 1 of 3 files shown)


src/capi.rs
```diff
@@ -356,6 +379,42 @@ pub unsafe extern fn rav1e_config_default() -> *mut Config {
   Box::into_raw(c)
 }
 
+/// Setup a single pass rate control
+///
```
dwbuiten, line 383: This is extremely confusing.

Does this mean 1-pass ABR *only*? As in, if I call this function, I can never do a second pass? If so, this is a bad API, and doesn't account for future uses of e.g. CRF first pass, followed by a second pass.

Or does it mean the first of two passes?

What ar its requirements and restrictions on what must be set?

src/capi.rs
```diff
@@ -356,6 +379,42 @@ pub unsafe extern fn rav1e_config_default() -> *mut Config {
...
+#[no_mangle]
+pub unsafe extern fn rav1e_config_set_rc_single_pass(cfg: *mut Config) {
+  (*cfg).cfg.rate_control.summary = None;
+}
+
+/// Setup a second pass rate control using the provided summary
+///
```
dwbuiten, line 390: Hiw exactly are we supposed to pass only the summary? FIrst pass stats files will be written as a single file blob. How is a user suppose to parse out only the summary and pass it? Rav1e provides absolutely no functions to parse these, and *no* documentation of the pass data format. Is it up to the use to make their own parseable format with stats "packets"?

src/capi.rs
```diff
@@ -356,6 +379,42 @@ pub unsafe extern fn rav1e_config_default() -> *mut Config {
...
+  if summary.is_none() {
+    -1
+  } else {
+    (*cfg).cfg.rate_control.summary = summary;
+
+    0
+  }
+}
+
+/// Request to emit pass data
```
dwbuiten, line 407: Related to the comment on the `rav1e_config_set_rc_single_pass`, the requirements and restrictions on this fuction *must* be documented. How to call functions and in what order, and what depends on what?

src/capi.rs
```diff
@@ -738,6 +797,81 @@ pub unsafe extern fn rav1e_twopass_out(ctx: *mut Context) -> *mut Data {
...
+  Frame,
+}
+
+/// Return the first pass data
+///
+/// Call it after rav1e_receive_packet().
+///
+/// use rav1e_data_unref() to free the data.
+///
+/// It will return a `RcDataKind::Summary` once the encoder is flushed.
```
dwbuiten, line 824: Are we expected to write this `Summary` data at the end, since no function was added for getting the size of the summary data? Or do we use the old/existing one?

I also think the old ratecontrol API should be removed while this one is added, as discussed on the watercooler. *Especially* if they share functions, since that is very confusing.

src/capi.rs
```diff
@@ -738,6 +797,81 @@ pub unsafe extern fn rav1e_twopass_out(ctx: *mut Context) -> *mut Data {
...
+  }));
+
+  kind
+}
+
+/// Number of pass data packets required to progress the encoding process.
+///
+pub unsafe extern fn rav1e_rc_second_pass_data_required(
+  ctx: *const Context,
+) -> i32 {
```
dwbuiten, line 847: This is not useful. Rav1e does not provide a way to packetize an existing stats file, not does it provide a documentation of the stats file format.

Did we really decide to leave it up to the user to make up their own packetized stats file format so it can read in separate stats packets and their types? That's awful.

src/capi.rs
```diff
@@ -738,6 +797,81 @@ pub unsafe extern fn rav1e_twopass_out(ctx: *mut Context) -> *mut Data {
...
+/// Frame-specific Packets only.
+///
+/// Call it before receive_packet()
+///
+/// Returns:
+/// - `0` on success,
+/// - `> 0` if the input queue is full
+/// - `< 0` on unrecoverable failure
+pub unsafe extern fn rav1e_rc_send_pass_data(
+  ctx: *mut Context, data: *const u8, len: size_t,
```
dwbuiten, line 861: See previous comments on packetization.

src/capi.rs
lu-zero: It simply reset to our default, single pass rate control with no packet data to emit or consume.

src/capi.rs
lu-zero: Is it up to the user to decide what to do with the data provided.
You get its type, dimension and size. How it is stored is not a concern of the library.

src/capi.rs
lu-zero: It doesn't immediately trigger a validation so you can call it anytime, same as `rav1e_config_set_rc_single_pass`

src/capi.rs
lu-zero: I will remove them. 

Every time you call this function you get a packet of data. Once the encoder if flushed you receive the Summary and you can write it back up or in a separate file or at the end as you like.

src/capi.rs
lu-zero: We did. I/O should not concern the encoder library.

src/capi.rs
dwbuiten: That is too rigid. It doesn't allow for arbitrary first passes, like the CRF first pass I descibed above, as far as I can tell, which is important to support.

src/capi.rs
dwbuiten: *HARD( disagree. Like, extremely hard*. This is obscenely bad, in my opinion. It just shunts more boilerplate onto the user, and *no* other encoding library does it like this.

src/capi.rs
dwbuiten: I do not mean just validation. There is no documentation provided at all on how to use this API correctly currently. A user most figure out the (IMO very non-straightforward) API calls, in what order, and which to call, to us the API. I am not a proponent of this.

src/capi.rs
dwbuiten: It doesn't need to be I/O. It could be a buffer. A user *SHOULD NOT* have to figure out how to packetize and write their data. It should be a buffer that we pass, and the encode consumes, and says how much it consumes.

I have an extremely hard stance on this. It's fricking stupid to force this on a user.

src/capi.rs
lu-zero: This function has to be used ONLY if you happen to have a Configuration set up for a second pass encoding and you want to reuse it to spawn a single pass encoding context.

Another way to support this can be:
``` c
rav1e_config_set_rc_summary(rac, NULL, 0);
```
But feels fairly uglier.

Configuration is just a stateless factory of Context: Changing the configuration has no impact to the Contexts spawned from it in the past.

tdaede: If you skip this function and instead just use rav1e_config_set_rc_summary(rac, NULL, 0) - do you actually need to call that? Or is the default configuration, e.g. I set up a final pass, but don't provide a rc summary, then it's assumed to be single pass?

lu-zero: The latter. By default it is single pass, you do not have to call any function to configure it if you did not set a summary before.

lu-zero: As discussed the function got removed, moving the reset capability to 
``` c
rav1e_config_set_rc_summary(rac, NULL, 0);
```

src/capi.rs
lu-zero: I changed it as you asked for. Now the functions update the buffer and size on success or return the expected size on failure.

src/capi.rs
lu-zero: We can add more explanations in general.

src/capi.rs
lu-zero: Removed.

src/capi.rs
lu-zero: I'll document it better, see the test for an example on how to use it.

src/capi.rs
lu-zero: Addressed

kodabb:
@daemon404 can you give this a final read when you have a chance please?

dwbuiten:
@kodabb i can, but you pinged the wrong person

lu-zero:
Shall we shelve the old api removal?

lu-zero:
Api removal shelved, @dwbuiten is it ok to merge?

PR #2359, opened 2020-05-31
Unbreak the rate control

### LINKED ISSUE #2354: The cli is not returning the error to the os
It has the unpleasant side effect of silently accepting breaks in the cli tests among many things.

Yesterday I fixed it, I'll send a PR once the tests are also fixed.

lu-zero:
The integration tests were not reporting the errors due the incomplete error management in the CLI.

This set addresses the errors and make sure the tests work as intended.

Also fixes #2354.

### DIFF
--- a/tests/binary.rs
+++ b/tests/binary.rs
@@ -42,7 +42,7 @@ mod binary {
 
   fn get_common_cmd(outfile: &PathBuf) -> Command {
     let mut cmd = get_rav1e_command();
-    cmd.args(&["--bitrate", "1000"]).arg("-o").arg(outfile);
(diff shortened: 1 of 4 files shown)


tests/binary.rs
```diff
@@ -30,14 +30,14 @@ mod binary {
   #[cfg(not(windows))]
   fn get_rav1e_command() -> Command {
     let mut cmd = Command::cargo_bin("rav1e").unwrap();
-    cmd.env_clear();
+    cmd.arg("-y").env_clear();
```
shssoichiro, line 33: I think it makes more sense to put the arg under `get_common_cmd`

lu-zero: Sure, and done :)

PR #2361, opened 2020-05-31
arm: tables: Fix the symbol redefinition error

### LINKED ISSUE #2352: Aarch64 asm isn't buildable using clang
On a system with clang as main cc:

```
running: "cc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-I" "." "-I" "/home/lu_zero/rav1e/target/release/build/rav1e-b17e0a559c668c01/out" "-Wall" "-Wextra" "-o" "/home/lu_zero/rav1e/target/release/build/rav1e-b17e0a559c668c01/out/src/arm/64/ipred.o" "-c" "src/arm/64/ipred.S"
cargo:warning=src/arm/tables.S:44:1: error: invalid symbol redefinition
cargo:warning=rav1e_mc_subpel_filters:
cargo:warning=^
cargo:warning=src/arm/tables.S:125:1: error: invalid symbol redefinition
cargo:warning=rav1e_filter_intra_taps:
cargo:warning=^
cargo:warning=src/arm/tables.S:171:1: error: invalid symbol redefinition
cargo:warning=rav1e_sgr_x_by_x:
cargo:warning=^
cargo:warning=src/arm/tables.S:195:1: error: invalid symbol redefinition
cargo:warning=rav1e_mc_warp_filter:
cargo:warning=^
cargo:warning=src/arm/tables.S:302:1: error: invalid symbol redefinition
cargo:warning=rav1e_sm_weights:
cargo:warning=^
cargo:warning=src/arm/tables.S:326:1: error: invalid symbol redefinition
```

vibhoothiiaanand:
Compiling with clang-6, clang-9 there seems symbol redefinition
error:
cargo:warning=src/arm/tables.S:44:1: error: invalid symbol
redefinition
cargo:warning=rav1e_mc_subpel_filters:
cargo:warning=^

The symbols are actually being redefined which GCC did not catch
while clang did, so by exporting the constant which is defined in
the src/arm/arm.S for all macros fixes the issue.

Tested with clang-6, 9, GCC 7.5

Also have compiled all 3 individually to make sure there are no build failure

Closes #2352

### DIFF
--- a/src/arm/tables.S
+++ b/src/arm/tables.S
@@ -38,9 +38,7 @@
 
 #include "src/arm/asm.S"
 
-.global X(mc_subpel_filters)
-.hidden X(mc_subpel_filters)
-const X(mc_subpel_filters), align=3
+const mc_subpel_filters, export=1, align=3
 .byte  0,   1,  -3,  63,   4,  -1,   0,   0 /* REGULAR */
 .byte  0,   1,  -5,  61,   9,  -2,   0,   0
 .byte  0,   1,  -6,  58,  14,  -4,   1,   0
@@ -118,9 +116,7 @@ const X(mc_subpel_filters), align=3
 .byte  0,   0,   1,  17,  31,  15,   0,   0
 endconst
 
-.global X(filter_intra_taps)
-.hidden X(filter_intra_taps)
-const X(filter_intra_taps), align=4
+const filter_intra_taps, export=1,  align=4
 .byte  -6,  10,  -5,   2,  -3,   1,  -3,   1 /* 0 */
 .byte  -4,   6,  -3,   2,  -3,   2,  -3,   1
 .byte   0,   0,  10,   0,   1,  10,   1,   2
@@ -163,9 +159,7 @@ const X(filter_intra_taps), align=4
 .byte  14,   0,  12,   0,  11,   0,   9,   0
 endconst
 
-.global X(sgr_x_by_x)
-.hidden X(sgr_x_by_x)
-const X(sgr_x_by_x), align=4
+const sgr_x_by_x, export=1, align=4
 .byte 255, 128,  85,  64,  51,  43,  37,  32,  28,  26,  23,  21,  20,  18,  17
 .byte  16,  15,  14,  13,  13,  12,  12,  11,  11,  10,  10,   9,   9,   9,   9
 .byte   8,   8,   8,   8,   7,   7,   7,   7,   7,   6,   6,   6,   6,   6,   6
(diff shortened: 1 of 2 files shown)


src/arm/tables.S
```diff
@@ -40,8 +40,7 @@
 
 .global X(mc_subpel_filters)
 .hidden X(mc_subpel_filters)
-const X(mc_subpel_filters), align=3
-X(mc_subpel_filters):
+const X(mc_subpel_filters), export=1, align=3
```
mstorsjo, line 43: No, this is wrong.

To fix the symbol redefinition problem, just remove all the `X(symbolname):` lines. Do that in one single commit first. They are redundant and clang is correctly pointing that out. This part doesn't relate to my patch at all.

Then my patch allows you to get rid of three things:
1) The `.global X(symbolname)`
2) The `.hidden X(symbolname)`
3) The `X()` wrapper in the const macro invocation.

So the correct version of this PR would be:
1) Remove superfluous symbols
2) Add my patch for adding the export parameter
3) Remove all `.global`, `.hidden` and `X()` in const invocations, and add `export=1` in the const invocations.

Using `const X(foo), export=1` is wrong, as `export=1` should imply the same prefix as `X()` adds. If it doesn't, there's something missing still in my patch.

vibhoothiiaanand: Thank you for the review,
I have updated the patchset based on your suggestion, can you please check again.

src/arm/tables.S
```diff
@@ -41,7 +41,6 @@
 .global X(mc_subpel_filters)
 .hidden X(mc_subpel_filters)
 const X(mc_subpel_filters), align=3
-X(mc_subpel_filters):
 .byte  0,   1,  -3,  63,   4,  -1,   0,   0 /* REGULAR */
```
mstorsjo, line 42: For the commit message of this one, I'd write the last section like this:

"The const macro implicitly defines a label, so the manually defined ones are duplicates."

src/arm/tables.S
```diff
@@ -38,9 +38,7 @@
 
 #include "src/arm/asm.S"
```
mstorsjo, line 40: Here, I'd write

```
arm: tables: Use the const macro's new export parameter

This avoids having to manually use platform specific assembler attributes for marking the symbols as exported.
```

vibhoothiiaanand:
> The changes look correct now, but I have two suggestions for the commit messages, making them (IMO) easier to understand.

Sure, they seem better than the current one,

mstorsjo:
LGTM, FWIW.

Issue #2351, opened 2020-05-30
hbd satd simd for x86 has high compile time impact


KyleSiefring:
Compile time with the next/prev 2 commits.

| Commit Hash                              | Full Build Time (Release) | Partial Build Time (Release) | Binary Size (Release) | Full Build Time (Debug) | Partial Build Time (Debug) | Binary Size (Debug) |
|------------------------------------------|---------------------------|------------------------------|-----------------------|-------------------------|----------------------------|---------------------|
| 97f642ec8e815b1d905b7b650f5a32903efcc67e | 255.92298901              | 171.193918164                | 28554560              | 67.018268034            | 9.55852379                 | 56658088            |
| 7d60a68f2c6d4447c6a1421abbc74eb3f49057a4 | 258.180816743             | 167.130605781                | 28562696              | 68.686580745            | 6.350572624                | 56659232            |
| 4dfda58791ab7c2406bffc78cd25602fde49feac | 263.215725554             | 175.597711184                | 29097776              | 70.264331135            | 7.117904333                | 57477808            |
| 8fa818bce51d42d180cf567d51377aa45b7aca6d | 263.776486008             | 176.768010131                | 29097776              | 69.681629602            | 6.908273139                | 57477808            |
| cd2aadf4908469ca302de1d335c23bc899f053c2 | 264.656863175             | 175.759415416                | 29097776              | 70.394743488            | 7.192520752                | 57477808            |

PR #2363, opened 2020-06-01
Reduce compile time for hbd satd intrinsics

KyleSiefring:
5m 00 to 4m 47 on awcy. 30188832 to 29804944 bytes for release build.


KyleSiefring:
Closes #2351

PR #2364, opened 2020-06-01
Clean up recon_intra.rs, which are ported from libaom

ycho:
- Remove left over libaom code when poring to rav1e
- Uncomment some array declarations, which are not used now but required
for PARTITION_VERT_A and _B in the future.


ycho:
> Seems fine to me, thank you :)

Thanks!

PR #2367, opened 2020-06-03
Bump the Cargo.lock

lu-zero:
This Cargo.lock is know to work with rustc 1.38.0.
The previous was not because of the doctest cfg key being still unstable on this version. `regex-1.3.9` addresses this.


lu-zero:
We can release as 0.3.3 so there isn't confusion regarding on what is what probably.

PR #2369, opened 2020-06-05
Prepare for release

lu-zero:
Release 0.3.2 started pinning the dependencies using Cargo.lock.
Release 0.3.3 pins a newer regex version to support older (1.38.0) compiler
versions.


lu-zero:
Once it lands https://github.com/xiph/rav1e/pull/2371 will be part of this PR

Issue #2311, opened 2020-05-17
Corrupt bitstream: multi-pass encoding


mikoim:
I encoded video with target bitrate mode to compare quality and performance. rav1e seems to output corrupt bitstream in specific parameters.

```bash
for TILE in 1 2 4 8 16 32 64
do
    for SPEED in 0 1 2 3 4 5 6 7 8 9 10
    do
        echo "--speed $SPEED --tiles $TILE"
        rav1e --output /dev/null --bitrate 3000 --first-pass $SPEED-$TILE.stat --speed $SPEED --tiles $TILE input.y4m
        rav1e --output $SPEED-$TILE.ivf --bitrate 3000 --second-pass $SPEED-$TILE.stat --speed $SPEED --tiles $TILE input.y4m
    done
done
```

## Playable?
| tiles/speed 	|      0     	|      1     	|      2     	|      3     	|      4     	|      5     	|      6     	|      7     	|      8     	|      9     	| 10 	|
|:-----------:	|:----------:	|:----------:	|:----------:	|:----------:	|:----------:	|:----------:	|:----------:	|:----------:	|:----------:	|:----------:	|:--:	|
|      1      	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|  ✅ 	|
|      2      	|      ❌     	|      ❌     	|      ❌     	|      ❌     	|      ❌     	|      ❌     	| not tested 	| not tested 	| not tested 	| not tested 	|  ✅ 	|
|      4      	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|  ✅ 	|
|      8      	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|  ✅ 	|
|      16     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|  ✅ 	|
|      32     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|      ✅     	|  ✅ 	|
[...]

Issue #2212, opened 2020-03-30
Crash when using 4 tiles for 1080p 4:2:2 input (but not with 8 tiles)


gnafuthegreat:
Description:
When encoding 1080p 4:2:2 video (even a single black frame) with 4 tiles, rav1e crashes.  With 8 tiles, the encoding finishes normally.

Command line:
rav1e -o /dev/null --tiles 4 black-frame.y4m

Result with debug build and RUSTBACKTRACE=full:
[rav1e-422-crash.txt](https://github.com/xiph/rav1e/files/4402954/rav1e-422-crash.txt)

```
Backtrace (most recent call first):
  File "<unknown>", line 0, in core::panicking::panic
  File "/var/home/gideon/rav1e/src/lrf.rs", line 1416, in rav1e::lrf::RestorationState::new
    uv_unit_log2 - uv_sb_v_log2,
  File "/var/home/gideon/rav1e/src/encoder.rs", line 384, in rav1e::encoder::FrameState<T>::new_with_frame
    let rs = RestorationState::new(fi, &frame);
  File "/var/home/gideon/rav1e/src/api/internal.rs", line 215, in rav1e::api::internal::FrameData<T>::new
    let fs = FrameState::new_with_frame(&fi, frame);
  File "/var/home/gideon/rav1e/src/api/internal.rs", line 426, in rav1e::api::internal::ContextInner<T>::set_frame_properties
    self.frame_data.insert(output_frameno, FrameData::new(fi, frame.clone()));
  File "/var/home/gideon/rav1e/src/api/internal.rs", line 744, in rav1e::api::internal::ContextInner<T>::compute_frame_invariants
    while self.set_frame_properties(self.next_lookahead_output_frameno).is_ok()
  File "/var/home/gideon/rav1e/src/api/internal.rs", line 348, in rav1e::api::internal::ContextInner<T>::send_frame
    self.compute_frame_invariants();
  File "/home/gideon/.cargo/registry/src/github.com-1ecc6299db9ec823/rayon-core-1.7.0/src/thread_pool/mod.rs", line 110, in rayon_core::thread_pool::ThreadPool::install::{{closure}}
    self.registry.in_worker(|_, _| op())
  File "/home/gideon/.cargo/registry/src/github.com-1ecc6299db9ec823/rayon-core-1.7.0/src/registry.rs", line 447, in rayon_core::registry::Registry::in_worker_cold::{{closure}}::{{closure}}
    op(&*worker_thread, true)
[...]

PR #2371, opened 2020-06-05
Fix for issue 2311: Corrupt bitstream: multi-pass encoding

xiphmont:
The problem had nothing to do with multipass.  It's related to issue 
#2212.

Specifically, the LRF code correctly fixes up the UV LRU size for
4:2:2, but does not reconcile Y.  Because only 4:2:0 can have
different LRU sizes for Y and UV, it leads to a desync.  The fix is to
treat the corrected 4:2:2 UV LRU size value as a maximum, and when Y
and UV differ, set Y and UV to min(Y, UV).


xiphmont:
> Please move `Fixes #2311` in the commit message body instead of the subject.
> If you have the spec references handy could you please add it to the comments?

done and done.

PR #2375, opened 2020-06-08
Initial wasm-pack support

lu-zero:
This is the bare minimum to have `wasm-pack build` and `wasm-pack test` running.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,8 +112,13 @@ features = ["parallel"]
 [target.'cfg(unix)'.dependencies]
 signal-hook = { version = "0.1.9", optional = true }
 
+[target.'cfg(not(target_arch="wasm32"))'.dev-dependencies]
+assert_cmd = { version = "1.0" }
+
+[target.'cfg(target_arch="wasm32")'.dev-dependencies]
+wasm-bindgen-test = "0.2"
+
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


lu-zero:
Ideally the jsapi could live in a separate crate as well, but @Urhengulas can decide.

Urhengulas:
I prefer having it in an own crate, because it separates some concerns and make it way cleaner imo.

lu-zero:
Go for it :) You may use this as base or redo the changes to account for a separate rav1e crate

Cargo.toml
```diff
@@ -112,8 +112,13 @@ features = ["parallel"]
 [target.'cfg(unix)'.dependencies]
 signal-hook = { version = "0.1.9", optional = true }
 
+[target.'cfg(not(target_arch="wasm32"))'.dev-dependencies]
+assert_cmd = { version = "1.0" }
```
Urhengulas, line 116:
```toml
[target.'cfg(not(target_arch="wasm32"))'.dev-dependencies]
assert_cmd = { version = "1.0" }
```

I think I don't need this, when having it in a separate crate, do I?

lu-zero: You should not.

lu-zero:
#2382 should take care of it.

PR #2365, opened 2020-06-02
Deprecate wiki

Urhengulas:
# Description

This PR deprecates the wiki, by moving or dropping the files to the main repo.

Additionally the documentation in the repo gets a cleanup. Also it gets tweaked for potential [GitHub-Pages](https://pages.github.com/) Support (example: https://urhengulas.github.io/rav1e)

Fixes #2346 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactoring (non-breaking change which doesn't change functionality
- [x] Documentation update

## Open TODOs
- [ ] Handle `Quality-&-Speed-Features.md` (waiting for feedback)
- [x] Handle `Roadmap.md` -> drop it + @negge will go through it and convert potential open questions to issues

# How Has This Been Tested?

- [x] `README.md` and `docs/*` look good and render correctly
- [x] Looks good on gh-pages

PR #2382, opened 2020-06-09
minimal jsapi with wasm_bindgen

### LINKED ISSUE #2381: Write a jsapi.rs
- [ ] write the API using [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen)
- [ ] write tests for the API using [wasm-bindgen-test](https://rustwasm.github.io/docs/wasm-bindgen/wasm-bindgen-test/index.html)

Open question: separate-crate or same-crate?

Urhengulas:
Proof of concept for #2381

### DIFF
--- a/rav1e_js/Cargo.toml
+++ b/rav1e_js/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "rav1e_js"
+version = "0.1.0"
+authors = ["Urhengulas <johann.hemmann@code.berlin>"]
+edition = "2018"
+description = "JavaScript bindings for rav1e"
+repository = "https://github.com/xiph/rav1e/"
+license= "BSD-2-Clause"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[features]
+default = ["console_error_panic_hook"]
+
+[dependencies]
+rav1e = { path = ".." }
+wasm-bindgen = "0.2.63"
+web-sys = {version = "0.3", features = ["console"]}
+
+# The `console_error_panic_hook` crate provides better debugging of panics by
+# logging them with `console.error`. This is great for development, but requires
+# all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
+# code size when deploying.
+console_error_panic_hook = { version = "0.1.6", optional = true }
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3.13"
--- a/rav1e_js/src/lib.rs
+++ b/rav1e_js/src/lib.rs
@@ -1,0 +1,94 @@
+// Copyright (c) 2020, The rav1e contributors. All rights reserved
+//
+// This source code is subject to the terms of the BSD 2 Clause License and
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+mod utils;
+
+use rav1e::config::SpeedSettings;
+use rav1e::prelude::*;
+
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+extern {
+  #[wasm_bindgen(catch)]
+  fn alert(s: &str) -> Result<(), JsValue>;
+}
+
+/// Encode the same tiny blank frame 30 times
+#[wasm_bindgen]
+pub fn simple_encoding(repeats: u32) {
+  let msg = simple_encoding_without_alert(repeats);
+
+  match alert(&*msg) {
+    Ok(_) => {}
+    Err(_) => log!("Error calling alert()"),
+  };
+}
+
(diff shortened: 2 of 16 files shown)


Urhengulas:
> I am thinking should we have `rav1e_` prefix for the folder? since we are inside root(rav1e/)

True. `wasm` would be enough, cause you still get the name via path (`rav1e/wasm`).

But I am still not 100% sure about the name. Maybe sth like `jsapi`, `npm`, `wasm_bindgen` is better.

rav1e_wasm/src/lib.rs
```diff
@@ -0,0 +1,56 @@
+mod utils;
```
vibhoothiiaanand, line 1: Please add the license headers to the newly created files

Urhengulas: Which one?
* rav1e contributors 2017-2020 (e.g. `src/context/block_unit.rs`)
* rav1e contributors 2017-2019 (e.g. `build.rs`)
* AOM 2001-2016 + rav1e contributors 2017-2019 (e.g. `src/capi.rs`)

I guess rav1e contributors 2017-2020?

rav1e_wasm/README.md
```diff
@@ -0,0 +1,34 @@
+# rav1e_wasm
+
+- [🚴 Usage](#-usage)
```
vibhoothiiaanand, line 3: Do we require to have this table here,
Seems to be a stray change for a small readme, I would avoid it

rav1e_wasm/README.md
```diff
@@ -0,0 +1,34 @@
...
+  - [Example Website](#example-website)
+- [🔋 Batteries Included](#-batteries-included)
+
+## 🚴 Usage
```
vibhoothiiaanand, line 7: Can we avoid the emojis also?

rav1e_wasm/README.md
```diff
@@ -0,0 +1,34 @@
...
+wasm-pack test --headless --firefox
+```
+
+### Example Website
+```bash
+rav1e_wasm$ wasm-pack build
+rav1e_wasm$ cd www-example/
+rav1e_wasm/www$ npm install
+rav1e_wasm/www$ npm run start
+# website served at localhost:8080
```
vibhoothiiaanand, line 23: Either keep it or remove it, having commented one seems odd.

rav1e_wasm/src/lib.rs
Urhengulas: Gotcha

rav1e_wasm/README.md
Urhengulas: It's there because I would like to establish consistent styling for all the docs. It matches the style of #2365.

But actually we could move this to the main README, since it is so short

rav1e_wasm/README.md
Urhengulas: Sure

rav1e_wasm/README.md
vibhoothiiaanand: Sure, 
Once that PR is merged, we can have it

rav1e_wasm/README.md
Urhengulas: > Also, we would require the CI integration for the npm thing for making this patchset complete

I am a bit unsure about how to tag the package. Should I just always increment the version? Or is it possible to use the commit-hash as version?

Best handling imo would be to keep the rav1e-rs and rav1e-npm version in sync, but I don't really see how this is possible.

Urhengulas:
> Also, we would require the CI integration for the npm thing for making this patchset complete

I am a bit unsure about how to tag the package. Should I just always increment the version? Or is it possible to use the commit-hash as version?

Best handling imo would be to keep the rav1e-rs and rav1e-npm version in sync, but I don't really see how this is possible.

Urhengulas:
Or @vibhoothiiaanand, one step back: Do you "just" want to run the tests, linting etc., or also publish it to npm.

vibhoothiiaanand:
@Urhengulas Yes, that would be fine for the initial set, basic lint checks, tests, build check etc, 
which would be fairly good, we can think and land the publishing to npm when the library is more matured.

rav1e_js/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+#[wasm_bindgen]
+extern {
+  fn alert(s: &str);
+}
+
+const LIMIT: u32 = 30;
+
+/// Encode the same tiny blank frame 30 times
+#[wasm_bindgen]
+pub fn simple_encoding() {
```
lu-zero, line 26: This could be a test

Urhengulas: Good idea

rav1e_js/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "rav1e_js"
+version = "0.1.0"
+authors = ["Urhengulas <johann.hemmann@code.berlin>"]
+edition = "2018"
+
```
vibhoothiiaanand, line 6: Nit: Extra Whitespace

vibhoothiiaanand:
Also, you can squash the relevant changes into different commits accordingly without breaking the build.
You can confirm it using `git rebase --exec -i `

PR #2386, opened 2020-06-12
Run grcov on stable

Luni-4:
In this way it shouldn't be no more necessary to pin a nightly version for `grcov`


lu-zero:
Looks like nasm-rs 0.1.8 has problems with certain version of nasm.

https://github.com/medek/nasm-rs/pull/31 should address it.

Luni-4:
@lu-zero 

I rebased this PR, so it works now

PR #2388, opened 2020-06-13
Correctly chunk sse calc in importance block sizes

KyleSiefring:
A previous patch tried to do this, but missed the 4x4 case on subsampled
frames. Roughly 2% speedup on the default speed level.

https://beta.arewecompressedyet.com/?job=ref_c1432f0%402020-06-13T00%3A43%3A41.614Z&job=sse4x4%402020-06-13T00%3A42%3A15.960Z

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -240,8 +240,8 @@ pub fn sse_wxh<T: Pixel, F: Fn(Area, BlockSize) -> DistortionScale>(
 
   // To bias the distortion correctly, compute it in blocks up to the size
   // importance block size in a non-subsampled plane.
-  let imp_block_w = IMPORTANCE_BLOCK_SIZE.min(w);
-  let imp_block_h = IMPORTANCE_BLOCK_SIZE.min(h);
+  let imp_block_w = IMPORTANCE_BLOCK_SIZE.min(w << src1.plane_cfg.xdec);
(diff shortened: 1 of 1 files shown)


src/rdo.rs
```diff
@@ -240,8 +240,8 @@ pub fn sse_wxh<T: Pixel, F: Fn(Area, BlockSize) -> DistortionScale>(
 
   // To bias the distortion correctly, compute it in blocks up to the size
   // importance block size in a non-subsampled plane.
-  let imp_block_w = IMPORTANCE_BLOCK_SIZE.min(w);
-  let imp_block_h = IMPORTANCE_BLOCK_SIZE.min(h);
+  let imp_block_w = IMPORTANCE_BLOCK_SIZE.min(w << src1.plane_cfg.xdec);
+  let imp_block_h = IMPORTANCE_BLOCK_SIZE.min(h << src1.plane_cfg.ydec);
```
ycho, line 244: I think this change is wrong because the w and h is already adjusted for chroma sampling when they are passed to sse_wxh( ) when processing chroma cases.

I've just run before and after your commit on awcy for speed 0, so let's wait for the result/
https://beta.arewecompressedyet.com/?job=test_master_db69fb4f637a_1f_s0%402020-06-16T19%3A33%3A58.377Z&job=test_master_0680e52a8_1f_s0%402020-06-16T19%3A34%3A36.480Z

I found that your awcy result for default speed 6 already tells that.

Issue #2360, opened 2020-05-31
Cdef apply assembly isn't being run


KyleSiefring:
18450431e83a3407f539a2abb6130d10306b9f9e disabled cdef assembly by providing u16 pixels.

xiphmont:
Kyle: Can you be more specific?  cdef always used u16 in these specific pathways in practice, it just hadn't hardwired it.

shssoichiro:
The input has always been u16, but the output has been generic between u8 and u16. Now the output is hardcoded to u16, which we don't have ASM for and is causing some significant slowdowns for 8-bit encodes.

xiphmont:
Ah.  OK, on it.  I dislike the explosion of generic combinatorics, but that'll get fixed when CDEF no longer requires 16 bit input.

xiphmont:
Current status: Watercooler meeting 7/10/2020 made it known that dav1d now has new assembly for CDEF that removes the 16-bit input restriction (eg, code no longer has to flag border pixels using an in-band value in the frame data).  We should adopt this new assembly immediately (along with equivalent Rust) so that there's no longer any need to go back and forth between Pixel:T and Pixel:u16 throughout the loop filter RDO loop.  Getting the new assembly in is a little work, but then we get a permanent fix for 2360 almost free, and get cleaner code as well.

vibhoothiiaanand:
@xiphmont Hi,
We are not using any of dav1d's CDEF patches as rav1e has diverged, 
We are having our own CDEF assembly from @KyleSiefring AFAIK,
https://github.com/xiph/rav1e/pull/1581
I was also checking history: https://github.com/xiph/rav1e/commits/master/src/x86/cdef_sse.asm
https://github.com/xiph/rav1e/commits/master/src/x86/cdef.asm

When we are also merging the dav1d assembly by release, we are skipping CDEF. So not sure how straightforward will can adopt dav1d's assembly.

Maybe, @barrbrain can comment bit more as he has done dav1d porting for x86 more, IIRC dav1d has more/different filters than us currently.

KyleSiefring:
@xiphmont This is the current function definition for cdef in dav1d. https://code.videolan.org/videolan/dav1d/-/blob/master/src/cdef.h#L54

It looks like the optimization is implemented only when none of the edge flags are set. It's also in place and has pointers to the top and left pixels.

PR #2390, opened 2020-06-14
WIP: PlaneRegion extension, deployment, buffer work, and fix for #2360

xiphmont:
This patch set is a multi-pronged approach at improving buffer usage in the loop filters by using a single consistent region interface (PlaneRegion), extending the PlaneRegion to add functionality previously only available from PlaneSlice while offering better boundary checking, eliminating a couple unneded copies/temp planes, and restoring the loop filter RDO loop to (mostly) Pixel<T> such that 8 bit assembly comes back into play along with reduced memory bandwidth pressure.


src/tiling/plane_region.rs
```diff
@@ -416,6 +507,43 @@ impl<'a, T: Pixel> PlaneRegionMut<'a, T> {
...
+      phantom: PhantomData,
+    }
+  }
+
+  #[inline(always)]
+  // as with subregion_mut above, but allows re-expanding the region.
+  // This will _not_ allow expanding a region beyond the original
+  // rectangle created with new() or from_slice().  As such, it
+  // protects the original Tile/Frame boundaries.
+  pub fn superregion_mut(&self, area: Area) -> PlaneRegionMut<'_, T> {
```
KyleSiefring, line 520: &mut self???

src/tiling/plane_region.rs
```diff
@@ -115,6 +144,7 @@ pub struct PlaneRegion<'a, T: Pixel> {
   pub plane_cfg: &'a PlaneConfig,
   // private to guarantee borrowing rules
   rect: Rect,
+  max_rect: Rect,
```
KyleSiefring, line 147: I already have concerns about the amount of data being passed around with PlaneRegions. You should create a separate struct for when you actually need this data.

xiphmont: I'm not at all sure what the concern is... could you elaborate?  the 16 bytes for a rect saves a few large plane copies and provides an addiitonal measure of limits checking as now PlaneRegions can be used where before it's needed to copy into a frame or use slices (that can only enforce beginning/end).

The only decent way I see to keep this simple is to extend the abstraction transparently.  The goal here is to make mistakes harder.

xiphmont: Oh!  Or do you mean 'PlaneRegion wasn't meant for this' and you're concerned it's being--- perverted....

PR #2400, opened 2020-06-16
Inline various code to improve non-lto performance

KyleSiefring:
Improves non-lto performance by 15-20%

### DIFF
--- a/src/api/config/encoder.rs
+++ b/src/api/config/encoder.rs
@@ -185,7 +185,10 @@ impl EncoderConfig {
   }
 
   /// Is temporal RDO enabled ?
+  #[inline]
   pub fn temporal_rdo(&self) -> bool {
+    // Note: This function is called frequently, unlike most other functions here.
+
     // `compute_distortion_scale` computes a scaling factor for the distortion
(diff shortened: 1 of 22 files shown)


KyleSiefring:
https://beta.arewecompressedyet.com/?job=disable-lto%402020-06-15T23%3A34%3A03.118Z&job=ref_rerun_0680e52a%402020-06-16T14%3A39%3A27.313Z&job=lto-inline6%402020-06-16T16%3A34%3A57.856Z

shssoichiro:
Does this have an impact on non-lto compile time? The reason I ask is because `#[inline]` in Rust is a bit of a misnomer. This annotation just says "This function is allowed to be inlined across crate boundaries" (which would include between rav1e lib and rav1e bin). Which is essentially what LTO does, here we're just saying only specific functions need to be analyzed for that. (The equivalent to C's inline is `#[inline(always)]`.)

KyleSiefring:
disable-lto@2020-06-15T23:34:03.118Z and disable-lto-inline6@2020-06-16T16:00:59.781Z have the same compile times.

Note: I originally put up the wrong link.

PR #2401, opened 2020-06-16
Remove lto from the release profile

KyleSiefring:
The performance difference has shrunk substantially adding inline to a
bunch of functions. The performance difference with or without lto is
about 4 seconds on the slowest clip/qp on a standard awcy run
(MINECRAFT, objective-1-fast, qp 80). The compile time difference is 42
seconds.

If it's possible to have good performance with lto off, then optimizing
without it will provide a good development profile. Going forward, it
would be advisable to use lto to check for performancem problems. It
may be possible to create a seperate profile between dev and release.
AWCY would need to be modified to use this and also work with old
versions of rav1e.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,11 +149,9 @@ debug = true
 codegen-units = 1  # if > 1 enables parallel code generation which improves
                    # compile times, but prevents some optimizations.
                    # Passes `-C codegen-units`. Ignored when `lto = true`.
-lto = true
 
 [profile.bench]
 codegen-units = 1
-lto = true
 
 [workspace]
 members = [".", "ivf", "rav1e_js", "v_frame"]


lu-zero:
Please mention https://github.com/rust-lang/cargo/issues/6988

lu-zero:
Using rustc 1.44.0:

On x86_64 (macos):
with lto:
```
real	3m52.024s
user	9m18.176s
sys	0m17.777s
```
w/out lto:
```
real	3m0.522s
user	8m21.834s
sys	0m15.617s
```

On aarch64(linux):
with lto:
```
real    7m11.829s
user    13m57.967s
sys     0m15.912s
```
w/out lto:
```
real    5m47.165s
user    12m34.384s
sys     0m15.241s
```

I didn't run a benchmark yet

lu-zero:
On Arm lto is about `2%` faster now.

PR #2405, opened 2020-06-17
Remove limit on codegen units for release profile

KyleSiefring:
Decreases compile time on AWCY from ~4 mins to ~2.25 mins. Increases run
time roughly 4% and 10 seconds on the most intense clip/qp (MINECRAFT @
80).

Testing also shows that codegen-units=1 actaully causes a substantial
performance decrease when iterative compilation is enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -143,12 +143,8 @@ opt-level = 1
 
 [profile.release]
 debug = true
-codegen-units = 1  # if > 1 enables parallel code generation which improves
-                   # compile times, but prevents some optimizations.
-                   # Passes `-C codegen-units`. Ignored when `lto = true`.
 
 [profile.bench]
-codegen-units = 1
 
 [workspace]
 members = [".", "ivf", "rav1e_js", "v_frame"]


KyleSiefring:
> As the `[profile.bench]` section is empty after this change, it could be removed.

Next step is going to be to add incremental compilation, so I decided to leave it in.

KyleSiefring:
AWCY run for reference. https://beta.arewecompressedyet.com/?job=codegen-units%402020-06-17T14%3A13%3A14.505Z&job=35e904e-ref%402020-06-17T14%3A24%3A44.831Z

vibhoothiiaanand:
Without this arm encoding will be slow as I said in calls and IRC.

KyleSiefring:
Okay, it sounds like the difference is similar to x86.

<mindfreeze> master: 0.149, 5:35
<mindfreeze> without codegen-unit: 0.146, 5:41
<mindfreeze> config: default speed with  tile-rows 2  50 frames(ducks_take_off)

vibhoothiiaanand:
On quick testing with arm, it is giving similar results as x86
Encoder config: `time cargo run --release ducks_take_off_420_720p50.y4m -s 10 -o /dev/null --tile-rows 2 -l 30`
TC: `nightly-aarch64-unknown-linux-gnu

Speed 10, 30 frames, s10, tile-rows 2 | FPS | Encoding Time
-- | -- | --
master | 0.535 fps | 0:56:27
without codegen-unit | 0.515 | 0:58:40


Default speed, 100 frames, tile-rows 2 | FPS | Encoding Time
-- | -- | --
master | 0.149 | 5:35
without codegen-unit | 0.146 | 5:41

Clip Ducks Take off 720

PR #2406, opened 2020-06-18
Add a script to benchmark build times

KyleSiefring:


KyleSiefring:
Actually, I'm not sure my touch command is working...

tools/bench_build.sh
```diff
@@ -0,0 +1,27 @@
...
+# Move to the correct directory
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "${DIR}/.."
+
+echo dev profile: full build time
+cargo clean
+time cargo build -q
+
+echo;echo;
+echo dev profile: touch src/lib.rs
```
lu-zero, line 15: It shouldn't do much. 
If you add from the end a `pub const FOO: usize = 42;` (and then remove it after the build completes) does it have different results?

lu-zero:
`cargo +nightly rustc -- -Z self-profile` might be of use.

See [here](https://github.com/rust-lang/measureme) for more information

PR #2409, opened 2020-06-18
Add a mean to use an external threadpool

lu-zero:


lu-zero:
@shssoichiro this should simplify a lot the single-instance rav1e-by-gop.

PR #2410, opened 2020-06-18
Deploy pre-releases periodically

### LINKED ISSUE #2404: Deploy pre-releases periodically
Currently, the pre-release cycle is more or less paused, and also the appveyor builds,
So to get the binaries easier, we need to deploy pre-releases periodically as discussed in the IRC on every Tuesday at 21:00 (UTC) which would solve the issue.

Another way to tackle this would be building and generating artifacts for the builds from the master periodically like the script.

Reference:
#2397
#2402

Luni-4:
This PR fixes #2404 

A pre-release is deployed on every Tuesday at 21:00 (UTC) automatically. If an error occurs during the build process, the pre-release can also be deployed manually, that's why the old method has been kept.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,6 +1,9 @@
 name: deploy
 
 on:
+  schedule:
+    # Run every Tuesday at 21:00 (UTC)
+    - cron:  '0 21 * * TUE'
   push:
     tags:
       - 'v*.*.*'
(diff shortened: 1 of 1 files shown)


Luni-4:
> Looks nice, but wouldn't be enough to schedule the tagging and let the original action pick up from there?

I don't know how to schedule the tagging, because the `tags` marker seems to be used ONLY for `push` and `pull_request` events.

Issue #2380, opened 2020-06-09
subregion in TileBlocks(Mut) inconsistent with PlaneRegion


KyleSiefring:
https://github.com/xiph/rav1e/blob/master/src/tiling/tile_blocks.rs#L75

This implementation looks different than what PlaneRegion does. In PlaneRegion the functions are defined as `pub fn subregion(&self, area: Area) -> PlaneRegion<'_, T>` and `pub fn subregion_mut(&mut self, area: Area) -> PlaneRegionMut<'_, T>`. In this case, the lifetime for the returned region is related to &mut self instead of `'a` (a seperate issue was created for this: https://github.com/xiph/rav1e/issues/2379). The other difference is that there are different implementations for subregion and subregion_mut for regions.

Issue #2379, opened 2020-06-09
subregion fn in TileBlocks/TileBlocksMut has broken borrow checking


KyleSiefring:
I don't think this should work: https://github.com/KyleSiefring/rav1e/commit/dfc441b28b011cd319d073eaade58acba5e2a346

Region: https://github.com/xiph/rav1e/blob/master/src/tiling/tile_blocks.rs#L75

This implementation looks different than what PlaneRegion does. In PlaneRegion the functions are defined as `pub fn subregion(&self, area: Area) -> PlaneRegion<'_, T>` and `pub fn subregion_mut(&mut self, area: Area) -> PlaneRegionMut<'_, T>`. In this case, the lifetime for the returned region is related to &mut self instead of `'a`. Unfortunately, changing this causes code to break.

PR #2412, opened 2020-06-18
Fix broken borrow checker for tile_blocks

KyleSiefring:
The borrow rules that were being used were too permissive. To make the
existing code work, it was necessary to split make additional changes.
Instead of calling count_lrf_switchable from ContextWriter, we make it a
function of CDFContext and call it from there.

#2379
#2380

### DIFF
--- a/src/context/frame_header.rs
+++ b/src/context/frame_header.rs
@@ -9,6 +9,45 @@
 
 use super::*;
 
+impl CDFContext {
+  // rather than test writing and rolling back the cdf, we just count Q8 bits using the current cdf
+  pub fn count_lrf_switchable(
+    &self, w: &dyn Writer, rs: &TileRestorationState,
+    filter: RestorationFilter, pli: usize,
+  ) -> u32 {
+    let nsym = &self.lrf_switchable_cdf.len() - 1;
+    match filter {
+      RestorationFilter::None => {
+        w.symbol_bits(0, &self.lrf_switchable_cdf[..nsym])
+      }
+      RestorationFilter::Wiener { .. } => {
+        unreachable!() // for now, not permanently
+      }
+      RestorationFilter::Sgrproj { set, xqd } => {
+        // Does *not* use 'RESTORE_SGRPROJ' but rather just '2'
+        let rp = &rs.planes[pli];
+        let mut bits = w.symbol_bits(2, &self.lrf_switchable_cdf[..nsym])
+          + ((SGRPROJ_PARAMS_BITS as u32) << OD_BITRES);
+        for i in 0..2 {
+          let s = SGRPROJ_PARAMS_S[set as usize][i];
+          let min = SGRPROJ_XQD_MIN[i] as i32;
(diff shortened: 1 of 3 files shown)


src/context/frame_header.rs
```diff
@@ -171,42 +210,6 @@ impl<'a> ContextWriter<'a> {
       }
     }
   }
-  // rather than test writing and rolling back the cdf, we just count Q8 bits using the current cdf
-  pub fn count_lrf_switchable(
-    &self, w: &dyn Writer, rs: &TileRestorationState,
-    filter: RestorationFilter, pli: usize,
-  ) -> u32 {
-    let nsym = &self.fc.lrf_switchable_cdf.len() - 1;
```
lu-zero, line 179: what about keeping it and call `self.fc.count_lrf_switchable` from there?

KyleSiefring: Do you mean so that we don't have to write cw.fc.count_lrf_switchable when not constrained by the borrow checker?

lu-zero:
yes, what is it telling you?

src/context/frame_header.rs
```diff
@@ -171,41 +210,12 @@ impl<'a> ContextWriter<'a> {
...
-              max + 1,
-              SGRPROJ_PRJ_SUBEXP_K,
-              rp.sgrproj_ref[i] as i32,
-            );
-          }
-        }
-        bits
-      }
-    }
+    self.fc.count_lrf_switchable(w, rs, filter, pli);
```
lu-zero, line 218:
```suggestion
    self.fc.count_lrf_switchable(w, rs, filter, pli)
```

src/context/frame_header.rs
lu-zero: yes :)

src/rdo.rs
```diff
@@ -2220,7 +2220,7 @@ pub fn rdo_loop_decision<T: Pixel>(
                       pli,
                     );
                     rate += if fi.sequence.enable_restoration {
-                      cw.count_lrf_switchable(
+                      cw.fc.count_lrf_switchable(
```
lu-zero, line 2223: so this shouldn't be needed anymore.

KyleSiefring: ???
```
error[E0502]: cannot borrow `*cw` as immutable because it is also borrowed as mutable
    --> src/rdo.rs:2223:23
     |
2041 |   let mut tileblocks_subset = cw.bc.blocks.subregion_mut(
     |                               ------------ mutable borrow occurs here
...
2223 |                       cw.count_lrf_switchable(
     |                       ^^ immutable borrow occurs here
...
2317 |             tileblocks_subset.set_cdef(loop_sbo, best_new_index as u8);
     |             ----------------- mutable borrow later used here
```

lu-zero:
then it is fine

src/rdo.rs
lu-zero: It can wait later. let's land it.

PR #2415, opened 2020-06-19
Update libaom

Luni-4:
Libraries are installed correctly, but tests fail.


lu-zero:
You must update the aom-sys crate to 2.0.

lu-zero:
https://github.com/xiph/rav1e/pull/2415 should go in first.

Luni-4:
> #2415 should go in first.

Ouch, I forgot to add that! Yep, it should go in first

Luni-4:
@lu-zero 

It doesn't work the same

Luni-4:
Close in favor of #2323

PR #2419, opened 2020-06-20
Jsapi: Implement missing EncoderConfig methods

Urhengulas:
## rav1e_js
* add `EncoderConfig.setColorDescription(...)`
* add `EncoderConfig.setMasteringDisplay(...)`
* `encoder`
    * Pass `EncoderConfig` as ref to constructor
      --> This enables still using the config afterwards
    * Rename [FrameEncoder, VideoEncoder]::fromEncoderConfig(...) to ::new(...)
    * Rm `::default()` from `FrameEncoder`, `VideoEncoder`
* `mod`: `pub use encoder::VideoEncoder`
* `tests`: Clean `utils::simple_encoding`
* bump version to `0.2.1`

### DIFF
--- a/rav1e_js/src/encoder_config.rs
+++ b/rav1e_js/src/encoder_config.rs
@@ -107,15 +107,27 @@ impl EncoderConfig {
   //   self.clone()
   // }
 
-  // /// High dynamic range mastering display color volume
-  // ///
-  // /// As defined by CIE 1931
-  // pub fn setMasteringDisplay(
-  //   &mut self, mastering_display: MasteringDisplay,
-  // ) -> Self {
-  //   self.conf.mastering_display = Some(mastering_display);
-  //   self.clone()
-  // }
+  /// High dynamic range mastering display color volume
+  ///
+  /// As defined by CIE 1931
+  ///
+  /// * primaries (red, green, blue): `ChromaticityPoint`
+  /// * `white_point`: `ChromaticityPoint`,
+  /// * `max_luminance`: 24.8 fixed-point maximum luminance in candelas per square meter
+  /// * `min_luminance`: 18.14 fixed-point minimum luminance in candelas per square meter
+  pub fn setMasteringDisplay(
+    &mut self, primaries_red: ChromaticityPoint,
+    primaries_green: ChromaticityPoint, primaries_blue: ChromaticityPoint,
+    white_point: ChromaticityPoint, max_luminance: u32, min_luminance: u32,
+  ) -> Self {
+    self.conf.mastering_display = Some(MasteringDisplay {
+      primaries: [primaries_red, primaries_green, primaries_blue],
+      white_point,
+      max_luminance,
+      min_luminance,
+    });
+    self.clone()
+  }
 
   /// HDR content light parameters.
   ///
--- a/rav1e_js/src/encoder_config.rs
+++ b/rav1e_js/src/encoder_config.rs
@@ -99,13 +99,126 @@ impl EncoderConfig {
     self.clone()
   }
 
-  // /// Content color description (primaries, transfer characteristics, matrix).
-  // pub fn setColorDescription(
-  //   &mut self, color_description: ColorDescription,
-  // ) -> Self {
-  //   self.conf.color_description = Some(color_description);
-  //   self.clone()
-  // }
+  /// Content color description (primaries, transfer characteristics, matrix).
+  ///
+  /// Please supply following enums as input:
+  /// * `color_primaries`: `ColorPrimaries`
+  /// * `transfer_characteristics`: `TransferCharacteristics`
+  /// * `matrix_coefficients`: `MatrixCoefficients`
+  pub fn setColorDescription(
+    &mut self, color_primaries: ColorPrimaries,
+    transfer_characteristics: TransferCharacteristics,
+    matrix_coefficients: MatrixCoefficients,
+  ) -> Self {
+    // these huge match statements are a workaround, they can be removed as
+    // soon https://github.com/rustwasm/wasm-bindgen/issues/2273 gets solved
+    let color_primaries = match color_primaries {
+      ColorPrimaries::BT709 => rav1e::color::ColorPrimaries::BT2020,
+      ColorPrimaries::Unspecified => rav1e::color::ColorPrimaries::Unspecified,
+      ColorPrimaries::BT470M => rav1e::color::ColorPrimaries::BT470M,
+      ColorPrimaries::BT470BG => rav1e::color::ColorPrimaries::BT470BG,
+      ColorPrimaries::BT601 => rav1e::color::ColorPrimaries::BT601,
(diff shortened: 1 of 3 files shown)


src/api/color.rs
```diff
@@ -41,33 +41,34 @@ pub use v_frame::pixel::ChromaSampling;
...
+  Unspecified = 0,
   /// BT.709
   BT709 = 1,
-  /// Unspecified, must be signaled or inferred outside of the bitstream
-  Unspecified,
-  /// BT.470 System M (historical)
-  BT470M = 4,
   /// BT.470 System B, G (historical)
-  BT470BG,
+  BT470BG = 2,
```
vibhoothiiaanand, line 55: AFAIK these numbers are defined as per spec `ISO/IEC 23091-4/ITU-T H.273`
Here 2 is unspecified, 3 is reserved so you cannot keep numbers like that IIUC.

src/api/color.rs
```diff
@@ -81,45 +82,46 @@ impl Default for ColorPrimaries {
 /// Supported Transfer Characteristics
 ///
 /// As defined by “Transfer characteristics” section of ISO/IEC 23091-4/ITU-TH.273.
+#[wasm_bindgen]
 #[derive(
   ArgEnum, Debug, Clone, Copy, PartialEq, FromPrimitive, Serialize, Deserialize,
 )]
 #[repr(C)]
 pub enum TransferCharacteristics {
+  /// Unspecified, must be signaled or inferred outside of the bitstream
```
vibhoothiiaanand, line 91: Same goes for this also

src/api/color.rs
```diff
@@ -139,31 +142,31 @@ pub enum MatrixCoefficients {
   /// Identity matrix
   Identity = 0,
   /// BT.709
-  BT709,
+  BT709 = 1,
   /// Unspecified, must be signaled or inferred outside of the bitstream.
-  Unspecified,
+  Unspecified = 2,
```
vibhoothiiaanand, line 147: Here again

src/api/color.rs
Urhengulas: Okay. That is basically what my question is about.

The problem is, that I either need to specify no or all options in the enum with a number.
I will have a look into the spec.

src/api/color.rs
```diff
@@ -41,33 +41,34 @@ pub use v_frame::pixel::ChromaSampling;
 /// Supported Color Primaries
 ///
 /// As defined by “Color primaries” section of ISO/IEC 23091-4/ITU-T H.273
+#[wasm_bindgen]
```
lu-zero, line 44: wasm_bindgen should not require this, can you poke upstream and ask them to fix that in the macro?

Urhengulas:
Good idea

src/api/color.rs
Urhengulas: Will do

Urhengulas: https://github.com/rustwasm/wasm-bindgen/issues/2273

lu-zero:
The version should be `0.3` since you are breaking the API by removing `fromEncoderConfig` and `default`.

rav1e_js/src/encoder_config.rs
```diff
@@ -270,3 +395,109 @@ impl EncoderConfig {
     self.clone()
   }
 }
+
+/// Supported Color Primaries
+///
+/// As defined by “Color primaries” section of ISO/IEC 23091-4/ITU-T H.273
+#[wasm_bindgen]
+pub enum ColorPrimaries {
```
vibhoothiiaanand, line 403: Not sure why are we re-defining here, 
We already have it in public API right, 
https://github.com/xiph/rav1e/blob/425c5ca9ccb08ff64c81f495e7fe109a891cae4f/src/api/color.rs#L48-L73

rav1e_js/src/encoder_config.rs
```diff
@@ -270,3 +395,109 @@ impl EncoderConfig {
...
+  SMPTE432,
+  /// EBU Tech. 3213-E
+  EBU3213,
+}
+
+/// Supported Transfer Characteristics
+///
+/// As defined by “Transfer characteristics” section of ISO/IEC 23091-4/ITU-TH.273.
+#[wasm_bindgen]
+pub enum TransferCharacteristics {
```
vibhoothiiaanand, line 434: Same comment as previous

rav1e_js/src/encoder_config.rs
```diff
@@ -270,3 +395,109 @@ impl EncoderConfig {
...
+  /// SMPTE ST 428
+  SMPTE428,
+  /// BT.2100 HLG (Hybrid Log Gamma), ARIB STD-B67
+  HLG,
+}
+/// Matrix coefficients
+///
+/// As defined by the “Matrix coefficients” section of ISO/IEC 23091-4/ITU-TH.273.
+#[wasm_bindgen]
+pub enum MatrixCoefficients {
```
vibhoothiiaanand, line 474: Ditto as previous

rav1e_js/src/encoder_config.rs
vibhoothiiaanand: https://docs.rs/rav1e/0.3.3/rav1e/color/enum.ColorPrimaries.html

Urhengulas: It is not possible to export these enums. See https://github.com/rustwasm/wasm-bindgen/issues/2273 for the problem.

Therefore I'm having a copy of them without annotated discriminants as a workaround until it is possible to export them.

src/api/color.rs
```diff
@@ -226,6 +227,16 @@ pub struct ChromaticityPoint {
   pub y: u16,
 }
 
+#[wasm_bindgen]
+impl ChromaticityPoint {
+  /// Chromaticity coordinates as defined by CIE 1931, expressed as 0.16
+  /// fixed-point values.
+  #[wasm_bindgen(constructor)]
+  pub fn constructor(x: u16, y: u16) -> Self {
```
lu-zero, line 235: please add a `#[cfg()]` to make so it is not added to the non-js API, why is that needed btw?

rav1e_js/src/encoder_config.rs
lu-zero: Please add a note about it.

src/api/color.rs
Urhengulas: So the js user can add a new `ChromaticityPoint` via:
```ts
let a = new ChromaticityPoint(0, 0);
```

Urhengulas:
Alex Crichton confirmed that there is currently no better way than our current workaround for the enum-problem (https://github.com/rustwasm/wasm-bindgen/issues/2273#issuecomment-685965892).
So I would propose to actually go with it for now. Even if it is kinda not beautiful...

barrbrain:
If this is still relevant, the PR should be ported to https://github.com/rust-av/rav1e_js

PR #2423, opened 2020-06-29
rav1e_js: New demo website with react

Urhengulas:
This PR rebuilds the example `rav1e_js` example website with `react.js` and makes it a bit more visual.


lu-zero:
image

Might be nicer to format the config and maybe add a flush button. With some css it would be a great demo :)

Urhengulas:
I actually like plain HTML for the demo purpose, but I will have a look into this :D

lu-zero:
The `receive packet` button is grayed out. It should not.

Urhengulas:
> The `receive packet` button is grayed out. It should not.

It should only be grayed out when there are no frames yet to be encoded.

lu-zero:
it does not work this way right now :)

rav1e_js/www/src/App.tsx
```diff
@@ -0,0 +1,118 @@
...
+			console.log(p.display());
+			setEncoderStatus("");
+
+			// append new packet
+			const newPackets = Array.from(packets);
+			newPackets.push(p);
+			setPackets(newPackets);
+		} catch (e) {
+			if (e === "encoded") {
+				receivePacket();
```
vibhoothiiaanand, line 54: I think we can show the encoder status on receiving a packet so it will be clear, right now when we are receiving frames properly, it is blank and only when an error is coming, we are showing there something, 
I was thinking to have something like Encoding or some text there
For example
`Encoder Status: %x frames encoded`
Which would be pretty needed instead of having blank there,
When we have error it is clear but on postive cases, it is not very clear why that text exists there.
The frame details are being shown in the bottom, so for better UX and easier understanding of people, I think we can have something here and handle things in nicer way.

rav1e_js/www/src/App.tsx
```diff
@@ -0,0 +1,118 @@
...
+			} else if (e === "need more data") {
```
vibhoothiiaanand, line 55: Also, I am not quite sure why we are checking error like this,
can we use the standard way like `RA_ENCODER_STATUS_NEED_MORE_DATA` for here and all the other places?

rav1e_js/www/src/App.tsx
```diff
@@ -0,0 +1,118 @@
...
+				setEncoderStatus(e + " (send more frames or flush the encoder)");
+			}
+		}
```
vibhoothiiaanand, line 58: Once we finish the encoding the encoder status should show completed encoding/limit reached based on `EncoderStatus::LimitReached` status.
Right now, the end-user does not know whether the encoder status have reached there or not, it is more or less a guesstimate now

rav1e_js/www/src/App.tsx
```diff
@@ -0,0 +1,118 @@
...
+							<td><strong>Details</strong></td>
+						</tr>
+						<tr>
+							<td>Send Frame</td>
+							<td>Send a new (blank) frame to the encoder.</td>
+							<td></td>
+						</tr>
+						<tr>
+							<td>Flush</td>
+							<td>Close the encoder.</td>
```
vibhoothiiaanand, line 95:
```suggestion
							<td>Siginals the End of the Video.</td>
```
As per encoder.rs 
` Flushing signals the end of the video. After the encoder has been flushed, no additional frames are accepted.`
I am not quite sure if Close the encoder is the right word to use here.

rav1e_js/www/src/App.tsx
Urhengulas: What do you mean with that?

Urhengulas: To elaborate: We are simply throwing the string value of the `EncoderStatus` to js. And the string value for e.g. `EncoderStatus::NeedMoreData` is `"need more data"`. So we are catching this on the js-side and handle it.

rav1e_js/www/src/App.tsx
Urhengulas: Added more verbose displaying of EncoderStatus

Urhengulas:
Squashed all the commits into one :)

PR #2436, opened 2020-07-07
Unbreak reservoir-frame-delay

### LINKED ISSUE #2433: --reservoir-frame-delay triggers another rc failure
In order to trigger that 92 frames of Bosphorus are enough. 

The RC does not move to ready even if `rc_second_pass_data_required()` packets are passed to the RC in some corner cases.

lu-zero:
Do not assume the value returned by rc_second_pass_data_required is reliable

Receiving non-show_frame pass data may not reduce the number of packets
required.

Also set the current metrics correctly.

Fixes #2433

### DIFF
--- a/src/api/context.rs
+++ b/src/api/context.rs
@@ -390,7 +390,10 @@ impl<T: Pixel> Context<T> {
     }
   }
 
-  /// Number of pass data packets required to progress the encoding process.
+  /// Lower bound number of pass data packets required to progress the
+  /// encoding process.
+  ///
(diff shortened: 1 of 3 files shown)


lu-zero:
```
00:06 <+derf> lu_zero: Sorry, you failed my "this is simple enough to review the diff" threshold, and I had to start actually reading the current state of the code, which is
              starting to look nothing like the code I originally wrote.
00:06 <+derf> So it took longer.
00:06 <+derf> But r+.
```

PR #2443, opened 2020-07-13
CI: Update nasm

Luni-4:


Luni-4:
> Seems to be fine as CI is going green :)
> 
> Thanks for the patch

Thanks to you for your review! :)

PR #2444, opened 2020-07-13
Add x86 simd for weighted sse computation

KyleSiefring:


src/x86/sse.asm
```diff
@@ -0,0 +1,510 @@
...
+    movq               rax, m1
+    RET
+
+INIT_XMM ssse3
+cglobal weighted_sse_4x16, 6, 7, 7, \
+        src, src_stride, dst, dst_stride, scale, scale_stride, \
+        h
+    mova                m0, [addsub]
+    pxor                m6, m6
+    mov                 hd, 1
```
cyanreg, line 207: Just get rid of the loop here. You save a register and a (maybe well) predicted branch.
You can also just macro what's inside the loop and call it twice, would be the same number of lines.

src/x86/sse.asm
```diff
@@ -0,0 +1,510 @@
...
+
+%macro WEIGHTED_SSE 2 ; w, h
+%if %1 == 8
+%if %2 == 4
+; Use scale_stride's register to store src_stride3
+cglobal weighted_sse_%1x%2, 6, 7, 5, \
+        src, src_stride, dst, dst_stride, scale, \
+        src_stride3, dst_stride3
+%else
+cglobal weighted_sse_%1x%2, 6, 9, 6, \
```
cyanreg, line 335: You can have C-like expressions when declaring the number of registers.
e.g. cglobal weighted_sse_%1x%2, 6, 9 + 2*(%1 == 32 ? 9 >> 1 : 2), 6 + 4*(%2 == 32), and so on.
So you can clean up the ifdeffery somewhat.

src/x86/sse.asm
```diff
@@ -0,0 +1,510 @@
...
+; This source code is subject to the terms of the BSD 2 Clause License and
+; the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+; was not distributed with this source code in the LICENSE file, you can
+; obtain it at www.aomedia.org/license/software. If the Alliance for Open
+; Media Patent License 1.0 was not distributed with this source code in the
+; PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+%include "ext/x86/x86inc.asm"
+
+%if ARCH_X86_64
```
cyanreg, line 12: Don't think this is needed, most of the functions use less than 8 registers.
You can have a separate WEIGHTED_SSE X, Y if block for those that go above 8.
But no strong opinion on this if it makes the rust code more complicated.

src/x86/sse.asm
```diff
@@ -0,0 +1,510 @@
...
+SECTION .text
+
+%define m(x) mangle(private_prefix %+ _ %+ x %+ SUFFIX)
+
+; Consolidate scaling and rounding to one place so that it is easier to change.
+
+%macro SSE_SCALE_4X4 0
+    ; Multiply and shift using scalar code
+    mov             scaled, [scaleq]
+    imul               rax, scaleq
```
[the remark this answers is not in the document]
KyleSiefring: scale is stored as 32 bits so this would load two elements.

src/x86/sse.asm
KyleSiefring: I don't think much assembly is hooked up for x86-32, but I'll gate that won't work on x86 for when we get around to it.

src/x86/sse.asm
KyleSiefring: Done.

src/x86/sse.asm
cyanreg: No, this should work on yasm and ancient nasm versions as well. Its used pretty extensively in ffmpeg.
Regardless, we don't support yasm or any old versions of nasm.

src/x86/sse.asm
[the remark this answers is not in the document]
cyanreg: `imul eax, dword [scaleq]` will work then.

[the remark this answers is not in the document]
KyleSiefring: Is there a way to get this to work with 64 bit multiplies?

btw, I'm probably going to change scale to 16 bits at some point anyways.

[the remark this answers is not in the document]
cyanreg: Operands must match, so unfortunately no. Still, 32 bits is a lot of range.

PR #2396, opened 2020-06-15
Enable open partition on frame boundary

ycho:
Issue #2166

- AV1 spec allows to use a partition that straddle on frame boundary if
  certain condition met, i.e, 'Not Split' is allowed if more than half of
  the partition size is available (based on coded frame size, i.e. multipe
  of 8 pixels) in either horizontal or vertical.

- For top-down partition search (i.e current speed levels 2~10),
  use bottomup on frame boundary

- Functon for computing visible area of a partition or a tx block based
  on actual frame size (not a coded size) is added and in general it should
  be only used for a compute distortion. Hence, do not use it for adjusting the block size.
  One good critical example would be that a coefficient block size, when it is encoded, is not affected by visibility.

- CFL specially requires to read outside the coded frame, when it compute average
  and obtain ac components for luma block, i.e. luma_ac().

PR #2445, opened 2020-07-13
Ground work to allow rav1e to access PlaneRegion memory that are outside of coded frame size

ycho:
The goal of this work is to establish the ground so that rav1e can encode blocks that straddle on bottom or right frame boundaries (i.e. #2396).
I thought it would better approach to first land this groundwork then land the open-partition feature,  #2396.
Because, that way we can ensure the groundwork itself does not affect the integrity of current master without any open-partition code mixed.
With this PR only, it is expected to not change coding behavior of current master.

Until it is finished, it will fail many of rav1e's internal cargol tests, mostly when comparing encoder and decoder outputs, not because the enc and dec's outputs really unmatched but more likely the dimension of plane when it is declared is not correctly changed.
When compare the two reconstructed outputs from rav1e and decoder that run on command line and not via cargo test , I see that they are matched.
Also, at the moment I create this PR, I don't observe rust panic caused by memory access violation (i.e. accessing outside the bounding box defined by PlaneRegion (+ subregion) ).

### DIFF
--- a/doc/TILE_ON_FRAME_BOUNDARY.md
+++ b/doc/TILE_ON_FRAME_BOUNDARY.md
@@ -1,0 +1,36 @@
+# Tile on a Frame boundary
+
+## Accessing Tiled-pixels outside of a frame
+
+Tiled input and output planes (type PlaneRegion) are configured in order to
+allow memory access of its pixels that are located outside of frame boundary.
+This configuration is only effective when the tile is adjacent to right or bottom frame boundary and
+a partition block straddles on any frame boundary.
+
+The type names of tiled input and output in the codebase is defined in:
+
+```
+pub struct TileStateMut<'a, T: Pixel> {
+  ...
+  pub input: &'a Frame<T>,     // the whole frame
+  pub rec: TileMut<'a, T>,
+  ...
+}
+```
+
+## How bounding box is enforced to a PlaneRegion and now it allows to access outside of frame pixels
+
+A memory access to tiled input and output pixels in a _tile_ 
+is bounded by 'rect' field of each plane (of type `PlaneRegion`) of the `input_tile` or `rec` of `TileStateMut`.
+
+The bounding box dimension of a PlaneRegion.rect, i.e. PlaneRegion.rect.width and .height is ROUNDED UP to SuperBlock size,
+which now allows accessing the pixels outside the right or bottom frame boundary and belongs to a partition that straddle on any frame boundary.
+
+Previously, the bounding box has prohibited accessing those pixels. 
+
+## When does rav1e access outside frame tiled-input and output pixels?
+
+A CfL requires to read outside the coded frame (luma only), when it computes average
+and obtain ac components for luma block, i.e. `luma_ac()`.
+Hence, the prediction of a luma plane should be performed for all the pixels in a block
+(more precisely, all required tx-blocks in a partition), whether the predicted pixel is inside or outside the frame.
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -607,6 +607,11 @@ pub fn get_intra_edges<T: Pixel>(
         enable_intra_edge_filter && p_angle > 90 && p_angle < 180;
     }
 
+    let rect_w =
+      dst.rect().width.min(dst.plane_cfg.width - dst.rect().x as usize);
+    let rect_h =
+      dst.rect().height.min(dst.plane_cfg.height - dst.rect().y as usize);
+
     // Needs left
     if needs_left {
       if x != 0 {
@@ -670,8 +675,8 @@ pub fn get_intra_edges<T: Pixel>(
         partition_bo.0.x > 0
       };
 
-    let right_available = x + tx_size.width() < dst.rect().width;
-    let bottom_available = y + tx_size.height() < dst.rect().height;
+    let right_available = x + tx_size.width() < rect_w;
+    let bottom_available = y + tx_size.height() < rect_h;
 
     let scaled_partition_size =
       supersample_chroma_bsize(partition_size, plane_cfg.xdec, plane_cfg.ydec);
@@ -692,7 +697,7 @@ pub fn get_intra_edges<T: Pixel>(
           plane_cfg.xdec,
           plane_cfg.ydec,
         ) {
-        tx_size.width().min(dst.rect().width - x - tx_size.width())
+        tx_size.width().min(rect_w - x - tx_size.width())
       } else {
         0
       };
@@ -728,7 +733,7 @@ pub fn get_intra_edges<T: Pixel>(
           plane_cfg.xdec,
           plane_cfg.ydec,
         ) {
-        tx_size.height().min(dst.rect().height - y - tx_size.height())
+        tx_size.height().min(rect_h - y - tx_size.height())
       } else {
         0
       };
--- a/src/tiling/tile_state.rs
+++ b/src/tiling/tile_state.rs
@@ -136,11 +136,15 @@ impl<'a, T: Pixel> TileStateMut<'a, T> {
       height % MI_SIZE == 0,
       "Tile width must be a multiple of MI_SIZE"
     );
+
+    let sb_rounded_width = width.align_power_of_two(sb_size_log2);
+    let sb_rounded_height = height.align_power_of_two(sb_size_log2);
+
     let luma_rect = TileRect {
       x: sbo.0.x << sb_size_log2,
       y: sbo.0.y << sb_size_log2,
-      width,
-      height,
+      width: sb_rounded_width,
+      height: sb_rounded_height,
     };
     let sb_width = width.align_power_of_two_and_shift(sb_size_log2);
     let sb_height = height.align_power_of_two_and_shift(sb_size_log2);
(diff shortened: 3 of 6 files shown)


v_frame/src/frame.rs
```diff
@@ -27,8 +27,9 @@ impl<T: Pixel> Frame<T> {
     width: usize, height: usize, chroma_sampling: ChromaSampling,
     luma_padding: usize,
   ) -> Self {
-    let luma_width = width.align_power_of_two(3);
-    let luma_height = height.align_power_of_two(3);
+    //TODO: If SB size is 128x128, 7 should be used instead of 6
```
vibhoothiiaanand, line 30: Hi,

I was thinking whether we should keep separate commit for v_frame changes as it is a separate module and we do maintain releases for them.
If it is split as an extra commit for v_frame changes, we can make things easier,

But I am not sure if it will break the intermediate commits or not, if that is the case, we should be having this style only,

ycho:
One tip for me: To test one failed test,

env RUST_BACKTRACE=full RAV1E_CPU_TARGET=rust cargo test --features=decode_test_dav1d dimension_32x32_dav1d

ycho:
Hmm, after I see compare_pic( ) (used during test, comparing encoder and decoder output frames),  I start to think that the idea of rounding up input and output frame (i.e. their planes) size to SB might be bad.

ycho:
In Irc #daala,
xiphmont, Another thing I am seeing now that is surely related to work you did before/do later is the use of bounding box in cdef_filter_tile( ) via rect( ).width|height.

<codeview> This makes me look again what I think so far is reasonable. I.e. TileRect is rounded up to SB size (thus its input and output (i.e. rec) tiled planes (i.e. PlaneRegion) ) but ts.width|height is not rounded up and keep coded w and h of tile.)  With my approach, In cdef_filter_tile( ),  "rec_w = rec.planes[p].rect().width" cause w and h include outside of coded frame w and h, which is wrong with current code of the  function.

<codeview> So, I am basically needing more time on pondering which way is more correct, tiled plane following coded w and h as current master or extend it include the block crossing frame boundaries. The most basic clue to me is that, as with the case how CFL works with block straddle on frame boundaries, if a block being encoded for any prediction mode requires to read/write pixels outside coded frame, the area should be included  as a plane.

ycho:
YAY~~~~~~~!!!

v_frame/src/frame.rs
ycho: This change has been removed.

ycho:
There is no change in coding efficiency or encoding speed.
[awcy, -s 6](https://beta.arewecompressedyet.com/?job=master_c89bf8bd_s6_low_latency%402020-07-21T08%3A39%3A47.175Z&job=fix_access_outside_frame_s6_low_latency%402020-07-21T08%3A41%3A00.423Z)

ycho:
With my most recent commit(s), I have focused on correctly replace the existing bounding 'rectangle' under PlaneRegion (either from input or output tile planes) with TileState.input_tile|res.plane[ ].plane_cfg.width|height or TIleState.width|height if possible. In some cases, correct bounding rectangle requires to apply the origin of rect, i.e. rect.x|y, because the origin of rectangle can be off the (0,0), as in two cases I see, 1) via subregion(), 2) in cargo test, 'tile_encoding_with_stretched_restoration_units' .

ycho:
As soon as this PR has landed, I will rebase my open-partition PR https://github.com/xiph/rav1e/pull/2396 on it then ask reviewers, esp. @lu_zero, to review again.

src/tiling/tiler.rs
```diff
@@ -460,9 +460,9 @@ pub mod test {
     assert_eq!((32, 0, 32, 32), rect(&tile.planes[2]));
 
     let tile = &tile_states[2].rec; // the top-right tile
```
KyleSiefring, line 462: Change comments above.

ycho: Sure, which format of doc would you think best?  New *.md file under doc directory?

src/partition.rs
```diff
@@ -670,8 +670,13 @@ pub fn get_intra_edges<T: Pixel>(
         partition_bo.0.x > 0
       };
 
-    let right_available = x + tx_size.width() < dst.rect().width;
-    let bottom_available = y + tx_size.height() < dst.rect().height;
+    let rect_w =
```
ycho, line 673: The rect_w, rect_h are derived from coded tile size. I.e. exclude the out of frame region because dst.rect().width|height now include out of frame region to access pixels belong to a block on frame boundary.

ycho: > The rect_w, rect_h are derived from coded tile size. I.e. exclude the out of frame region because dst.rect().width|height now include out of frame region to access pixels belong to a block on frame boundary.

I might want to add member functions under rect of PlaneRegion, which returns these two values.

src/tiling/tiler.rs
KyleSiefring: These are the comments I meant. https://github.com/xiph/rav1e/pull/2445/files#diff-61a9dba6df92b3be4f74e67feaa6cd10R448

I think a new md file would be great.

src/tiling/tile_state.rs
```diff
@@ -126,24 +126,24 @@ impl IndexMut<usize> for MiTileState {
 impl<'a, T: Pixel> TileStateMut<'a, T> {
   pub fn new(
     fs: &'a mut FrameState<T>, sbo: PlaneSuperBlockOffset,
-    sb_size_log2: usize, width: usize, height: usize,
+    sb_size_log2: usize, w: usize, h: usize,
```
barrbrain, line 129: It appears that reverting the rename of the `width` and `height` parameters will result in a smaller, more focused set of changes to this function.

ycho: Yes, that is surely right. I also wanted to change as small as possible but it seems the lines 142 and 143 prohibits it by a rust grammer. I did end up compile errors. I thought one.possible way to keep formal parameter names width and height here would be change the field names width,height of TileRect struct.

vibhoothiiaanand: We could just keep as
Width.... instead of explicitly assigning right? 
I think if we are passing all args we could do like that 
If so it might work, not so sure

ycho: By introducing two new variables to keep width and height that are rounded up to SB, I think I can keep existing width and height. Let me try..

src/tiling/tile_state.rs
```diff
@@ -136,11 +136,15 @@ impl<'a, T: Pixel> TileStateMut<'a, T> {
       height % MI_SIZE == 0,
       "Tile width must be a multiple of MI_SIZE"
     );
+
+    let sb_rounded_width = width.align_power_of_two(sb_size_log2);
```
ycho, line 140: Yeap,  how dumb I am I have not done this way from the start and just use them to change two lines below!

src/tiling/tiler.rs
ycho: I've pushed document for this PR, under name "doc/TILE_ON_FRAME_BOUNDARY.md"
Please let me know if you'd like more revisions or it looks fine for landing.

ycho:
@xiphmont, @lu_zero, Your opinions are greatly appreciately as well!

xiphmont:
If I'm reading the code correctly, this has no effect whatsoever except at the right and bottom visible frame boundaries.  Tile access still does not overlap.

So... Not the way I'm chasing it in my Rectangle Reform patches, but this is simple, tiny, and works.  I see no reason not to land it ASAP.

r+me

doc/TILE_ON_FRAME_BOUNDARY.md
```diff
@@ -0,0 +1,35 @@
...
+## Accessing Tiled-pixels outside of a frame
+
+Tiled input and output planes (type PlaneRegion) are configured in order to
+allow memory access of its pixels that are located outside of frame boundary.
+This configuration is only effective when the tile is adjacent to right or bottom frame boundary and
+a partition block straddles on any frame boundary.
+
+The type names of tiled input and output in the codebase is defined in:
+
+"'TileStateMut' struct"
```
vibhoothiiaanand, line 12: Nit: Can we use code and code blocks for this part so it looks nicer.

doc/TILE_ON_FRAME_BOUNDARY.md
```diff
@@ -0,0 +1,35 @@
...
+"  ..."
+"  pub input: &'a Frame<T>,     // the whole frame"
+"  pub rec: TileMut<'a, T>,"
+"  ..."
+"}"
+
+## How bounding box is enforced to a PlaneRegion and now it allows to access outside of frame pixels
+
+A memory access to tiled input and output pixels in a _tile_ 
+is bounded by 'rect' field of each plane (of type 'PlaneRegion') of the 'input_tile' or 'rec' of 'TileStateMut'.
```
vibhoothiiaanand, line 23:
```suggestion
is bounded by 'rect' field of each plane (of type `PlaneRegion`) of the `input_tile` or `rec` of `TileStateMut`.
```

So it will be better when we render the markdown :)

doc/TILE_ON_FRAME_BOUNDARY.md
ycho: Sur- Is ``` for code block?

doc/TILE_ON_FRAME_BOUNDARY.md
ycho: Done. Thanks!

src/tiling/tiler.rs
lu-zero: @KyleSiefring is that enough for you?

PR #2450, opened 2020-07-15
draw-importance: Rewrite CLI with batch mode

vibhoothiiaanand:
This commit refactors the draw-importance tool with multiple features like
+ Use click module for CLI parsing
+ Add --help for getting help
+ Add --verbose mode to get detailed output
+ Add --raw mode to get raw values
+ Add --path mode for batch processing of multiple files in a nice way
+ Uses regex pattern to matx the strings
The default mode where $program $hres.png $imps.bin is now replaced with the --input mode.

Screenshots:

- Help Screen
`./draw-importance.py --help`:
Screenshot 2020-07-15 at 5 05 25 PM

- Single File Mode
`tools/draw-importances.py --input tools/19-hres.png tools/19-imps.bin`
Screenshot 2020-07-15 at 7 50 53 PM

- Batch Mode

`tools/draw-importances.py --path /Volumes/buildBox/rav1e/rav1e/tools/ --verbose`
Screenshot 2020-07-15 at 7 52 42 PM

### DIFF
--- a/tools/draw-importances.py
+++ b/tools/draw-importances.py
@@ -1,55 +1,187 @@
 #!/usr/bin/env python3
 import struct
 import sys
+import os
 from os.path import splitext
+import datetime
 import numpy as np
 from PIL import Image, ImageDraw
 from matplotlib import pyplot as plt
+import click
+import glob
+import re
+
+@click.command()
+@click.option('--path',  type = str,
+        help = "Absolute File locations of extracted hres.png and imps.bin ")
+@click.option('--input', help="Pass both hres.png and imps.bin",
+                default=[None] * 2, type=click.Tuple([str, str]))
+
+@click.option('--verbose', is_flag=True, help="Will print verbose messages.")
+@click.option('--raw', is_flag=True, help="Print RAW Data of the bin.")
 
 
-# Renders block importances output by ContextInner::compute_lookahead_data().
-# Usage:
-#        cargo run --features=dump_lookahead_data <input.y4m> -o /dev/null
-#        draw-importances.py <i-hres.png> <i-imps.bin>
-#        will output i-imps.png.
+def blockImportance(input, verbose, path, raw):
+    """
+    CLI tool for extracting rav1e's Block Importance
+
+    Renders block importances output by
+    ContextInner::compute_lookahead_motion_vectors()
+    ContextInner::compute_block_importances().
+
+    Run rav1e binary with dump_lookahead_data feature flag to dump:\t\t
+    For eg: `cargo run --features=dump_lookahead_data <input.y4m> -o /dev/null`
+
+    Two modes available:\n
+    Mode 1: Simple file, passing one hres.png, and one imps.bin \t\t
+    Mode 2: Batch file processing for multiple hres and imps data\n
+    Output files will be saved in out/YYYYMMDD_HH folder in -imps.png files.
+
+    """
+
+    if input == (None, None) and path == None :
+        raise click.BadParameter("Please use either --input or --path method, for more information use --help.")
+
+    if verbose:
+        click.echo("Input given: "+ str(input))
+        click.echo("Path given: " + str(path))
+
+    if input != (None, None):
+        if verbose:
+            click.secho("MODE 1", fg="yellow")
+        with open(input[1], 'rb') as f:
+            contents = f.read()
+
+        rows, cols = struct.unpack('qq', contents[:16])
+        imps = np.frombuffer(contents[16:],
+                            dtype=np.float32).reshape((rows, cols))
+
+        if raw:
+            click.secho("imps data after processing:", fg="red")
+            click.echo(imps)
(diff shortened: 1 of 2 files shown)


tools/draw-importances.py
```diff
@@ -1,55 +1,180 @@
 #!/usr/bin/env python3
 import struct
 import sys
+import os
 from os.path import splitext
+import datetime
 import numpy as np
 from PIL import Image, ImageDraw
 from matplotlib import pyplot as plt
+import click
```
YaLTeR, line 10: Is this module bundled with default Python 3 installs?

tools/draw-importances.py
```diff
@@ -1,55 +1,180 @@
...
 
-with open(sys.argv[2], 'rb') as f:
-    contents = f.read()
+def blockImportance(input, verbose, path, raw):
+    """
+    CLI tool for extracting rav1e's Block Importance
 
-rows, cols = struct.unpack('qq', contents[:16])
-imps = np.frombuffer(contents[16:], dtype=np.float32).reshape((rows, cols))
+    Renders block importances output by ContextInner::compute_lookahead_data().
```
YaLTeR, line 28:
```suggestion
    Renders block importances output by ContextInner::compute_lookahead_motion_vectors() ContextInner::compute_block_importances().
```

tools/draw-importances.py
```diff
@@ -1,55 +1,180 @@
...
-frame = frame.resize((frame.width * frame_size_multiplier, frame.height * frame_size_multiplier))
-frame = frame.convert(mode='RGB')
+    """
+    if verbose:
+        click.echo("Input given: "+ str(input))
+        click.echo("Path given: " + str(path))
 
-mv_original_block_size = 8 // 2 # The importances are in 8×8 blocks, but we use half-resolution images.
-mv_block_size = mv_original_block_size * frame_size_multiplier
+    rx = re.compile(r'(\d\d-imps)|(\d-imps)')
```
YaLTeR, line 43: I'm not sure this is the best way of doing this (besides, this works only for 1 and 2 digits). Could you simply use something like `basename()` and `splitext()` to get the *-imps part of the name?

tools/draw-importances.py
```diff
@@ -1,55 +1,180 @@
...
-            draw.rectangle((top_left, bottom_right), fill=(int(imp / max_imp * 255), 0, 0, 128))
+        if raw:
+            click.secho("imps data after processing:", fg="red")
+            click.echo(imps)
+
+    elif path != None:
+        if verbose:
+            click.secho("MODE 2", fg="yellow")
+        png_list = glob.glob(str(path)+"/*hres.png")
+        bin_list = glob.glob(str(path)+"/*imps.bin")
```
YaLTeR, line 63: To prevent errors I think it's best to get only one of these lists from a glob, and then string replace `imps` with `hres` or vice versa to get the second list. This way you will get an error upfront if you have a different amount of hres and imps files or if some of them are missing (rather than it silently using wrong files to produce wrong results).

tools/draw-importances.py
```diff
@@ -1,55 +1,180 @@
...
+    if raw:
+        click.secho("List of Figure Elements: "+str(fig_list))
+
+    if verbose:
+        if path == None:
+            click.secho("File name is "+ splitext(input[1])[0]+ ".png",
+                                                            fg="blue")
+        print("Folder Path: "+ os.getcwd()+"/"+ folder_path)
+
+    click.secho("File is sucessfully saved.", fg="green")
```
YaLTeR, line 176: Maybe "Processing done."? Since it can process multiple files.

tools/draw-importances.py
vibhoothiiaanand: Nop this is not bundled by default in python3, 
We require to install, either we can add a README.md in the [src/tools](https://github.com/xiph/rav1e/tree/master/tools), explaining the scripts in a short and simple way, 
or we can leave it as it is, since it is mostly used for internal evaluation, 
EIther way I am fine :)

tools/draw-importances.py
vibhoothiiaanand: Sure :)

tools/draw-importances.py
YaLTeR: Do we use click anywhere else in the repo? If not maybe it'd be better to stick to argparse?

YaLTeR: Actually since we already use a few packages like numpy and matplotlib it should be fine. Took me a bit to realize, these have just become so standard to me that I don't even think about them. 😅

tools/draw-importances.py
vibhoothiiaanand: Yes makes sense

tools/draw-importances.py
vibhoothiiaanand: indeed, makes sense and lot more easier

PR #2451, opened 2020-07-15
Channel API

lu-zero:
This set provides a channel-based API that should make much simpler the normal usage.

TODO:
- [x] Basic encoding API
  - [x] Works
  - [x] Produces the same output for the same input
- [x] Multipass encoding API
   - [x] Single pass
   - [x] Second pass
     - [x] Produces the same output for the same input
   - [x] Multipass

Open questions:
- Fit a `last-error` inside the ReceiverPacket abstraction and a mean to retrieve it
- For multipass the api can be a single function and use the Config rc information to decide if it has to be multipass/firstpass/secondpass/normal

Solves: #689

(apparently github got some consistency issue with #2118 so I ended up opening this)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,6 +73,7 @@ ivf = { version = "0.1", path = "ivf/", optional = true }
 v_frame = { version = "0.1", path = "v_frame/" }
 av-metrics = { version = "0.5.1", optional = true }
 rayon = "1.0"
+crossbeam = "0.7"
 toml = { version = "0.5", optional = true }
 arrayvec = { version = "0.5", features = ["array-sizes-33-128"] }
 thiserror = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ readme = "README.md"
 repository = "https://github.com/xiph/rav1e/"
 autobenches = false
 autobins = false
+default-run = "rav1e"
 
 [features]
 unstable = []
@@ -129,6 +130,11 @@ rand_chacha = "0.2"
 
 [[bin]]
 name = "rav1e"
+required-features = ["binaries"]
+bench = false
+
+[[bin]]
+name = "rav1e-ch"
 required-features = ["binaries"]
 bench = false
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ default-run = "rav1e"
 
 [features]
 unstable = []
+channel-api = ["crossbeam"]
 decode_test = ["aom-sys"]
 decode_test_dav1d = ["dav1d-sys"]
 binaries = [
@@ -74,7 +75,7 @@ ivf = { version = "0.1", path = "ivf/", optional = true }
 v_frame = { version = "0.1", path = "v_frame/" }
 av-metrics = { version = "0.5.1", optional = true }
 rayon = "1.0"
-crossbeam = "0.7"
+crossbeam = { version = "0.7", optional = true }
 toml = { version = "0.5", optional = true }
 arrayvec = { version = "0.5", features = ["array-sizes-33-128"] }
 thiserror = "1.0"
@@ -135,7 +136,7 @@ bench = false
 
 [[bin]]
 name = "rav1e-ch"
-required-features = ["binaries"]
+required-features = ["binaries", "channel-api", "unstable"]
 bench = false
 
 [lib]
--- a/README.md
+++ b/README.md
@@ -100,8 +100,15 @@ cargo build --release
 Experimental API and Features can be enabled by using the `unstable` feature.
 
 ```sh
-cargo build --features unstable
+cargo build --features <feature>,unstable
 ```
+
+#### Current unstable features
+- Channel API:
+```sh
+cargo build --features channel-api,unstable
+```
+
 
 Those Features and API are bound to change and evolve, do not rely on them staying the same over releases.
 
--- a/src/api/channel.rs
+++ b/src/api/channel.rs
@@ -1,0 +1,517 @@
+// Copyright (c) 2018-2020, The rav1e contributors. All rights reserved
+//
+// This source code is subject to the terms of the BSD 2 Clause License and
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+#![allow(missing_docs)]
+
+use crate::api::color::*;
+use crate::api::config::*;
+use crate::api::context::RcData;
+use crate::api::internal::ContextInner;
+use crate::api::util::*;
+
+use bitstream_io::*;
+use crossbeam::channel::*;
+
+use crate::encoder::*;
+use crate::frame::*;
+use crate::rate::RCState;
+use crate::rayon::{ThreadPool, ThreadPoolBuilder};
+use crate::util::Pixel;
+
+use std::io;
+use std::sync::Arc;
+
+/// Endpoint to send previous-pass statistics data
+pub struct RcDataSender {
+  sender: Sender<RcData>,
+  limit: u64,
+  count: u64,
+}
+
+impl RcDataSender {
+  fn new(limit: u64, sender: Sender<RcData>) -> RcDataSender {
+    Self { sender, limit, count: 0 }
+  }
+  pub fn try_send(
+    &mut self, data: RcData,
+  ) -> Result<(), TrySendError<RcData>> {
+    if self.limit <= self.count {
+      Err(TrySendError::Full(data))
+    } else {
+      let r = self.sender.try_send(data);
+      if r.is_ok() {
+        self.count += 1;
+      }
+      r
+    }
+  }
+  pub fn send(&mut self, data: RcData) -> Result<(), SendError<RcData>> {
+    if self.limit <= self.count {
+      Err(SendError(data))
+    } else {
+      let r = self.sender.send(data);
+      if r.is_ok() {
+        self.count += 1;
+      }
+      r
+    }
+  }
+  pub fn len(&self) -> usize {
+    self.sender.len()
+  }
+  pub fn is_empty(&self) -> bool {
+    self.sender.is_empty()
+  }
+
+  // TODO: proxy more methods
+}
+
+/// Endpoint to receive current-pass statistics data
+pub struct RcDataReceiver(Receiver<RcData>);
+
+impl RcDataReceiver {
+  pub fn try_recv(&self) -> Result<RcData, TryRecvError> {
+    self.0.try_recv()
+  }
+  pub fn recv(&self) -> Result<RcData, RecvError> {
+    self.0.recv()
+  }
+  pub fn len(&self) -> usize {
+    self.0.len()
+  }
+  pub fn is_empty(&self) -> bool {
+    self.0.is_empty()
+  }
+  pub fn iter(&self) -> Iter<RcData> {
+    self.0.iter()
+  }
+}
+
+pub type PassDataChannel = (RcDataSender, RcDataReceiver);
+
+pub type FrameInput<T> = (Option<Arc<Frame<T>>>, Option<FrameParameters>);
+
+/// Endpoint to send frames
+pub struct FrameSender<T: Pixel> {
+  sender: Sender<FrameInput<T>>,
+  config: Arc<EncoderConfig>,
+  limit: u64,
+  count: u64,
+}
+
+// Proxy the crossbeam Sender
+//
+// TODO: enforce the limit
+impl<T: Pixel> FrameSender<T> {
+  fn new(
+    limit: u64, sender: Sender<FrameInput<T>>, config: Arc<EncoderConfig>,
+  ) -> FrameSender<T> {
+    Self { sender, config, limit, count: 0 }
+  }
+  pub fn try_send<F: IntoFrame<T>>(
+    &mut self, frame: F,
+  ) -> Result<(), TrySendError<FrameInput<T>>> {
+    if self.limit <= self.count {
+      Err(TrySendError::Full(frame.into()))
+    } else {
+      let r = self.sender.try_send(frame.into());
+      if r.is_ok() {
+        self.count += 1;
+      }
+      r
+    }
+  }
+
+  pub fn send<F: IntoFrame<T>>(
+    &mut self, frame: F,
+  ) -> Result<(), SendError<FrameInput<T>>> {
+    if self.limit <= self.count {
+      Err(SendError(frame.into()))
+    } else {
+      let r = self.sender.send(frame.into());
+      if r.is_ok() {
+        self.count += 1;
+      }
+      r
+    }
+  }
+  pub fn len(&self) -> usize {
+    self.sender.len()
+  }
+  pub fn is_empty(&self) -> bool {
+    self.sender.is_empty()
+  }
+  // TODO: proxy more methods
+}
+
+// Frame factory
+impl<T: Pixel> FrameSender<T> {
+  /// Helper to create a new frame with the current encoder configuration
+  #[inline]
+  pub fn new_frame(&self) -> Frame<T> {
+    Frame::new(
+      self.config.width,
+      self.config.height,
+      self.config.chroma_sampling,
+    )
+  }
+}
+
+/// Endpoint to receive packets
+pub struct PacketReceiver<T: Pixel> {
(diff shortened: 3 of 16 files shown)


src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+use crate::rate::RCState;
+use crate::rayon::{ThreadPool, ThreadPoolBuilder};
+use crate::util::Pixel;
+
+use std::io;
+use std::sync::Arc;
+
+/// Endpoint to send previous-pass statistics data
+pub struct RcDataSender {
+  sender: Sender<RcData>,
```
shssoichiro, line 31: Could we use the `bounded` helper in crossbeam to avoid having to manually implement senders and receivers? It seems a lot of this code is boilerplate and would be provided by crossbeam.

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+  }
+  // TODO: proxy more methods
+}
+
+// Frame factory
+impl<T: Pixel> FrameSender<T> {
+  /// Helper to create a new frame with the current encoder configuration
+  #[inline]
+  pub fn new_frame(&self) -> Frame<T> {
+    Frame::new(self.enc.width, self.enc.height, self.enc.chroma_sampling)
```
shssoichiro, line 161: This will probably need to be changed given the removal of the Frame::new API.

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+
+  /// Create a multipass encoder channel
+  ///
+  /// The `PacketReceiver<T>` may block if not enough pass statistics data
+  /// are sent through the `PassDataSender` endpoint
+  ///
+  /// Drop the `FrameSender<T>` endpoint to flush the encoder.
+  /// The last buffer in the PassDataReceiver is the summary of the whole
+  /// encoding process.
+  pub fn new_multipass_channel<T: Pixel>(
```
shssoichiro, line 519: What is the difference between this and the first+second pass channels? Is the idea that this one would be able to run both passes simultaneously? Or simply a simplified API but still running sequentially?

src/api/channel.rs
lu-zero: No since those endpoint need to do more than that: 
- error reporting
- `new_frames()`
- `container_sequence_header()`

src/api/channel.rs
lu-zero: It is removed only from the public API, but I had to change rav1e-ch.rs due to that.

src/api/channel.rs
lu-zero: This is to run both passes simultaneously, you take from a previous first pass and emit a new first pass.

Potentially we can have  
``` rust
fn new_multipass_channel<T: Pixel>() -> Result<(VideoDataChannel<T>, Option<RcDataReceiver, Option<RcDataSender>), InvalidConfig>
```
that allocates the right combination of channels based on its configuration.

This is one of the open topics.

src/rate.rs
```diff
@@ -1471,6 +1474,12 @@ impl RCState {
     );
     cur_pos =
       self.buffer_val(self.prev_metrics.log_scale_q24 as i64, 4, cur_pos);
+    eprintln!(
+      "show_frame {} fti {} log_scale_q24 {}",
+      self.prev_metrics.show_frame,
```
Urhengulas, line 1479: Maybe have:
```rust
let prev_metrics = self.prev_metrics;
```
in the beginning of the function to make it a bit less verbose?

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
+// Copyright (c) 2017-2019, The rav1e contributors. All rights reserved
```
Urhengulas, line 1:
```suggestion
// Copyright (c) 2020, The rav1e contributors. All rights reserved
```

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
...
+use crate::common::*;
+use crate::error::*;
+use crate::stats::*;
+use rav1e::prelude::*;
+
+use crate::decoder::{Decoder, FrameBuilder, VideoDetails};
+use crate::muxer::*;
+use std::fs::File;
+use std::io::{Read, Seek, Write};
+use std::sync::Arc;
```
Urhengulas, line 50: Also maybe structure the imports into two blocks. One with crate, the other without.

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
...
+
+impl<T: Pixel> FrameBuilder<T> for FrameSender<T> {
+  fn new_frame(&self) -> Frame<T> {
+    FrameSender::new_frame(self)
```
Urhengulas, line 54:
```suggestion
    self.new_frame()
```

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
...
+          cfg.new_channel().map_err(|e| e.context("Invalid setup"))?;
+        (channel, (None, None))
+      }
+    };
+
+  let y4m_details = source.input.get_video_details();
+
+  crossbeam::thread::scope(move |s| -> Result<(), CliError> {
+    // Receive pass data
+    let a = s.spawn(move |_| -> Result<(), CliError> {
```
Urhengulas, line 166: Maybe not use `a` as variable name?

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
...
+              })?;
+            }
+          }
+        }
+      }
+      Ok(())
+    });
+
+    // Send frames
+    let b = s.spawn(move |_| -> Result<(), CliError> {
```
Urhengulas, line 218: Maybe not use `b` as variable name?

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
...
+      while source.read_frame(&mut send_frame, y4m_details) {}
+
+      // send_frame.result()
+      Ok(())
+    });
+
+    // Send pass data
+    let c = s.spawn(move |_| -> Result<(), CliError> {
```
Urhengulas, line 226: Maybe not use `c` as variable name?

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
...
+            e.context("Unable to read the two-pass data file.")
+          })?;
+        }
+      }
+
+      Ok(())
+    });
+
+    // Receive Packets
+    let d = s.spawn(move |_| -> Result<(), CliError> {
```
Urhengulas, line 261: Maybe not use `d` as variable name?

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
...
+  });
+
+  run().map_err(|e| {
+    error::print_error(&e);
+    Box::new(e) as Box<dyn std::error::Error>
+  })
+}
+
+fn init_logger() {
+  use std::str::FromStr;
```
Urhengulas, line 334: why having the use statement in the function and not on top?

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
...
+    // output to stdout
+    .chain(std::io::stderr())
+    .apply()
+    .unwrap();
+}
+
+cfg_if::cfg_if! {
+  if #[cfg(any(target_os = "windows", target_arch = "wasm32"))] {
+    fn print_rusage() {
+      eprintln!("Windows benchmarking is not supported currently.");
```
Urhengulas, line 379: Maybe add sth about wasm in the Error message also?

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
...
+      .with_colorspace(y4m_dec.get_colorspace())
+      .write_header(rec)
+      .unwrap(),
+    ),
+    None => None,
+  };
+
+  match video_info.bit_depth {
+    8 | 10 | 12 => {}
+    _ => return Err(CliError::new("Unsupported bit depth")),
```
Urhengulas, line 438: Are there other bit depths? :exploding_head:

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,574 @@
...
+  cli.io.output.write_header(
+    video_info.width,
+    video_info.height,
+    cli.enc.time_base.den as usize,
+    cli.enc.time_base.num as usize,
+  );
+
+  info!(
+    "Using y4m decoder: {}x{}p @ {}/{} fps, {}, {}-bit",
+    video_info.width,
```
Urhengulas, line 518: Maybe also move these attributes out to make it less verbose?

```rust
let width = video_info.width;
// ...
```

src/bin/rav1e-ch.rs
shssoichiro: Theoretically input could be 16- or 32-bit. But rav1e doesn't support it. (I think even AV1 doesn't support it? Haven't checked.)

lu-zero: Yep

src/bin/rav1e-ch.rs
lu-zero: because it is used only in one place.

src/bin/rav1e-ch.rs
lu-zero: Unrelated to this patch, but it could be added to the rav1e cli on master as well.

Cargo.toml
```diff
@@ -19,9 +19,11 @@ readme = "README.md"
 repository = "https://github.com/xiph/rav1e/"
 autobenches = false
 autobins = false
+default-run = "rav1e"
 
 [features]
 unstable = []
+channel-api = ["crossbeam", "unstable"]
```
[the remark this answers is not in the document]
lu-zero: it should go the other way round, good catch!

README.md
```diff
@@ -100,9 +100,16 @@ cargo build --release
...
 
 ```sh
-cargo build --features unstable
+cargo build --features <feature>,unstable
 ```
 
+#### Current unstable features
+- Channel API:
+```sh
+cargo build --features channel-api,unstable
```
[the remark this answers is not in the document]
lu-zero: Ideally you have to explicitly ask for the unstable feature so you won't enable it by mistake.

src/rate.rs
lu-zero: it is a leftover, thanks for spotting. It will be removed.

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+
+use crate::api::context::RcData;
+use crate::encoder::*;
+use crate::frame::*;
+use crate::rate::RCState;
+use crate::rayon::{ThreadPool, ThreadPoolBuilder};
+use crate::util::Pixel;
+
+use std::io;
+use std::sync::Arc;
```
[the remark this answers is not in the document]
lu-zero: I prefer to keep std and crate::api aside, RcData should got up.

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+}
+
+impl RcDataSender {
+  fn new(limit: u64, sender: Sender<RcData>) -> RcDataSender {
+    Self { sender, limit, count: 0 }
+  }
+  pub fn try_send(
+    &mut self, data: RcData,
+  ) -> Result<(), TrySendError<RcData>> {
+    if self.limit < self.count {
```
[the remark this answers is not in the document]
lu-zero: It is not, good catch :)

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+      Err(TrySendError::Full(data))
+    } else {
+      let r = self.sender.try_send(data);
```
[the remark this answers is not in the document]
lu-zero: it makes the code more verbose.

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+      if r.is_ok() {
+        self.count += 1;
+      }
+      r
+    }
+  }
+  pub fn send(&mut self, data: RcData) -> Result<(), SendError<RcData>> {
+    if self.limit < self.count {
```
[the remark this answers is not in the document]
lu-zero: same error, thank you for spotting.

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+      Err(SendError(data))
+    } else {
+      let r = self.sender.send(data);
```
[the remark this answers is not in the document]
lu-zero: It makes the code more verbose.

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+  pub fn iter(&self) -> Iter<RcData> {
+    self.0.iter()
+  }
+
+  pub fn summary_size(&self) -> usize {
+    crate::rate::TWOPASS_HEADER_SZ
+  }
+}
+
+pub type PassDataChannel = (RcDataSender, RcDataReceiver);
```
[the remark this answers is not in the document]
lu-zero: I prefer to keep it closer to where the data types are defined.

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+
+pub type FrameInput<T> = (Option<Arc<Frame<T>>>, Option<FrameParameters>);
+
+/// Endpoint to send frames
+pub struct FrameSender<T: Pixel> {
```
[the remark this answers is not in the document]
lu-zero: We have another kind of sender and receiver pairs so I preferred to be explicit.

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+  sender: Sender<FrameInput<T>>,
+  enc: Arc<EncoderConfig>,
```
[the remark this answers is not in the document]
lu-zero: I can rename it to `config`.

src/api/channel.rs
```diff
@@ -0,0 +1,649 @@
...
+    let seq = Sequence::new(&self.enc);
+
+    sequence_header_inner(&seq).unwrap()
+  }
+}
+
+/// A channel modeling an encoding process
+pub type VideoDataChannel<T> = (FrameSender<T>, PacketReceiver<T>);
+
+impl Config {
```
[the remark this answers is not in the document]
lu-zero: it has to be here. This way the Config extension for the `channel-api` feature doesn't require additional `cfg` guards.

src/api/channel.rs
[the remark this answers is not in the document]
Urhengulas: I actually think it makes it easier to read. But maybe just a preference.

README.md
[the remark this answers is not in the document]
Urhengulas: Maybe you could also call the feature `unstable-channel-api` and thereby don't need the separate `unstable` feature?

And as soon it is stable you call rename it back to `channel-api`.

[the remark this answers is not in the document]
lu-zero: I'd rather not go that route, but it depends on how many unstable features we have that would make sense to be used together.

src/bin/rav1e-ch.rs
vibhoothiiaanand: +1 for that, 
```suggestion
    let pass_data = s.spawn(move |_| -> Result<(), CliError> {
```

Cargo.toml
```diff
@@ -73,6 +75,7 @@ ivf = { version = "0.1", path = "ivf/", optional = true }
 v_frame = { version = "0.1", path = "v_frame/" }
 av-metrics = { version = "0.5.1", optional = true }
 rayon = "1.0"
+crossbeam = { version = "0.7", optional = true }
```
vibhoothiiaanand, line 78: Current crossbeam version is 0.7.3, not sure if going to latest upstream version is better

lu-zero: It will pick the latest from 0.7 whichever it is :)

src/bin/rav1e-ch.rs
```diff
@@ -0,0 +1,575 @@
...
+  exit_requested: Arc<std::sync::atomic::AtomicBool>,
+}
+
+impl<D: Decoder> Source<D> {
+  cfg_if::cfg_if! {
+    if #[cfg(all(unix, feature = "signal-hook"))] {
+      fn new(limit: usize, input: D) -> Self {
+        let exit_requested = {
+          use std::sync::atomic::*;
+          let e = Arc::new(AtomicBool::from(false));
```
vibhoothiiaanand, line 72: Nit: I think it will be lot easier to read if we some better var name had to read entire function to get full idea ^^;

lu-zero: It is the same code present in the normal cli, I would not add spurious changes that would make unifying the cli later harder.

vibhoothiiaanand: makes sense

lu-zero:
Now with the commits collapsed without the interims :)

PR #2454, opened 2020-07-16
Fix quantizer index clamping

YaLTeR:
Fixes panics on saturating_add(100) and corruption on saturating_sub(100) to the dc qi.

Should I remove these debug asserts too?
https://github.com/YaLTeR/rav1e/blob/9721626144b00c377530c1dd0c9199040ef36a58/src/encoder.rs#L940-L943


tdaede:
Those debug asserts seem redundant with the new ones you added?

YaLTeR:
I've removed them.

PR #2457, opened 2020-07-17
Run tests in Github Actions with ASM disabled

### LINKED ISSUE #2452: Travis tests in CI does not seem to include "RAV1E_CPU_TARGET=rust" run
I suspect this because my pending WIP PR #2396 shows numerous cargo test fails when run locally with
RAV1E_CPU_TARGET=rust, but Travis CI only shows the fail for arm target, and furthermore only for the function that does not have asm code.

@barrbrain

shssoichiro:
Closes #2452

### DIFF
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -52,6 +52,7 @@ jobs:
          - 1.43.1-tests
          - aom-tests
          - dav1d-tests
+         - no-asm-tests
          - grcov-coveralls
          - bench
          - doc
@@ -67,6 +68,8 @@ jobs:
          - conf: aom-tests
            toolchain: stable
          - conf: dav1d-tests
+           toolchain: stable
+         - conf: no-asm-tests
            toolchain: stable
          - conf: grcov-coveralls
            toolchain: stable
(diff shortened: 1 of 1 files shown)


shssoichiro:
Here's my personal opinion, and I'm open to discussion on this:
1. We should be avoiding adding to Travis runs. We currently only have Travis as part of our pipeline because Github Actions does not yet support ARM. Our goal is to move off of Travis completely once possible.
2. Given that these tests are running without any platform-specific code, it seems unnecessary to also run them on ARM. Any native Rust code which passes on x86 but fails on ARM would be a compiler bug.

vibhoothiiaanand:
> Here's my personal opinion, and I'm open to discussion on this:
> 
> 1. We should be avoiding adding to Travis runs. We currently only have Travis as part of our pipeline because Github Actions does not yet support ARM. Our goal is to move off of Travis completely once possible.
> 2. Given that these tests are running without any platform-specific code, it seems unnecessary to also run them on ARM. Any native Rust code which passes on x86 but fails on ARM would be a compiler bug.

I agree with both.

.github/workflows/rav1e.yml
```diff
@@ -94,6 +97,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - name: Set env vars
```
Luni-4, line 100: Can we use this approach

```yaml
    - name: Set no-asm-tests environment variables
      if: matrix.conf == 'no-asm-tests'
      run: |
        echo "::set-env name=RAV1E_CPU_TARGET::rust"
```

instead of an external action?

shssoichiro: Sure, I just used the recommendation from the Github Actions docs. This is honestly better.

shssoichiro:
Thanks! Need one more review, since luni does not have merge permissions. https://screen.ac/d5uW7rd2

PR #2458, opened 2020-07-17
api: internal: Export frame_subtype and update script to store it.

vibhoothiiaanand:
Right now we are exporting/storing width, height, block_importance,
for exhaustive analysis we need frame_subtype.
The get_frame_subtype from encoder.rs, will give the type of frame:
 - if key_frame, then I(0) frame
 - else, it will P(1) frame + pyramid level depth.

We are currently not using B0, B1 Frames for rav1e, but with help
of pyramid level value we can get the index.

Edit:
CLI Output preview:
<Details>

```bash
vibhoothiiaanand@derpBox /V/b/r/r/t/o/dump (draw-importance) [1]>
/Volumes/buildBox/rav1e/rav1e/tools/draw-importances.py --path .  --verbose

Input given: (None, None)
Path given: .
MODE 2
Frame Type List: [2, 3, 2, 3, 3, 2, 3, 0, 3, 1, 3, 1, 3, 1, 1, 3, 2, 3, 2, 3]

 png list: ['./6-hres.png', './7-hres.png', './14-hres.png', './15-hres.png', './19-hres.png', './18-hres.png', './1-hres.png', './0-hres.png', './13-hres.png', './12-hres.png', './17-hres.png', './16-hres.png', './5-hres.png', './4-hres.png', './8-hres.png', './9-hres.png', './10-hres.png', './11-hres.png', './2-hres.png', './3-hres.png']
```
</Details>

Edit 2:
With the third commit 5c7b96d003c53dd4389689b5c9a30e6a6008b757, the calculation of the frame_subtype in the script is rapid, earlier it was taking some time/script is running till the images are exported, we do not need that on normal runtime.

Edit 3:
Sorry for multiple edits, wanted to make it proper, 
Now it is matured enough for better data collection,
New changes after Version 2.
+ Introduced CSV Option to Export as CSV 

Right now, when we do large processing of data, we are not storing
the mean importance and frame type anywhere, so we require to save
it to a file, by having an export option to csv, this makes things
easier.

To use: append --csv option along with other modes.
This is only applicable to batch mode

+ api: internal: Have zero-padded file name

Currently, we are having natural path names, when we do sorting and
further analysis, it becomes a bit problematic as the natural sort is not
robust. By having zero paddings to the file name, both lexicographic
and natural sort are equal.

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -1043,6 +1043,7 @@ impl<T: Pixel> ContextInner<T> {
         let mut buf = vec![];
         buf.write_u64::<NativeEndian>(fi.h_in_imp_b as u64).unwrap();
         buf.write_u64::<NativeEndian>(fi.w_in_imp_b as u64).unwrap();
+        buf.write_u64::<NativeEndian>(fi.get_frame_subtype() as u64).unwrap();
         for y in 0..fi.h_in_imp_b {
           for x in 0..fi.w_in_imp_b {
             buf
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -593,7 +593,7 @@ impl<T: Pixel> ContextInner<T> {
         plane.cfg.height as u32,
         |x, y| image::Luma([plane.p(x as usize, y as usize).as_()]),
       )
-      .save(format!("{}-qres.png", fi.input_frameno))
+      .save(format!("{:010}-qres.png", fi.input_frameno))
       .unwrap();
       let plane = &fs.input_hres;
       image::GrayImage::from_fn(
@@ -601,7 +601,7 @@ impl<T: Pixel> ContextInner<T> {
         plane.cfg.height as u32,
         |x, y| image::Luma([plane.p(x as usize, y as usize).as_()]),
       )
-      .save(format!("{}-hres.png", fi.input_frameno))
+      .save(format!("{:010}-hres.png", fi.input_frameno))
       .unwrap();
     }
 
@@ -677,7 +677,8 @@ impl<T: Pixel> ContextInner<T> {
           buf.write_i16::<NativeEndian>(mv.col).unwrap();
         }
       }
-      ::std::fs::write(format!("{}-mvs.bin", fi.input_frameno), buf).unwrap();
+      ::std::fs::write(format!("{:010}-mvs.bin", fi.input_frameno), buf)
+        .unwrap();
     }
 
     // Set lookahead_rec_buffer on this FrameInvariants for future
@@ -1053,7 +1054,7 @@ impl<T: Pixel> ContextInner<T> {
               .unwrap();
           }
         }
-        ::std::fs::write(format!("{}-imps.bin", fi.input_frameno), buf)
+        ::std::fs::write(format!("{:010}-imps.bin", fi.input_frameno), buf)
           .unwrap();
       }
     }
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -30,6 +30,9 @@ use arrayvec::ArrayVec;
 use log::Level::Info;
 use std::cmp;
 use std::collections::{BTreeMap, BTreeSet};
+use std::env;
+use std::fs;
+use std::path::PathBuf;
 use std::sync::Arc;
 
 /// The set of options that controls frame re-ordering and reference picture
@@ -433,6 +436,21 @@ impl<T: Pixel> ContextInner<T> {
     Ok(())
   }
 
+  #[allow(unused)]
+  pub fn build_dump_properties() -> PathBuf {
+    let mut data_location = PathBuf::new();
+    if env::var_os("RAV1E_DATA_PATH").is_some() {
+      data_location.push(&env::var_os("RAV1E_DATA_PATH").unwrap());
+      fs::create_dir_all(data_location.clone()).unwrap();
+      data_location
+    } else {
+      data_location.push(&env::current_dir().unwrap());
+      data_location.push(".lookahead_data");
+      fs::create_dir_all(data_location.clone()).unwrap();
+      data_location
+    }
+  }
+
   fn build_frame_properties(
     &mut self, output_frameno: u64,
   ) -> Result<FrameInvariants<T>, EncoderStatus> {
@@ -587,21 +605,24 @@ impl<T: Pixel> ContextInner<T> {
 
     #[cfg(feature = "dump_lookahead_data")]
     {
+      let data_location = Self::build_dump_properties();
       let plane = &fs.input_qres;
+      let mut file_name = format!("{:010}-qres", fi.input_frameno);
       image::GrayImage::from_fn(
         plane.cfg.width as u32,
         plane.cfg.height as u32,
         |x, y| image::Luma([plane.p(x as usize, y as usize).as_()]),
       )
-      .save(format!("{:010}-qres.png", fi.input_frameno))
+      .save(data_location.join(file_name).with_extension("png"))
       .unwrap();
       let plane = &fs.input_hres;
+      file_name = format!("{:010}-hres", fi.input_frameno);
       image::GrayImage::from_fn(
         plane.cfg.width as u32,
         plane.cfg.height as u32,
         |x, y| image::Luma([plane.p(x as usize, y as usize).as_()]),
       )
-      .save(format!("{:010}-hres.png", fi.input_frameno))
+      .save(data_location.join(file_name).with_extension("png"))
       .unwrap();
     }
 
@@ -651,7 +672,8 @@ impl<T: Pixel> ContextInner<T> {
     #[cfg(feature = "dump_lookahead_data")]
     {
       use crate::partition::RefType::*;
-
+      let data_location = Self::build_dump_properties();
+      let file_name = format!("{:010}-mvs", fi.input_frameno);
       let second_ref_frame = if !self.inter_cfg.multiref {
         LAST_FRAME // make second_ref_frame match first
       } else if fi.idx_in_group_output == 0 {
@@ -677,8 +699,11 @@ impl<T: Pixel> ContextInner<T> {
           buf.write_i16::<NativeEndian>(mv.col).unwrap();
         }
       }
-      ::std::fs::write(format!("{:010}-mvs.bin", fi.input_frameno), buf)
-        .unwrap();
+      ::std::fs::write(
+        data_location.join(file_name).with_extension("bin"),
+        buf,
+      )
+      .unwrap();
     }
 
     // Set lookahead_rec_buffer on this FrameInvariants for future
@@ -1042,6 +1067,8 @@ impl<T: Pixel> ContextInner<T> {
       {
         use byteorder::{NativeEndian, WriteBytesExt};
         let mut buf = vec![];
+        let data_location = Self::build_dump_properties();
+        let file_name = format!("{:010}-imps", fi.input_frameno);
         buf.write_u64::<NativeEndian>(fi.h_in_imp_b as u64).unwrap();
         buf.write_u64::<NativeEndian>(fi.w_in_imp_b as u64).unwrap();
         buf.write_u64::<NativeEndian>(fi.get_frame_subtype() as u64).unwrap();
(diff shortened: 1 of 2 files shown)


tools/draw-importances.py
```diff
@@ -75,21 +81,25 @@ def blockImportance(input, verbose, path, raw):
...
                 col_list.append(cols)
+                frame_type_list.append(frame_type)
+                mean_list.append(np.mean(imps))
 
         if raw:
             click.secho("The full imps data after processing:", fg="red")
             click.echo(imps_list)
 
     if verbose and path != None:
+        click.echo("Frame Type List: "+ str(frame_type_list))
```
barrbrain, line 98: This output of data is inconsistent with the lines below.
```suggestion
        click.secho("Frame Type List: "+ str(frame_type_list))
```

tools/draw-importances.py
```diff
@@ -136,45 +147,46 @@ def blockImportance(input, verbose, path, raw):
                                                     bbox_inches='tight')
 
     else:
```
barrbrain, line 149: By using `elif` here, we may avoid indenting this block.

tools/draw-importances.py
```diff
@@ -52,10 +53,13 @@ def blockImportance(input, verbose, path, raw):
...
 
-        rows, cols = struct.unpack('qq', contents[:16])
-        imps = np.frombuffer(contents[16:],
+        rows, cols, frame_type = struct.unpack('qqq', contents[:24])
+        imps = np.frombuffer(contents[24:],
                             dtype=np.float32).reshape((rows, cols))
 
+        if verbose:
+            click.echo("Frame Type: "+ str(frame_type))
+            click.echo("Mean Average: "+ str(np.mean(imps)))
```
barrbrain, line 62:
```suggestion
            click.echo("Mean Importance: "+ str(np.mean(imps)))
```

vibhoothiiaanand:
Open Question:
So right now exporting to CSV, we are using Numpy modules So we are having accurate precision, while when we use Pandas, we are having extra precision longer values, not sure if that is required or not. I have tried to have it without introducing additional module only for this, but did not work out. If we require extra longer float precision, we can use that.


`2.5084376335144043` with pandas data frame.
`2.508438` with numpy(currently) 

Would like to hear from @barrbrain

barrbrain:
The input data is in `f32` so 7 digits precision should be enough.

src/api/internal.rs
```diff
@@ -433,6 +437,15 @@ impl<T: Pixel> ContextInner<T> {
     Ok(())
   }
 
+  /*
+  #[cfg(feature = "dump_lookahead_data")]
```
lu-zero, line 441: Why this?

src/api/internal.rs
```diff
@@ -587,21 +600,25 @@ impl<T: Pixel> ContextInner<T> {
 
     #[cfg(feature = "dump_lookahead_data")]
     {
+      let out_dir = &env::var("CARGO_MANIFEST_DIR").unwrap();
+      let data_location = env::var("RAV1E_DATA_PATH")
```
lu-zero, line 604: You should use Path::join() when dealing with paths.

vibhoothiiaanand:
Clarification:
When we use PathBuf as such without converting to string like: 
https://github.com/xiph/rav1e/blob/1475fda3721f02e95ce0d1d83a3d0cf2b007f63f/src/api/internal.rs#L600

It will cause some issues as PathBuf default formatting is not available, we are doing with the help of ` {:?}`
so it will be resolving with `" "` for first and no quotes for the second  like: `"/playground/target/out/"0000000001`

So we have two options as far as I understand 
- Convert it to os_string `into_os_string()` then into `string()` then `unwrap()` it.
- Convert it to `as_path()` and `display()` then `to_strings()`
as_path reference: https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.as_path
https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.display
Also display method gas it is "Returns an object that implements Display for safely printing paths that may contain non-Unicode data."
I thought to go for the second option as it seems to be more convenient, 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9cc1a6b77e95b1894b7a3762edba39b3

Would be happy to hear thoughts and if some better way is there too.

src/api/internal.rs
vibhoothiiaanand: I had to reintroduce this, since L598
```diff
    };
+     #[cfg(feature = "dump_lookahead_data")]
+     let data_location =
+       Self::build_dump_properties().as_path().display().to_string();
```

without the flag, clippy says variable is not used and advises to remove if we keep that as only optional then the function is warned as not used, so kept flag, I think it will reduce the overhead marginally and users/devs who do not use this feature, do not need this.

vibhoothiiaanand: I have used `#[allow(unused)]` to fix this

src/api/internal.rs
```diff
@@ -433,6 +436,24 @@ impl<T: Pixel> ContextInner<T> {
...
+    let mut out_dir = PathBuf::new();
+    let mut data_location = PathBuf::new();
+    out_dir.push(&env::var_os("CARGO_MANIFEST_DIR").unwrap());
+
+    if env::var_os("RAV1E_DATA_PATH").is_some() {
+      data_location.push(&env::var_os("RAV1E_DATA_PATH").unwrap());
+      fs::create_dir_all(data_location.clone()).unwrap();
+      data_location
+    } else {
+      data_location.push(&env::var_os("CARGO_MANIFEST_DIR").unwrap());
```
YaLTeR, line 450: Not sure if this is the best idea since:
- You have to use `cargo` to get the manifest dir environment variable (so just running the binary built with data dumping will panic out of the box). Idk if it's a big deal considering this is a debug feature anyway.
- The `target/out` folder might not make the most sense for this sort of output if we consider its usual subfolders:
  - `debug`, `release`
  - `<target triple>` for building with `--target=triple`
  - `rls` for RLS
  - `doc` for `cargo doc`

src/api/internal.rs
```diff
@@ -433,6 +436,24 @@ impl<T: Pixel> ContextInner<T> {
     Ok(())
   }
 
+  #[allow(unused)]
+  pub fn build_dump_properties() -> PathBuf {
+    let mut out_dir = PathBuf::new();
+    let mut data_location = PathBuf::new();
+    out_dir.push(&env::var_os("CARGO_MANIFEST_DIR").unwrap());
```
YaLTeR, line 443: I guess this is left-over?

src/api/internal.rs
```diff
@@ -587,21 +608,23 @@ impl<T: Pixel> ContextInner<T> {
 
     #[cfg(feature = "dump_lookahead_data")]
     {
+      let data_location =
+        Self::build_dump_properties().as_path().display().to_string();
```
YaLTeR, line 612: It would be more correct to avoid path-to-regular-string conversion and instead use `Path::join()`. `ImageBuffer::save()` does accept an `AsRef<Path>` rather than a `&str`. This way there will be no issues with invalid-UTF8 paths.

lu-zero:
> Would be happy to hear thoughts and if some better way is there too.

Never use `format!()` that way always use `Path::join()` :)

src/api/internal.rs
vibhoothiiaanand: Ah yes, thanks for the catch

src/api/internal.rs
vibhoothiiaanand: Ah yes, 
and this feature is not enabled by default tho, anyone needs to compile with `--feature` flag to have this,

and yes, this is not a portable solution tbh, so I am modifying it to something like this

```diff
-      data_location.push(&env::var_os("CARGO_MANIFEST_DIR").unwrap());
-      data_location.push("target/out/");
+      data_location.push(&env::current_dir().unwrap());
+      data_location.push("out/");
       fs::create_dir_all(data_location.clone()).unwrap();
```

So it creates out/ folder where the data will be dumped, which seems to be more reasonable. 
Is that ok?

vibhoothiiaanand:
> Never use `format!()` that way always use `Path::join()` :)

I have updated with required changes, now it is better and also used `with_extension("X")` for having extensions :)

src/api/internal.rs
YaLTeR: Maybe something more telling instead of `out`?

src/api/internal.rs
```diff
@@ -587,21 +605,24 @@ impl<T: Pixel> ContextInner<T> {
...
+      let data_location = Self::build_dump_properties();
       let plane = &fs.input_qres;
+      let mut file_name = format!("{:010}-qres", fi.input_frameno);
       image::GrayImage::from_fn(
         plane.cfg.width as u32,
         plane.cfg.height as u32,
         |x, y| image::Luma([plane.p(x as usize, y as usize).as_()]),
       )
-      .save(format!("{}-qres.png", fi.input_frameno))
+      .save(Path::new(&data_location).join(file_name).with_extension("png"))
```
lu-zero, line 616: PathBuf has a [join](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.join).

src/api/internal.rs
```diff
@@ -587,21 +605,24 @@ impl<T: Pixel> ContextInner<T> {
...
       .unwrap();
       let plane = &fs.input_hres;
+      file_name = format!("{:010}-hres", fi.input_frameno);
       image::GrayImage::from_fn(
         plane.cfg.width as u32,
         plane.cfg.height as u32,
         |x, y| image::Luma([plane.p(x as usize, y as usize).as_()]),
       )
-      .save(format!("{}-hres.png", fi.input_frameno))
+      .save(Path::new(&data_location).join(file_name).with_extension("png"))
```
lu-zero, line 625: Same comment here.

src/api/internal.rs
```diff
@@ -677,7 +699,11 @@ impl<T: Pixel> ContextInner<T> {
           buf.write_i16::<NativeEndian>(mv.col).unwrap();
         }
       }
-      ::std::fs::write(format!("{}-mvs.bin", fi.input_frameno), buf).unwrap();
+      ::std::fs::write(
+        Path::new(&data_location).join(file_name).with_extension("bin"),
```
lu-zero, line 703: And here.

src/api/internal.rs
```diff
@@ -1052,8 +1081,11 @@ impl<T: Pixel> ContextInner<T> {
               .unwrap();
           }
         }
-        ::std::fs::write(format!("{}-imps.bin", fi.input_frameno), buf)
-          .unwrap();
+        ::std::fs::write(
+          Path::new(&data_location).join(file_name).with_extension("bin"),
```
lu-zero, line 1085: And here

src/api/internal.rs
vibhoothiiaanand: Not sure what to keep, one option would be getting filename from binary or something like that, but I think it maybe overkill for debug feature because,
+ This is only fall-back case when user do have env var,
+ Right now it is making out dir from $PWD.

On quick conversation in channel,

Chat Logs:

```irc
18:00 <lu_zero> this feature is a debug feature
18:00 <lu_zero> and you could restrict its scope as much as needed
18:00 <lu_zero> in the end you want to not clutter the root-directory
18:01 <lu_zero> so ideally you could just make it always output to .{descriptive-name}/
18:02 <lu_zero> and have an env-override in case somebody wants to explicitly give it a full path
18:02 <lu_zero> so you do not need to check for cargo-specific env
18:02 <lu_zero> or such
18:02 <mindfreeze> right now, we are not having any cargo-specific ones
18:03 <lu_zero> just by default write to ${PWD}/.some-name-you-decide
18:03 <mindfreeze> also updated the patchset with join and with_extension, please do check it :) 
```

src/api/internal.rs
vibhoothiiaanand: Fixed, for a moment, I did not think that way :) 
Thanks

src/api/internal.rs
vibhoothiiaanand: Now it is with `.lookahead_data` as discussed in IRC :)

vibhoothiiaanand:
Rebased to latest master and also a slight modification to CSV storing commit with the addition of file_name as arg instead of a predefined name. So batch processing will be easier.

vibhoothiiaanand:
22:16 <YaLTeR> ./shrug i mean it's not a big deal so if you're okay with the dot then you can merge it
Ivan has given +1 for changes via IRC, so going forward :)

PR #2459, opened 2020-07-17
Use CARGO_CFG_TARGET_OS where needed

### LINKED ISSUE #2435: Undefined reference to asm functions cross compiling from macOS to mingw-w64
I'm trying to compile the latest rav1e, but I'm getting loads of undefined references to asm functions:
```
ld: asm/x86/predict.rs:169: undefined reference to `rav1e_ipred_z1_avx2'
ld: asm/x86/predict.rs:175: undefined reference to `rav1e_ipred_z3_avx2'
ld: asm/x86/predict.rs:179: undefined reference to `rav1e_ipred_smooth_avx2'
...
ld: asm/x86/predict.rs:212: undefined reference to `rav1e_ipred_v_ssse3'
ld: asm/x86/predict.rs:224: undefined reference to `rav1e_ipred_smooth_h_ssse3'
ld: asm/x86/predict.rs:227: undefined reference to `rav1e_ipred_paeth_ssse3'
...
```

I'm not familiar with Rust enough to ensure the correct library is being linked in, but it does appear that these functions in particular are `extern {}` defined.

I attempted to build with: `cargo build --release` using Rust 1.44.1

lu-zero:
Fixes #2435.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -80,7 +80,7 @@ fn build_nasm_files() {
   config_file.write(b"	%define PIC 1\n").unwrap();
   config_file.write(b" %define STACK_ALIGNMENT 16\n").unwrap();
   config_file.write(b" %define HAVE_AVX512ICL 0\n").unwrap();
-  if cfg!(target_os = "macos") {
+  if env::var("CARGO_CFG_TARGET_OS").unwrap() == "macos" {
     config_file.write(b" %define PREFIX 1\n").unwrap();
   }
 


build.rs
```diff
@@ -80,7 +80,7 @@ fn build_nasm_files() {
   config_file.write(b"	%define PIC 1\n").unwrap();
   config_file.write(b" %define STACK_ALIGNMENT 16\n").unwrap();
   config_file.write(b" %define HAVE_AVX512ICL 0\n").unwrap();
-  if cfg!(target_os = "macos") {
```
tdaede, line 83: is the target_os cfg deprecated now? did we have documentation for it?

lu-zero: It is not deprecated, it is the build-time information for build.rs, so it is always the host os.
to have the target_os for what we are building for you have to use the env variable.

PR #2462, opened 2020-07-21
Ensure all mv predictors are converted to fullpel

KyleSiefring:
Subpel motion vectors were being created when seeding the full search.
Later, when doing subpel search, the mvs retained the fractional
component from the initial estimate.

Also prune zero mvs from coarse motion vectors. This is a side effect
of refactoring the code.

AWCY run: https://beta.arewecompressedyet.com/?job=rerun-master-96a439c1f360b510651eb7aedf39c24458e02003%402020-07-21T17%3A06%3A11.036Z&job=quantize_fullpel_fix2%402020-07-21T16%3A48%3A53.265Z

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -86,11 +86,22 @@ pub fn get_subset_predictors<T: Pixel>(
 ) -> ArrayVec<[MotionVector; 17]> {
   let mut predictors = ArrayVec::<[_; 17]>::new();
 
-  // Zero motion vector
+  // Add a candidate predictor, aligning to fullpel and filtering out zero mvs.
+  let add_cand = |predictors: &mut ArrayVec<[MotionVector; 17]>,
+                  cand_mv: MotionVector| {
+    let cand_mv = cand_mv.quantize_to_fullpel();
+    if !cand_mv.is_zero() {
+      predictors.push(cand_mv)
+    }
+  };
+
+  // Zero motion vector, don't use add_cand since it skips zero vectors.
   predictors.push(MotionVector::default());
 
   // Coarse motion estimation.
-  predictors.extend(cmvs.into_iter().map(MotionVector::quantize_to_fullpel));
+  for mv in cmvs {
+    add_cand(&mut predictors, mv);
+  }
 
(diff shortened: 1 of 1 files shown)


tdaede:
Does not using zero mvs as candidates make sense? Currently we can always code them as GLOBALMV and try the separately, but would not being able to choose them in the search mess up the predictors for other searches?

KyleSiefring:
> Does not using zero mvs as candidates make sense? Currently we can always code them as GLOBALMV and try the separately, but would not being able to choose them in the search mess up the predictors for other searches?

We do use 0 mvs, it's just that we prune duplicates of that vector from the list. It's a little arbitrary, but that's what the existing code did.

src/me.rs
```diff
@@ -86,35 +86,40 @@ pub fn get_subset_predictors<T: Pixel>(
...
+  // Add a candidate predictor, aligning to fullpel and filtering out zero mvs.
+  let add_cand = |predictors: &mut ArrayVec<[MotionVector; 17]>,
+                  cand_mv: MotionVector| {
+    let cand_mv = cand_mv.quantize_to_fullpel();
+    if !cand_mv.is_zero() {
+      predictors.push(cand_mv)
+    }
+  };
+
+  // Zero motion vector, don't use add_cand since it skips zero vectors.
```
rzumer, line 98: Why not make add_cand push the MV without the `quantize_to_fullpel()` call? That seems more straightforward than making it a special case at this level.

tdaede:
It would definitely improve subpel search speed but the difference is probably a lot smaller for fullpel search.
The rationale for not seeding the subpel search with subpel-accurate predictors is that the subpel search pattern doesn't make sense to start at a subpel offset?

KyleSiefring:
> It would definitely improve subpel search speed but the difference is probably a lot smaller for fullpel search.
> The rationale for not seeding the subpel search with subpel-accurate predictors is that the subpel search pattern doesn't make sense to start at a subpel offset?

These are seeding the fullpel search.

PR #2463, opened 2020-07-21
Reduce input size of the test 'odd_size_frame_with_full_rdo'

ycho:
I am HAPPY to release this PR that will save precious time of every rav1e developer! :)


tdaede:
Any reason you didn't just replace the 512 with 64?

ycho:
> Any reason you didn't just replace the 512 with 64?

Yes, both +1 or 64-1 are very peculiar test as you replied in #daala, and I expect +1 is as tricky as 64-1.

PR #2464, opened 2020-07-22
Change rust backtrace to full

ycho:


ycho:
> Can you please remove other changes in the commit related to open-partition

Yeap- just did that!

ycho:
> Nit: It is actually CI and not Travis, updating commit message would be better imo
> Travis is only used for ARM Tests, and this is actually GitHub Actions if you are looking for the term replacing Travis :)

Oh, thanks for comment, I changed it.

shssoichiro:
I couldn't find very good (read: any) documentation on how RUST_BACKTRACE=full works. The one mention I found in a PR shows that it _is_ case-sensitive and needs to be lowercase. However, that code no longer seems to be in master? So I'm not even sure if `full` does anything different than `1` anymore. I think the only way to verify is to test and see if the backtrace becomes more verbose.

ycho:
> I couldn't find very good (read: any) documentation on how RUST_BACKTRACE=full works. The one mention I found in a PR shows that it _is_ case-sensitive and needs to be lowercase. However, that PR is from 4 years ago and that code no longer seems to be in master? So I'm not even sure if `full` does anything different than `1` anymore. I think the only way to verify is to test and see if the backtrace becomes more verbose.

Yes, that is valid point. The only source that made me try write this PR was a part of the error message from github CI, saying ... try "RUST_BACKTRACE=FULL" to see full details (not exact copy of the message, since at this point all those errors are fixed fortunately and cannot access the error message anymore).
In addition, two CI tests (both for mac os) seems not proceeding with unknown reason to me.

As you proposed, one (and only) way is to land this PR then submit some test PR (which fails run tests)!

ycho:
> > I couldn't find very good (read: any) documentation on how RUST_BACKTRACE=full works. The one mention I found in a PR shows that it _is_ case-sensitive and needs to be lowercase. However, that PR is from 4 years ago and that code no longer seems to be in master? So I'm not even sure if `full` does anything different than `1` anymore. I think the only way to verify is to test and see if the backtrace becomes more verbose.
> 
> Yes, that is valid point. The only source that made me try write this PR was a part of the error message from github CI, saying ... try "RUST_BACKTRACE=FULL" to see full details (not exact copy of the message, since at this point all those errors are fixed fortunately and cannot access the error message anymore).
> In addition, two CI tests (both for mac os) seems not proceeding with unknown reason to me.
> 
> As you proposed, one (and only) way is to land this PR then submit some test PR (which fails run tests)!

Regarding the difference between full and FULL, 
honestly I am now even confused whether it was RUST_BACKTRACE=full or RUST_BACKTRACE=FULL.

lu-zero:
[this](https://doc.rust-lang.org/std/backtrace/index.html) seems to state that the variable has to be set with something, no matter what.

shssoichiro:
I read some of the RFC linked in std::backtrace:

> Today, the RUST_BACKTRACE controls backtraces generated by panics. After this RFC, it also controls backtraces generated in the standard library. Two additional variables will be added: RUST_PANIC_BACKTRACE and RUST_LIB_BACKTRACE: these will independently override the behavior of RUST_BACKTRACE for backtraces generated for panics and from the std API.

So it sounds like actually this is more complicated than it needs to be, for whatever reason that the Rust language team decided... and that still fails to actually answer whether "full" does anything different from "1". But it does sound like the code that handles "panic" backtraces is not the same as std::backtrace::Backtrace and is found somewhere else, which I haven't found yet.

barrbrain:
I couldn't find any documented methods along the implementation path but here it is in libstd:
https://github.com/rust-lang/rust/blob/1.45.0/src/libstd/sys_common/backtrace.rs#L163-L173

ycho:
> I couldn't find any documented methods along the implementation path but here it is in libstd:
> https://github.com/rust-lang/rust/blob/1.45.0/src/libstd/sys_common/backtrace.rs#L163-L173

Thanks for finding and proving this info. So, would we want then merge this PR anyway w/o testing it elsewhere?

shssoichiro:
We'll want to change it to lowercase. The code barrbrain linked shows that it is case sensitive and only matches on the exact value "full".

ycho:
> We'll want to change it to lowercase. The code barrbrain linked shows that it is case sensitive and only matches on the exact value "full". Any other value except 0 (including "FULL") will give short backtraces.

Oh, there it needs. I did a wrong guess that there would a step to convert to lowercase.

ycho:
> We'll want to change it to lowercase. The code barrbrain linked shows that it is case sensitive and only matches on the exact value "full". Any other value except 0 (including "FULL") will give short backtraces.

Done.

PR #2466, opened 2020-07-23
Format tools

Luni-4:
Run `black` and `isort` on the `Python` scripts. Add a toml file to set the parameters used by the format tools.

### DIFF
--- a/tools/draw-importances.py
+++ b/tools/draw-importances.py
@@ -1,28 +1,39 @@
 #!/usr/bin/env python3
+
+import datetime
+import glob
+import os
+import re
 import struct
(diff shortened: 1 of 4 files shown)


Luni-4:
> Thanks for this,
> 
> This is a missed thing ,
> 
> So I wonder how in future will this work?

Run `black tools` to format the code and `isort -rc tools` to sort out imports. If you want to set parameters both for `isort` and `black` tools, change the `pyproject.toml` file. Let me update it with some configurations for `isort`

Luni-4:
Ok, now it should be ready.

PR #2471, opened 2020-07-27
Reduce space to store Block struct

KyleSiefring:
Reduces size from 96 bytes to 36 bytes. A single block is used for every
4x4 of the current frame.

### DIFF
--- a/src/context/block_unit.rs
+++ b/src/context/block_unit.rs
@@ -167,11 +167,11 @@ pub struct Block {
   pub ref_frames: [RefType; 2],
   pub mv: [MotionVector; 2],
   // note: indexes are reflist index, NOT the same as libaom
-  pub neighbors_ref_counts: [usize; INTER_REFS_PER_FRAME],
+  pub neighbors_ref_counts: [u8; INTER_REFS_PER_FRAME],
   pub cdef_index: u8,
   pub bsize: BlockSize,
-  pub n4_w: usize, /* block width in the unit of mode_info */
-  pub n4_h: usize, /* block height in the unit of mode_info */
+  pub n4_w: u8, /* block width in the unit of mode_info */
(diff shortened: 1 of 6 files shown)


KyleSiefring:
@shssoichiro A new AWCY run and I think it was just noise. :(

shssoichiro:
I wonder if the criterion benchmarks would be useful here?

src/context/block_unit.rs
```diff
@@ -167,11 +167,11 @@ pub struct Block {
...
   pub mv: [MotionVector; 2],
   // note: indexes are reflist index, NOT the same as libaom
-  pub neighbors_ref_counts: [usize; INTER_REFS_PER_FRAME],
+  pub neighbors_ref_counts: [u8; INTER_REFS_PER_FRAME],
   pub cdef_index: u8,
   pub bsize: BlockSize,
-  pub n4_w: usize, /* block width in the unit of mode_info */
-  pub n4_h: usize, /* block height in the unit of mode_info */
+  pub n4_w: u8, /* block width in the unit of mode_info */
+  pub n4_h: u8, /* block height in the unit of mode_info */
```
barrbrain, line 174: I'm not sure exactly what is happening with padding in this struct. When I tried replacing `n4_w` and `n4_h` with functions of `bsize`, `size_of<Block>()` was reduced to 32.

barrbrain: The padding was due to the widest alignment requirement, `enum TxSize` which has `#[repr(C)]` and is thus 4-bytes wide. This is a remnant from when rav1e was still linked to libaom, see https://github.com/xiph/rav1e/commit/4745e694ba3895e921475e24eb2e6e4d18899094.

tdaede:
I was able to measure a tiny speed improvement on the benchmarks.

Issue #2461, opened 2020-07-20
Prevent passing Frames with the wrong dimensions and/or chroma sampling


lu-zero:
As in the code snippet from #2460 we do not validate that the frame provided has the correct dimensions since we assume the user uses `Context::new_frame()` to create a new frame with the correct settings.

In order to add validation we either need to store the configuration in the `Frame`.

Urhengulas:
I'd like to take this on, also because I'd like access to the configuration for the jsapi.

PR #2472, opened 2020-07-27
Remove rav1e::Frame::new()

lu-zero:
It is a big footgun (#2461) and should be avoided.


lu-zero:
Thank you, it landed yesterday by mistake.

PR #2473, opened 2020-07-27
Use new, non-broken me predictors

KyleSiefring:
The old predictors were sampling without regards to block size. All of
the samples would be taken from 4x4 blocks adjacent or on the upper left
corner of the currently tested block. The top right subset B (EPZS,
current frame) predictor and right and bottom subset C (prev frame) were
sampling entirely incorrect locations. The other predictors were biased
to that upper left corner.

Add an additional top left predictor.

| -0.3292 | -0.3730 | -0.1695 |  -0.3330 | -0.3641 | -0.3603 |    -0.3046 |

### DIFF
--- a/src/mc.rs
+++ b/src/mc.rs
@@ -23,30 +23,11 @@ use crate::tiling::*;
 use crate::util::*;
 
 use simd_helpers::cold_for_target_arch;
-use std::ops;
 
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 pub struct MotionVector {
   pub row: i16,
   pub col: i16,
-}
-
-impl ops::Add<MotionVector> for MotionVector {
-  type Output = MotionVector;
-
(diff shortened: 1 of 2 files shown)


KyleSiefring:
| -0.2775 | -0.3811 | -0.1415 |  -0.2733 | -0.2422 | -0.2178 |    -0.2871 |

I made some changes that reduced some of my gains.

KyleSiefring:
> [07 27 13:40:06] \<derf\> maseck, soichiro: It looks okay to me, but as a comment on pre-existing code, is it possible to use named constants for these parameters to ArrayVec?

I think this would be best to deal with when resolving the underlying problem here. (not checking a threshold)

KyleSiefring:
The man who chronically can't log into github says

> \<derf\> maseck: Anyway, it still looks fine.

PR #2474, opened 2020-07-27
Refactor me.rs

KyleSiefring:


src/me.rs
```diff
@@ -165,395 +172,361 @@ pub fn get_subset_predictors<T: Pixel>(
...
-      &ts.input.planes[0],
+      ts,
+      org_region,
       &rec.frame.planes[0],
-      &predictors,
-      fi.sequence.bit_depth,
-      pmv,
+      tile_bo_adj,
       lambda,
+      pmvs.iter().cloned().filter_map(identity).collect(),
```
shssoichiro, line 323: What is the purpose of the filter_map here?

src/me.rs
```diff
@@ -165,395 +172,361 @@ pub fn get_subset_predictors<T: Pixel>(
...
+  let org_region =
+    &ts.input_hres.region(Area::StartingAt { x: po.x, y: po.y });
+
+  let tile_mvs = &ts.mvs[ref_frame.to_index()].as_const();
+  let frame_ref =
+    fi.rec_buffer.frames[fi.ref_frames[0] as usize].as_ref().map(Arc::as_ref);
+
+  let mut predictors = get_subset_predictors::<T>(
+    tile_bo_adj,
+    pmvs.iter().cloned().filter_map(identity).collect(),
```
shssoichiro, line 422: Also here (filter_map)

KyleSiefring: Collecting the non None values of an array into an ArrayVec<[MotionVector, ?]>. This code is only marked new because I removed a trait thus decreasing indentation. That has been intentionally put in a separate commit to make it easier to review.

shssoichiro: Thanks, for some reason it looked new when I was looking at the diff.

PR #2480, opened 2020-08-03
Jsapi update ci

Urhengulas:
This PR has two purposes:
* basic checking of the javascript in `rav1e_js/www/`
* prepare for a continuous deployment of the rav1e_js example website

## Changes
* ~~Upload artifact 'rav1e_js-pkg'~~
* Install and build javascript 
* Fix `wasm-pack test --headless --firefox`

### DIFF
--- a/.github/workflows/js_api.yml
+++ b/.github/workflows/js_api.yml
@@ -25,6 +25,7 @@ jobs:
             SCCACHE_CACHE_SIZE: 300M
             SCCACHE_DIR: /home/runner/.cache/sccache
             SCCACHE_IDLE_TIMEOUT: 0
+            RAV1E_JS_DIR: rav1e_js
 
         runs-on: ubuntu-latest
 
@@ -84,13 +85,13 @@ jobs:
             # BUILD & PACK
             - name: Build
               if: matrix.conf == 'build'
-              working-directory: rav1e_js
+              working-directory: ${{ env.RAV1E_JS_DIR }}
               # --dev needed, because of https://github.com/rustwasm/wasm-pack/issues/886
               run: wasm-pack build --dev
 
             - name: Pack
               if: matrix.conf == 'build'
-              working-directory: rav1e_js
+              working-directory: ${{ env.RAV1E_JS_DIR }}
               run: wasm-pack pack
 
             # TESTING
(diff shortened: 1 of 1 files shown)


Urhengulas:
> Seems good to me too :)
> 
> Perhaps we could convert the `deploy` job in a new configuration of the `build-pack-test` job, called `deploy`, in this way it isn't necessary to upload/download artifacts. What do you think?
> 
> Does the `deploy` job need to be run for `pull_request` events also? If not, we need to move the entire job in a new `yml` file.

The thing is, though it is called "deploy" it also does linting for js, which is also useful on PRs. But you are probably right, that this should be a configuration of the other job :)

Luni-4:
> The thing is, though it is called "deploy" it also does linting for js, which is also useful on PRs. But you are probably right, that this should be a configuration of the other job :)

I see, so let's convert it in a configuration :)

Urhengulas:
I've decided to add the `yarn install && yarn build` to the "build" configuration, because to run yarn we need to build the wasm before anyways.

Additionally I also fixed `wasm-pack test --headless --firefox` (https://github.com/rustwasm/wasm-bindgen/issues/2261).

.github/workflows/js_api.yml
```diff
@@ -16,8 +16,10 @@ jobs:
                 include:
                     - conf: firefox
                       headless: --headless
+                      driver: --geckodriver ./geckodriver
                     - conf: chrome
                       headless: --headless
+            fail-fast: false
```
Luni-4, line 22: it would be better to remove this line, in this way GitHub cancels all in-progress jobs if any matrix job fails.

.github/workflows/js_api.yml
```diff
@@ -101,10 +103,19 @@ jobs:
               with:
                   node-version: 14
 
+            # needed because of # https://github.com/rustwasm/wasm-bindgen/issues/2261
+            - name: Install geckodriver
+              if: matrix.conf == 'firefox'
+              working-directory: ${{ env.RAV1E_JS_DIR }}
+              run: |
+                  wget https://github.com/mozilla/geckodriver/releases/download/v0.26.0/geckodriver-v0.26.0-linux64.tar.gz
```
Luni-4, line 111: Add the URL to a `LINK` env variable like [here](https://github.com/xiph/rav1e/blob/master/.github/workflows/rav1e.yml#L106)

.github/workflows/js_api.yml
Urhengulas: I know. But because of the firefox error I noticed it is actually useful to run all configurations to know if the problem is with only one or all of them.

.github/workflows/js_api.yml
Urhengulas: Okay :+1: 

That's actually way cleaner

.github/workflows/js_api.yml
Luni-4: Aren't the configurations run in parallel? Can't we look at the GUI to detect which configuration has failed?

Urhengulas: They are run, but because of scheduling sometimes a configuration didn't start yet, so it will get canceled if another one fails.

Urhengulas: E.g. Firefox fails --> cancels chrome even, if it would have been fine

Urhengulas:
> Seems good to me now, thanks!

Thanks for the review :P

Luni-4:
> Thanks for the review :P

Thanks to you for your work! :P

PR #2483, opened 2020-08-06
tools: submit_awcy: Submit runs by commit ID

vibhoothiiaanand:
With this commit, users can submit runs to AWCY with
custom commit ID instead of choosing head which is the
current behaviour.
Usage: submit_awcy.py <normal commands> -commit $ID

### DIFF
--- a/tools/submit_awcy.py
+++ b/tools/submit_awcy.py
@@ -55,6 +55,7 @@ parser = argparse.ArgumentParser(
     description="Submit test to arewecompressedyet.com"
 )
 parser.add_argument("-branch", default=None)
+parser.add_argument("-commit", default=None)
 parser.add_argument("-prefix", default=None)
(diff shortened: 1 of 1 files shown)


tools/submit_awcy.py
```diff
@@ -78,12 +79,20 @@ def f(*popenargs, **kwargs):
...
+        .strip()
+        .decode("utf-8")
+    )
+else:
+    commit = args.commit
+    short_arg = "git rev-parse --short " + args.commit
+    short = (
+        subprocess.check_output(short_arg, shell=True).strip().decode("utf-8")
+    )
+
```
YaLTeR, line 95: This could be extracted down from both branches

vibhoothiiaanand: I tried to address this, can you please check agian

PR #2484, opened 2020-08-06
Optimize predictors for multipass motion estimation

KyleSiefring:
Since we perform multiple passes during motion estimation, it makes
sense to sample the bottom and right edges for predictors.
| -0.3470 | -0.4163 | -0.4998 |  -0.3425 | -0.3951 | -0.3994 |    -0.4004 |

https://beta.arewecompressedyet.com/?job=master-b17d5009f7e539aba2e1abde9cbb5203f1aca3f3&job=me_port_improved_predictors2%402020-08-06T17%3A44%3A58.052Z

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -90,13 +90,13 @@ pub fn get_subset_predictors<T: Pixel>(
   tile_bo: TileBlockOffset, cmvs: ArrayVec<[MotionVector; 7]>,
   tile_mvs: &TileMotionVectors<'_>, frame_ref_opt: Option<&ReferenceFrame<T>>,
   ref_frame_id: usize, bsize: BlockSize,
-) -> ArrayVec<[MotionVector; 16]> {
-  let mut predictors = ArrayVec::<[_; 16]>::new();
+) -> ArrayVec<[MotionVector; 17]> {
+  let mut predictors = ArrayVec::<[_; 17]>::new();
   let w = bsize.width_mi();
   let h = bsize.height_mi();
 
   // Add a candidate predictor, aligning to fullpel and filtering out zero mvs.
-  let add_cand = |predictors: &mut ArrayVec<[MotionVector; 16]>,
+  let add_cand = |predictors: &mut ArrayVec<[MotionVector; 17]>,
                   cand_mv: MotionVector| {
     let cand_mv = cand_mv.quantize_to_fullpel();
     if !cand_mv.is_zero() {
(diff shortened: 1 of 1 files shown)


src/me.rs
```diff
@@ -90,13 +90,13 @@ pub fn get_subset_predictors<T: Pixel>(
   tile_bo: TileBlockOffset, cmvs: ArrayVec<[MotionVector; 7]>,
   tile_mvs: &TileMotionVectors<'_>, frame_ref_opt: Option<&ReferenceFrame<T>>,
   ref_frame_id: usize, bsize: BlockSize,
-) -> ArrayVec<[MotionVector; 16]> {
-  let mut predictors = ArrayVec::<[_; 16]>::new();
+) -> ArrayVec<[MotionVector; 17]> {
+  let mut predictors = ArrayVec::<[_; 17]>::new();
```
lu-zero, line 94: If you are going to change it often make it a type so you would not have to keep in mind to change it in two places.

src/me.rs
```diff
@@ -112,26 +112,49 @@ pub fn get_subset_predictors<T: Pixel>(
     add_cand(&mut predictors, mv);
   }
 
-  // EPZS subset A and B predictors.
```
ycho, line 115: Any reason you didn't want to keep this line of EPZS comment?

tdaede:
The only thing I don't like about this is that it removes the distinction of which predictors are from a previous pass and which are from fully RDO'd neighbor blocks, because both are stored in tile_mvs.

ycho:
> The only thing I don't like about this is that it removes the distinction of which predictors are from a previous pass and which are from fully RDO'd neighbor blocks, because both are stored in tile_mvs.

Yeap I understand. I think it can only be knowable at best from implicit information, unless there is separate storage or indicator flag.

src/me.rs
```diff
@@ -112,26 +112,49 @@ pub fn get_subset_predictors<T: Pixel>(
...
 
+  // Sample the center of the current block.
+  add_cand(
+    &mut predictors,
+    tile_mvs[tile_bo.0.y + clipped_half_h][tile_bo.0.x + clipped_half_w],
+  );
+
+  // Sample the middle of all blocks bordering this one.
+  // Note: If motion vectors haven't been precomputed to a given blocksize, then
+  // then the right and bottom edges will be duplicates of the center predictor
```
ycho, line 128: 'then' repeated.

PR #2485, opened 2020-08-06
Revise EPZS for open-partition

ycho:
- Fix to not use ts.input.plane[] directly and instead use
  ts.input_tile[].
- Candidate position for EPZS motion estimation must be clipped to frame
  size if a block straddle on frame boundary.


ycho:
There is no change with coding efficiency.
[awcy -s 6](https://beta.arewecompressedyet.com/?job=master-66dc853284fddd9efcd1aacae593a913ed3432e0&job=revise_epzs_for_openpartition_7eca57314eec45c07%402020-08-06T23%3A23%3A46.609Z)

PR #2487, opened 2020-08-07
Do not use C-impossible types in the API

### LINKED ISSUE #2441: rav1e_config_set_mastering_display arguments aren't C-ABI safe
The current nightly rustc is marking it as faulty eventually and cbindgen will reject it [soon](https://github.com/eqrion/cbindgen/issues/528).

It should take a `*const rav1e::ChromaticityPoint` instead.

lu-zero:
Fixes #2441.


src/capi.rs
```diff
@@ -562,12 +562,16 @@ pub unsafe extern fn rav1e_config_set_content_light(
 /// Returns a negative value on error or 0.
 #[no_mangle]
 pub unsafe extern fn rav1e_config_set_mastering_display(
-  cfg: *mut Config, primaries: [rav1e::ChromaticityPoint; 3],
+  cfg: *mut Config, primaries: *const rav1e::ChromaticityPoint,
```
YaLTeR, line 565: Maybe `*const [rav1e::ChromaticityPoint; 3]` and then just `let primaries = *primaries;`? Does it work like this?

lu-zero: It would work now but there is no ABI-guarantee it keeps working in the future. rust-arrays do not have a C equivalent

YaLTeR: I was reading the [unsafe code guidelines](https://rust-lang.github.io/unsafe-code-guidelines/layout/arrays-and-slices.html) the other day and it says:
> If the element type is repr(C) the layout of the array is guaranteed to be the same as the layout of a C array with the same element type.

and has a pointer-to-array-in-FFI example.

lu-zero: Sounds like we can get away with it in this case, to be seen if cbindgen maps it correctly.

lu-zero: It produces
```
const RaChromaticityPoint (*primaries)[3]
```

YaLTeR: I... don't actually know if this is what we need or one level of indirection more. 😀 Frequent issue with some of the C declarations...

lu-zero: we do not need that, but since the layout is guaranteed to be correct within rust I can simplify the code doing an evil cast :P. 

With [this](https://github.com/eqrion/cbindgen/pull/545) we get

``` c
int rav1e_config_set_mastering_display(RaConfig *cfg,
                                       const RaChromaticityPoint primaries[3],
                                       RaChromaticityPoint white_point,
                                       uint32_t max_luminance,
                                       uint32_t min_luminance);
```

from

``` rust
/// cbindgen:ptrs-as-arrays=[[primaries;3]]
#[no_mangle]
pub unsafe extern fn rav1e_config_set_mastering_display(
  cfg: *mut Config, primaries: *const rav1e::ChromaticityPoint,
  white_point: rav1e::ChromaticityPoint, max_luminance: u32,
  min_luminance: u32,
) -> c_int
```

YaLTeR: Not a huge fan of repeating the array length in a comment, but I suppose this'll do.

Issue #2453, opened 2020-07-15
Visible desyncs in some 4K clips


rzumer:
Still researching the cause, and will update this issue as I figure it out.

Commit: https://github.com/xiph/rav1e/commit/01052fc7121114904501f81f52a234c96373829d
Sample: https://drive.google.com/file/d/1UrXkq8pDKBFhq8wojWd7W_VOBOIY-pb2/view?usp=sharing (3840x2160)
Command lines and output:
```
$ cargo run --bin rav1e --release -- surfing_50f.y4m -s 10 -r surfing_50f_rec.y4m -o surfing_50f.ivf
$ dav1d -i surfing_50f.ivf -o surfing_50f_dec.y4m
$ cmp <(tail -n+2 surfing_50f_dec.y4m) <(tail -n+2 surfing_50f_rec.y4m)
/dev/fd/63 /dev/fd/62 differ: byte 318238388, line 19122
```

Blocky artifacts pop in and out in the decoded output, so this affects image quality significantly. Definitely should block any release, assuming it is not also present in past ones. Visible artifacts appear concentrated in the top-right corner of the visible part of the frame in this sample.

Occurs in both constant quantizer and single-pass bitrate target mode. Could reproduce down to speed 7, with very visible artifacts at levels 9 and 10.

rzumer:
Bisect points to 5d7599c9bbdcbd1ff44d100d806003bd7021544d.

ycho:
Hey- if possible, could you revert the suspected commit 5d7599c9bbdcbd1ff44d100d806003bd7021544d and test please? That way, we can confirm that the bisected commit is the sole cause.

rzumer:
@ycho As expected, reverting fixes the desync.

ycho:
> @ycho As expected, reverting fixes the desync.

Oh, indeed. Thanks a lot for testing this again!

ycho:
@barrbrain , Hey- I think this issue is wanting your help! :)

rzumer:
Forgot to mention in writing, the issue is present in v0.3.2. In the future I recommend avoiding merging non-bugfixes for patch releases.

rzumer:
We found that enabling CDEF while disabling loop restoration (done at speed level 10) is another cause. It's likely that there is a bug at that level that was exposed by the RDO commit.

ycho:
Please test with applying https://github.com/xiph/rav1e/pull/2488 , since this can be possible culprit as @barrbrain stated.

[...]

PR #2488, opened 2020-08-09
Fix tx size context update for skipped intra block

ycho:
This bug has not affected current master because the RDO mode decision
does not include testing a skip case for intra mode partition.

### DIFF
--- a/src/context/block_unit.rs
+++ b/src/context/block_unit.rs
@@ -324,7 +324,7 @@ impl<'a> BlockContext<'a> {
     let n4_h = bsize.height_mi();
 
     let (tx_w, tx_h) = if skip {
-      (n4_w as u8, n4_h as u8)
+      ((n4_w * MI_SIZE) as u8, (n4_h * MI_SIZE) as u8)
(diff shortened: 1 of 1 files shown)


barrbrain:
This is possibly related to #2453, which is a desync case at speeds 9 & 10. These are the only presets for which `enable_inter_tx_split` is set.

ycho:
> This is possibly related to #2453, which is a desync case at speeds 9 & 10. These are the only presets for which `enable_inter_tx_split` is set.

@rzumer,

Looking again, I think what you think is possible.
Since, this bug affect when both 'fi.tx_mode_select' and 'bsize == BlockSize::BLOCK_4X4 || (is_inter && skip)' are true.
I.e. for 1) 4x4 in intra frame, speed 0~5 (tx split enabled for intra block mode in intra frame)
or 2) skipped inter mode block in inter frames, speed 9,10 (tx split enabled for inter frame).

If any intra coded partition that has incorrect above or left tx_size contexts affected by either of above two cases will fail.

PR #2490, opened 2020-08-10
Support max_width/max_height options.

### LINKED ISSUE #2467: API to set sequence header max resolution

tdaede:
Fixes #2467


rzumer:
It could be useful to state in the doc comments what implications these options have on the decoded bitstream. Without reading the spec it's hard to tell what they might be used for.

src/capi.rs
```diff
@@ -632,6 +632,12 @@ unsafe fn option_match(
   match key {
     "width" => enc.width = check_frame_size(value.parse().map_err(|_| ()))?,
     "height" => enc.height = check_frame_size(value.parse().map_err(|_| ()))?,
+    "max_width" => {
+      enc.max_width = check_frame_size(value.parse().map_err(|_| ()))?
+    }
+    "max_height" => {
+      enc.max_heightheight = check_frame_size(value.parse().map_err(|_| ()))?
```
kodabb, line 639: `max_heightheight`?

src/bin/common.rs
```diff
@@ -655,6 +655,9 @@ fn parse_config(matches: &ArgMatches<'_>) -> Result<EncoderConfig, CliError> {
   cfg.tile_cols = matches.value_of("TILE_COLS").unwrap().parse().unwrap();
   cfg.tile_rows = matches.value_of("TILE_ROWS").unwrap().parse().unwrap();
 
+  cfg.max_width = matches.value_of("MAX_WIDTH").unwrap().parse().unwrap();
+  cfg.max_height = matches.value_of("MAX_HEIGHT").unwrap().parse().unwrap();
+
```
kodabb, line 660: should this fail if max_width/max_height are < or > than width/height? (can't tell which sets of parameters determines the amount of pixels to be displayed)

tdaede: yes, less than. max_width and max_height are basically to set the sequence options so you can create a compatible sequence header for switch frames.

tdaede: added validation in Config::validate()

tdaede:
Added some doc comments.

lu-zero:
it needs a rebase.

src/encoder.rs
```diff
@@ -182,8 +182,12 @@ pub struct Sequence {
...
 impl Sequence {
   pub fn new(config: &EncoderConfig) -> Sequence {
-    let width_bits = 32 - (config.width as u32).leading_zeros();
-    let height_bits = 32 - (config.height as u32).leading_zeros();
+    let max_width =
+      if config.max_width > 0 { config.max_width } else { config.width };
+    let max_height =
+      if config.max_height > 0 { config.max_height } else { config.height };
+    let width_bits = 32 - (max_width as u32).leading_zeros();
+    let height_bits = 32 - (max_height as u32).leading_zeros();
```
jamrial, line 190: Fix this while at it by making it max_{width,height} - 1, same as it was done in 325ae51558cec45e7c7e96b0121c3b966cdeca7c, so once you rebase this PR to solve the conflicts in src/header.rs you don't introduce a regression.

src/header.rs
```diff
@@ -868,10 +868,8 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
   fn write_frame_size<T: Pixel>(
     &mut self, fi: &FrameInvariants<T>,
   ) -> io::Result<()> {
-    // width_bits and height_bits will have to be moved to the sequence header OBU
-    // when we add support for it.
-    let width_bits = 32 - (fi.width as u32).leading_zeros();
-    let height_bits = 32 - (fi.height as u32).leading_zeros();
+    let width_bits = fi.sequence.num_bits_width;
+    let height_bits = fi.sequence.num_bits_height;
```
jamrial, line 872: You should also write fi.sequence.max_frame_width and fi.sequence.max_frame_height instead of fi.width and fi.height within this function, now that they may differ.

src/encoder.rs
```diff
@@ -223,8 +227,8 @@ impl Sequence {
       color_description: config.color_description,
       mastering_display: config.mastering_display,
       content_light: config.content_light,
-      max_frame_width: config.width as u32,
-      max_frame_height: config.height as u32,
+      max_frame_width: config.max_width as u32,
+      max_frame_height: config.max_height as u32,
```
jamrial, line 231: Should be max_width and max_height

vibhoothiiaanand:
Now it requires some rebase as #2520 is merged

vibhoothiiaanand (commented):
Now it requires some rebase as #2520 is merged

jamrial:
Also, the spec states that if reduced_still_picture_header is enabled (currently always forced for still_picture), dimensions in the sequence header and the frame header must be the same, so you should check the user provided arguments to ensure this is the case.

lu-zero:
This patch needs are rebase, should we land it before 0.5?

tdaede:
This has rotted pretty badly, it'll need some work but I would like it in.

tdaede:
A long time coming, but I've rebased this and fixed the nits.

src/api/config/encoder.rs
```diff
@@ -213,7 +221,7 @@ impl EncoderConfig {
     // has the property that the scaled distortion of a 2Nx2N block is always
     // equal to the sum of the scaled distortions of the NxN sub-blocks it's
     // made of, this is a necessary property to be able to do RDO between
-    // multiple partition sizes properly. Unfortunately, when tx domain
+    // multiple partition sizes properly. Unfortunately, when tx domains
```
rzumer, line 224: What's this change for? At first glance it seems like singular domain is grammatically correct.

rzumer:
LGTM besides one comment

jamrial:
If i set a max_width value lower than the width of the actual frames (e.g. a 720x480 y4m source encoded with --max-width=64) i get this error:

```
!! Invalid configuration: invalid width 720 (expected >= 16, <= 65535)
Error: Generic { msg: "Invalid configuration", e: "invalid width 720 (expected >= 16, <= 65535)" }
```

Which is confusing and wrong.

tdaede:
Would you prefer that the error message for width/height be updated to indicate that it should be less than 65535 or max_width, or that a new error message for explicitly max_width be created? (and if so, we could technically error either for the width or max_width setting?)

FWIW for the S-Frame content that this is useful for right now, the muxing happens outside of rav1e, so it itself doesn't have to see the width/height change dynamically.

jamrial:
The first option sounds good. Basically, replacing the hardcoded 65535 with something like min(65535, max_width).

tdaede:
(we should probably remove that redundant CLI error printing)

jamrial:
`!! Invalid configuration: invalid width 720 (expected >= 16, <= 720)`

But 720 is <= 720.

Also, if i passed --max_width=64, it should report `!! Invalid configuration: invalid width 720 (expected >= 16, <= 64)` for an input video with width 720.

tdaede:
Ah indeed I forgot to update a parameter of the format string. Now displays:
```
!! Invalid configuration: invalid width 1280 (expected >= 16, <= 720)
Error: Generic { msg: "Invalid configuration", e: "invalid width 1280 (expected >= 16, <= 720)" }
```

jamrial:
Should be good then. Thanks.

PR #2492, opened 2020-08-11
Lf asm reenable

xiphmont:
This set of patches is the sum set of changes involved in re-enabling use of the existing assembly for CDEF and the other loop filters.

The ASM fell unused when joint RDO of loop filters in rdo.rs went to a fully Pixel:u16 pipeline internally, partly because CDEF needed edge information passed in as part of the pixel data (via out of range flag values).  We only have assembly for the LBD case right now, so the u16 piepline rendered the assembly useless and caused a decent-sized speed regression.

Refactoring loop filter RDO to a Pixel:T pipeline (which will be u8 for LBD input) required refactoring CDEF to a more dav1d-like abstraction, which is a useful long-term goal anyway.  The refactor also eliminated a few large temporary buffer and copying; more copying will fall away once we adopt the actual dav1d assembly (which will require one change to dav1d:  separate src/dest arguments to the ASM routines)

The first commit in this patch set is a functional change to the clipping behavior of CDEF as used in LF RDO; the metrics show no effective change across objective-1-fast.  Subsequent patches are bit-exact.


src/asm/x86/cdef.rs
```diff
@@ -58,59 +58,83 @@ pub(crate) unsafe fn cdef_filter_block<T: Pixel>(
...
+      copy
+    };
+    match T::type_enum() {
+      PixelType::U8 => {
+        match CDEF_FILTER_FNS[cpu.as_index()][decimate_index(xdec, ydec)] {
+          Some(func) => {
+            // current cdef_filter_block asm does 16->8 for historical
+            // reasons.  Copy into tmp space for now (also handling
+            // padding) until asm is updated
+            let tmpstride = 16; /* 128-bit alignment greater than 2 * (8>>xdec) + 2 */
```
lu-zero, line 83: We should probably factor out all the alignment so bumping to avx512 or similar wide-vector-isa will not bite us.

xiphmont: The alignment there is only to further a hack that had *better* be short-lived-- I don't want that copy/temp space at all.  All the dav1d asm handles the border cases internally, our current implementation doesn't.  So that's literally some duct-tape to keep ours running for right now.

lu-zero:
https://github.com/xiph/rav1e/commit/ff4b17b57d18f4e0d81811f6281acfdda485adfa should fix some of the crashes, I opened a PR on your repo already :)

PR #2494, opened 2020-08-13
Avoid choosing unrepresentable motion vectors

### LINKED ISSUE #2453: Visible desyncs in some 4K clips

barrbrain:
Fixes #2453.

### DIFF
--- a/src/me.rs
+++ b/src/me.rs
@@ -68,7 +68,7 @@ pub struct MVSearchResult {
   cost: u64,
 }
 
-const fn get_mv_range(
+fn get_mv_range(
   w_in_b: usize, h_in_b: usize, bo: PlaneBlockOffset, blk_w: usize,
   blk_h: usize,
 ) -> (isize, isize, isize, isize) {
@@ -83,7 +83,14 @@ const fn get_mv_range(
     * (8 * MI_SIZE) as isize
     + border_h;
 
-  (mvx_min, mvx_max, mvy_min, mvy_max)
+  // <https://aomediacodec.github.io/av1-spec/#assign-mv-semantics>
+  use crate::context::{MV_LOW, MV_UPP};
+  (
+    mvx_min.max(MV_LOW as isize + 1),
+    mvx_max.min(MV_UPP as isize - 1),
+    mvy_min.max(MV_LOW as isize + 1),
+    mvy_max.min(MV_UPP as isize - 1),
+  )
 }
 
 pub fn get_subset_predictors<T: Pixel>(
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -767,6 +767,16 @@ pub fn rdo_tx_size_type<T: Pixel>(
 }
 
 #[inline]
+fn dmv_in_range(mv: MotionVector, ref_mv: MotionVector) -> bool {
+  let diff_row = mv.row as i32 - ref_mv.row as i32;
+  let diff_col = mv.col as i32 - ref_mv.col as i32;
+  diff_row >= MV_LOW
+    && diff_row <= MV_UPP
+    && diff_col >= MV_LOW
+    && diff_col <= MV_UPP
+}
+
+#[inline]
 fn luma_chroma_mode_rdo<T: Pixel>(
   luma_mode: PredictionMode, fi: &FrameInvariants<T>, bsize: BlockSize,
   tile_bo: TileBlockOffset, ts: &mut TileStateMut<'_, T>,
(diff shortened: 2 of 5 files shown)


src/rdo.rs
```diff
@@ -782,6 +793,29 @@ fn luma_chroma_mode_rdo<T: Pixel>(
...
+    } else {
+      [MotionVector::default(); 2]
+    };
+
+    if (luma_mode == PredictionMode::NEWMV
+      || luma_mode == PredictionMode::NEW_NEWMV
+      || luma_mode == PredictionMode::NEW_NEARESTMV)
+      && !mv_in_range(mvs[0], ref_mvs[0])
+    {
+      return;
```
ycho, line 808: Would there be case that all inter mode candidates, i.e. MVs tested by "sorted.iter().take(num_modes_rdo).for_each( ...
...
luma_chroma_mode_rdo( )..."
are out of range and return, thus all inter modes are dropped?

To avoid such case, are we able to revise motion_estimation() clip the MV to valid range, if the clipped is not completely bad choice.

src/rdo.rs
```diff
@@ -766,6 +766,17 @@ pub fn rdo_tx_size_type<T: Pixel>(
   (best_tx_size, best_tx_type)
 }
 
+#[inline]
+fn mv_in_range(mv: MotionVector, ref_mv: MotionVector) -> bool {
+  const MAX_MV_MAG: i32 = 1 << 14;
```
ycho, line 771: so this constant is for 1/8 pixel, Q3, right? So, in integer, 2048 pixel positions.

rzumer:
Thank you for looking into this issue and finding the root cause!
I started some runs to compare how currently desyncing clips (surfing and slow_pan) compare between this fix and a 4-tile configuration: http://34.68.179.107/?job=barrbrain-fix%402020-08-13T07%3A05%3A37.291Z&job=barrbrain-fix-4tiles%402020-08-13T07%3A06%3A14.046Z

barrbrain:
I started runs with the tighter constraint enforced by libaom: http://34.68.179.107/?job=barrbrain-fix%402020-08-13T08%3A24%3A28.052Z&job=barrbrain-fix-4tiles%402020-08-13T08%3A25%3A07.981Z

src/rdo.rs
barrbrain: As far as I can tell, constraining the motion-estimation is necessary since libaom enforces bounds on the absolute motion vector values. I have added this constraint.

src/rdo.rs
barrbrain: I identified `context::MV_UPP` as the name for this constant.

src/me.rs
```diff
@@ -83,7 +84,12 @@ const fn get_mv_range(
     * (8 * MI_SIZE) as isize
     + border_h;
 
-  (mvx_min, mvx_max, mvy_min, mvy_max)
+  (
```
KyleSiefring, line 87: I believe it's possible to jump motion vectors off of adjacent motion vectors/global motion. It would be nice to document this when are restricting the range of all motion vectors.

barrbrain: I'd like to get clarity on this constraint, particularly to confirm if it is normative in the spec. I will confirm which version of libaom I was looking at tomorrow.

barrbrain: The relevant logic in libaom begins here: https://aomedia.googlesource.com/aom/+/v2.0.0/av1/decoder/decodemv.c#1101
Note that `read_mv()` sums the delta and reference in the output: https://aomedia.googlesource.com/aom/+/v2.0.0/av1/decoder/decodemv.c#878
The `is_mv_valid()` test is applied to the output: https://aomedia.googlesource.com/aom/+/v2.0.0/av1/decoder/decodemv.c#1206

Finally, found the relevant part of the spec: https://aomediacodec.github.io/av1-spec/#assign-mv-semantics

barrbrain:
Confirmed that squashing and tidying up gave identical results: http://34.68.179.107/?job=barrbrain-fix%402020-08-13T09%3A07%3A33.936Z&job=barrbrain-fix-squashed%402020-08-14T02%3A14%3A10.332Z

PR #2499, opened 2020-08-17
Initial port to tracing

lu-zero:
It behaves as fern since no changes to the logging are present, but should let you use the tracing api if you need something more refined.

It clashes with the `tracing` feature that enables hawktracer and that's in itself unfortunate.


lu-zero:
Done in 1a774396a70b9003f0fec1eabd04de345349ea32

PR #2500, opened 2020-08-17
Substantially rework initial motion estimation

KyleSiefring:
* Removes consideration of various coarse motion vectors from the list
of candidate motion vectors. This seems to be the largest source of
regression.
* Limit use of fullsearch behind a threshold value determined by the sad
of candidate motion vectors.
* Consolidate logic to handle subsampling and partioning the superblock.

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019, The rav1e contributors. All rights reserved
+// Copyright (c) 2018-2020, The rav1e contributors. All rights reserved
 //
 // This source code is subject to the terms of the BSD 2 Clause License and
 // the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
@@ -636,7 +636,7 @@ impl<T: Pixel> ContextInner<T> {
         input_hres: fs.input_hres.clone(),
         input_qres: fs.input_qres.clone(),
         cdfs: fs.cdfs,
-        frame_mvs: fs.frame_mvs.clone(),
+        frame_me_stats: fs.frame_me_stats.clone(),
         output_frameno,
         segmentation: fs.segmentation,
       });
@@ -667,7 +667,7 @@ impl<T: Pixel> ContextInner<T> {
     compute_motion_vectors(fi, fs, &self.inter_cfg);
 
     // Save the motion vectors to FrameInvariants.
-    fi.lookahead_mvs = fs.frame_mvs.clone();
+    fi.lookahead_me_stats = fs.frame_me_stats.clone();
 
     #[cfg(feature = "dump_lookahead_data")]
     {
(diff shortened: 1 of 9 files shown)


src/encoder.rs
```diff
@@ -723,10 +719,10 @@ impl<T: Pixel> FrameInvariants<T> {
       tx_mode_select: false,
       default_filter: FilterMode::REGULAR,
       invalid: false,
-      lookahead_mvs: {
+      lookahead_me_stats: {
         let mut vec = Vec::with_capacity(REF_FRAMES);
```
lu-zero, line 723: maybe `vec![FrameMEStats::new(); REF_FRAMES]` while at it?

src/me.rs
```diff
@@ -93,126 +251,184 @@ fn get_mv_range(
...
   }
 
+  let median = if corner != BlockCorner::INIT {
+    // Sample the center of the current block.
+    Some(process_cand(
+      tile_me_stats[tile_bo.0.y + clipped_half_h]
+        [tile_bo.0.x + clipped_half_w],
+    ))
+  } else {
+    if subset_b.len() != 3 {
```
lu-zero, line 348: would read better as 
``` rust
} else if subset_b.len() == 3 {
  ...
  Some(...)
} else {
  None
}
```

src/me.rs
```diff
@@ -315,9 +329,14 @@ fn get_subset_predictors<T: Pixel>(
       tile_me_stats[tile_bo.0.y - 1][tile_bo.0.x + clipped_half_w],
     ));
   }
+
+  // I tried sampling far right and far bottom edges, but this had worse results
```
vibhoothiiaanand, line 333: Hi,
It may be better if the documentation is in third-person or more of technical one instead of being in first person stating as I... etc

KyleSiefring:
I think something went wrong with speed 10...

KyleSiefring:
<derf> maseck: My only remaining question is if you really meant > y_start or >= y_start.
<derf> Other than that, it LGTM, assuming you're going to squash and link an AWCY run.

src/api/internal.rs
```diff
@@ -855,14 +855,14 @@ impl<T: Pixel> ContextInner<T> {
       unique_indices.iter().for_each(|&(mv_index, rec_index)| {
         // Use rec_buffer here rather than lookahead_rec_buffer because
         // rec_buffer still contains the reference frames for the current frame
-        // (it's only overwritten when the frame is encoded), while
+        // (it's only overwritten when the frame is encoded), while/`
```
tmatth, line 858: nit: typo at the end of the line here

src/me.rs
```diff
@@ -18,54 +19,223 @@ use crate::mc::MotionVector;
...
+    // Will either process starting at the first block or only on the edges.
+    for y in (
+      // Don't skip the vertical edge.
+      if vert_edge { 0 } else { y_start as isize }
+        ..=h_in_b as isize - mv_size as isize
+    )
+      .step_by(mv_size)
+    {
+      // Process unprocessed columns and the horizontal edge.
+      let x_start = if !(edge_mode && vert_edge) || y as usize >= y_start {
```
barrbrain, line 181: > \<derf\> maseck: My only remaining question is if you really meant `> y_start` or `>= y_start`.

KyleSiefring: Yeah, I meant >= instead of > so I changed it.

barrbrain (approved):
Passing on derf's approval.

PR #2505, opened 2020-08-20
Support arbitrary SAR anamorphic video

### LINKED ISSUE #2504: Output of encoding small_input.y4m is stretched by players
To reproduce:
1. run ./build.sh as of #2503 
1. Use your favourite player to play any of rec_file.y4m, enc_file.ivf or dec_file.y4m

Expected:
Should be played with the same PAR as small_input.y4m

Actual:
It's rotated

My best guess is that we are not preserving the `A16:9` info from small_input y4m.

rzumer:
This adds an aspect ratio input parameter
to the Rust and C APIs, computes the render
size by upscaling one dimension, and codes
it in-bitstream, as well as in Y4M output.

The default aspect ratio is parsed from
Y4M input, or set to 1:1 if absent.

Note that neither aomdec nor dav1d will propagate the bitstream render size to the Y4M files they output.
It doesn't seem to be honored by ffmpeg with aomdec either, but it might be with dav1d (haven't tested it myself, but the code seems present). Reconstruction output is affected as well.
I think we can close #2504 with this, as it implements everything that should be needed for decoders/players to correctly output/display the encoded video.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ bitstream-io = "0.8"
 cfg-if = "0.1"
 clap = { version = "2", optional = true, default-features = false }
 libc = "0.2"
-y4m = { version = "0.6", optional = true }
+y4m = { version = "0.7", optional = true }
 backtrace = { version = "0.3", optional = true }
 num-traits = "0.2"
 num-derive = "0.3"
--- a/src/api/config/encoder.rs
+++ b/src/api/config/encoder.rs
@@ -29,6 +29,8 @@ pub struct EncoderConfig {
   pub width: usize,
   /// Height of the frames in pixels.
   pub height: usize,
+  /// Sample aspect ratio (for anamorphic video).
+  pub sample_aspect_ratio: Rational,
   /// Video time base.
   pub time_base: Rational,
 
@@ -129,6 +131,8 @@ impl EncoderConfig {
     EncoderConfig {
       width: 640,
       height: 480,
+      sample_aspect_ratio: Rational { num: 1, den: 1 },
+      time_base: Rational { num: 1, den: 30 },
 
       bit_depth: 8,
       chroma_sampling: ChromaSampling::Cs420,
@@ -144,8 +148,6 @@ impl EncoderConfig {
 
       error_resilient: false,
       switch_frame_interval: 0,
-
-      time_base: Rational { num: 1, den: 30 },
 
       min_key_frame_interval: 12,
       max_key_frame_interval: 240,
--- a/src/api/config/mod.rs
+++ b/src/api/config/mod.rs
@@ -37,6 +37,12 @@ pub enum InvalidConfig {
   /// The height is invalid.
   #[error("invalid height {0} (expected >= 16, <= 32767)")]
   InvalidHeight(usize),
+  /// Aspect ratio numerator is invalid.
+  #[error("invalid aspect ratio numerator {0} (expected > 0)")]
+  InvalidAspectRatioNum(usize),
+  /// Aspect ratio denominator is invalid.
+  #[error("invalid aspect ratio denominator {0} (expected > 0)")]
+  InvalidAspectRatioDen(usize),
   /// RDO lookahead frame count is invalid.
   #[error(
     "invalid rdo lookahead frames {actual} (expected <= {max} and >= {min})"
@@ -238,6 +244,17 @@ impl Config {
     }
     if config.height < 16 || config.height > u16::max_value() as usize {
       return Err(InvalidHeight(config.height));
+    }
+
+    if config.sample_aspect_ratio.num == 0 {
+      return Err(InvalidAspectRatioNum(
+        config.sample_aspect_ratio.num as usize,
+      ));
+    }
+    if config.sample_aspect_ratio.den == 0 {
+      return Err(InvalidAspectRatioDen(
+        config.sample_aspect_ratio.den as usize,
+      ));
     }
 
     if config.rdo_lookahead_frames > MAX_RDO_LOOKAHEAD_FRAMES
--- a/src/api/test.rs
+++ b/src/api/test.rs
@@ -1802,6 +1802,7 @@ fn log_q_exp_overflow() {
   let enc = EncoderConfig {
     width: 16,
     height: 16,
+    sample_aspect_ratio: Rational::new(1, 1),
     bit_depth: 8,
     chroma_sampling: ChromaSampling::Cs420,
     chroma_sample_position: ChromaSamplePosition::Unknown,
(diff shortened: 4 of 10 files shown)


src/api/config/encoder.rs
```diff
@@ -29,6 +29,8 @@ pub struct EncoderConfig {
   pub width: usize,
   /// Height of the frames in pixels.
   pub height: usize,
+  /// Display aspect ratio (for anamorphic video).
+  pub aspect_ratio: Rational,
```
dwbuiten, line 33: This needs to specify if it's a SAR or DAR.

rzumer: Clarified it.

rzumer:
Needs a revision to propagate SAR rather than DAR. Please hold...

Cargo.toml
```diff
@@ -59,7 +59,7 @@ bitstream-io = "0.8"
 cfg-if = "0.1"
 clap = { version = "2", optional = true, default-features = false }
 libc = "0.2"
-y4m = { version = "0.6", optional = true }
+y4m = { git = "https://github.com/image-rs/y4m.git", optional = true }
```
vibhoothiiaanand, line 62: Can we try to poke y4m-rs, people to cut a new release, not sure if it is in the schedule for them, it would be nice to have it from crates instead of git, 
Maybe @Kagami  could give us some insights over the release cycle of y4m-rs

Kagami: I've published y4m 0.7.0.

dwbuiten:
Thinking more about this, shouldn't it be per-frame?

rzumer:
I think we can leave that as future work.

rzumer:
I moved the parameters to `FrameInvariants` where they should be. I first thought you were suggesting to make aspect ratio variable per-frame via the API.

dwbuiten:
@rzumer I was indeed saying it should be per-frame in the API. Aspect ratio may change midstream, and should be supported either via frame properties, or via encoder-reconfig, I believe.

rzumer:
Then as mentioned earlier I prefer it to be left as future work considering how rare of a use case it is.

jamrial:
Maybe this should instead set frame_size_override_flag to 1 on inter frames and take the frame and render dimension values from one of the frames they reference in the ref_frame_idx[] array (The frame_size_with_refs() function defined in Section 5.9.7).

That way you'll save 4 bytes per frame header used to code the same render dimension values.

rzumer:
@jamrial That works as long as frame size is static. We can probably assume it is for now though.

jamrial:
Afailk, it can also work for non static frame sizes by ensuring a given inter frame references a frame with the exact same values it would otherwise write. If none of the seven potential reference frames share the exact same values, then they are explicitly written instead.

rzumer:
I see, I'll look into adding that but I also don't think it's a blocker for merging.

jamrial:
If you'd rather not bother looking at the values for each of the seven reference frame since for now the sizes are fixed, then it's a matter of setting frame_size_override_flag to 1 and implementing frame_size_with_refs() as a single bit found_ref to signal that the first reference frame should be used by the decoder to infer this frame's dimensions.

rzumer:
After looking into it I don't think this frame_size_override has the same effect as overriding the render_size. Seems to me that frame size is related to the number of luma samples, while render size signals players to rescale the frame when displaying it. Unless this part of the spec is inconsistently written and poorly defined, and the aomdec behavior matches what you're implying, I'd keep the current implementation.

jamrial:
No, it doesn't have the same effect. frame_size_override_flag means that either frame dimensions, render dimensions, or both, may be coded in the frame or should be inferred from a reference frame. For this specific case, frame dimensions and render dimensions will be constant for the entire video sequence, so setting both frame_size_override_flag and render_and_frame_size_different to 1 and then taking the first reference frame's dimensions is the same as signaling only render_and_frame_size_different and coding the render dimensions in the frame, except it doesn't use four bytes for it.

See the code below

```
diff --git a/src/header.rs b/src/header.rs
index 194c0510..7c66fe98 100644
--- a/src/header.rs
+++ b/src/header.rs
@@ -538,7 +538,14 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
       //self.write(frame_id_len, fi.current_frame_id);
     }

-    let mut frame_size_override_flag = false;
+    let render_and_frame_size_different = fi.render_width != fi.width as u32
+      || fi.render_height != fi.height as u32;
+
+    // This assumes frame dimensions are the same as sequence header dimensions,
+    // and that only render dimensions may be different but nonetheless constant
+    // for the entire video sequence
+    let mut frame_size_override_flag = render_and_frame_size_different && !fi.intra_only
+      && !fi.error_resilient;
     if fi.frame_type == FrameType::SWITCH {
       frame_size_override_flag = true;
     } else if fi.sequence.reduced_still_picture_hdr {
@@ -606,8 +613,6 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
       if fi.sequence.enable_superres {
         unimplemented!();
       }
-      let render_and_frame_size_different = fi.render_width != fi.width as u32
-        || fi.render_height != fi.height as u32;
       self.write_bit(render_and_frame_size_different)?;
       if render_and_frame_size_different {
         self.write(16, fi.render_width - 1);
@@ -640,7 +645,11 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
       }

       if !fi.error_resilient && frame_size_override_flag {
-        unimplemented!();
+        //TODO: check reference frames to compare frame dimensions
+        self.write_bit(true)?; // found_ref[0]
+        if fi.sequence.enable_superres {
+          unimplemented!();
+        }
       } else {
         if frame_size_override_flag {
           self.write_frame_size_override(fi);
@@ -648,9 +657,6 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
         if fi.sequence.enable_superres {
           unimplemented!();
         }
-        let render_and_frame_size_different = fi.render_width
-          != fi.width as u32
-          || fi.render_height != fi.height as u32;
         self.write_bit(render_and_frame_size_different)?;
         if render_and_frame_size_different {
           self.write(16, fi.render_width - 1);
```

rzumer:
I mean, I'm fine with this change but it seems like it should be separate from this PR.

jamrial:
Sure, i can send a PR for it after this one is merged.

Also, here's a complex sample where frame dimensions vary between frames (But render dimensions are constant), and uses the frame_size_with_refs() logic to avoid coding the values in the bitstream when a reference frame shares the same values, if you're curious.

https://code.videolan.org/videolan/dav1d/uploads/e41cf2423c6ae93e2a0a4fc9ae543730/resize.ivf

dwbuiten:
> Then as mentioned earlier I prefer it to be left as future work considering how rare of a use case it is, unless there is already a framework for setting parameters frame by frame (I can't see anything in the `api/config` module).

I think it should be doable with an encoder reconfig API, so doing it eoncig level should be OK, I suppose. Color properties are done this way too.

dwbuiten (approved):

PR #2506, opened 2020-08-21
Map chroma transform types for inter

### LINKED ISSUE #2122: Tx split for inter mode only works when reduce_tx_set is enabled
Investigating this and the most possible reason seems the tx_type for chroma is not derived correctly when tx-split happens in inter mode partition for certain lager partition sizes.

    let uv_tx_type =
      if partition_has_coeff { tx_type } else { TxType::DCT_DCT }; // if inter mode, uv_tx_type == tx_type

barrbrain:
Note that https://github.com/xiph/rav1e/commit/22bf11b7fc5ec47c2f31716fc916f8c3a5c11817 disabled RDO of transform type for inter blocks. To verify this change I reverted that commit, forced `reduced_tx_set_preset()` to `false` and forced `enable_inter_tx_split_preset()` to `true`. i.e.: https://github.com/barrbrain/rav1e/commit/ab278ccc2ddcc0bf1d2959ec28e1d786952f07cc
The only test failure I was able to induce was an assertion in `test_encode_decode::chroma_sampling_444_dav1d`.

This implementation uses the same simplification as in dav1d to remove the indirection via `TxSet`. Fixes #2122.

### DIFF
--- a/src/transform/mod.rs
+++ b/src/transform/mod.rs
@@ -71,6 +71,29 @@ pub enum TxType {
   H_FLIPADST = 15,
 }
 
+impl TxType {
+  /// Compute transform type for inter chroma.
+  ///
+  /// <https://aomediacodec.github.io/av1-spec/#compute-transform-type-function>
+  #[inline]
+  pub fn uv_inter(self: Self, uv_tx_size: TxSize) -> Self {
+    use TxType::*;
+    if uv_tx_size.sqr_up() == TX_32X32 {
+      match self {
+        IDTX => IDTX,
+        _ => DCT_DCT,
+      }
+    } else if uv_tx_size.sqr() == TX_16X16 {
+      match self {
+        V_ADST | H_ADST | V_FLIPADST | H_FLIPADST => DCT_DCT,
+        _ => self,
+      }
+    } else {
+      self
+    }
+  }
+}
+
 /// Transform Size
 #[derive(Copy, Clone, Debug, PartialEq, PartialOrd, Eq, Ord)]
 pub enum TxSize {
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -2230,7 +2230,11 @@ pub fn write_tx_tree<T: Pixel>(
   bw_uv /= uv_tx_size.width_mi();
   bh_uv /= uv_tx_size.height_mi();
 
-  let uv_tx_type = if partition_has_coeff { tx_type } else { TxType::DCT_DCT }; // if inter mode, uv_tx_type == tx_type
+  let uv_tx_type = if partition_has_coeff {
+    tx_type.uv_inter(uv_tx_size)
+  } else {
+    TxType::DCT_DCT
+  };
 
   for p in 1..3 {
     ts.qc.update(


src/transform/mod.rs
```diff
@@ -71,6 +71,29 @@ pub enum TxType {
   H_FLIPADST = 15,
 }
```
barrbrain, line 73: That this function is fully covered by our basic encode-decode tests suggests either:
- Coverage is misleading
- The content in our tests is resulting in RDO decisions that hide potential desyncs

https://coveralls.io/builds/33004278/source?filename=src/transform/mod.rs#L74

tdaede: Do you mean coverage of the uv_tx_size branch, or of the match arms? I'd be a bit suspicious of the latter being accurate.

src/transform/mod.rs
```diff
@@ -71,6 +71,29 @@ pub enum TxType {
...
 
+impl TxType {
+  /// Compute transform type for inter chroma.
+  ///
+  /// <https://aomediacodec.github.io/av1-spec/#compute-transform-type-function>
+  #[inline]
+  pub fn uv_inter(self: Self, uv_tx_size: TxSize) -> Self {
```
ycho, line 79: Thanks for catching the reason and fixes this!
I indeed missed a step to check the tx_size for each plane as stated in spec. (So, more tx_type is available for uv channels, since its tx_size is smaller after chroma sampling except 444)
I see that there are three cases for uv_tx_size:  1) 32x32, 2) 16x16, 3) all others.
I think 3) is meant for uv_tx_size < 16x16.
What would happen if uv_tx_size = 64x64? This happens for chroma sampling 444, right?
I guess this might explain the failure with your 444 test, if it still fails.

ycho: As you referred, https://aomediacodec.github.io/av1-spec/#compute-transform-type-function has
"txSzSqrUp > TX_32X32", case which returns DCT_DCT.

src/transform/mod.rs
```diff
@@ -71,6 +71,29 @@ pub enum TxType {
...
+        IDTX => IDTX,
+        _ => DCT_DCT,
+      }
+    } else if uv_tx_size.sqr() == TX_16X16 {
+      match self {
+        V_ADST | H_ADST | V_FLIPADST | H_FLIPADST => DCT_DCT,
+        _ => self,
+      }
+    } else {
+      self
```
ycho, line 92: I mean, uv_tx_size is 64x64 (when input is chroma subsampling 444) falls at here, which is wrong, correct? Becase only dct_dct is allowed for 64x64.

barrbrain: Because only DCT is allowed for 64x64, both for Y and UV, `self` is correct.

ycho: That is right, but the tx_loop in rdo_tx_type... ( ) only iterate for dct_dct? If so, this is fine.

barrbrain: Implicitly, yes. We filter the candidate `TxType` values by membership of the `TxSet`.
https://github.com/xiph/rav1e/blob/p20200825/src/rdo.rs#L1682-L1686
Conversely, if we didn't do this then we would be testing invalid transform types for the luma plane.

ycho: Got it. I double checked that, inside pub fn rdo_tx_type_decision( ),
it checks
    if av1_tx_used[tx_set as usize][tx_type as usize] == 0 {
      continue;
    }

so, the tx_type loop does not test other than dct_dct for 64x64.

src/transform/mod.rs
```diff
@@ -71,6 +71,29 @@ pub enum TxType {
...
+  /// Compute transform type for inter chroma.
+  ///
+  /// <https://aomediacodec.github.io/av1-spec/#compute-transform-type-function>
+  #[inline]
+  pub fn uv_inter(self: Self, uv_tx_size: TxSize) -> Self {
+    use TxType::*;
+    if uv_tx_size.sqr_up() == TX_32X32 {
+      match self {
+        IDTX => IDTX,
+        _ => DCT_DCT,
```
barrbrain, line 84: The particular set of cases that I encountered in `test_encode_decode::chroma_sampling_444_dav1d` was:
```
uv_tx_size.sqr_up() == TX_32X32 && tx_type != IDTX && tx_type != DCT_DCT
```
e.g.:
```
uv_tx_size == TX_32X32 && tx_type == DCT_ADST
```

PR #2511, opened 2020-08-23
Swap the vendored ivf crate for av-ivf

shssoichiro:


dwbuiten:
Do *really* need to make like 20 lines of IVF writing code a dependency? Really?

lu-zero:
The question is if we want to add mkv(rawvideo) input and mkv(av1) output later. I'd do that in other applications and keep this one as simple as it is.

kodabb:
I know this is just a draft, but I'd recommend a brief explanation in the pr description

dwbuiten:
I don't thing either of those belong in rav1e, but in uses of rav1e's API.

For better or for worse, IVF is basically AV1's "elementary stream" format for all intents and purposes, and it should be in ra1ve.

The other part I object to is the big intrusive change to use this new rational type from the dependency all over the place. We should *not* be doing, especially in our API. We should own our API types, and the fundemental types we use inside rav1e, like rationals.

tdaede:
My original concern was using rust-av types in the public API - basically because rust-av wraps rav1e, I wanted to keep the dependency one-way to make updating either easier. For the binary I don't care as much either way.

PR #2519, opened 2020-08-29
Update nasm-rs

kornelski:
This fixes version parsing problem that nasm-rs 0.1.8 had


kornelski:
The build failure to install nasm look suspicious, but couldn't possibly be related to nasm-rs. It was probably a network error?

barrbrain:
There was a CI issue fetching the nasm package. After rebase, the new CI config should be effective.

lu-zero (approved):
Thank you!

PR #2521, opened 2020-08-30
CI: Update nasm to 2.15.04

barrbrain:
Note that homebrew has moved on to 2.14.05 already.


barrbrain (commented):
This `nasm` release might be buggy. See https://bugzilla.nasm.us/show_bug.cgi?id=3392715
We don't tickle this bug for now because we don't compile `x86_32` assembly.

PR #2522, opened 2020-08-30
Test fix

kornelski:
`cargo test --all` used to fail on this test. I've also added `--all` to travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ jobs:
       rust: stable
       arch: arm64
       script:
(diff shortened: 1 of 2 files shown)


lu-zero (approved):
Thank you, @Luni-4 maybe we could add it to the github actions as well.

Luni-4:
> Thank you, @Luni-4 maybe we could add it to the github actions as well.

Sure! I think it would be better to use `--workspace` option though because the `--all` one is deprecated.

PR #2523, opened 2020-08-30
Interpret 0:0 SAR as 1:1

rzumer:
Missing change for #2505, which causes panics when trying to encode clips specifying an SAR of 0:0.

### DIFF
--- a/src/api/config/mod.rs
+++ b/src/api/config/mod.rs
@@ -246,12 +246,16 @@ impl Config {
       return Err(InvalidHeight(config.height));
     }
 
-    if config.sample_aspect_ratio.num == 0 {
+    if config.sample_aspect_ratio.num == 0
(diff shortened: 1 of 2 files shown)


lu-zero:
I wonder if wouldn't be simpler to fix the `0 / 0` input to `1 / 1` when the context is created.

tmatth:
> I wonder if wouldn't be simpler to fix the `0 / 0` input to `1 / 1` when the context is created.

Merged as is since existing master was broken (thanks for the heads up @KyleSiefring), but such a patch would be fine too.

rzumer:
In my opinion it's the Y4M files that are broken. An SAR of 0:0 simply does not make sense. I would agree with correcting the SAR at the Y4M decoder level and reverting the API-level changes.

PR #2529, opened 2020-09-04
Change the minimum frame size from 16x16 to 1x1

### LINKED ISSUE #2518: Support inputs smaller than 16 pixels
Currently rav1e refuses to encode inputs smaller than 16px in width or height. This is problematic for AVIF support, because tiny images like that are sometimes used.

If it's not too hard, could you remove the limitation?

ycho:
Resolve #2518 

~~Note that this change violates the constraint of AV1 spec:
"FrameWidth is greater than or equal to 16"
"FrameHeight is greater than or equal to 16"
in https://aomediacodec.github.io/av1-spec/#levels~~

### DIFF
--- a/src/api/config/mod.rs
+++ b/src/api/config/mod.rs
@@ -233,10 +233,16 @@ impl Config {
 
     let config = &self.enc;
 
-    if config.width < 16 || config.width > u16::max_value() as usize {
+    if (config.still_picture && config.width < 1)
+      || (!config.still_picture && config.width < 16)
+      || config.width > u16::max_value() as usize
+    {
       return Err(InvalidWidth(config.width));
     }
-    if config.height < 16 || config.height > u16::max_value() as usize {
+    if (config.still_picture && config.height < 1)
+      || (!config.still_picture && config.height < 16)
+      || config.height > u16::max_value() as usize
+    {
       return Err(InvalidHeight(config.height));
     }
 


ycho:
I wanted to provide this change to help rav1e encoder fans who want to encode 15x15 or smaller size image.
Probably we want to enable this change as run time option, not a default condition.

ycho:
Changed so that it only allow for still picture mode.
"--still-picture" should be used.

vibhoothiiaanand (approved):
Seems good to me
Right with this change, we are not violating spec conditions for AV1.
AVIF mandates keeping still_picture flag if it is AVIF image
"The AV1 Image Item should have its still_picture flag set to 1."
https://rawcdn.githack.com/AOMediaCodec/av1-avif/67a92add6cd642a8863e386fa4db87954a6735d1/index.html#av1-image-sequence
As per Spec of AVIF, there is no hard limit on frame size, so this should cover case in a nicer way.

rzumer:
1. AVIF streams are standard AV1, so just because a bitstream conformance requirement is not mentioned in the AVIF specification does not mean it does not apply.
2. AVIF streams "should" use the still picture mode but it is not mandated. The still picture modes does not define whether or not an AV1 stream is "AVIF" or not.

I don't have a strong opinion anyway on violating the standard, but just pointing out that it is still a violation.

lu-zero:
sub-16pixels images cannot be just achieved at container level anyway?

ycho:
>     1. AVIF streams are standard AV1, so just because a bitstream conformance requirement is not mentioned in the AVIF specification does not mean it does not apply.
> 
>     2. AVIF streams "should" use the still picture mode but it is not mandated. The still picture modes does not define whether or not an AV1 stream is "AVIF" or not.
> 
> 
> I don't have a strong opinion anyway on violating the standard, but just pointing out that it is still a violation.

I think it is a valid concern.

I also read this from the av1 spec,
https://aomediacodec.github.io/av1-spec/#general-sequence-header-obu-semantics,
"still_picture equal to 1 specifies that the coded video sequence contains only one coded frame. still_picture equal to 0 specifies that the coded video sequence contains one or more coded frames."

Hence, until further justification or revision of av1 spec on the topic is announced, I would hold this PR and thus I am closing it at the moment.

rzumer:
> "still_picture equal to 1 specifies that the coded video sequence contains only one coded frame. still_picture equal to 0 specifies that the coded video sequence contains one or more coded frames."

That part I did not know about. I don't think we enforce it with `still_picture` enabled right now.

lu-zero:
With Open Partition doesn't that produce the the correct bitstream?

lu-zero (approved):
It seems fine to me.

vibhoothiiaanand:
Can we land this ?

ycho:
> Can we land this ?

Not sure! :)

rzumer (approved):
This should be fine and legal as long as we code an unconstrained level, which we currently always do.

PR #2532, opened 2020-09-08
Filter input Y4M SAR

rzumer:
This fixes the minimum SAR to 1:1 so that
Y4M files signaling 0:0 SAR can be processed,
without allowing illegal SAR values at the
API level or handling them from the encoder.

In practice this slightly simplifies the SAR handling in the `encoder` module, as well as validation in the API `config` module, while making the Y4M `decoder` module responsible for filtering the input SAR. A validation error will be returned if the SAR is set to 0:0 at the API level.

As suggested in the comments of #2523.

### DIFF
--- a/src/api/config/mod.rs
+++ b/src/api/config/mod.rs
@@ -257,16 +257,12 @@ impl Config {
       return Err(InvalidHeight(config.height));
     }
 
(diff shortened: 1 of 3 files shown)


lu-zero (approved):
Thank you!

PR #2489, opened 2020-08-10
Update paste dependency to 1.0, arbitrary to 0.4, console to 0.12 and…

sdroege:
… semver to 0.10

PR #2535, opened 2020-09-10
Implement Structure-Aware Fuzzing

barrbrain:
This is necessary after #2489, which replaced usage of `RingBuffer` by `Unstructured`. We unwrap an `Error` when the fuzz input is small. We should rather implement `Arbitrary` and raise the `Error` there.


lu-zero (approved):
Looks fine to me, thank you a lot!

PR #2537, opened 2020-09-11
Fix and add test case for 4:2:2 fuzzing failure

barrbrain:
```
running 1 test
Overrun in OBU bit buffer into next OBU
test test_encode_decode::fuzz_422_77x17_tile_cols_2_dav1d ... FAILED

failures:

---- test_encode_decode::fuzz_422_77x17_tile_cols_2_dav1d stdout ----
thread 'test_encode_decode::fuzz_422_77x17_tile_cols_2_dav1d' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
<1
>0

', src/test_encode_decode/mod.rs:144:7
```

### DIFF
--- a/src/tiling/tiler.rs
+++ b/src/tiling/tiler.rs
@@ -110,6 +110,12 @@ impl TilingInfo {
       tile_width_sb_pre
     };
 
+    let cols = (frame_width_sb + tile_width_sb - 1) / tile_width_sb;
+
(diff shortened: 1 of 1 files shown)


lu-zero (approved):
Looks good to me, thank you :)

barrbrain (commented):
> [00:35] \<derf\> Yeah, I mean, I am okay with this patch landing as-is.

PR #2548, opened 2020-09-17
CI: Build fuzzers with stable rustc

barrbrain:
Removes our one remaining dependency on a pinned nightly toolchain.


lu-zero (approved):

lu-zero (approved):
Thank you for the documentation update on top :)

PR #2549, opened 2020-09-17
Fix still picture frame limiting in the API

rzumer:
The limit was overridden at the wrong place. The `simple_encoding` example hangs trying to queue up more than one frame if `still_picture` is set in the encoder configuration.

This returns `Err(EncoderStatus::EnoughData)` when one frame has already been processed in still picture mode, which allows `simple_encoding` to run through completely, while any frame beyond the first one is simply ignored.


lu-zero (approved):
It seems good, thank you.

PR #2550, opened 2020-09-19
fuzzer: Increase coverage of encode and encode_decode

barrbrain:
Increase the value ranges for fields of DecodeTestParameters. Ensure the still_picture limit is respected.
Use a single thread for encode_decode - This helps to avoid over-allocating threads when running with many cores.

### DIFF
--- a/src/test_encode_decode/mod.rs
+++ b/src/test_encode_decode/mod.rs
@@ -191,7 +191,9 @@ fn setup_encoder<T: Pixel>(
   enc.tile_rows = 1 << tile_rows_log2;
   enc.still_picture = still_picture;
 
-  let cfg = Config::new().with_encoder_config(enc);
+  let threads = if cfg!(fuzzing) { 1 } else { 2 };
+
+  let cfg = Config::new().with_encoder_config(enc).with_threads(threads);
 
   cfg.new_context().unwrap()
 }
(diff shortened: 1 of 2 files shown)


src/test_encode_decode/mod.rs
```diff
@@ -191,7 +191,7 @@ fn setup_encoder<T: Pixel>(
   enc.tile_rows = 1 << tile_rows_log2;
   enc.still_picture = still_picture;
 
-  let cfg = Config::new().with_encoder_config(enc);
+  let cfg = Config::new().with_encoder_config(enc).with_threads(1);
```
lu-zero, line 194: would be better to make it 2 or such for now. I'll lazy-allocate a shared pool and use it for all the tests in the future.

barrbrain: Testing and fuzzing probably require different solutions since the first is single-process and the second can be multi-process. Maybe a good compromise is a simple `if cfg!(fuzzing) { 1 } else { 2 }`.

barrbrain:
The following fields are not yet fuzzed:
```rust
pub struct EncoderConfig {
  ...
  pub sample_aspect_ratio: Rational,
  pub color_description: Option<ColorDescription>,
  pub mastering_display: Option<MasteringDisplay>,
  pub content_light: Option<ContentLight>,
  pub switch_frame_interval: u64,
  pub enable_timing_info: bool,
}
```

When `sample_aspect_ratio` is sufficiently large it can cause the calculated `render_width` to be out of range.
e.g. `width: 7680, sample_aspect_ratio: Rational::new(9, 1),` implies `render_width = 69120` which exceeds the 16-bit range.

Also, `enable_timing_info` is mutually exclusive with `still_picture`.

lu-zero (approved):
Looks good :)

PR #2551, opened 2020-09-20
Avoid use of Plane::p in Plane::downsample

shssoichiro:
Plane::p accesses pixels in a fashion that the compiler
cannot assume to be linear. This is harmful to both
compiler optimization and cache linearity.

Although we cannot completely eliminate this problem
since two rows must be accessed at a time for downsampling,
we can alleviate it by accessing the data slice directly.

Improves performance of av-scenechange by 15%.
Did not benchmark rav1e but presumably helps by 1-2% there also,
since this code is shared between both applications.

### DIFF
--- a/v_frame/src/plane.rs
+++ b/v_frame/src/plane.rs
@@ -424,7 +424,6 @@ impl<T: Pixel> Plane<T> {
     }
   }
 
-  #[allow(clippy::needless_range_loop)]
   pub fn downsampled(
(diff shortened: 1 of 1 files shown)


shssoichiro:
Would be great if we could do a point release of v_frame with this patch, since it's such a big performance boost for av-scenechange.

lu-zero (approved):
Looks good and I'm fine in making the point release for v_frame, thank you a lot for looking into it!

PR #2556, opened 2020-09-22
Write Frame OBUs instead of a pair of Frame Header and Tile Group OBUs

jamrial:
For non show_existing_frame frames, and seeing rav1e currently doens't offer
the option to write more than one Tile Group per frame, it's more
efficient to use Frame OBUs instead of a pair of Frame Header and Tile Group
OBUs.

This saves a few bytes per frame by removing the overhead from the superfluous
OBU header.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1007,65 +1007,21 @@ pub fn write_temporal_delimiter(packet: &mut dyn io::Write) -> io::Result<()> {
   Ok(())
 }
 
-fn write_obus<T: Pixel>(
-  packet: &mut dyn io::Write, fi: &FrameInvariants<T>, fs: &FrameState<T>,
-  inter_cfg: &InterConfig,
+fn write_key_frame_obus<T: Pixel>(
+  packet: &mut dyn io::Write, fi: &FrameInvariants<T>, obu_extension: u32,
 ) -> io::Result<()> {
-  let obu_extension = 0 as u32;
-
   let mut buf1 = Vec::new();
-
-  // write sequence header obu if KEY_FRAME, preceded by 4-byte size
-  if fi.frame_type == FrameType::KEY {
-    let mut buf2 = Vec::new();
-    {
-      let mut bw2 = BitWriter::endian(&mut buf2, BigEndian);
-      bw2.write_sequence_header_obu(fi)?;
-      bw2.write_bit(true)?; // trailing bit
-      bw2.byte_align()?;
-    }
-
-    {
-      let mut bw1 = BitWriter::endian(&mut buf1, BigEndian);
-      bw1.write_obu_header(ObuType::OBU_SEQUENCE_HEADER, obu_extension)?;
-    }
-    packet.write_all(&buf1).unwrap();
-    buf1.clear();
-
-    {
-      let mut bw1 = BitWriter::endian(&mut buf1, BigEndian);
-      bw1.write_uleb128(buf2.len() as u64)?;
-    }
-    packet.write_all(&buf1).unwrap();
-    buf1.clear();
-
-    packet.write_all(&buf2).unwrap();
-    buf2.clear();
-
-    if fi.sequence.content_light.is_some() {
-      let mut bw1 = BitWriter::endian(&mut buf1, BigEndian);
-      bw1.write_metadata_obu(ObuMetaType::OBU_META_HDR_CLL, fi.sequence)?;
-      packet.write_all(&buf1).unwrap();
-      buf1.clear();
-    }
-
-    if fi.sequence.mastering_display.is_some() {
-      let mut bw1 = BitWriter::endian(&mut buf1, BigEndian);
-      bw1.write_metadata_obu(ObuMetaType::OBU_META_HDR_MDCV, fi.sequence)?;
-      packet.write_all(&buf1).unwrap();
-      buf1.clear();
-    }
-  }
-
   let mut buf2 = Vec::new();
   {
     let mut bw2 = BitWriter::endian(&mut buf2, BigEndian);
-    bw2.write_frame_header_obu(fi, fs, inter_cfg)?;
+    bw2.write_sequence_header_obu(fi)?;
+    bw2.write_bit(true)?; // trailing bit
+    bw2.byte_align()?;
   }
 
   {
     let mut bw1 = BitWriter::endian(&mut buf1, BigEndian);
-    bw1.write_obu_header(ObuType::OBU_FRAME_HEADER, obu_extension)?;
+    bw1.write_obu_header(ObuType::OBU_SEQUENCE_HEADER, obu_extension)?;
   }
   packet.write_all(&buf1).unwrap();
   buf1.clear();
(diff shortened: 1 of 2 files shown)


jamrial:
Could someone take a look at this?

src/encoder.rs
```diff
@@ -3449,9 +3419,37 @@ pub fn encode_show_existing_frame<T: Pixel>(
   fi: &FrameInvariants<T>, fs: &mut FrameState<T>, inter_cfg: &InterConfig,
 ) -> Vec<u8> {
   debug_assert!(fi.show_existing_frame);
+  let obu_extension = 0 as u32;
```
tmatth, line 3422: Any reason for not using `0u32` ?

src/encoder.rs
```diff
@@ -3484,6 +3482,8 @@ pub fn encode_frame<T: Pixel>(
 ) -> Vec<u8> {
   debug_assert!(!fi.show_existing_frame);
   debug_assert!(!fi.invalid);
+  let obu_extension = 0 as u32;
```
tmatth, line 3485: idem

tmatth (commented):
LGTM, just some minor style questions, but someone more familiar with OBUs should take a look (@rzumer ?)

src/encoder.rs
jamrial: Not familiar enough with Rust, so i wasn't aware that's an option

lu-zero: It can be `0`, the type should inferred w/out external help.

src/encoder.rs
```diff
@@ -3449,9 +3419,37 @@ pub fn encode_show_existing_frame<T: Pixel>(
...
   debug_assert!(fi.show_existing_frame);
+  let obu_extension = 0;
+
   let mut packet = Vec::new();
 
-  write_obus(&mut packet, fi, fs, inter_cfg).unwrap();
+  if fi.frame_type == FrameType::KEY {
+    write_key_frame_obus(&mut packet, fi, obu_extension).unwrap();
+  }
+  let mut buf1 = Vec::new();
```
rzumer, line 3429: Why are those two defined and then later redefined in unconditional inner blocks?

src/encoder.rs
```diff
@@ -3449,9 +3419,37 @@ pub fn encode_show_existing_frame<T: Pixel>(
   fi: &FrameInvariants<T>, fs: &mut FrameState<T>, inter_cfg: &InterConfig,
 ) -> Vec<u8> {
   debug_assert!(fi.show_existing_frame);
+  let obu_extension = 0;
```
rzumer, line 3422: What is this constant's purpose?

src/header.rs
```diff
@@ -843,7 +843,6 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
     if fi.large_scale_tile {
       unimplemented!();
     }
-    self.write_bit(true)?; // trailing bit
```
rzumer, line 846: Why is this removed?

src/encoder.rs
```diff
@@ -3492,22 +3492,33 @@ pub fn encode_frame<T: Pixel>(
...
-    bw1.write_obu_header(ObuType::OBU_TILE_GROUP, 0).unwrap();
+    bw1.write_obu_header(ObuType::OBU_FRAME, obu_extension).unwrap();
   }
   packet.write_all(&buf1).unwrap();
   buf1.clear();
 
   {
     let mut bw1 = BitWriter::endian(&mut buf1, BigEndian);
-    bw1.write_uleb128(tile_group.len() as u64).unwrap();
+    bw1.write_uleb128((buf2.len() + tile_group.len()) as u64).unwrap();
```
rzumer, line 3514: I can't tell what's going on here. It is writing data from `buf1` but for the length of `buf2`. Why?

src/encoder.rs
```diff
@@ -3492,22 +3492,33 @@ pub fn encode_frame<T: Pixel>(
   }
   let tile_group = encode_tile_group(fi, fs, inter_cfg);
 
-  write_obus(&mut packet, fi, fs, inter_cfg).unwrap();
+  if fi.frame_type == FrameType::KEY {
+    write_key_frame_obus(&mut packet, fi, obu_extension).unwrap();
+  }
```
rzumer, line 3497: Would like another line break after this to separate steps and make reading this easier.

src/encoder.rs
```diff
@@ -3449,9 +3419,37 @@ pub fn encode_show_existing_frame<T: Pixel>(
...
+
+  {
+    let mut bw1 = BitWriter::endian(&mut buf1, BigEndian);
+    bw1.write_obu_header(ObuType::OBU_FRAME_HEADER, obu_extension).unwrap();
+  }
+  packet.write_all(&buf1).unwrap();
+  buf1.clear();
+
+  {
+    let mut bw1 = BitWriter::endian(&mut buf1, BigEndian);
```
rzumer, line 3444: Here there are two consecutive uses of `bw1`, they can probably be merged.

src/encoder.rs
rzumer, line 3429: Not really in scope for this PR since it matches the previous state of the code, but I'd like some clarity about why we have two buffers and writers, through renaming the variables or comments. It's not clear at a glance why this code is written the way it is.

src/encoder.rs
jamrial: I moved it around when i refactored these functions. It looked superfluous to me as well, but i figured leaving it alone was the better idea, in case verbosity was preferred.

src/header.rs
jamrial: Because it's no longer needed. That bit is the trailing bit at the end of a Frame Header to signal the end of the entire OBU, and we don't write those anymore. Frame OBUs are composed by a Frame Header followed immediately by a Tile Group merged into a single OBU.

See section 5.10 in the spec.

src/encoder.rs
rzumer: I see, I missed its definition in the previous version of the code. I don't have a strong opinion on how to handle it.

src/encoder.rs
jamrial: It's the same as before this patch, and i can't seem to find a way to merge them.

rzumer: Okay, not important for this patch then, I will look into it on my own.

src/encoder.rs
jamrial: Done

rzumer:
AWCY run results: http://35.223.44.203/?job=master%402020-11-18T21%3A18%3A35.622Z&job=obu-optimization%402020-11-18T21%3A29%3A25.667Z

jamrial:
I was not expecting changes in encoding metrics. This simply saves a few overhead header bytes per frame.

rzumer:
I can't say why the metrics might have changed, and the analyzer is not hooked up to this AWCY instance, but I started a new run prior to the last commit on master which is not merged into your branch, in case it caused unexpected coding changes: http://35.223.44.203/?job=master-minus-one%402020-11-18T22%3A27%3A40.206Z

rzumer (approved):

rzumer:
Just to follow up on the initial results, it turned out the "master" reference did not point to the origin repository as expected.

PR #2557, opened 2020-09-23
always set frame_size_override_flag when creating new inter frames

jamrial:
This prevents the value from the previous frame cloned at the beginning of new_inter_frame() from being used if none of the existing checks succeeds.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -819,19 +819,20 @@ impl<T: Pixel> FrameInvariants<T> {
     fi.error_resilient =
       if fi.frame_type == FrameType::SWITCH { true } else { error_resilient };
 
-    // force frame_size_with_refs() code path if render size != frame size
-    if fi.frame_type == FrameType::INTER
+    fi.frame_size_override_flag = if fi.frame_type == FrameType::SWITCH {
+      true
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -819,18 +819,20 @@ impl<T: Pixel> FrameInvariants<T> {
...
 
-    // force frame_size_with_refs() code path if render size != frame size
-    if fi.frame_type == FrameType::INTER
-      && !fi.error_resilient
-      && fi.render_and_frame_size_different
-    {
-      fi.frame_size_override_flag = true;
-    }
-
     if fi.frame_type == FrameType::SWITCH {
```
barrbrain, line 822: We may make the intent explicit by additionally rewriting to the form:
```rust
fi.frame_size_override_flag = if fi.frame_type == FrameType::SWITCH {
...
};
```

jamrial:
This has been marinating for a while now. Can it be merged?

barrbrain (approved):
This change makes it easier to reason about the value of `frame_size_override_flag`.
I have been mostly offline since my initial review comments. Thank you for the reminder.

PR #2560, opened 2020-09-29
Detect more invalid configurations

### LINKED ISSUE #2559: Encoder configuration admits invalid combinations
A few assertions were encountered while expanding fuzzing coverage. When `sample_aspect_ratio` is sufficiently large it can cause the calculated `render_width` to be out of range. e.g. `width: 7680, sample_aspect_ratio: Rational::new(9, 1),` implies `render_width = 69120` which exceeds the 16-bit range. Also, `enable_timing_info` is mutually exclusive with `still_picture`. These are not identified in `Config::validate()` and trip assertions when writing headers.

rzumer:
This closes #2559.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -599,12 +599,14 @@ impl<T: Pixel> FrameInvariants<T> {
 
     let sar = config.sample_aspect_ratio.as_f64();
     let (render_width, render_height) = if sar > 1.0 {
-      ((width as f64 * sar).round() as u32, height as u32)
+      (((width as f64 * sar).round() as u32), height as u32)
     } else {
-      (width as u32, (height as f64 / sar).round() as u32)
+      (width as u32, ((height as f64 / sar).round() as u32))
     };
     let render_and_frame_size_different =
       render_width != width as u32 || render_height != height as u32;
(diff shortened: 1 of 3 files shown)


src/encoder.rs
```diff
@@ -599,12 +599,22 @@ impl<T: Pixel> FrameInvariants<T> {
...
     let (render_width, render_height) = if sar > 1.0 {
-      ((width as f64 * sar).round() as u32, height as u32)
+      (((width as f64 * sar).round() as u32), height as u32)
     } else {
-      (width as u32, (height as f64 / sar).round() as u32)
+      (width as u32, ((height as f64 / sar).round() as u32))
     };
     let render_and_frame_size_different =
       render_width != width as u32 || render_height != height as u32;
+    assert!(render_width > 0, "rendered frame width is 0");
```
lu-zero, line 608: might be better to move all of this during validation/context-allocation.

rzumer: Render width and height is per-frame.

rzumer: Done.

lu-zero (approved):
Looks nice, thank you!

PR #2564, opened 2020-10-03
Add warning if chromaticity coords are out of range

### LINKED ISSUE #2555: rav1e silently clips out of range values passed to mastering-display
`rav1e input.y4m --mastering-display="G(1.1,1.1)B(1.1,1.1)R(1.1,1.1)WP(1.1,1.1)L(10,1)" -o output.ivf`

Passing out of range values like 1.1 in the above example results in the Metadata OBU containing 65535 as value for all primary_chromaticity_x and primary_chromaticity_y fields, which is the max value that can be written and the equivalent of 0.9999847412109375 in 0.16 fixed point format, as per the spec.

rav1e should warn or abort if the values passed for R, G, B and WP are outside of the 0.0-0.9999847412109375 range (or maybe just 0.0-1.0).

matthew-jack:
Hello! This is my first PR here and thought this would be a good issue.

Fixes #2555

Ran cargo fmt && cargo clippy && cargo test

Edits and feedback are welcome.
Thanks

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -604,6 +604,26 @@ fn parse_config(matches: &ArgMatches<'_>) -> Result<EncoderConfig, CliError> {
         f64
       )
       .expect("Cannot parse the mastering display option");
+
+    /* AV1 spec sec. 6.7.4 "Metadata high dynamic range mastering display color volume semantics"
+     * specifies chromaticity coords as 0.16 fixed-point numbers, which have a max float value
+     * of 0.9999847412109375 (rounding to 1).
+     */
+    let chromaticity_range = 0.0..=1.0;
+    if !chromaticity_range.contains(&g_x)
+      || !chromaticity_range.contains(&g_y)
+      || !chromaticity_range.contains(&b_x)
+      || !chromaticity_range.contains(&b_y)
+      || !chromaticity_range.contains(&r_x)
+      || !chromaticity_range.contains(&r_y)
+      || !chromaticity_range.contains(&wp_x)
+      || !chromaticity_range.contains(&wp_y)
+    {
+      warn!(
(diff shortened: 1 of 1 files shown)


lu-zero:
Thank you! if you could put the commit message in multiple lines github would pick up the issue, I let @rzumer give his opinion on the patch.

src/bin/common.rs
```diff
@@ -604,6 +604,22 @@ fn parse_config(matches: &ArgMatches<'_>) -> Result<EncoderConfig, CliError> {
         f64
       )
       .expect("Cannot parse the mastering display option");
+
```
vibhoothiiaanand, line 607: Maybe add a reference to spec,
I think it is "6.7.4. Metadata high dynamic range mastering display color volume semantics"

Where it defines the details for chromaticity X and Y

vibhoothiiaanand (commented):
Thanks for the patchset, 
I have one suggestion,
Would be nice if you could make the commit message shorter,
From 
"Add runtime warning if chromaticity coordinates are out of bound and will be trimmed"

to something like 
"bin: Warn when passing out of range Chroma Values"

And in the description, bit more explanation would be nice along with issue number :)

matthew-jack:
Forgot the note about the spec in the warning! Adding that now.

src/bin/common.rs
matthew-jack: I added a reference in a comment and in the warning but not sure if you want it both places.

lu-zero:
@rzumer do you have opinions?

rzumer (approved):
It looks good to me.

vibhoothiiaanand (approved):
+1 from me too, 

Thank you for the contribution

Issue #2516, opened 2020-08-26
Copyright headers not up to date


KyleSiefring:

rzumer:
Keeping them up to date is a constant battle to the point I wonder if they're even needed.

KyleSiefring:
I think debian (???) requires this. It should be possible to set up a test for this,

PR #2565, opened 2020-10-07
Update from copyright headers

darakian:
Simple update to resolve
https://github.com/xiph/rav1e/issues/2516

Generated with
```
sed -i 's/Copyright (c) 2019,/Copyright (c) 2020,/' $(grep 'Copyright \(c\) 2019,' -l)
```

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018, The rav1e contributors. All rights reserved
+// Copyright (c) 2017-2020, The rav1e contributors. All rights reserved
 //
 // This source code is subject to the terms of the BSD 2 Clause License and
 // the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
--- a/benches/dist.rs
+++ b/benches/dist.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018, The rav1e contributors. All rights reserved
+// Copyright (c) 2017-2020, The rav1e contributors. All rights reserved
 //
 // This source code is subject to the terms of the BSD 2 Clause License and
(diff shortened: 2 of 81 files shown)


vibhoothiiaanand (commented):
So, in my opinion, we should update headers of files which was having 201X and if they are modified in 2020, to update with 201X-2020 like that, and also see if we are missing any headers on any files.

darakian:
Ok. Easy enough to change :)

I'll modify these to read as `Copyright (c) 2019-2020, ...`. Should I update more than simply the 2019 headers? I assumed that if they were older then there was a reason.

vibhoothiiaanand:
> Ok. Easy enough to change :)
> 
> I'll modify these to read as `Copyright (c) 2019-2020, ...`. Should I update more than simply the 2019 headers? I assumed that if they were older then there was a reason.

Need to see properly, some/many of 2019 files may not be touched in 2020 too, so have to manually check with git blame or something similar to get to know if they are modified.

darakian:
> Need to see properly, some/many of 2019 files may not be touched in 2020 too, so have to manually check with git blame or something similar to get to know if they are modified.

Oh I see. So only the files which were touched should be updated. Ok, give me a bit and I'll update the commit.

darakian:
Ok @vibhoothiiaanand I've updated to only update files touched this year. It looks like all copyrights were either already ranges ending in 2018 or 2019 or were the single year 2019. I've included my scripts in the second and third commits to aid in automating this in the future :)

lu-zero (approved):
I'm fine with this, @vibhoothiiaanand do you want something more?

vibhoothiiaanand (approved):
Seems good to me as well, 
We can re-use the scripts, thanks for that, I haven't look closely and overall these seems to be correct.

vibhoothiiaanand:
Thank you for the patchset

PR #2568, opened 2020-10-09
Cleanups based on clippy suggestions

darakian:
The first fix here is simply to add a clippy annotation to the
`get_subset_predictors` function to allow stable sorts. I suspect
order matters in these sorts hence this annotation.

second is a simple use of the `matches!` macro

and the third is a removal of a Self type.

Clippy links:
https://rust-lang.github.io/rust-clippy/master/index.html#match_like_matches_macro
https://rust-lang.github.io/rust-clippy/master/index.html#needless_arbitrary_self_type
https://rust-lang.github.io/rust-clippy/master/index.html#stable_sort_primitive

### DIFF
--- a/src/lrf.rs
+++ b/src/lrf.rs
@@ -139,13 +139,7 @@ impl Default for RestorationFilter {
 impl RestorationFilter {
   pub fn notequal(self, cmp: RestorationFilter) -> bool {
     match self {
-      RestorationFilter::None {} => {
-        if let RestorationFilter::None {} = cmp {
-          false
-        } else {
-          true
-        }
-      }
+      RestorationFilter::None {} => !matches!(cmp, RestorationFilter::None {}),
       RestorationFilter::Sgrproj { set, xqd } => {
         if let RestorationFilter::Sgrproj { set: set2, xqd: xqd2 } = cmp {
           !(set == set2 && xqd[0] == xqd2[0] && xqd[1] == xqd2[1])
--- a/src/me.rs
+++ b/src/me.rs
@@ -365,8 +365,8 @@ fn get_subset_predictors<T: Pixel>(
       subset_b.iter().map(|&a| a.row).collect();
     let mut cols: ArrayVec<[i16; 3]> =
       subset_b.iter().map(|&a| a.col).collect();
-    rows.as_mut_slice().sort();
-    cols.as_mut_slice().sort();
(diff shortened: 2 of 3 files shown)


lu-zero (approved):
It looks good to me, @barrbrain do we need the stable sort?

shssoichiro:
The [clippy docs](https://rust-lang.github.io/rust-clippy/master/index.html#stable_sort_primitive) state that this lint should only trigger when sorting primitives, in which case an unstable sort is always fine. If the lint is triggering for some other case, that seems like a bug we should report to clippy.

darakian:
> The [clippy docs](https://rust-lang.github.io/rust-clippy/master/index.html#stable_sort_primitive) state that this lint should only trigger when sorting primitives, in which case an unstable sort is always fine. If the lint is triggering for some other case, that seems like a bug we should report to clippy.

It simply came up as a warning. Given that element 1 is indexed I assume the order matters and the suppression of the warning seemed like a good idea.

src/lrf.rs
```diff
@@ -139,13 +139,7 @@ impl Default for RestorationFilter {
...
   pub fn notequal(self, cmp: RestorationFilter) -> bool {
     match self {
-      RestorationFilter::None {} => {
-        if let RestorationFilter::None {} = cmp {
-          false
-        } else {
-          true
-        }
-      }
+      RestorationFilter::None {} => matches!(cmp, RestorationFilter::None {}),
```
lu-zero, line 142: @barrbrain pointed out there is a `!` missing here

darakian: Updated. Sorry about that and thanks for catching it.

src/me.rs
```diff
@@ -284,6 +284,7 @@ impl MotionEstimationSubsets {
   }
 }
 
+#[allow(clippy::stable_sort_primitive)]
```
shssoichiro, line 287: > barrbrain> we don't need a stable sort for finding the median of primitive values

So these could actually be fixed instead of annotating them with allow

shssoichiro (changes requested):

src/me.rs
darakian: Awesome. I've pushed a new commit which uses unstable sorting.

shssoichiro (approved):

PR #2572, opened 2020-10-14
CI: Avoid deploying automatically binaries on forks

Luni-4:
This PR avoids deploying automatically the binaries on forks.


lu-zero (approved):
Sure

PR #2573, opened 2020-10-14
CI: Add a way to skip CI

Luni-4:
This PR skips a CI check when a commit message contains the `[skip ci]` tag.

### DIFF
--- a/.github/workflows/js_api.yml
+++ b/.github/workflows/js_api.yml
@@ -28,6 +28,10 @@ jobs:
             SCCACHE_IDLE_TIMEOUT: 0
             RAV1E_JS_DIR: rav1e_js
 
+        if: >-
+         (github.event_name == 'push' && !endsWith(github.event.head_commit.message, 'CI: skip')) ||
+         (github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.names, 'skip-ci'))
+
         runs-on: ubuntu-latest
 
         steps:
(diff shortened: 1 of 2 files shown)


vibhoothiiaanand:
I wonder why do we need such thing ?

tmatth:
> I wonder why do we need such thing ?

To save time/CO2 if CI isn't required (WIPs, etc.)

Luni-4:
> > I wonder why do we need such thing ?
> 
> To save time/CO2 if CI isn't required (WIPs, etc.)

Yep, exactly

lu-zero (changes requested):
I'd rather use the pull request tags instead.

Luni-4:
> I'd rather use the pull request tags instead.

Good idea! I will implement that then, thanks!

Luni-4:
@lu-zero 

Now it works both with `push` and `pull_request` events. [Here](https://github.com/Luni-4/rav1e/commit/8bec2b7d2d2f22b2cb738eb39e0aa225c3f2912c) you can find a typical commit message that could be used, in this way we don't ruin our history.

vibhoothiiaanand (approved):
I think this should implement what @lu-zero advised,

lu-zero (approved):
Would be nicer with the in-commit message as `ci: skip` (since some viewer render `key: value` in git commits)

```
Commit subject

This is the commit body
with 
CI: skip
Co-Autored-by: somebody
...
```

Luni-4:
@lu-zero 

Done

vibhoothiiaanand (approved):
LGTM :)

PR #2574, opened 2020-10-15
Implement max bitrate option for quantizer mode

shssoichiro:
- Adds a new `--max-bitrate` CLI option to specify an upper local bitrate limit when in quantizer encoding mode. For example, specifying `--quantizer 40 --max-bitrate 15000` will use a base quantizer of 40, but will raise the quantizer in areas where a quantizer of 40 would cause the encode to exceed 15 Mbps.

- Currently not implemented for bitrate mode because that would require a more significant rework of rate control. Bitrate mode will ignore the max-bitrate parameter for the time being.


shssoichiro:
Closing in favor of #2578

PR #2575, opened 2020-10-15
Arm cdef

xiphmont:
Add the interface layer needed to use the current (0.7) dav1d CDEF assembly for aarch64 unmodified in rav1e.  Both 8-bit and HBD are fully enabled.

Testing in progress. Should it pass, this is a full cut of the needed changes.

### DIFF
--- a/src/asm/aarch64/cdef.rs
+++ b/src/asm/aarch64/cdef.rs
@@ -13,27 +13,26 @@ use crate::frame::*;
 use crate::tiling::PlaneRegionMut;
 use crate::util::*;
 
+type CdefPaddingFn = unsafe extern fn(
+  tmp: *mut u16,
+  src: *const u8,
+  src_stride: isize,
+  left: *const [u8; 2],
+  top: *const u8,
+  h: i32,
+  edges: isize,
+);
+
 type CdefFilterFn = unsafe extern fn(
   dst: *mut u8,
   dst_stride: isize,
   tmp: *const u16,
-  tmp_stride: isize,
   pri_strength: i32,
   sec_strength: i32,
   dir: i32,
   damping: i32,
-);
-
-type CdefFilterHBDFn = unsafe extern fn(
-  dst: *mut u16,
-  dst_stride: isize,
-  tmp: *const u16,
-  tmp_stride: isize,
-  pri_strength: i32,
-  sec_strength: i32,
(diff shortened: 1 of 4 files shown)


xiphmont:
OK, looks like the only test failing is the wasm-pack test that mindfreeze mentioned is broken.  This is ready for review.

src/asm/aarch64/cdef.rs
```diff
@@ -0,0 +1,407 @@
...
+    call_rust(&mut copy.as_region_mut());
+    copy
+  };
+  match T::type_enum() {
+    PixelType::U8 => {
+      match (
+        CDEF_PAD_FNS[cpu.as_index()][decimate_index(xdec, ydec)],
+        CDEF_FILTER_FNS[cpu.as_index()][decimate_index(xdec, ydec)],
+      ) {
+        (Some(pad), Some(func)) => {
```
lu-zero, line 100: pad and filter are always implemented together, ideally we could make a single dispatch instead of two.

lu-zero (approved):
It looks fine.

I just have a nit, but we can check once it is landed if doing that gives a little speed boost.

xiphmont:
> It looks fine.
> 
> I just have a nit, but we can check once it is landed if doing that gives a little speed boost.

FWIW, I originally did it that way, but the cpu_feature code wouldn't accept a dispatch that wasn't cdecl (ie, it wouldn't take a rust call).

lu-zero:
> > It looks fine.
> > I just have a nit, but we can check once it is landed if doing that gives a little speed boost.
> 
> FWIW, I originally did it that way, but the cpu_feature code wouldn't accept a dispatch that wasn't cdecl (ie, it wouldn't take a rust call).

`#[no_mangle]` and `extern fn` would fix that.

lu-zero:
please rebase it

vibhoothiiaanand (commented):
These are invoking the CDEF assembly functions, there is no impact over performance, but on seeing memory allocation, we are seeing decrease in total usage of memory by CDEF functions.
So would be nicer to test on wider range of clips to see if we have any impact for assembly codepath integration.

lu-zero:
Tested on the honeycomb with all the speed levels, between 0.7% and 1.7% overall speedup using 16 threads.

PR #2576, opened 2020-10-15
Remove the chromedriver for now, it is apparently broken

lu-zero:


Luni-4:
Closed in favor of #2577

PR #2578, opened 2020-10-15
Ensure that min and max quantizer are respected

shssoichiro:
When specifying e.g. `--bitrate 10000 --min-quantizer 100`, the min-quantizer setting did not appear to take effect. For example, on a test clip of BBB 720p first 500 frames, the average QPs were 20 and 68 for I and P frames respectively. After this PR, the QPs are flat as would be expected of a `--quantizer 100` encode (since bitrate 10000 would, in this case, always want a lower quantizer).

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -1102,7 +1102,7 @@ impl<T: Pixel> ContextInner<T> {
     }
   }
 
-  fn encode_packet(
+  pub(crate) fn encode_packet(
     &mut self, cur_output_frameno: u64,
   ) -> Result<Packet<T>, EncoderStatus> {
     if self.frame_data.get(&cur_output_frameno).unwrap().fi.show_existing_frame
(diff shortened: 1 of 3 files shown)


barrbrain:
This does ensure that min/max quantizer are hard limits. Considering that the same parameters are used to set the bounds of the rate-control search, the cases where they are not respected should be interesting. I need to read through again to understand clearly.

lu-zero:
Maybe we should put an integration test for them.

shssoichiro:
@lu-zero I added some tests to verify this

lu-zero (approved):
It seems fine to me.

shssoichiro:
Summary of conversation with Tim on IRC:

This is fine as a baseline for now. We want to add to this later to include things like temporal RDO into the calculation. This can probably be done without majorly changing the approach here.

PR #2587, opened 2020-11-09
Simplify internals

lu-zero:
First set of tiny changes of many. The final plan is to have a function that outputs the subgop as whole and possibly make so we do not need do not need `garbage_collect` anymore.

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -219,6 +219,9 @@ impl<T: Pixel> FrameData<T> {
   }
 }
 
+type FrameQueue<T> = BTreeMap<u64, Option<Arc<Frame<T>>>>;
+type FrameDataQueue<T> = BTreeMap<u64, FrameData<T>>;
+
 // the fields pub(super) are accessed only by the tests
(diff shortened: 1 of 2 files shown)


src/api/internal.rs
```diff
@@ -265,24 +268,25 @@ impl<T: Pixel> ContextInner<T> {
...
       frame_count: 0,
       limit: None,
-      inter_cfg: InterConfig::new(enc),
+      inter_cfg,
       output_frameno: 0,
       frames_processed: 0,
       frame_q: BTreeMap::new(),
       frame_data: BTreeMap::new(),
-      keyframes: BTreeSet::new(),
+      keyframes: [0].iter().cloned().collect(),
```
rzumer, line 281: How is this a simplification?

lu-zero: it avoids a branch in the send_frame(). I can make setting BTreeSet in two lines that are less surprising.

barrbrain (approved):
These small changes look like marginal improvements but they help to reduce the entanglement of encoder configuration.

PR #2590, opened 2020-11-16
Move the tiling information in the Sequence, and compute it only once

lu-zero:


shssoichiro (approved):
Seems like a good idea

PR #2591, opened 2020-11-17
Use env var $GITHUB_PATH instead of add-path

vibhoothiiaanand:
As of last week, GitHub's add-path command has been deprecated.


Luni-4:
@vibhoothiiaanand 

Could you please update `deploy.yml` as well?

Luni-4 (approved):
Thanks a lot! :)

Perhaps we should squash the commits and then we are ready to go

lu-zero (approved):

PR #2593, opened 2020-11-17
Enable Thin LTO

shssoichiro:
Provides minor build time improvements on all platforms.
Provides minor runtime improvements on x86_64.
Provides significant runtime improvements (~13%) on ARM.


lu-zero (approved):
\o/

PR #2594, opened 2020-11-18
Optimize initialization of residual array

shssoichiro:
This initialization was a hotspot detected in profiling.
This change avoids initializing the unused parts of the array,
and avoids re-initializing the used parts since they will be set
anyway via the call to `diff`.

Provides approximately 3% speedup on 8-bit and 1% speedup on 10-bit.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1216,10 +1216,7 @@ pub fn encode_tx_block<T: Pixel>(
   }
 
   let coded_tx_area = av1_get_coded_tx_size(tx_size).area();
-  let mut residual_storage: Aligned<[i16; 64 * 64]> =
-    Aligned::new([0i16; 64 * 64]);
-  // TODO(yushin): When tx size != visible size, instead of init whole residual block as zeros everytime,
-  // consider doing so only pixels outside of visible frame.
+  let mut residual_storage: Aligned<[i16; 64 * 64]> = Aligned::uninitialized();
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -1251,6 +1248,16 @@ pub fn encode_tx_block<T: Pixel>(
       visible_tx_h,
     );
   }
+  unsafe {
```
lu-zero, line 1251: init_slice_repeat_mut doesn't do the same but using safe code?

shssoichiro: init_slice_repeat_mut works on a slice of MaybeUninit, but I can do this with a for loop (which is what init_slice_repeat_mut does internally) and it compiles down to a memset.

Interestingly, using `slice.iter_mut.for_each()` does not compile down to a memset. but using `for a in slice.iter_mut()` does. Weird.

lu-zero: I mean using the MaybeUninit array + unit_slice_repeat_mut so we get the correct slice. This version is already nice but if it works the same I'd keep the code more uniform.

shssoichiro: I think the complexity around using init_slice_repeat_mut here is that we initialize part of the array with real values while running the `diff` function. At which point we already need a `&[i16]`. So it's simpler, in my opinion, to not use MaybeUninit in this case.

lu-zero (approved):
Ok.

PR #2595, opened 2020-11-18
Make the scenechange detector publicly usable

shssoichiro:
The intent is to stop av-scenechange from having
to duplicate all of the internals of rav1e's
scenechange detection.
This makes all necessary items public and interoperable
with av-scenechange, so that av-scenechange can pull in
rav1e as a dependency and always be up to date
with its scenechange detection.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ bench = []
 check_asm = []
 capi = []
 tracing = ["rust_hawktracer"]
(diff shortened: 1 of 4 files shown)


lu-zero:
I'd put it under the unstable feature and/or a specific feature.

lu-zero (approved):
It is fine to me.

PR #2598, opened 2020-11-25
CI: Update libdav1d to 0.8.0-dmo1

barrbrain:
A new dav1d release is out, let's pull it into CI before the current release is dropped.


vibhoothiiaanand (approved):
LGTM

PR #2600, opened 2020-11-25
Optimise NEON assembly for get_sad for widths 32, 64 and 128

barrbrain:
Reduces execution time for these functions by between 35% and 42%.
Binary size of the SAD module is reduced by 23%, to less than 1kB.

Reorder widths 4 and 8 for consistency with other widths.

### DIFF
--- a/src/arm/64/sad.S
+++ b/src/arm/64/sad.S
@@ -42,11 +42,11 @@ function sad4x4_neon, export=1
         sxtw            x3,  w3
         mov             w4,  #4
 L(sad_w4):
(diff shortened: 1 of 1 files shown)


lu-zero (approved):
Looks fine to me.

rzumer:
What is the overall improvement in encoding time?

barrbrain:
> What is the overall improvement in encoding time, on average?

I haven't benchmarked that yet but I think it shouldn't be more than 0.5%.

PR #2604, opened 2020-12-05
Run keyframe cost estimations in parallel

shssoichiro:
Intra cost and inter cost estimation account for 99%
of the time spent in keyframe detection (except at speed 10).
These two computations can be run in parallel,
thus providing a 30-40% speedup on keyframe detection
and a 1-3% speedup overall (depends on speed level, etc).

### DIFF
--- a/src/scenechange/mod.rs
+++ b/src/scenechange/mod.rs
@@ -245,22 +245,29 @@ impl SceneChangeDetector {
         has_scenecut: delta >= threshold,
       }
     } else {
-      let intra_costs =
-        estimate_intra_costs(&*frame2, self.bit_depth, self.cpu_feature_level);
-      let intra_cost = intra_costs.iter().map(|&cost| cost as u64).sum::<u64>()
-        as f64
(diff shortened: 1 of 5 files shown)


src/scenechange/mod.rs
```diff
@@ -245,22 +245,29 @@ impl SceneChangeDetector {
...
-        / intra_costs.len() as f64;
-
-      let inter_costs = estimate_inter_costs(
-        frame2,
-        frame1,
-        self.bit_depth,
-        self.encoder_config,
-        self.sequence.clone(),
+      let frame2_ref2 = Arc::clone(&frame2);
+      let (intra_cost, inter_cost) = rayon::join(
```
lu-zero, line 249: you should use `crate::rayon` and no-op `join()` to stay compatible for the js bindings.

lu-zero (changes requested):

lu-zero (approved):
It seems fine to me.

PR #2607, opened 2020-12-09
Unify register assignment in NEON assembly for get_sad

barrbrain:
Easier to read and reduces compressed compiled object size by 10%.


shssoichiro:
It seems like something isn't quite happy here: https://travis-ci.org/github/xiph/rav1e/builds/748500930

barrbrain:
> do we have other places that would enjoy ldp?

For sure, most of the wider distortion kernels. I will revisit them when we have a complete set of distortion assembly.

Issue #151, opened 2018-03-22
Support lossless mode!


ycho:

ycho:
Won't do this.

tdaede:
Someone had an interesting use case of encoding MAME files, e.g. og.kervella.org/pt.7z

This is low priority but no reason not to implement it.

ycho:
Sure, rav1e always can encode losslessly. I believe AV1 do that with only using 4x4 DCTs as transform, correct?

tdaede:
There's a special 4x4 transform that we currently don't have. That's the main thing missing.

cryptomilk:
I've implemented support for importing and exporting AVIF images in [darktable](https://darktable.org/) using [libavif](https://github.com/AOMediaCodec/libavif/). I would be interested to have lossless support using rav1e.

https://blog.cryptomilk.org/2019/11/26/avif-image-support-for-darktable/

cryptomilk:
We have added support for importing and exporting AVIF files in [darktable](https://darktable.org) which will be released with version 3.0.1. It would be great to have this fixed in rav1e soon.

eclipseo:
Any chance of this issue seeing  some love?

CartoonFan:
I'd like to voice my support for this feature as well :heart:

CartoonFan:
Just now seeing the recent commit here? Has there been any progress on lossless encoding? :eyes:

PR #2608, opened 2020-12-09
Make EncoderConfig non_exhaustive

est31:
This allows changing it in the future without performing an API breaking
change.


lu-zero:
It is probably a good idea, but the current implementation in rust seems to trip in exactly the pattern that makes possible to write code w/out caring about additional fields...

est31:
Oh indeed, I've thought the ..Default::default pattern still works with non_exhaustive, but seems it's disallowed. Apparently so that you are allowed to add private fields in the future...

Should I close this?

lu-zero:
At least for now, it seems that it has more cons than pros sadly =/

est31:
Ok then! Btw, I'm interested in this struct because of #151.

PR #2609, opened 2020-12-09
Simplify Hawktracer boilerplate

shssoichiro:
Hawktracer already has a feature, "profiling_enabled",
which controls whether it performs profiling or performs a no-op.
With the feature disabled, it performs the equivalent of our
noop_proc_macro. Therefore, we can remove our extra layer of
boilerplate.


lu-zero (approved):

lu-zero:
Sounds a good idea. Thank you :)

PR #2613, opened 2020-12-24
Fix 16-bit overflow in rav1e_sad_16x16_hbd_ssse3.

negge:
Accumulate SAD in 32 bits as 16 * 16 * 1023 = 261888 exceeds a 16-bit value.


lu-zero (approved):
good catch, we should update the tests to avoid those problems ^^;

PR #2614, opened 2020-12-24
Disable NEON assembly that fail check_asm.

negge:


lu-zero (approved):
I confirmed the issue. please land it

PR #2615, opened 2020-12-24
Add AVX2 put 8tap HBD functions

negge:


lu-zero (approved):
The rust side is fine for me. You tested the assembly with `features=check_asm` already so it shouldn't have problems.

PR #2618, opened 2020-12-26
CI: Use ilammy/setup-nasm@v1 to install nasm

EwoutH:
Replaces the proprietary script with the [ilammy/setup-nasm@v1](https://github.com/ilammy/setup-nasm) GitHub Action.


Luni-4 (approved):
Thanks a lot for your changes @EwoutH! 

The code is much clear now!

lu-zero (approved):
I guess we can land it.

PR #2619, opened 2020-12-26
Enable AVX2 avg HBD function.

negge:


lu-zero (approved):
Looks good if check_asm agrees

PR #2621, opened 2020-12-28
Fix assembly issues not found in our CI.

negge:
These errors would have been found if we had regression tests on MacOS
 and a similar register clobber checker to the one in dav1d.


KyleSiefring (approved):
LGTM

PR #2622, opened 2020-12-28
AVX2 blend HBD functions

negge:


barrbrain (approved):
The table is correctly translated; the code passes checkasm and the dav1d regression suite.

PR #2627, opened 2020-12-30
Prepare to release v_frame

lu-zero:
It is needed by the stand-alone rav1e.js


vibhoothiiaanand (approved):
LGTM

barrbrain:
Given the build errors, I guess we might also need to adjust the dependencies of the dependent modules.

lu-zero:
Ideally we should bump av-metrics at the same time.

lu-zero:
And consider spinning off v_frame so we can avoid this dependency mess.

vibhoothiiaanand (approved):
Now that, the CI goes green, we can land this possibly

PR #2630, opened 2020-12-31
Fix clippy lints

Luni-4:
This PR fixes `clippy` lints introduced by `Rust 1.49`


lu-zero (approved):
Thank you a lot!

PR #2636, opened 2021-01-12
Bump dav1d version

lu-zero:


.github/workflows/rav1e.yml
```diff
@@ -141,11 +141,11 @@ jobs:
         matrix.conf == 'grcov-coveralls' || matrix.conf == 'fuzz' || matrix.conf == 'no-asm-tests'
       env:
         LINK: https://www.deb-multimedia.org/pool/main/d/dav1d-dmo
-        DAV1D_VERSION: 0.8.0-dmo1
+        DAV1D_VERSION: 0.8.1-dmo1~bpo10+1
```
jamrial, line 144: There's no 0.8.1-dmo1~bpo10+1 on the above link, only 0.8.1-dmo1

Try that, then dcf911325699d93a90818e16736e2c93b29d8e7538c1545accd3b25c610876c0 for DAV1D_DEV_SHA256 and 06f51b9660d413417827270b298e2ad541bd8ddaae7e027ebcb6bb7b6b1ad006 for DAV1D_LIB_SHA256.

lu-zero: Thank you! I was too tired to spot the typo.

vibhoothiiaanand (approved):
LGTM as tests passes

barrbrain (commented):
We should also update `.travis/install-dav1d.sh` as we still run that.

lu-zero:
Strangely travis seems happy with what we have...

PR #2642, opened 2021-01-15
Fix documentation speed 10

BlueSwordM:
Correcting a comment in the speedsettings documentation about speed 10 using minimum block sizes of 64x64, which was changed to a minimum and maximum block size of 32x32 in this commit:
https://github.com/xiph/rav1e/commit/8d3affb


negge (commented):
Would it be clearer to say "fixed block size 32x32," ?

BlueSwordM:
True, that would clearer.

negge:
Can you squash into one commit and push -f the branch?

BlueSwordM:
Sorry if it took a while, but it's done now.

negge (approved):
LGTM

PR #2643, opened 2021-01-20
x86: cdef: Add SIMD implementation of cdef_dir for 16bpc

negge:
```
Relative speed-ups over C code (compared with gcc-9.3.0):

                                           C       ASM
cdef_dir_16bpc_avx2:                   534.2      72.5     7.36x
cdef_dir_16bpc_ssse3:                  534.2     104.8     5.10x
cdef_dir_16bpc_ssse3 (x86-32):         854.1     116.2     7.35x
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -99,6 +99,9 @@ fn build_nasm_files() {
     "src/x86/satd.asm",
     "src/x86/sse.asm",
     "src/x86/cdef.asm",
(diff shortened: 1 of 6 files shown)


lu-zero (approved):
It seems good to me

lu-zero (approved):
It is not changed from when I looked at it. Land it anytime.

PR #2647, opened 2021-01-23
Add few missing cdfs to the map

lu-zero:
We should make sure that all the fields are instrumented


tmatth:
nit: re the commit message, CDFs (CDEF is unrelated here).

lu-zero:
btw help checking all are covered would be welcome.

barrbrain:
> btw help checking all are covered would be welcome.

The following shows that `tx_size_cdf` and `txfm_partition_cdf` are uncovered.
```bash
for CDF in `grep -o 'pub .*_cdf[^:]*' src/context/cdf_context.rs | cut -f2 -d' '`
do grep -q "${CDF}_start =" src/context/cdf_context.rs || echo "$CDF"
done
```

barrbrain (approved):

PR #2649, opened 2021-01-23
Remove the js api from the main tree

### LINKED ISSUE #2553: Move `rav1e_js` out of the main encoder tree
Hi,

Right now rav1e_js living in the main tree is superfluous to the encoder and its functionality and also additional CI run-time, 
I would like to propose to move `$root/rav1e_js` as a whole to a separate repo. So it is better and cleaner.
Either we could keep it in rust-av or maybe in Xiph itself, I think it might be better if it lives in rust-av like rav1e-by-gop and later moves to Xiph if it is matured and required.

lu-zero:
It lives in a separate repository now.

Closes #2553


vibhoothiiaanand (approved):
Hi,

Thanks for the patch, 

I can confirm that the standalone rav1e_js works without issue
https://github.com/rust-av/rav1e_js

Also, I have also confirmed the CI works fine,

PR #2663, opened 2021-02-06
Shrink CDFs by merging counter to final symbol position

barrbrain:
A similar approach is used in dav1d and this enables some improvements to the CDF log.

### DIFF
--- a/src/context/block_unit.rs
+++ b/src/context/block_unit.rs
@@ -485,19 +485,19 @@ impl<'a> BlockContext<'a> {
 
 #[derive(Clone, Copy)]
 pub struct NMVComponent {
-  pub classes_cdf: [u16; MV_CLASSES + 1],
-  pub class0_fp_cdf: [[u16; MV_FP_SIZE + 1]; CLASS0_SIZE],
-  pub fp_cdf: [u16; MV_FP_SIZE + 1],
-  pub sign_cdf: [u16; 2 + 1],
-  pub class0_hp_cdf: [u16; 2 + 1],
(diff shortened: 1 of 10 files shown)


src/context/block_unit.rs
```diff
@@ -689,7 +687,7 @@ impl<'a> ContextWriter<'a> {
     if bs.cfl_allowed() {
       symbol_with_update!(self, w, uv_mode as u32, cdf);
     } else {
-      symbol_with_update!(self, w, uv_mode as u32, &mut cdf[..UV_INTRA_MODES]);
+      symbol_with_update!(self, w, uv_mode as u32, &mut cdf[..INTRA_MODES]);
```
barrbrain, line 690: This may seem a little odd. Where the original code had `UV_INTRA_MODES`, it ought to read `INTRA_MODES + 1` because the additional element here is the CDF counter rather than the chroma-from-luma symbol. As with other edits in this patch, we are subtracting 1 now that the trailing zero is removed. Thus `UV_INTRA_MODES` becomes `INTRA_MODES`.

barrbrain:
There is a change to `UV_INTRA_MODES` that may seem a little odd. Where the original code had `UV_INTRA_MODES`, it ought to read `INTRA_MODES + 1` because the additional element here is the CDF counter rather than the chroma-from-luma symbol. As with other edits in this patch, we are subtracting 1 now that the trailing zero is removed. Thus `UV_INTRA_MODES` becomes `INTRA_MODES`.

lu-zero (approved):

PR #2664, opened 2021-02-08
Partition CDF log by size

barrbrain:
The vast majority of writes are to small CDFs, so we may partition the log to reduce redundant storage. Ordering the CDFs by size ensures that however we split rollback write sizes, the writes will not overlap between regions.

### DIFF
--- a/src/context/block_unit.rs
+++ b/src/context/block_unit.rs
@@ -484,20 +484,28 @@ impl<'a> BlockContext<'a> {
 }
 
 #[derive(Clone, Copy)]
+#[repr(C)]
 pub struct NMVComponent {
-  pub classes_cdf: [u16; MV_CLASSES],
-  pub class0_fp_cdf: [[u16; MV_FP_SIZE]; CLASS0_SIZE],
-  pub fp_cdf: [u16; MV_FP_SIZE],
   pub sign_cdf: [u16; 2],
   pub class0_hp_cdf: [u16; 2],
   pub hp_cdf: [u16; 2],
   pub class0_cdf: [u16; CLASS0_SIZE],
   pub bits_cdf: [[u16; 2]; MV_OFFSET_BITS],
+
+  pub class0_fp_cdf: [[u16; MV_FP_SIZE]; CLASS0_SIZE],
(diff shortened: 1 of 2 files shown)


barrbrain:
Abridged transcript of review discussion on IRC:
> [11:36] \<derf> "Ordering the CDFs by size ensures that however we split rollback
[11:36] \<derf> write sizes, the writes will not overlap between regions."
[11:37] \<derf> This commit does not do that, though?
...
[11:38] \<derf> Like, even just in NMVComponent, class0_fp_cdf has size 4 CDFs, followed by sign_cdf (size 2).
[11:39] \<derf> const MV_FP_SIZE: usize = 4;
...
[11:42] \<derf> I mean, the statement is still a little weird, since you are going to have to add padding in the right places in addition to the re-ordering
  (which you do for your desired split, but not for arbitrary splits).
[11:43] \<barrbrain> yeah, padding is required since there are multiple ordered regions
[11:44] \<derf> But I think padding is still required even if they are strictly ordered.
[11:44] \<barrbrain> if each region ended with the global maximum size, padding would not be required
[11:44] \<derf> I guess that is true, yes.
...
[11:58] \<barrbrain> Okay, I've addressed the NMVContext comments
...
[12:08] \<derf> I'm not wild about using a tuple for the CDF log checkpoint instead of separate named fields, but I guess I can see why you did it.
[12:08] \<derf> And the rest LGTM.
[12:09] \<derf> I won't insist on you changing it, though.

lu-zero:
It fixes the speed regression I was seeing on a specific aarch64 platform on higher speed levels:

image

lu-zero:
On x86_64 it has consistent results both on stable and nightly:
image

barrbrain:
The benchmarks indicate overall that this change is either performance-neutral or offers a marginal increase in speed. The notable exception is that it resolves a performance regression in a preceding PR on a specific architecture. The primary purpose is to reduce the log size and we see that this can be achieved without trading away any speed. A smaller log may pave the way for future improvements such as stack allocation. We are now at the point that cache behavior becomes the dominant concern.

lu-zero (approved):

Issue #2646, opened 2021-01-22
Unaligned memmove is > 20% of encode time at -s0


negge:
```
$ perf record target/release/rav1e Bosphorus_1920x1080_120fps_420_8bit_YUV.y4m -o /dev/null --tiles 4 --threads 4 --limit 64 -s0
Couldn't synthesize bpf events.
>  Using y4m decoder: 1920x1080p @ 30/1 fps, 4:2:0, 8-bit
>  Encoding settings: keyint_min=12 keyint_max=240 quantizer=100 bitrate=0 min_quantizer=0 low_latency=false tune=Psychovisual rdo_lookahead_frames=40 min_block_size=4x4 max_block_size=64x64 multiref=true fast_deblock=false reduced_tx_set=false tx_domain_distortion=false tx_domain_rate=false encode_bottomup=true rdo_tx_decision=true prediction_modes=Complex-All include_near_mvs=true no_scene_detection=false cdef=true use_satd_subpel=true non_square_partition=true enable_timing_info=false fine_directional_intra=true
>  CPU Feature Level: AVX2
>  Using 4 tiles (2x2)
>  encoded 64/64 frames, 0.100 fps, 2342.27 Kb/s, est. size: 0.60 MB, est. time: 0s,  elap. time: 10m 38s                        
>  ----------
>  Key frame:             1 | avg QP:  79.00 | avg size:  104170 B
>  Inter frame:          63 | avg QP: 138.00 | avg size:    8260 B
>  Intra only frame:      0 | avg QP:   0.00 | avg size:       0 B
>  Switching frame:       0 | avg QP:   0.00 | avg size:       0 B
>  ----
[ perf record: Woken up 1229 times to write data ]
[ perf record: Captured and wrote 308.871 MB perf.data (8096427 samples) ]

$ perf report
# Overhead  Command  Shared Object       Symbol                                                                                                                                                                   
# ........  .......  ..................  .........................................................................................................................................................................
#
    20.33%  rav1e    libc-2.30.so        [.] __memmove_avx_unaligned_erms
    11.39%  rav1e    rav1e               [.] rav1e::encoder::encode_tx_block
     9.44%  rav1e    rav1e               [.] <rav1e::ec::WriterBase<S> as rav1e::ec::Writer>::symbol_with_update
     7.59%  rav1e    rav1e               [.] rav1e::rdo::cdef_dist_wxh_8x8
[...]

PR #2666, opened 2021-02-09
Reduce size of BlockContextCheckpoint

barrbrain:
Saving and restoring over 4KiB at a time is excessive considering the vast majority of rollbacks change less than 10 bytes.
We may store only the data for the current superblock in the checkpoint. This further advances issue #2646.

### DIFF
--- a/src/context/block_unit.rs
+++ b/src/context/block_unit.rs
@@ -211,13 +211,15 @@ impl Default for Block {
 
 #[derive(Clone)]
 pub struct BlockContextCheckpoint {
+  x: usize,
+  chroma_sampling: ChromaSampling,
   cdef_coded: bool,
-  above_partition_context: [u8; PARTITION_CONTEXT_MAX_WIDTH],
+  above_partition_context: [u8; MIB_SIZE >> 1],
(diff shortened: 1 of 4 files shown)


src/context/block_unit.rs
```diff
@@ -256,25 +258,61 @@ impl<'a> BlockContext<'a> {
...
+      &self.above_partition_context[(bx >> 1)..][..(MIB_SIZE >> 1)],
+    );
+    bcc
+      .above_tx_context
+      .copy_from_slice(&self.above_tx_context[bx..][..MIB_SIZE]);
+    let num_planes =
+      if chroma_sampling == ChromaSampling::Cs400 { 1 } else { 3 };
+    for p in 0..num_planes {
+      let xdec = (p > 0 && chroma_sampling != ChromaSampling::Cs444) as usize;
+      unsafe {
```
lu-zero, line 286: I'd rather avoid this if it has a measurable impact split the unchecked from the copy_from_slice please.

lu-zero (approved):
Looks good, thank you!

PR #2667, opened 2021-02-10
Use explicit table lookup for ac_q and dc_q

barrbrain:
* Resolve a surprising compiler decision to invoke memmove.
* Calculate indices in a way that rustc may elide bounds checks.


lu-zero (changes requested):
Removing the `&` from `&table` should have the same effect.

lu-zero (approved):
Please change the first commit message to 
```
Correct a misuse of pattern in a let statement

The expression required the compiler to make a copy of the array referenced
```

The rest looks good to land.

PR #2668, opened 2021-02-12
Fix clippy warnings

Luni-4:
This PR fixes `clippy` warnings introduced by `Rust 1.50`


barrbrain (approved):
Ignoring https://rust-lang.github.io/rust-clippy/master/#wrong_self_convention for now is a good call, the fix is to rename a utility function called in many places. The other fix is well contained.

lu-zero (approved):
Looks fine to me, maybe we could suppress the lint for `rdo_partition_none` instead.

@barrbrain do you have opinions?

PR #2669, opened 2021-02-12
Parallel temporal rdo update

lu-zero:
Initial naive implementation, it nearly halves the `compute_block_importances` execution time according to hawktracer.
It is ~9% faster for 1080p at speed 6.


barrbrain (approved):
Looks good given that stray lines in the last 2 patches get cleaned up before merge.

PR #2673, opened 2021-02-16
Refactor CDF access so that size is always statically known

barrbrain:
While removing explicit slices for CDF access, it was shown that some were not ordered by size in `CDFContext`. This violates the invariant introduced in PR #2664, to guarantee non-overlapping writes at rollback. This change ensures that the invariant holds.
As the CDF expressions passed to `symbol_with_update!` all become mutable references to arrays, we can specialise `symbol_with_update()` for small numbers of CDF symbols. This yields a speed-up on the order of 1.5%.


src/context/cdf_context.rs
```diff
@@ -76,8 +76,9 @@ pub struct CDFContext {
 
   pub kf_y_cdf: [[[u16; INTRA_MODES]; KF_MODE_CONTEXTS]; KF_MODE_CONTEXTS],
   pub y_mode_cdf: [[u16; INTRA_MODES]; BLOCK_SIZE_GROUPS],
+  pub uv_mode_cdf: [[u16; INTRA_MODES]; INTRA_MODES],
 
-  pub uv_mode_cdf: [[[u16; UV_INTRA_MODES]; INTRA_MODES]; 2],
+  pub uv_mode_cfl_cdf: [[u16; UV_INTRA_MODES]; INTRA_MODES],
```
barrbrain, line 90: Missing additional entry in `CDFContext::build_map()`.

lu-zero (approved):
Looks good to me.

PR #2674, opened 2021-02-19
per speed rdo-lookahead-frames

master-of-zen:
## Motivation:
Make speed settings faster with reasonable trade-off on different rdo-lookahead-frames values and suggest new defaults for each speed

### Recommended
* `rdo-lookahead-frames` 10 for speeds 10,9.
* `rdo-lookahead-frames` 20 for speeds 8,7,6.
* `rdo-lookahead-frames` 30 for speeds 5,4,3.
* `rdo-lookahead-frames` 40 for speeds 2,1,0.

### DIFF
--- a/src/api/config/speedsettings.rs
+++ b/src/api/config/speedsettings.rs
@@ -211,14 +211,12 @@ impl SpeedSettings {
 
   /// Set default rdo-lookahead-frames for different speed settings
   pub fn rdo_lookahead_frames(speed: usize) -> usize {
-    if speed == 10 || speed == 9 {
-      10
-    } else if speed <= 8 || speed >= 6 {
-      20
-    } else if speed <= 5 || speed >= 3 {
-      30
-    } else {
-      40
+    match speed {
+      9..=10 => 10,
+      6..=8 => 20,
+      3..=5 => 30,
+      0..=2 => 40,
+      _ => 40,
     }
   }
 
--- a/src/api/config/speedsettings.rs
+++ b/src/api/config/speedsettings.rs
@@ -209,6 +209,19 @@ impl SpeedSettings {
     speed <= 1
   }
 
+  /// Set default rdo-lookahead-frames for different speed settings
+  pub fn rdo_lookahead_frames(speed: usize) -> usize {
+    if speed == 10 || speed == 9 {
+      10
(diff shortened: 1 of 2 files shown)


src/api/config/speedsettings.rs
```diff
@@ -209,6 +209,19 @@ impl SpeedSettings {
     speed <= 1
   }
 
+  /// Set default rdo-lookahead-frames for different speed settings
+  pub fn rdo_lookahead_frames(speed: usize) -> usize {
+    if speed == 10 || speed == 9 {
```
lu-zero, line 214: a [match](https://doc.rust-lang.org/rust-by-example/flow_control/match.html) would be clearer.

lu-zero (commented):
It looks interesting. It is fine with me, possibly with the match replacing the if-nest.

src/api/config/speedsettings.rs
master-of-zen: Done https://github.com/xiph/rav1e/pull/2674/commits/e015505686e81a1d20191c8033a125edb2f84260

src/bin/common.rs
```diff
@@ -214,7 +214,8 @@ pub fn parse_cli() -> Result<CliOptions, CliError> {
     )
     .arg(
       Arg::with_name("RDO_LOOKAHEAD_FRAMES")
-        .help("Number of frames encoder should lookahead for RDO purposes [default: 40]\n")
+        .help("Number of frames encoder should lookahead for RDO purposes\n\
+        [defaults for speeds: 10,9 - 10, 8,7,6 - 20, 5,4,3 - 30, 2,1,0 - 40]\n")
```
vibhoothiiaanand, line 218:
```suggestion
        [defaults for speeds: 10,9 - 10; 8,7,6 - 20; 5,4,3 - 30; 2,1,0 - 40]\n")
```
I would suggest having something like a semi-colon or something else to distinguish b/w levels and value, as it is confusing for end-users.

vibhoothiiaanand (commented):
Thanks for the patch, I have a small nit, apart from that LGTM.

Also, if we modify the commit message to be a bit descriptive, will be nicer{you could squash both and have patchset cleaner}

lu-zero (approved):
Thank you a lot.

PR #2677, opened 2021-02-25
Make C API installation instructions usable

rzumer:
The Rust toolchain is not installed globally by default, so it is impossible to use
Cargo to install a library unless installing another toolchain for root, messing with
the root environment so the whole user toolchain can be located, or installing locally
and then manually moving the output. This fixes the installation instructions to use
the simplest of those solutions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,14 +127,15 @@ The resulting binary will not work on cpus that do not sport the same set of SIM
 
 > **NOTE** : You may use `rustc --print target-cpus` to check if the cpu is supported, if not `-C target-cpu=native` would be a no-op.
 
-### Building the C-API
+### Building the C API
 **rav1e** provides a C-compatible set of library, header and pkg-config file.
 
 To build and install it you can use [cargo-c](https://crates.io/crates/cargo-c):
 
 ```sh
-cargo install cargo-c
-cargo cinstall --release
+$ cargo install cargo-c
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -127,14 +127,15 @@ The resulting binary will not work on cpus that do not sport the same set of SIM
...
 **rav1e** provides a C-compatible set of library, header and pkg-config file.
 
 To build and install it you can use [cargo-c](https://crates.io/crates/cargo-c):
 
 ```sh
-cargo install cargo-c
-cargo cinstall --release
+$ cargo install cargo-c
+$ cargo cinstall --destdir=$HOME/librav1e --release
+# mv $HOME/librav1e/* / && rmdir $HOME/librav1e
```
lu-zero, line 138: sudo might be mentioned.

tdaede: Using mv like this won't yield correct permissions unfortunately.

We could just suggest an install into ~/.local

tdaede: The canonical way to copy with correct permissions is /usr/bin/install, but it unfortunately can't copy recursively.

I had some incantation that made cinstall work as well but I can't find it now.

rzumer: `#` implies root, on my end moving the files with sudo seemed to set the owner to root and permissions also seem to match other libraries. I didn't run exactly those lines though but I also didn't change permissions or owner manually.

PR #2685, opened 2021-03-09
Use the current threadpool if no thread nor threadpool are specified

### LINKED ISSUE #2684: Allow to use the global threadpool
Currently, the encoder only allows to create its own dedicated threadpool and use it for encoding operations.

However, in a larger app it would be useful to reuse the globally preconfigured threadpool so that 1) same config could be used for all threads and 2) threads could be shared not just within the encoder itself, but with other tasks in the app equally.

More generally, does the lib part of the encoder really need an API to create its own threadpool? It seems like something that belongs to the application code, as it's normally configured globally for all threads for reasons outlined above.

lu-zero:
Fixes #2684.

- [x] Implement the global threadpool support 
- [x] Document the behavior

### DIFF
--- a/src/api/channel.rs
+++ b/src/api/channel.rs
@@ -238,16 +238,18 @@ pub type VideoDataChannel<T> = (FrameSender<T>, PacketReceiver<T>);
 impl Config {
   fn setup<T: Pixel>(
     &self,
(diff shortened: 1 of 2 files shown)


barrbrain (approved):
There is a typo in "behavior" in the second commit message. Looks good otherwise.

PR #2686, opened 2021-03-10
Use global threadpool unless configured otherwise

### LINKED ISSUE #2684: Allow to use the global threadpool

RReverser:
Follow-up to #2685, using the global threadpool in one more place.

Fixes #2684.

### DIFF
--- a/src/api/config/mod.rs
+++ b/src/api/config/mod.rs
@@ -253,11 +253,13 @@ impl Config {
     let inner = self.new_inner()?;
     let config = *inner.config;
     let pool = if let Some(ref p) = self.pool {
-      p.clone()
-    } else {
+      Some(p.clone())
+    } else if self.threads != 0 {
       let pool =
         ThreadPoolBuilder::new().num_threads(self.threads).build().unwrap();
-      Arc::new(pool)
+      Some(Arc::new(pool))
+    } else {
+      None
     };
 
     Ok(Context { is_flushing: false, inner, pool, config })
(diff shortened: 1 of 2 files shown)


lu-zero (changes requested):
Looks fine to me, just change the documentation accordingly :)

RReverser:
> Looks fine to me, just change the documentation accordingly :)

I made a minor nitpick change to the doc from #2685 - it said "if global pool is present", but the thing is, it's always present (it's lazily created by Rayon).

But other than that, I don't think there's any specific docs to update here since fields it uses come from the same `with_threads` / `with_thread_pool` methods you already documented in your PR?

RReverser:
Also extracted both usages to a single `Config::new_thread_pool` helper, since they're meant to return the same thing.

src/api/config/mod.rs
```diff
@@ -153,7 +153,7 @@ impl Config {
   /// components in the encoder.
   ///
   /// If it is left unset `new_context()` will create a new default one
-  /// `new_*_channel()` will use the current global threadpool if present.
+  /// `new_*_channel()` will use the default global threadpool.
```
lu-zero, line 156: With your change new_context() will use the global threadpool likewise. I can reword it myself once this lands.

lu-zero (approved):

src/api/config/mod.rs
RReverser: Done, WDYT?

lu-zero: Thank you :)

PR #2689, opened 2021-03-12
Refactor internals

lu-zero:
- [ ] unbreak the lookahead_size_properly_bounded tests.


barrbrain (approved):
No externally visible change in behavior and fewer paths were `fi.invalid` is true. :+1:

PR #2691, opened 2021-03-12
Document downsampled() behavior

master-of-zen:
Current `downsampled` is just downscale by factor of 2 and use given width and height for padding.

### DIFF
--- a/v_frame/src/plane.rs
+++ b/v_frame/src/plane.rs
@@ -424,8 +424,8 @@ impl<T: Pixel> Plane<T> {
     }
   }
 
-  /// Downscales plane by factor of 2, witdth and height for padding
-  pub fn downscale_by_2(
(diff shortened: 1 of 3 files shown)


v_frame/src/plane.rs
```diff
@@ -424,7 +424,8 @@ impl<T: Pixel> Plane<T> {
     }
   }
 
-  pub fn downsampled(
```
lu-zero, line 427: would be better to keep the old name to not cause an hard API break.

lu-zero (changes requested):

v_frame/src/plane.rs
master-of-zen: :+1:

barrbrain (approved):

lu-zero (approved):

PR #2697, opened 2021-03-22
Minimal of by_gop parallel encoding

lu-zero:
It uses the internals as they are so there is some overhead due unneeded
computation we will remove later.

- [x] properly restrict the pool usage
- [x] part of the channel-api
  - Only single pass rate-control for now.
- [x] rav1e-ch integration

### DIFF
--- a/src/api/config/mod.rs
+++ b/src/api/config/mod.rs
@@ -128,6 +128,9 @@ pub struct Config {
   pub(crate) threads: usize,
   /// Shared thread pool
   pub(crate) pool: Option<Arc<ThreadPool>>,
+  #[cfg(feature = "unstable")]
+  /// Number of parallel encoding slots
+  pub(crate) slots: usize,
 }
 
 impl Config {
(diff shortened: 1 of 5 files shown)


src/api/config/mod.rs
```diff
@@ -128,6 +128,8 @@ pub struct Config {
   pub(crate) threads: usize,
   /// Shared thread pool
   pub(crate) pool: Option<Arc<ThreadPool>>,
+  /// Number of parallel encoding slots
+  pub(crate) slots: usize,
```
barrbrain, line 132: We can safely excluded this field from stable API builds.
```suggestion
  #[cfg(feature = "unstable")]
  pub(crate) slots: usize,
```

src/bin/common.rs
```diff
@@ -369,6 +370,14 @@ pub fn parse_cli() -> Result<CliOptions, CliError> {
         .help("Overwrite output file.")
         .short("y")
     )
+    .arg(
+      Arg::with_name("SLOTS")
```
barrbrain, line 374: Perhaps this argument should be hidden for stable builds. I'm not sure how to ergonomically exclude it with this builder syntax.

lu-zero: I'll get that sorted.

barrbrain (approved):
This is a good baseline, contained by `#[cfg(feature = "unstable")]`. Prior to future enhancements, we might want at least to ensure builds with this feature in CI.

PR #2698, opened 2021-03-22
Move away the info logging

lu-zero:
Move all the info! lines in the context creation out, it makes a little simpler to rework the internals (see the by_gop poc PRs)


shssoichiro (approved):
Makes sense to me

PR #2699, opened 2021-03-22
Adding no-scene-detection as a CLI argument

BlueSwordM:
The reasoning for this change is that in short scenes without any significant scenes changes, the need for scene detection is null. Therefore, disabling scene detection has no visual or metric deficit, and gives a nice free speedup.

This is particularly useful in per GOP software where scene detection is done externally(like av1an or rav1e-by-gop soon), so this would result in essentially a free speedup.

Average speed difference for each speed preset tested(tested on own machine as AWCY has a large backlog, so numbers may not be indicative of a large dataset)
Speed 4 --- 5,5% speed increase
Speed 6 --- 16% speed increase
Speed 10 --- 9% speed increase

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -214,6 +214,13 @@ pub fn parse_cli() -> Result<CliOptions, CliError> {
         .alias("low_latency")
     )
     .arg(
+      Arg::with_name("NO_SCENE_DETECTION")
+        .help("Disables scene detection entirely\n\
+            Has a significant speed-to-quality trade-off in full encodes. Experimental for rav1e-by-gop")
+        .long("no-scene-detection")
+        .alias("no_scene_detection")
+    )
+    .arg(
       Arg::with_name("RDO_LOOKAHEAD_FRAMES")
         .help("Number of frames encoder should lookahead for RDO purposes\n\
         [default value for speed levels: 10,9 - 10; 8,7,6 - 20; 5,4,3 - 30; 2,1,0 - 40]\n")
@@ -704,6 +711,10 @@ fn parse_config(matches: &ArgMatches<'_>) -> Result<EncoderConfig, CliError> {
   }
 
   cfg.low_latency = matches.is_present("LOW_LATENCY");
+  // Disables scene_detection
+  if matches.is_present("NO_SCENE_DETECTION") {
+    cfg.speed_settings.no_scene_detection = true;
+  }
 
   Ok(cfg)
 }


lu-zero (changes requested):
it should stay in the speed settings, adding a cli override is fine though.

BlueSwordM:
Indeed, I do think it would be a better route to keep it as a speed setting with just a CLI override.

tdaede:
If you set a forced keyframe interval, do we turn this off as well? If not, should we do it automatically in that case?

master-of-zen:
@tdaede 
> If you set a forced keyframe interval, do we turn this off as well?

It handled differently.
scene detection is run on range between min and max keyframe distance, 
and with set forced keyframe placement ` -i X -I X` logic just don't allow scene detection to run 
[CODE](https://github.com/xiph/rav1e/blob/c474d191a06a7855c5f4261f79f4782f620b555f/src/scenechange/mod.rs#L92)
```rust
if distance < self.encoder_config.min_key_frame_interval {
      return false;
    }
    if distance >= self.encoder_config.max_key_frame_interval {
      return true;
    }
```
> If not, should we do it automatically in that case?

Yes, explicit is better than implicit

src/bin/common.rs
```diff
@@ -705,5 +712,8 @@ fn parse_config(matches: &ArgMatches<'_>) -> Result<EncoderConfig, CliError> {
 
   cfg.low_latency = matches.is_present("LOW_LATENCY");
 
+  // Disables scene-detection
+  cfg.no_scene_detection = matches.is_present("NO_SCENE_DETECTION");
```
master-of-zen, line 716: It better to do it in this way
If scene detection would be disabled for speed, it will be respected in future:
```rust 
  if matches.is_present("NO_SCENE_DETECTION") {
    cfg.no_scene_detection = true;
  } else {
    cfg.no_scene_detection = SpeedSettings::no_scene_detection(speed);
  }
```

tdaede:
This does make it redundant with setting min and max keyframe interval though, doesn't it? With this off, you'll always run at max keyframe interval. Is there some case where it isn't?

master-of-zen:
@BlueSwordM This is full fix
```diff
diff --git a/src/api/config/speedsettings.rs b/src/api/config/speedsettings.rs
index be2c9a62..0e8cf8de 100644
--- a/src/api/config/speedsettings.rs
+++ b/src/api/config/speedsettings.rs
@@ -240,7 +240,8 @@ impl SpeedSettings {
     speed <= 3
   }

-  const fn no_scene_detection_preset(_speed: usize) -> bool {
+  /// Currently scenedetection is enabled for all speeds
+  pub fn no_scene_detection_preset(_speed: usize) -> bool {
     false
   }

diff --git a/src/bin/common.rs b/src/bin/common.rs
index d5920e11..a714f5db 100644
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -713,7 +713,10 @@ fn parse_config(matches: &ArgMatches<'_>) -> Result<EncoderConfig, CliError> {
   cfg.low_latency = matches.is_present("LOW_LATENCY");

   // Disables scene-detection
-  cfg.no_scene_detection = matches.is_present("NO_SCENE_DETECTION");
+  if matches.is_present("NO_SCENE_DETECTION") {
+  } else {
+    cfg.no_scene_detection = SpeedSettings::no_scene_detection_preset(speed);
+  }

   Ok(cfg)
 }
```

master-of-zen:
@tdaede 
> This does make it redundant with setting min and max keyframe interval though, doesn't it?

Yes, but it implicit, and feels like hack.
Probably there should be explicit scene detection switch should not allow execution of scene detection, before and outside of min/max scene detection keyframe logic.

>With this off, you'll always run at max keyframe interval. Is there some case where it isn't?

I think yes, it will be always hard cut at max keyframe interval

BlueSwordM:
Did I change it correctly now @lu-zero?

lu-zero:
> Did I change it correctly now like you've asked @lu-zero?
> 
> ```
> if matches.is_present("NO_SCENE_DETECTION") {
>     cfg.no_scene_detection = true;
>   } else {
>     cfg.speed_settings.no_scene_detection =
>       matches.is_present("NO_SCENE_DETECTION");
>    }
> ```

No. Do not change anything on the EncoderContext. Nothing in the library crate should change. Simply override to true  `cfg.speed_settings.no_scene_detection` if the flag is passed to the cli.

BlueSwordM:
I see. I finally understand, thank you.
I will make the necessary changes.

lu-zero:
The changes I asked do not seem to be in this PR, did you forget to push them?

src/api/config/encoder.rs
```diff
@@ -229,6 +229,10 @@ impl fmt::Display for EncoderConfig {
       ("bitrate", self.bitrate.to_string()),
       ("min_quantizer", self.min_quantizer.to_string()),
       ("low_latency", self.low_latency.to_string()),
+      (
+        "no_scene_detection",
+        self.speed_settings.no_scene_detection.to_string(),
+      ),
```
lu-zero, line 235: This has to be removed.

src/api/config/speedsettings.rs
```diff
@@ -240,7 +240,8 @@ impl SpeedSettings {
     speed <= 3
   }
 
-  const fn no_scene_detection_preset(_speed: usize) -> bool {
+  /// Currently, scenedetection is enabled for all speeds
+  pub fn no_scene_detection_preset(_speed: usize) -> bool {
```
lu-zero, line 244: This should be removed.

src/capi.rs
```diff
@@ -688,6 +688,9 @@ unsafe fn option_match(
       enc.rdo_lookahead_frames = value.parse().map_err(|_| ())?
     }
     "low_latency" => enc.low_latency = value.parse().map_err(|_| ())?,
+    "no_scene_detection" => {
+      enc.no_scene_detection = value.parse().map_err(|_| ())?
+    }
```
lu-zero, line 693: this should be removed.

lu-zero (changes requested):
There are still some stray changes, keep only what's in `src/bin/common.rs`

lu-zero (approved):
Thank you :)

vibhoothiiaanand (approved):
Would be nice if the commit message is more descriptive for future reference without hopping to PR:)

PR #2704, opened 2021-03-29
Bump arrayvec to 0.6

shssoichiro:
This is the const generics release of arrayvec. It seems to give about a 1.5% global encode speed improvement. However, it requires a minimum of Rust 1.51.

Before merging:
- [x] Update AWCY to Rust 1.51


lu-zero (approved):
Looks good to me, @tdaede can you bump the compiler on awcy?

PR #2708, opened 2021-03-30
Use compile-time generics instead of dyn dispatch for writers

shssoichiro:
Profiling revealed that some very small Writer functions, such as `bit`, were not being inlined. These functions were sometimes used within loops and therefore the inability to inline resulted in inefficient code. This inability to inline was due to the writers being passed as dynamic objects at runtime. Using generics to monomorphise the functions at compile-time allows the writer functions to be inlined and produces more optimized code.

Testing shows that this produces an overall speed increase of 1-2% at all speed levels.

### DIFF
--- a/src/context/block_unit.rs
+++ b/src/context/block_unit.rs
@@ -695,8 +695,8 @@ impl<'a> ContextWriter<'a> {
     &self.fc.kf_y_cdf[above_ctx][left_ctx]
   }
 
-  pub fn write_intra_mode_kf(
-    &mut self, w: &mut dyn Writer, bo: TileBlockOffset, mode: PredictionMode,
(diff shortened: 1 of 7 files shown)


lu-zero (approved):
It does look good to me. Thank you :)

barrbrain (approved):
This also addresses an issue raised in #2673, that the `dyn` boundary prevents constant propagation (particularly for CDF sizes). One thing worth measuring before merging is impact to binary size.

shssoichiro:
With the current default settings in Cargo.toml, here is the size comparison for a stripped release build:

Before: 3565976 bytes
After: 3660184 bytes

So about 100kb increase.

lu-zero (commented):
The tradeoff is fine IMHO.

PR #2711, opened 2021-03-31
CI: Update to macOS environment, add Arm64 target

EwoutH:
Updates the GitHub Actions to macOS 11.0 environment and add aarch64-apple-darwin target for cargo and cargo-C builds.


lu-zero (approved):
It does look fine, thank you :)

PR #2719, opened 2021-04-16
Check available rows upfront in cdef_dist_wxh_8x8

barrbrain:
Admits full unrolling and autovec for the following loop.
At speed 2, this yields a 9.6% increase in encoding speed.


master-of-zen:
Simple trick, but quite powerful

tmatth (approved):
Is this the kind of the thing an improved compiler could do automatically?

PR #2720, opened 2021-04-21
CI: Update to cargo-c v0.8.0

EwoutH:
Update the CI config to use cargo-c [v0.8.0](https://github.com/lu-zero/cargo-c/releases/tag/v0.8.0).

Currently cargo-c updates automatically for the macOS runs by using Brew, but not on Linux and Windows. Maybe it would be useful to keep those versions in sync.


lu-zero (approved):
Looks good, thank you!

PR #2688, opened 2021-03-11
CI: Add Windows Arm64 job

EwoutH:
Adds a Windows job targeting the Arm64 platform.

Moves Nasm installation to ilammy/setup-nasm and upgrades to Cargo-C v0.7.3.

I'm still working on the deploy CI (on [this branch](https://github.com/EwoutH/rav1e/tree/win-arm64-ci)), that will follow in a separate PR.

Part of #2652.

PR #2721, opened 2021-04-21
CI: Windows: Actually build for specified target in matrix

EwoutH:
I screwed up in #2688.

Currently the Arm64 job in Windows installs the aarch64 target, but doesn't actually build for that target. In the `actions-rs/toolchain@v1` step, both the `override` and `default` [inputs](https://github.com/actions-rs/toolchain#inputs) are set to _true_, which I thought would enable build with this target directly. Turns out this isn't the case, only the _toolchain_ is made default and not the _target_.

This PR fixes that, but unfortunately now the Windows Arm64 build breaks.

### DIFF
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -446,6 +446,8 @@ jobs:
           target: ${{ matrix.target }}
           override: true
           default: true
+      - name: Set compiler & target environment variable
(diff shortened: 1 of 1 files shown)


lu-zero:
What is the assembler in the msvc to use in scenario? setting `CC="clang -target <the-right-triple>"` does fix it?

Luni-4:
What's the status of this PR @EwoutH?

PR #2722, opened 2021-04-21
Run cargo check for unstable features

### LINKED ISSUE #2723: Enable and build unstable features in CI Runs
Currently the CI do not build the unstable APIs for example, channel-api and gop example/poc, we need to extend the Ci to have them,

There can be multiple ways to do this, 
1. Have the default builds with unstable features, [would be time-effective].
2. Have separate builds with unstable features for different targets 
3. Run unstable APIs based on label approach where the PR has label of unstable, then CI picks it and build it.
There can be other way to solve this too, 
FWIW, In future, we will be having lot more unstable features, and we need CI to cover-them.

lu-zero:
Closes #2723

### DIFF
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -54,6 +54,7 @@ jobs:
           - cargo-c
           - check-no-default
           - check-extra-feats
+          - check-unstable-feats
           - fuzz
         include:
           - conf: beta-build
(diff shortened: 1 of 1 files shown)


vibhoothiiaanand (approved):
Thanks for this patch,
LGTM, this do add additional step of unstable features in ci and works as intended.

PR #2728, opened 2021-05-08
Wire in the scenechange detection

lu-zero:


barrbrain:
I tested with a random 3 minute video:

```
rav1e

>  Using 1 tile
>  encoded 4252 frames, 4.826 fps, 100.62 Kb/s, elap. time: 14m 41s                 
>  ----------
>  Key frame:            33 | avg QP: 233.00 | avg size:    2927 B
>  Inter frame:        4219 | avg QP: 255.00 | avg size:     505 B
>  Intra only frame:      0 | avg QP:   0.00 | avg size:       0 B
>  Switching frame:       0 | avg QP:   0.00 | avg size:       0 B
>  ----

rav1e-ch with --slots 2

>  Using 1 tile
>  encoded 4252 frames, 7.449 fps, 100.97 Kb/s, elap. time: 9m 30s                 
>  ----------
>  Key frame:            33 | avg QP: 233.00 | avg size:    2686 B
>  Inter frame:        4219 | avg QP: 255.00 | avg size:     509 B
>  Intra only frame:      0 | avg QP:   0.00 | avg size:       0 B
>  Switching frame:       0 | avg QP:   0.00 | avg size:       0 B
>  ----
```

The key-frame positions were not identical but the results are similar.

src/api/channel/by_gop.rs
```diff
@@ -34,50 +38,55 @@ impl<T: Pixel> SubGop<T> {
...
+    }
 
-    if len > self.pyramid_size {
+    if self.frames > self.pyramid_size {
+      self.frames = 0;
       Some((self.pyramid_size, new_gop))
     } else if new_gop {
-      Some((len - 1, true))
+      self.frames = 0;
+      Some((self.frames, true))
```
barrbrain, line 87: Note that `self.frames` is set to zero before being referenced in the return value. This is probably not correct.

lu-zero: you found the bug :)

lu-zero:
Now the output is bit-identical at least for the vimeo-set I used to test.

barrbrain (approved):

PR #2731, opened 2021-05-12
Remove the dimension validation when setting it in the C API

### LINKED ISSUE #2730: capi's check_frame_size() ignores still_picture mode
**Description:**

rav1e is supported in libavif via it's C API (`capi.rs`). When encoding single images (not sequences) with libavif, libavif will engage `still_picture` mode, but despite recent changes to rav1e to allow for images smaller than `16x16` when in `still_picture` mode, the `capi.rs` layer still restricts smaller images even when the `still_picture` flag is set.

**To Reproduce**
Attempt to encode an image smaller than 16x16 (tested with 1x1 and 3x3) with the C API and `still_picture` enabled. The failure path can be seen using libavif's `avifenc`, such as:

    avifenc -c rav1e 1x1.png 1x1.avif

I've verified if I loosen the restriction in `capi.rs`'s `check_frame_size()`, the encode succeeds. The results aren't ideal (see #151), but this ceases to be a failure case.

**Expected behavior**
Enabling `still_picture` via the C API should allow setting `"width"` and `"height"` (also via the C API) to values less than 16. 

Version: v0.4.1
Operating system: All known to work with libavif (Windows, Linux, macOS)

lu-zero:
The configuration is validated already on encoder instantiation.

Fixes #2730

### DIFF
--- a/src/capi.rs
+++ b/src/capi.rs
@@ -602,40 +602,6 @@ pub unsafe extern fn rav1e_config_unref(cfg: *mut Config) {
   }
 }
 
-fn tile_log2(blk_size: usize, target: usize) -> usize {
-  let mut k = 0;
-  while (blk_size << k) < target {
-    k += 1;
-  }
-  k
-}
-
-fn check_tile_log2(n: Result<usize, ()>) -> Result<usize, ()> {
-  match n {
-    Ok(n) => {
-      if ((1 << tile_log2(1, n)) - n) == 0 || n == 0 {
-        Ok(n)
-      } else {
-        Err(())
-      }
-    }
-    Err(e) => Err(e),
-  }
-}
-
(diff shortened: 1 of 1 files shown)


src/capi.rs
```diff
@@ -623,19 +623,6 @@ fn check_tile_log2(n: Result<usize, ()>) -> Result<usize, ()> {
   }
 }
 
-fn check_frame_size(n: Result<usize, ()>) -> Result<usize, ()> {
```
rzumer, line 626: Instead of removing this, you should set the minimum to 1. We still need to make sure that it fits into a u16 and that it's non-zero (and positive).

rzumer (approved):

rzumer (approved):

PR #2735, opened 2021-05-14
Pull check for allowed intrabc out of write_lrf().

negge:


lu-zero (approved):
It does look fine to me.

PR #2738, opened 2021-05-18
Fix clippy warnings from Rust nightly

shssoichiro:
These will be enabled by default in a future stable Rust,
so fixing them now keeps us ahead of the ball.

Relevant lints:
- used `assert_eq!` with a literal bool
- all if blocks contain the same code at the end
- called `is_none()` after searching an `Iterator` with `find`


lu-zero (approved):
Looks good.

PR #2740, opened 2021-05-20
Consider tune when setting speed presets

shssoichiro:
Temporal RDO depends on tx_domain_distortion being disabled.
Currently it's easy for a user of the API to accidentally
leave tx_domain_distortion enabled if they are using the speed presets.
This change should make it easier to keep settings in a consistent
state.


lu-zero:
Might be better to flip the priority between the two and document the behavior.

PR #2746, opened 2021-05-30
Add a cli option to use the high_bitdepth codepath for 8bit content

lu-zero:


barrbrain:
I was able to find desyncs by adapting the fuzzer to run 8-in-16. Here is one small such config:
```rust
DecodeTestParameters {
    w: 16,
    h: 31,
    speed: 6,
    q: 0,
    limit: 2,
    bit_depth: 8,
    chroma_sampling: Cs420,
    min_keyint: 1,
    max_keyint: 4,
    switch_frame_interval: 0,
    low_latency: false,
    error_resilient: false,
    bitrate: 11523,
    tile_cols_log2: 1,
    tile_rows_log2: 0,
    still_picture: false,
}
```

lu-zero:
Shall we refresh it and try it again ?

lu-zero:
Let's refresh this and see how much breaks in the process :)

barrbrain (approved):
I would somewhat prefer it be a hidden flag or have a "Here there be dragons" warning attached. We can fuzz and run through AWCY to verify.

lu-zero:
Rebased and now with the cli tests wired in.

Issue #2747, opened 2021-05-31
CPU Feature level being shown in version and help


vibhoothi:
Right now the CPU feature level is shown in both version and help page, which seems to not necessarily required, this came after 6cb66707f65949a5bdee9bb0ea8e3d28ba650306 cleanup, might be good for UX for not showing in the versions and help page of the encoder

PR #2748, opened 2021-05-31
Move the cpu feature info line

lu-zero:
Address #2747


vibhoothi (approved):
Checked locally works fine, thanks for the patch

PR #2751, opened 2021-06-03
Use the x86 8-bit CDEF assembly on edges

barrbrain:
The current workaround to use 16-bit intermediates also means that
we do not need to fall back to the Rust implementation on edges.


lu-zero (approved):
Sounds fine. Probably aarch64 should enjoy the same?

barrbrain:
> Sounds fine. Probably aarch64 should enjoy the same?

The current aarch64 integration implements padding in the Rust wrapper for both pixel types.

PR #2754, opened 2021-06-14
CI: Update libaom to 3.1.1-dmo0~bpo10+1

vibhoothi:


lu-zero:
you need to add libvmaf1 as well.

lu-zero (approved):

Issue #2662, opened 2021-02-03
Squared artefacts on image when converting to AVIF 


vatsaljrathod:
Getting square artefacts at the right edge of the image while converting jpg to avif using av1 compression. Adding some example it might be useful.

[Already checked with libheif,](https://github.com/strukturag/libheif/issues/424#issuecomment-770804317) as per them bug should be in encoding application.  One observation is, if we scale down speed setting then we are not getting these artefacts.

[squared_artefacts_samples.zip](https://github.com/xiph/rav1e/files/5920307/squared_artefacts_samples.zip)

adityapatadia:
I should mention that this issue only comes with version 0.4.0. Everything works fine with 0.3.4.

lu-zero:
Could you please provide more information on how you produced the file?

adityapatadia:
It was produces using https://github.com/strukturag/libheif library. That library was in turn used inside https://github.com/lovell/sharp to generate image.

When we downgraded Rav1e, the issue went away hence reported here.

lu-zero:
If you can provide the settings in use and, even better a command line, it would be great.

vatsaljrathod:
Apologies for the delayed response.

Here is the setting I have used. Here speed >5 is causing the issue. 
sharp: 0.27.0
libvips: 8.10.5
libheif: 1.10.0
```
let image = await sharp(inputImage)

  let resizeOptions = {
    kernel: 'lanczos3',
    fastShrinkOnLoad: true,
    width: 388,
    height: null,
    withoutEnlargement: true,
    fit: 'inside'
  };

  let heifOptions = {
    quality: 60,
    compression: 'av1',
    lossless: false,
    speed: 8
  }

  let imageBuffer = await image
    .rotate()
    .resize(resizeOptions)
    .sharpen()
    .heif(heifOptions)
    .toBuffer({resolveWithObject: true});
```

PR #2756, opened 2021-06-26
Maybe a new crates.io release?

kornelski:
There are significant unreleased quality/performance improvements. 

BTW: rav1e has a very long release cycle, IMHO way too slow for the pace of changes in rav1e and around AV1 in general. 

The long release cycle paradoxically makes stable versions of rav1e quite unattractive. The latest not-alpha v0.4 is too far behind in performance/quality to be useful. Currently only the unreleased version of rav1e is competitive with libaom for AVIF compression.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rav1e"
-version = "0.5.0-alpha"
+version = "0.5.0-alpha.2"
 authors = ["Thomas Daede <tdaede@xiph.org>"]
 edition = "2018"
 build = "build.rs"


lu-zero:
I think we could consider stable the new API and get the 0.5 out. We should discuss it the next meeting, but I was already planning to cut a beta next week.

shssoichiro:
I'm also +1 for more frequent point releases. In fact, I might say that we should replace the weekly pre-releases with official point releases including a push to crates.io.

lu-zero:
Ideally we do not have further breaking changes so we could push to crates.io point releases nearly from master. The `unstable` feature would shield users from experiments.

cryptomilk:
I would expect a bugfix release for issues like #2662 being fixed. However I will patch packages in openSUSE and Fedora for now.

It might be a good idea to establish a fixed release cycle. For Samba we are doing well with 6 month, same for darktable.

tdaede:
Now that our API is more stable I think a fixed release cycle is a great idea.

One thing I'd really love to have is a more in-depth test specifically to run on releases to avoid any sort of quality regressions - this would be the distinguishing feature between releases and the weekly builds. I don't think that should block 0.5, but I have some plans for the future.

eclipseo:
I too would welcome a new release, not an alpha one, but a proper one, if it's stable enough. I haven't pushed the 0.5 alpha to Fedora because of this, but the significant gains are so useful.

lu-zero:
We plan to land two features and close 0.5 between this week and the next one. @kornelski I drafted the support for the early exit so you can poke at it and tell me what you think and what you'd like to get as progress data btw.

kornelski:
I don't want any new features. There's a serious encoding bug in the 0.5, so I'd like any release of anything, right now. Features can wait.

barrbrain:
[0.5.0-beta](https://crates.io/crates/rav1e/0.5.0-beta) is now available on crates.io as a prerelease.
This should aid wider testing as we finalise 0.5.

tdaede:
I opened #2777 to set up time-based releases for 0.5.1 and onward.

PR #2758, opened 2021-07-08
Initialise residual when less than the transform width is visible

### LINKED ISSUE #2662: Squared artefacts on image when converting to AVIF

### LINKED ISSUE #2757: Incorrectly encoded edge of a frame
**Describe the bug**
Artifacts at the right edge of the frame:
Screen Shot 2021-06-27 at 14 35 02

**To Reproduce**
Encode resized (I've used `cavif --speed=10 -f image.jpg`)

**Expected behavior**
There should be no dashed lines at the edge.

**Required Information**
This has regressed in cfb9ef1773e8cd288523575febb944d6e3009733

barrbrain:
The input stride for forward transforms did not match the output stride of residual computation in this case. Extend the residual stride to the transform width and zero the non-visible portion.
Fixes #2662. Fixes #2757.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1209,12 +1209,20 @@ pub fn encode_tx_block<T: Pixel, W: Writer>(
       residual,
       &ts.input_tile.planes[p].subregion(area),
       &rec.subregion(area),
-      visible_tx_w,
+      tx_size.width(),
       visible_tx_h,
     );
+    if visible_tx_w < tx_size.width() {
+      for row in residual.chunks_mut(tx_size.width()).take(visible_tx_h) {
+        for a in &mut row[visible_tx_w..] {
+          *a = 0;
+        }
+      }
+    }
   }
-  let visible_area = visible_tx_w * visible_tx_h;
-  for a in residual[visible_area..].iter_mut() {
+  let initialized_area =
+    if visible_tx_w == 0 { 0 } else { tx_size.width() * visible_tx_h };
+  for a in residual[initialized_area..].iter_mut() {
     *a = 0;
   }
 


lu-zero (approved):

kornelski:
After using this commit, I've found a panic:

> thread '<unnamed>' panicked at 'attempt to negate with overflow', /Users/kornel/.cargo/registry/src/github.com-1ecc6299db9ec823/num-traits-0.2.14/src/sign.rs:75:1

      let eob_minus_two =
        scan[1..].iter().rposition(|&i| coeffs[i as usize].abs() >= deadzone);

<pre>
   3: <i16 as num_traits::sign::Signed>::abs
             at num-traits-0.2.14/src/sign.rs:49:41
   4: rav1e::quantize::QuantizationContext::quantize::{{closure}}
             at rav1e/941ac5e/src/quantize.rs:268:41
   5: <core::slice::iter::Iter<T> as core::iter::traits::iterator::Iterator>::rposition
             at core/src/slice/iter/macros.rs:310:24
   6: rav1e::quantize::QuantizationContext::quantize
             at rav1e/941ac5e/src/quantize.rs:268:9
   7: rav1e::encoder::encode_tx_block
             at rav1e/941ac5e/src/encoder.rs:1238:13
   8: rav1e::encoder::write_tx_blocks
             at rav1e/941ac5e/src/encoder.rs:2018:31
   9: rav1e::rdo::rdo_tx_type_decision
             at rav1e/941ac5e/src/rdo.rs:1721:7
  10: rav1e::rdo::rdo_tx_size_type
             at rav1e/941ac5e/src/rdo.rs:759:30
  11: rav1e::rdo::luma_chroma_mode_rdo::{{closure}}
             at rav1e/941ac5e/src/rdo.rs:856:32
  12: rav1e::rdo::luma_chroma_mode_rdo
             at rav1e/941ac5e/src/rdo.rs:945:5
  13: rav1e::rdo::intra_frame_rdo_mode_decision::{{closure}}
             at rav1e/941ac5e/src/rdo.rs:1491:5
</pre>

barrbrain:
> After using this commit, I've found a panic:

It looks like my change surfaced some bugs in (de-)quantization where coefficients are near the limits.

src/encoder.rs
```diff
@@ -1209,11 +1209,18 @@ pub fn encode_tx_block<T: Pixel, W: Writer>(
       residual,
```
ycho, line 1209: I think I did init the residue with zeros in case visible area is smaller than tx size.
I remember I did it when a memory is allocated for the residual, like as early as line 1184 above, i.e. instead of uninitialized, I probably used sth like initialized (if that was available? :) ).
Anyway, thanks for the fix!

src/encoder.rs
```diff
@@ -1209,11 +1209,18 @@ pub fn encode_tx_block<T: Pixel, W: Writer>(
...
+    if visible_tx_w < tx_size.width() {
+      for row in residual.chunks_mut(tx_size.width()).take(visible_tx_h) {
+        for v in &mut row[visible_tx_w..] {
+          *v = 0;
+        }
+      }
+    }
   }
-  let visible_area = visible_tx_w * visible_tx_h;
+  let visible_area = tx_size.width() * visible_tx_h;
```
ycho, line 1223: Probably better to use different variable name than 'visilble_area', such as 'tx_area'?

src/encoder.rs
```diff
@@ -1209,11 +1209,18 @@ pub fn encode_tx_block<T: Pixel, W: Writer>(
...
   for a in residual[visible_area..].iter_mut() {
```
ycho, line 1224: This for loop initializes the whole tx_size residual[ ] as zeros, correct? If so, isn't the computed diff (i.e. residue) values erased? So, I think this for loop can be put before diff( ) is performed, i.e. between line 1207 and 1208, and
lines 1215~1221 can be removed (I think this block of code is zeroing right side of non-visible area in tx area?)

src/encoder.rs
barrbrain: Thank you for taking a look. The primary issue was the inconsistency in stride. This was highlighted in the original TODO you left but that was removed when we reduced the zeroing. My first attempt at addressing the stride accidentally introduced an unzeroed region in the `visible_tx_w == 0` case. This is fortunately resolved now.

src/encoder.rs
barrbrain: This loop zeroes the area below the visible region, to the full width of the transform.

src/encoder.rs
```diff
@@ -1209,12 +1209,20 @@ pub fn encode_tx_block<T: Pixel, W: Writer>(
...
+        for a in &mut row[visible_tx_w..] {
+          *a = 0;
+        }
+      }
+    }
   }
-  let visible_area = visible_tx_w * visible_tx_h;
-  for a in residual[visible_area..].iter_mut() {
+  let initialized_area =
+    if visible_tx_w == 0 { 0 } else { tx_size.width() * visible_tx_h };
```
ycho, line 1224: I believe "visible_tx_w == 0" case should not reach here, bcz of "if visible_tx_w != 0 && visible_tx_h != 0 { " at line 1207?

barrbrain: The converse is true - because of that condition, nothing is initialised in the `visible_tx_w == 0` case.
Specifically, if `visible_tx_h != 0` is also true we need to be cautious not to assume any rows have been initialised.

PR #2760, opened 2021-07-09
Fix fall-through of x86 dispatch_predict_intra for CpuFeatureLevel::RUST

barrbrain:
Found by a failure of `asm::shared::predict::test::pred_matches_u8` when running:
```
RAV1E_CPU_TARGET=rust cargo test --release --lib
```


tmatth (approved):
Confirmed that thits fixes `RAV1E_CPU_TARGET=rust` locally.

PR #2761, opened 2021-07-12
Disable HBD asm without 8-bit support for 8-in-16 and enhance fuzzing

barrbrain:
Pass `check_asm` feature through to rav1e from fuzz crate. Disable HBD asm without 8-bit support for 8-in-16 on x86_64 and aarch64. Add a new fuzz target, `encode_decode_hbd`.


lu-zero (approved):
Looks nice :)

PR #2764, opened 2021-07-27
CI: Update to libaom to 3.1.2-dmo1

barrbrain:
Note that Travis scripts are not maintained as they are inactive.
If we decide to bring back Travis integration, we should upgrade the base image first.


lu-zero (approved):
Thank you!

PR #2765, opened 2021-07-30
Scene detection improvement

AleksandrGushchin:
Adjusting values of thresholds for both versions of the scene detection algorithm.

Improved metric for the fast version. The peaks of the metric are now more distinguishable from the other values, so the threshold cuts off only relevent values.


vibhoothi:
Thanks for the patch,
Do you have any early numbers  for this patch :)

AleksandrGushchin:
> Thanks for the patch,
> Do you have any early numbers for this patch :)

You can see a detailed analysis [here](https://dev.to/aleksandrgushchin/threshold-experiments-for-scene-change-detector-2hbp)

lu-zero:
Please make sure to run rustfmt before committing.

vibhoothi:
> > Thanks for the patch,
> > Do you have any early numbers for this patch :)
> 
> You can see a detailed analysis [here](https://dev.to/aleksandrgushchin/threshold-experiments-for-scene-change-detector-2hbp)

Great, I assume this analysis is done with new scene-change patches ie. 7970d35ebb1424358b837c6e19b0cecf3cc6f8df. Would be nice to add the commit hash of encoder to know.

src/scenechange/mod.rs
```diff
@@ -239,6 +237,14 @@ impl<T: Pixel> SceneChangeDetector<T> {
     let mut cloned_deque = self.score_deque.to_vec();
     cloned_deque.remove(self.deque_offset);
 
+
+    // Subtract the previous metric value from the current one
+    // It makes the peaks in the metric more distinctive
```
tdaede, line 242: I might like a bit more analysis of this. In particular, this effectively limits us to a scene cut only every other frame, correct?

AleksandrGushchin: The visualization you can see at the end of this [post](https://dev.to/aleksandrgushchin/threshold-experiments-for-scene-change-detector-2hbp). The new metric can be interpreted as the rate of change in dissimilarity metric between successive frames. Since the action inside the scene is usually monotonous, the dissimilarity metric stays about the same or changes at a small rate. However, there are scenes whose movement within a frame causes the dissimilarity metric to have high values, even exceeding the values of some scene changes.
The new metric solves this problem by decreasing the values on such scenes. At the same time dissimilarity metric peaks on frames with scene change remain high enough for detection with the threshold.

tdaede: Great, OK it does look like a big improvement.

Have you done a run on AWCY with vimeo-corpus-10s? I can help you do this if you like.

What is the intended behavior for the second frame in a scene, when the previous scene's score isn't available? This looks like it treats the previous frame as having a delta of 0. This seems like it'd effectively make a scenecut here more likely than average always.

Have you considered doing a delta of an average of x previous frames?

src/scenechange/mod.rs
```diff
@@ -367,7 +372,7 @@ impl<T: Pixel> SceneChangeDetector<T> {
           * (distance_from_keyframe - min_keyint) as f64
           / (max_keyint - min_keyint) as f64
     };
-    let threshold = intra_cost * (1.0 - bias);
+    let threshold = intra_cost * (1.0 - bias) / 2.2;
```
tdaede, line 375: You should be able to change THRESH_MAX rather than adding another constant here.

src/scenechange/mod.rs
AleksandrGushchin: > Have you done a run on AWCY with vimeo-corpus-10s? I can help you do this if you like.

Not yet, it would be helpful!

> What is the intended behavior for the second frame in a scene, when the previous scene's score isn't available? 

On the first few frames of the video the previous metric has low values, so subtracting one frame metric value from the other should not exceed the threshold. 

> Have you considered doing a delta of an average of x previous frames?

Not yet, because this new metric solved my problem, but I'll do experiments with it.

vibhoothi (commented):
Can you please rebase to latest master

lu-zero:
Please rebase, not re-merge.
Assuming origin is `xiph/rav1e`, from your branch
```
git fetch origin
git pull --rebase origin/master
```

vibhoothi (commented):
Hi,

I checking in s10 for clip, 
One interesting result is this,
https://beta.arewecompressedyet.com/?job=rav1e-s10-vimeo-new-scene-detection-2021-08-13_103818-cac79b5e&job=rav1e-s10-vimeo-new-scene-detection-2021-08-13_105536-cac79b5e
This is doing s10 runs twice, and for the clip slideshow, we got one extra keyframe with same configuration, this is worth investigating, on why same configuration is giving two different output. 

https://beta.arewecompressedyet.com/analyzer/?maxFrames=4&decoder=https://media.xiph.org/analyzer/inspect.js&decoderName=rav1e-s10-vimeo-new-scene-detection-2021-08-13_103818-cac79b5e&file=https://beta.arewecompressedyet.com/runs/rav1e-s10-vimeo-new-scene-detection-2021-08-13_103818-cac79b5e/vimeo-corpus-10s/slideshow.y4m-252.ivf&decoder=https://media.xiph.org/analyzer/inspect.js&decoderName=rav1e-s10-vimeo-new-scene-detection-2021-08-13_105536-cac79b5e&file=https://beta.arewecompressedyet.com/runs/rav1e-s10-vimeo-new-scene-detection-2021-08-13_105536-cac79b5e/vimeo-corpus-10s/slideshow.y4m-252.ivf

master-of-zen:
@AleksandrGushchin great work, looks good to me :+1:

master-of-zen (approved):

AleksandrGushchin:
Detailed description and analysis can be seen in a [blogpost](https://dev.to/aleksandrgushchin/new-scene-change-detector-version-4ja7)

src/api/config/speedsettings.rs
```diff
@@ -246,8 +246,14 @@ impl SpeedSettings {
     false
   }
 
-  const fn fast_scene_detection_preset(speed: usize) -> bool {
-    speed == 10
+  const fn fast_scene_detection_preset(speed: usize) -> SceneDetectionSpeed {
+    if speed <= 6 {
```
vibhoothi, line 250: Can we expose this option to CLI also, like users could explicitly set different speed detection mode irrespective of default modes? I think that might be a good addition

src/bin/common.rs
```diff
@@ -175,6 +175,13 @@ pub fn parse_cli() -> Result<CliOptions, CliError> {
         .takes_value(true)
         .default_value("6")
     )
+    .arg(
+      Arg::with_name("SCENE_CHANGE_DETECTION_SPEED")
+        .help("Speed level (0 is best quality, 2 is fastest)")
```
vibhoothi, line 180: Might be good if we can say current state in a nice and concise way, 
One idea I had was like this, feel free to edit and modify:) 
```suggestion
        .help("Specify Speed of scene-change detection, 0: best, 1: medium, 2: Fastest mode[default:  0 for s0-s6, 1 for s7-s9, 2 for s10]")
```

src/bin/common.rs
```diff
@@ -592,6 +604,16 @@ fn parse_config(matches: &ArgMatches<'_>) -> Result<EncoderConfig, CliError> {
     .unwrap_or_default();
 
   let mut cfg = EncoderConfig::with_speed_preset(speed);
+
+  if matches.occurrences_of("SCENE_CHANGE_DETECTION_SPEED") != 0 {
+    cfg.speed_settings.fast_scene_detection = match scene_detection_speed {
+      0 => SceneDetectionSpeed::Slow,
+      1 => SceneDetectionSpeed::Medium,
+      2 => SceneDetectionSpeed::Fast,
+      3..=u32::MAX => cfg.speed_settings.fast_scene_detection,
```
vibhoothi, line 613: Do we need to go to upperbound of u32, if doing that might be better to alert user by a warning that it is going to be fast mode(which is 2)? Or just restrict to 0,1,2. and throw error.

Cargo.toml
```diff
@@ -57,6 +57,7 @@ wasm = ["wasm-bindgen"]
 dump_lookahead_data = ["byteorder", "image"]
 
 [dependencies]
+histogram = "*"
```
shssoichiro, line 60: This needs to be set to an actual version (at least a semver-compatible major version). The reason is that crates can't be published to crates.io if they depend on a * version.

vibhoothi, line 60: Would be safer to have explicit version :)
```suggestion
histogram = "0.6.9"
```

vibhoothi (commented):
The patchset looks good, we need to do some more AWCY runs if we have get different results on same configuration.

shssoichiro:
@vibhoothi Do you have any changes that still need to be made? I see that this has one approval but wanted to ensure your changes were addressed before merging.

vibhoothi:
@shssoichiro before merging, I would like to get a review from @tdaede, and also AWCY runs with new changes since the new modes were not ran through AWCY.

vibhoothi (commented):
Please squash relevant changes and cleanup the commits.

vibhoothi:
I have submitted again new set of runs in AWCY, again the S10 is giving non-deterministic results for QP >172.  Would be helpful if that can be investigated. 

Secondly, there are lot of runs for current HEAD vs beta on AWCY, would be amazing if you could try to check them individually and drop a summary here, so will be helpful for others to see. For instance why clip X is behaving strange over otehrs at speed-preset Y. 

The AWCY gives us RD-curves, BD-rate tables, csv export per-qp level data, and also instnat inbrowser bitstream analyzer for all runs.

src/scenechange/mod.rs
```diff
@@ -319,6 +341,7 @@ impl<T: Pixel> SceneChangeDetector<T> {
     previous_keyframe: u64,
   ) -> ScenecutResult {
     let frame2_ref2 = Arc::clone(&frame2);
+
```
vibhoothi, line 344: Think this is extra whitespace added? 
```suggestion

```

lu-zero:
It needs a rebase, @AleksandrGushchin if you rebase it now I will land it today :)

src/bin/common.rs
```diff
@@ -175,6 +175,14 @@ pub fn parse_cli() -> Result<CliOptions, CliError> {
         .takes_value(true)
         .default_value("6")
     )
+    .arg(
+      Arg::with_name("SCENE_CHANGE_DETECTION_SPEED")
+        .help("Speed level for scene-change detection, 0: best quality, 1: speed-to-quality trade-off, 2: fastest mode\n\
+          [default:  0 for s0-s6, 1 for s7-s9, 2 for s10]")
+        .long("scd_speed")
+        .takes_value(true)
+        .default_value("2")
```
shssoichiro, line 184: I think we want to remove the default value here, because by default we want it set according to the speed level.

shssoichiro (changes requested):

src/bin/common.rs
AleksandrGushchin: It is set according to the speed level by default on line 607.
`cfg.speed_settings.fast_scene_detection` changes on lines 609-617 only when the user specifies a value of scene detection speed.

shssoichiro: Have we tested that? My concern was that `matches.occurrences_of("SCENE_CHANGE_DETECTION_SPEED")` would pick up the default value of "2" specified here. I'd like to verify that's not the case.

lu-zero: matches.occurrences_of("SCENE_CHANGE_DETECTION_SPEED") is 0 if no flags are passed. If that's your concern.

shssoichiro (approved):

PR #2769, opened 2021-08-06
Import x86 assembly for dav1d 0.9.1

shssoichiro:
The majority of this assembly targets high bit-depth encoding for SSSE3/SSE4.1, so this will primarily benefit older machines that do not have AVX2.

There are minor improvements to AVX2 which result in about a 1% speedup at speed 6.

With only SSE4.1 available, this provides a 24% speedup on high bit-depth content at speed 6.


negge:
What about the arm assembly?

shssoichiro:
I don't have an arm machine to test on. Usually mindfreeze handles those. From IRC:

10:04 AM <lu_zero> mindfreeze: willing to pick the aarch64 parts?
10:07 AM <mindfreeze> lu_zero: sure, I was checking the changes
10:07 AM <mindfreeze> https://code.videolan.org/videolan/dav1d/-/commits/0.9.1/src/arm
10:07 AM <mindfreeze> It looks like it is Filmgrain changes, we could import them, but it will be not useful
10:07 AM <mindfreeze> at this point

lu-zero (approved):
Thank you, let's land it :)

PR #2773, opened 2021-08-09
fix some warnings

remilauzier:


lu-zero (approved):
It does look good, I wonder why my rustc is not spotting them.

PR #2774, opened 2021-08-09
fix some warnings found by intellij idea

remilauzier:


lu-zero:
Thank you, could you please split the different changes?

remilauzier:
I am not sure how to split that without losing everything?

PR #2775, opened 2021-08-09
Replace some deprecated call

remilauzier:
Replace deprecated use of std::{number} and min/max_value


vibhoothi (commented):
Please split the commit message into commit message and body.

lu-zero (approved):
Looks good, thank you a lot!

PR #2776, opened 2021-08-10
Fix some warnings

remilauzier:
Most of it is to fix hidden anonymous lifetime that are deprecated and will be remove in rust 2021.

### DIFF
--- a/crates/vergen/src/lib.rs
+++ b/crates/vergen/src/lib.rs
@@ -83,6 +83,12 @@
 //! See the [generate_version_rs](crate::output::codegen::generate_version_rs) documentation
 //! if you wish to use this method.
 #![deny(
+  bare_trait_objects,
+  rust_2018_compatibility,
+  rust_2018_idioms,
+  nonstandard_style,
+  future_incompatible,
+  unused,
   missing_docs,
   missing_debug_implementations,
   missing_copy_implementations,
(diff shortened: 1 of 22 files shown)


crates/vergen/src/lib.rs
```diff
@@ -83,6 +83,12 @@
 //! See the [generate_version_rs](crate::output::codegen::generate_version_rs) documentation
 //! if you wish to use this method.
 #![deny(
+  bare_trait_objects,
+  rust_2018_compatibility,
+  rust_2018_idioms,
+  nonstandard_style,
+  future_incompatible,
+  unused,
```
tdaede, line 91: Isn't unused default?

lu-zero:
It does not seem to survive the CI, I'd not merge this until later.

crates/vergen/src/lib.rs
remilauzier: Not all the lints are deny by default.

remilauzier:
On my end, i can't see why the Ci hit the timeout. Do you see the error that did that on your end?

lu-zero:
It unhelpfully states
```
Run timed out after > 72 hours.
```

I'd postpone this patch till the compiler is more stable and the edition is set in stone, if it isn't something you already built upon other patches.

remilauzier:
Can you tried to restart the CI pass on your end, to see if the result will be different?

remilauzier:
I have made the CI pass.

PR #2788, opened 2021-08-31
Remove dead branches in scenecut code

shssoichiro:
This function never gets called if distance < min_keyint, so these two branches are unnecessary. (If distance == min_keyint, then the math works out so that the result equals THRESH_MIN.)


lu-zero (approved):
Looks fine to me.

shssoichiro:
Verified no change in BD rate on vimeo set: [AWCY](https://beta.arewecompressedyet.com/?job=master-sc%402021-08-31T13%3A06%3A11.743Z&job=sc-cleanup-1%402021-08-31T13%3A05%3A27.050Z)

PR #2792, opened 2021-09-01
Histogram fixup

lu-zero:


src/api/lookahead.rs
```diff
@@ -145,8 +145,10 @@ pub(crate) fn estimate_inter_costs_histogram_blocks<T: Pixel>(
         height: IMPORTANCE_BLOCK_SIZE,
       });
 
-      let mut histogram_org =
-        Histogram::configure().max_value(256).build().unwrap();
+      let mut histogram_org = Histogram::configure()
+        .max_value(i16::max_value() as u64)
```
tdaede, line 149: If this is really binning pixel values then the max should be 4096.

shssoichiro:
Verified this does fix the issue for HBD (with 4096 as the max value).

lu-zero:
Done :)

shssoichiro (approved):

PR #2794, opened 2021-09-07
Remove the histogram crate

shssoichiro:
It was extremely slow, especially at higher max values.
We can just calculate the mean directly by summing and division.

This version makes the slow scenechange only 15% slower than medium. (Approximately a 4x speedup compared to before.)
As part of a speed 6 encode, the difference in full encode time is 1-2% between slow and medium scenecut.

There is a slight AWCY difference, likely due to differences in how the means are rounded.
I don't feel that this is significant enough to hold up this PR, although I am open to opinions.

https://beta.arewecompressedyet.com/?job=master-vimeo%402021-09-06T18%3A36%3A16.994Z&job=no-histogram-try3%402021-09-06T19%3A26%3A04.175Z

### DIFF
--- a/src/api/lookahead.rs
+++ b/src/api/lookahead.rs
@@ -14,7 +14,6 @@ use crate::rayon::iter::*;
 use crate::tiling::{Area, TileRect};
 use crate::transform::TxSize;
 use crate::{Frame, Pixel};
-use histogram::Histogram;
 use rust_hawktracer::*;
 use std::sync::Arc;
 use v_frame::pixel::CastFromPrimitive;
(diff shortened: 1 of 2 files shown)


src/api/lookahead.rs
```diff
@@ -145,31 +144,30 @@ pub(crate) fn estimate_inter_costs_histogram_blocks<T: Pixel>(
...
-      let mut histogram_ref =
-        Histogram::configure().max_value(max_value).build().unwrap();
+      let mut histogram_ref_sum = 0f64;
       let iter_ref = region_ref.rows_iter();
       for row in iter_ref {
         for pixel in row {
-          let cur = i16::cast_from(*pixel);
-          histogram_ref.increment(cur as u64).unwrap();
+          let cur = u16::cast_from(*pixel);
+          histogram_ref_sum += cur as f64;
```
tdaede, line 163: a u64 should fully contain this without overflowing.

src/api/lookahead.rs
```diff
@@ -145,31 +144,30 @@ pub(crate) fn estimate_inter_costs_histogram_blocks<T: Pixel>(
...
         }
       }
 
-      let mean = (histogram_org.mean().unwrap() as i32
-        - histogram_ref.mean().unwrap() as i32)
-        .abs();
+      let mean = ((histogram_org_sum / count as f64)
```
tdaede, line 167: In the end it looks like this is effectively just doing the SSE of downscaled frames. Maybe we could just use our existing downscaling code.

shssoichiro: i.e. downscaled by IMPORTANCE_BLOCK_SIZE?

src/api/lookahead.rs
```diff
@@ -145,31 +144,30 @@ pub(crate) fn estimate_inter_costs_histogram_blocks<T: Pixel>(
...
-          histogram_ref.increment(cur as u64).unwrap();
+          let cur = u16::cast_from(*pixel);
+          histogram_ref_sum += cur as u64;
         }
       }
 
-      let mean = (histogram_org.mean().unwrap() as i32
-        - histogram_ref.mean().unwrap() as i32)
-        .abs();
+      let mean = ((histogram_org_sum as f64 / count as f64)
```
tdaede, line 167: You can do just an integer divide here as well (with adding half of count for rounding).

src/api/lookahead.rs
tdaede: Yeah I think ultimately that's the case, though I think our downscaling has SIMD so I wasn't totally sure.

tdaede (approved):

lu-zero:
This should land squashed.

PR #2798, opened 2021-09-08
Reuse allocations of `Plane<T>` for downscaling when using fast scene detection

redzic:


src/scenechange/mod.rs
```diff
@@ -308,8 +308,12 @@ impl<T: Pixel> SceneChangeDetector<T> {
       let frame2_scaled = frame2.planes[0].downscale(self.scale_factor);
       self.frame_buffer.push(frame2_scaled);
     } else {
-      self.frame_buffer.remove(0);
-      self.frame_buffer.push(frame2.planes[0].downscale(self.scale_factor));
+      // swap the first and second elements (instead of removing the first and then
+      // pushing a new frame) so that we can reuse the allocation of the `Plane<T>`
+      // inside `self.frame_buffer`
+      self.frame_buffer.swap(0, 1);
+      frame2.planes[0]
```
shssoichiro, line 315: I don't think this ever actually puts the downscaled frame2 onto the buffer, unless I'm mistaken. The frames in the buffer are still the previous two frames, just swapped in order, and frame2 is downscaled then thrown away.

shssoichiro (changes requested):

src/scenechange/mod.rs
redzic: The previous code removed the first element and then pushed the new frame into the buffer, which also causes the second element to become the first element, and then the new frame is the second element. Here, we are swapping the frames (which also achieves causes the second element to become the first element), but then are reusing the already-existing `Plane<T>` to store the downscaled frame, which is written to by `downscale_in_place` since it takes a mutable reference to the plane. I could also be mistaken, but I don't think the downscaled frame is thrown away.

shssoichiro: Ah, I see. downscale_in_place is taking the buffer as an argument. So I was reading the code wrong.

shssoichiro (approved):

master-of-zen:
LGTM

redzic:
I didn't like the fact that `frame_buffer` was a `Vec`, so I changed it to a fixed-size heap-allocated array, which also eliminates some bounds checks since the compiler statically knows the size of the buffer.

src/scenechange/mod.rs
```diff
@@ -200,7 +194,7 @@ impl<T: Pixel> SceneChangeDetector<T> {
 
     if scenecut {
       // Clear buffers and deque
-      self.frame_buffer.clear();
+      self.frame_buffer = None;
```
redzic, line 197: Actually, do we need to clear the frame buffer if a scenecut is found? This is something the original version did, but I don't know if it's actually necessary

shssoichiro: I think we do (and should) because after a scenecut is found, we're not guaranteed to look at the next frame (because of min scenecut distance). So we need to clear it to make sure the frames in the buffer remain contiguous.

src/scenechange/mod.rs
```diff
@@ -301,24 +295,29 @@ impl<T: Pixel> SceneChangeDetector<T> {
...
-      inter_cost: delta as f64,
+      ScenecutResult {
+        intra_cost: self.threshold as f64,
+        threshold: self.threshold as f64,
+        inter_cost: delta as f64,
+      }
+    } else {
+      // SAFETY: `self.frame_buffer` was just initialized in this function,
+      // and must be a `Some` variant
+      unsafe { std::hint::unreachable_unchecked() }
```
KyleSiefring, line 320: There shouldn't be that much overhead from get() compared to the other tasks of this function (downscaling).

KyleSiefring (commented):
I would avoid unsafe unless it's really critical to performance.

src/scenechange/mod.rs
```diff
@@ -301,24 +295,29 @@ impl<T: Pixel> SceneChangeDetector<T> {
...
+      frame_buffer.swap(0, 1);
+      frame2.planes[0]
+        .downscale_in_place(self.scale_factor, &mut frame_buffer[1]);
     } else {
-      self.frame_buffer.remove(0);
-      self.frame_buffer.push(frame2.planes[0].downscale(self.scale_factor));
+      self.frame_buffer = Some(Box::new([
+        frame1.planes[0].downscale(self.scale_factor),
+        frame2.planes[0].downscale(self.scale_factor),
+      ]));
```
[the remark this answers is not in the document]
redzic: I'm a little unsure what you mean -- what would it look like for the frame buffer to be extracted from the conditions? Do you mean for the `Box` to be reused even after it is cleared (which is what the original code does, since `.clear()`ing a `Vec` removes the elements, but keeps the capacity which will avoid future allocations caused by this branch)?

Or do you mean for `delta_in_planes` to be inside both branches, so as to avoid using `Option::unwrap` by being able to directly reference the original frame buffer?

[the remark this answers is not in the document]
KyleSiefring: Something like
```
    let frame_buffer = if let Some(frame_buffer) = self.frame_buffer.as_deref_mut() {
      frame_buffer.swap(0, 1);
      frame2.planes[0]
         .downscale_in_place(self.scale_factor, &mut frame_buffer[1]);
      &frame_buffer
    } else {
      self.frame_buffer.remove(0);
      self.frame_buffer.push(frame2.planes[0].downscale(self.scale_factor));
      let frame_buffer = Box::new([
        frame1.planes[0].downscale(self.scale_factor),
        frame2.planes[0].downscale(self.scale_factor),
      ]);
      self.frame_buffer = Some(frame_buffer);
      &frame_buffer
    }
```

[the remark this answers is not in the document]
redzic: That's a great idea that I didn't think of, but unfortunately it doesn't seem to work with the borrow checker. Doing this:
```rust
    let frame_buffer =
      if let Some(frame_buffer) = self.frame_buffer.as_deref_mut() {
        frame_buffer.swap(0, 1);
        frame2.planes[0]
          .downscale_in_place(self.scale_factor, &mut frame_buffer[1]);
        todo!()
      } else {
        let frame_buffer = Box::new([
          frame1.planes[0].downscale(self.scale_factor),
          frame2.planes[0].downscale(self.scale_factor),
        ]);
        self.frame_buffer = Some(frame_buffer);
        &frame_buffer
      };
```
causes the error:
```rust
error[E0382]: borrow of moved value: `frame_buffer`
   --> src/scenechange/mod.rs:310:9
    |
305 |         let frame_buffer = Box::new([
    |             ------------ move occurs because `frame_buffer` has type `Box<[v_frame::plane::Plane<T>; 2]>`, which does not implement the `Copy` trait
...
309 |         self.frame_buffer = Some(frame_buffer);
    |                                  ------------ value moved here
310 |         &frame_buffer
    |         ^^^^^^^^^^^^^ value borrowed here after move
```
The only way I could get it to work was by returning `self.frame_buffer.as_deref().unwrap()`, but that seems to defeat the point of returning a reference to the frame buffer within each branch, unless there is some other way to convince the compiler without `unsafe` code that `self.frame_buffer` within the branch is always `Some`.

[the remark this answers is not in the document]
redzic: Even if I solved the borrow checker-related errors for the branch that heap-allocates a new frame buffer by doing `delta_in_planes` inside the branch and using an early return, the other branch still gave similar (to the original posted) borrow checker errors. Even though I am essentially certain that the `std::hint::unreachable_unchecked` cannot actually be reached, I have replaced it with its safe counterpart `unreachable!` so as to avoid `unsafe` code, since the branch likely does not affect the performance much.

redzic:
I also changed the code to use a `bool` to keep track of whether the data in the frame buffer is initialized, so that "clearing" the buffer doesn't actually perform a deallocation, which allows the `Plane<T>` allocations to be completely reused for the entire scene detection, so that there are only ever 2 allocated planes for downscaled frames.

src/scenechange/mod.rs
redzic: I changed this to use `unreachable!()`, so it should be fine now.

src/scenechange/mod.rs
```diff
@@ -200,7 +199,9 @@ impl<T: Pixel> SceneChangeDetector<T> {
 
     if scenecut {
       // Clear buffers and deque
-      self.frame_buffer.clear();
+      if let Some((_, is_initialized)) = &mut self.frame_buffer {
+        *is_initialized = false;
+      }
```
shssoichiro, line 204: This issue has preceded this branch, but I think we also need to clear the frame buffer and score deque on line 147, where we check for max keyframe interval:

```
if distance >= self.encoder_config.max_key_frame_interval {
  return true;
}
```

Since that block returns immediately, I think it could cause some odd behavior if the max scenecut distance is reached because it won't clear the buffer and score queue.

redzic: I have added the buffer and score deque clears to the max key frame interval check.

lu-zero:
Do you want to land it before the beta is cut today?

redzic:
@lu-zero sure, I would like to land it before the beta

PR #2801, opened 2021-09-09
Write the 'monochrome' bit correctly

wantehchang:
The original code write the 'monochrome' bit with the value
seq.bit_depth == 1. I suspect this comes from misunderstanding what
monochrome means in AV1. In AV1, monochrome is what is usually known as
grayscale, rather than black and white. Therefore the correct check for
monochrome should be seq.chroma_sampling == ChromaSampling::Cs400.

Also change the comment "sample_position" to "chroma_sample_position"
because chroma_sample_position is the name of that field in the AV1
ISOBMFF specification.

### DIFF
--- a/src/api/channel/data.rs
+++ b/src/api/channel/data.rs
@@ -209,10 +209,10 @@ impl<T: Pixel> PacketReceiver<T> {
         bw.write_bit(false)?; // tier
         bw.write_bit(seq.bit_depth > 8)?; // high_bitdepth
         bw.write_bit(seq.bit_depth == 12)?; // twelve_bit
-        bw.write_bit(seq.bit_depth == 1)?; // monochrome
+        bw.write_bit(seq.chroma_sampling == ChromaSampling::Cs400)?; // monochrome
         bw.write_bit(seq.chroma_sampling != ChromaSampling::Cs444)?; // chroma_subsampling_x
         bw.write_bit(seq.chroma_sampling == ChromaSampling::Cs420)?; // chroma_subsampling_y
-        bw.write(2, 0)?; // sample_position
+        bw.write(2, 0)?; // chroma_sample_position
         bw.write(3, 0)?; // reserved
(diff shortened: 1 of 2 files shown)


rzumer (approved):

barrbrain:
I note that we currently have the correct logic in `write_color_config()` which is called by `UncompressedHeader::write_sequence_header_obu()`. See [src/header.rs:393-398](https://github.com/xiph/rav1e/blob/469220833/src/header.rs#L393-L398).
```rust
    let monochrome = seq.chroma_sampling == ChromaSampling::Cs400;
    if seq.profile == 1 {
      assert!(!monochrome);
    } else {
      self.write_bit(monochrome)?; // mono_chrome
    }
```

I am curious why we have 2 implementations of `sequence_header_inner()` but it makes sense that they are distinct from `UncompressedHeader::write_sequence_header_obu()`.

tdaede:
Maybe separately we can rename sequence_header_inner() to config_box_inner() or something to be less confusing. I would also be tempted to change the C API name but probably not worth the churn.

@wantehchang Could you rebase? I'll merge once it's rebased.

wantehchang:
@tdaede Hi Thomas, I think GitHub provides an option named "Rebase and merge". Do you see that option on this webpage?

If not, what are the git commands I can use to rebase this pull request? I am not very familiar with git or GitHub. Thanks!

tdaede:
Normally yes, the UI allows it, but I think it may have drifted enough to not work. On your set-monochrome-correctly branch:
```
git fetch origin
git rebase origin/master
git push -f <your_remote_name>
```
Otherwise I think I can do it too.

tdaede:
Actually I was able to do it locally.

wantehchang:
@tdaede I saw an "Update branch" button on the page. I pressed the button. Is this PR up to date now?

tdaede:
That one added a merge commit, but I was able to fix it. I'll land once the tests pass, and then we'll put out a release.

PR #2803, opened 2021-09-10
Add rec and source to capi.

tdaede:


src/capi.rs
```diff
@@ -218,7 +218,7 @@ impl EncContext {
   fn receive_packet(&mut self) -> Result<Packet, rav1e::EncoderStatus> {
     fn receive_packet<T: rav1e::Pixel>(
       ctx: &mut rav1e::Context<T>,
-    ) -> Result<Packet, rav1e::EncoderStatus> {
+    ) -> Result<Packet, rav1e::EncoderStatus> where FrameInternal: From<Arc<v_frame::frame::Frame<T>>> {
```
lu-zero, line 221: Why do you need to express that bound?

tdaede: I am not sure, it is pretty gross. The compiler suggested it to me and it doesn't work without. Is there a way to tell it that the From trait is implemented on all the FrameInternal variants?

lu-zero (approved):

src/capi.rs
lu-zero: Now I see why, either we do the blanket for every `T: Pixel` or keep this.

lu-zero (approved):
Looks fine.

lu-zero (approved):

PR #2807, opened 2021-09-15
cdef: enforce 8 element size for first_max_element

tmatth:
~This is not needed currently as cost is always 8 elements long but will guard against
unexpected breakage should the function be used on single element lists in the future.~

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -65,7 +65,7 @@ pub(crate) mod rust {
   ///
   /// Panics if `elems` is empty
   #[inline]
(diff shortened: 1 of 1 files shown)


redzic:
I think it would be better to just enforce that the function can only be called on 8-element lists by just taking a `&[i32; 8]` as the argument, as this also allows for better optimization

redzic:
Btw my PR (#2806) also already does this, and also updates the doc comment

tmatth:
> Btw my PR (#2806) also already does this, and also updates the doc comment

Ah ok will close.

PR #2808, opened 2021-09-16
Better autovectorization of `delta_in_planes` for faster fast scene detection

redzic:
By getting a subslice of the plane before the actual calculation of the
delta, the compiler is able to utilize SIMD instructions, but can't for
some reason when doing `.take()`. When compiling with AVX2 enabled, this
results in about a 5x speedup for `delta_in_planes`.

Difference in generated assembly (simplified for example):
https://godbolt.org/z/oTKe7n5dr

### DIFF
--- a/src/scenechange/mod.rs
+++ b/src/scenechange/mod.rs
@@ -426,10 +426,11 @@ impl<T: Pixel> SceneChangeDetector<T> {
     let lines = plane1.rows_iter().zip(plane2.rows_iter());
 
     for (l1, l2) in lines {
+      let l1 = l1.get(..plane1.cfg.width).unwrap_or(l1);
+      let l2 = l2.get(..plane1.cfg.width).unwrap_or(l2);
       let delta_line = l1
         .iter()
         .zip(l2.iter())
-        .take(plane1.cfg.width)
         .map(|(&p1, &p2)| {
           (i16::cast_from(p1) - i16::cast_from(p2)).abs() as u32
         })


src/scenechange/mod.rs
```diff
@@ -426,10 +426,11 @@ impl<T: Pixel> SceneChangeDetector<T> {
     let lines = plane1.rows_iter().zip(plane2.rows_iter());
 
     for (l1, l2) in lines {
+      let l1 = l1.get(..plane1.cfg.width).unwrap_or(l1);
```
lu-zero, line 429: `l1[..plane1.cfg.width]` produces worse code? if it does put a note to try `unwrap_unchecked()` once it is stable.

redzic: It does seem to generate slightly worse code, as it adds the bounds check and panicking code.

I can completely remove the bounds check, but it would probably be easier to do that with just `get_unchecked` which works on stable. Also side note, the equivalent of `unwrap_unchecked` can be used on stable by doing `unwrap_or_else(|| std::hint::unreachable_unchecked())`

shssoichiro (approved):
Impressive, especially given that fast scenechange is already ridiculously fast.

src/scenechange/mod.rs
shssoichiro: I'm actually slightly on the fence about using unsafe code here to eliminate the bounds check and panic code. Given that the bounds check occurs outside of the loop, the performance hit should be minimal. I might leave the final call to @lu-zero

redzic:
At least on a 720p source, the main bottleneck for fast scenecut actually seems to be downscaling. Maybe the downscaling factor per resolution should be adjusted? On the 720p source, this new fast version of `delta_in_planes` only accounted for about 3-4% of the runtime of `fast_scenecut`

src/scenechange/mod.rs
redzic: If we are going to remove the unsafe code my opinion on this would be to use the original solution (`l1.get(..plane1.cfg.width).unwrap_or(l1)`), since compared to the unsafe version it only adds 4 instructions (2 cmp and 2 cmov) to the beginning of each iteration over the lines (using [] causes a lot more, plus the panicking code). It is correct because in the case that that `plane1.cfg.width > l1.len()` (which I think is impossible anyway), you wouldn't be comparing the padding.

shssoichiro:
I'm open to that, if you want to do some testing and if changing the downscaling is an improvement, submit it as a separate PR. I think this one is good to go as an isolated change.

PR #2809, opened 2021-09-16
Do not downscale if `self.scale_factor == 1`

redzic:
This removes unnecessary allocations and overhead.

### DIFF
--- a/src/scenechange/mod.rs
+++ b/src/scenechange/mod.rs
@@ -27,12 +27,17 @@ pub struct SceneChangeDetector<T: Pixel> {
   /// scaling factor for fast scene detection
   scale_factor: usize,
   /// Frame buffer for scaled frames
-  frame_buffer: Option<(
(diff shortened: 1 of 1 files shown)


redzic:
This makes fast scenecut significantly faster on clips that use a downscale factor of 1, as it completely avoids the overhead of downscaling.

Tested on 240x240 clip:

Performance of `fast_scenecut` on the clip before this change:

image

Performance of `fast_scenecut` on the clip after this change:

image

So it is about 17x faster on that clip. There shouldn't be any chance in awcy.

tdaede (approved):
Looks OK.

On a side note, we use halfres and qres frames for the motion search prepasses later on. It would be nice if we could reuse the frames scaled here, which would make the downscales "free"

master-of-zen:
LGTM, great work @redzic :+1:

master-of-zen (approved):

PR #2811, opened 2021-09-18
Fix some new clippy lints

shssoichiro:
Some unused fields as well as Default impls that can be derived automatically.


src/scenechange/mod.rs
```diff
@@ -338,7 +338,6 @@ impl<T: Pixel> SceneChangeDetector<T> {
       let delta = self.delta_in_planes(&frame_buffer[0], &frame_buffer[1]);
 
       ScenecutResult {
-        intra_cost: self.threshold as f64,
```
tdaede, line 341: hmm this looks almost like a previous rebase error...

lu-zero (approved):
Seems fine if it survives the rebase.

shssoichiro:
Note that there is one remaining clippy warning. This is a false positive and I've report it upstream: https://github.com/rust-lang/rust-clippy/issues/7683

PR #2813, opened 2021-09-20
Merge medium and slow scenechange methods into one method

shssoichiro:
This changeset introduces a significant number of changes to the way that scene detection works internally.

- All frames are now analyzed. Previously, frames that were within `min-keyint` frames from a keyframe were not analyzed at all. Analyzing these frames, even if we know they will not be set as keyframes, improves the accuracy of both the histogram-based cost adjustments as well as flash detection.
- There are several fixes to how flash detection works, as it contained a number of bugs in the previous iteration.
- The histogram-based cost adjustments were improved to give more accurate results, especially in high detail scenes with motion. The threshold bias was adjusted to account for this.
- The last few frames of the video are now ignored, similar to a scene flash. This removes the issue where it was common for the last frame of the video to be set as a keyframe.
- The cost analysis is performed using a hybrid of the previous two methods. The previous "medium" algorithm, as the primary algorithm, will be used for flash detection and will have the cost adjustments applied to it. This was selected as the primary because the intra/inter-cost based algorithm was more accurate at detecting a number of scenarios such as pans, fades, and blur-ins. The previous "slow" algorithm will be used as a secondary check to filter out some false positives that were present in the "medium" algorithm.

Note that the fast algorithm is unchanged and mostly unaffected by this changeset. The only change affecting it is disallowing keyframes at the end of the video. [AWCY](https://beta.arewecompressedyet.com/?job=fast_sc_master%402021-09-20T14%3A32%3A39.865Z&job=fast_sc_sanity_run2%402021-09-21T01%3A44%3A38.657Z)

This changeset was primarily tested against the vimeo-corpus-10s set from AWCY, both within AWCY and manually. The manual testing involved checking each clip to verify that the selected keyframes matched what would be expected by a typical human viewer. For edge cases where it was not clear if a frame should be chosen, BD Rate was considered--if a keyframe improved BD Rate, it should be chosen, otherwise it should not be (for example, frame 31 in slideshow.y4m should be chosen, and frame 40 in ffmpeg.y4m should not be chosen). Manual verification confirmed perfect matches to the expected keyframes on all clips except for brothers.y4m, on which it was impossible to determine expected keyframe placement due to the nature of the video.

AWCY runs can be found here:
- Previous "slow" to new hybrid: [Link](https://beta.arewecompressedyet.com/?job=sc_baseline%402021-09-18T08%3A00%3A40.471Z&job=sc_testing_hybrid%402021-09-19T22%3A59%3A34.162Z)
- Previous "medium" to new hybrid: [Link](https://beta.arewecompressedyet.com/?job=sc_baseline_medium%402021-09-18T08%3A07%3A35.324Z&job=sc_testing_hybrid%402021-09-19T22%3A59%3A34.162Z)

Performance was tested against the 720p full video of Big Buck Bunny, showing the following:
- Fast: 650 fps
- Old Medium: 163 fps
- Old Slow: 272 fps (yes, it wasn't actually slower than medium, it turns out)
- New Method: 151 fps

This equates to a 7.5% drop compared to the previous medium method. This should be considered still fast enough to be useful at most speed levels, and as such, this changeset uses it as the default for speeds 0-9.

### DIFF
--- a/src/scenechange/mod.rs
+++ b/src/scenechange/mod.rs
@@ -18,6 +18,19 @@ use std::sync::Arc;
 use std::{cmp, u64};
 // use crate::api::*;
 
+// The fast implementation is based on a Python implementation at
+// https://pyscenedetect.readthedocs.io/en/latest/reference/detection-methods/.
+// The Python implementation uses HSV values and a threshold of 30. Comparing the
+// YUV values was sufficient in most cases, and avoided a more costly YUV->RGB->HSV
+// conversion, but the deltas needed to be scaled down. The deltas for keyframes
+// in YUV were about 1/3 to 1/2 of what they were in HSV, but non-keyframes were
+// very unlikely to have a delta greater than 3 in YUV, whereas they may reach into
+// the double digits in HSV.
+//
+// Experiments have shown that these thresholds are optimal.
+const FAST_THRESHOLD: f64 = 18.0;
+const IMP_BLOCK_DIFF_THRESHOLD: f64 = 7.0;
+
 /// Runs keyframe detection on frames from the lookahead queue.
 pub struct SceneChangeDetector<T: Pixel> {
   /// Minimum average difference between YUV deltas that will trigger a scene change.
@@ -43,7 +56,7 @@ pub struct SceneChangeDetector<T: Pixel> {
   /// Start deque offset based on lookahead
   deque_offset: usize,
   /// Scenechange results for adaptive threshold
-  score_deque: Vec<(f64, f64)>,
+  score_deque: Vec<ScenecutResult>,
   /// Number of pixels in scaled frame for fast mode
   pixels: usize,
   /// The bit depth of the video.
@@ -59,19 +72,6 @@ impl<T: Pixel> SceneChangeDetector<T> {
     encoder_config: EncoderConfig, cpu_feature_level: CpuFeatureLevel,
     lookahead_distance: usize, sequence: Arc<Sequence>,
   ) -> Self {
-    // This implementation is based on a Python implementation at
-    // https://pyscenedetect.readthedocs.io/en/latest/reference/detection-methods/.
-    // The Python implementation uses HSV values and a threshold of 30. Comparing the
-    // YUV values was sufficient in most cases, and avoided a more costly YUV->RGB->HSV
-    // conversion, but the deltas needed to be scaled down. The deltas for keyframes
-    // in YUV were about 1/3 to 1/2 of what they were in HSV, but non-keyframes were
-    // very unlikely to have a delta greater than 3 in YUV, whereas they may reach into
-    // the double digits in HSV.
-    //
-    // Experiments have shown that these thresholds is optimal.
-    const FAST_THRESHOLD: f64 = 18.0;
-    const SLOW_THRESHOLD: f64 = 7.0;
-
     let bit_depth = encoder_config.bit_depth;
     let speed_mode = if encoder_config.low_latency {
       SceneDetectionSpeed::Fast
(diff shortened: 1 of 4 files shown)


shssoichiro:
Commits will be squashed at the end.

src/scenechange/mod.rs
```diff
@@ -354,67 +393,64 @@ impl<T: Pixel> SceneChangeDetector<T> {
...
-        / (max_keyint - min_keyint) as f64;
-
-    // Adaptive threshold for medium version, static thresholf for the slow one
-    let threshold = if self.speed_mode == SceneDetectionSpeed::Medium {
-      intra_cost * (1.0 - bias)
-    } else {
-      self.threshold as f64
-    };
+    const BIAS: f64 = 0.7;
+    let threshold = intra_cost * (1.0 - BIAS);
```
tdaede, line 445: Did you find the ramp when we approach max keyframe interval to not be useful?

src/scenechange/mod.rs
```diff
@@ -241,64 +223,119 @@ impl<T: Pixel> SceneChangeDetector<T> {
...
+
+    // We use the histogram algorithm's cost metrics as a secondary algorithm
+    // because, although it struggles in certain scenarios such as
+    // finding the end of a pan, it is very good at detecting hard scenecuts
+    // or detecting if a pan exists.
+    // Because of this, we only consider a frame for a scenechange if
+    // the histogram algorithm is over the threshold either on this frame (hard scenecut)
+    // or within the past few frames (pan). This helps filter out a few false positives
+    // produced by the cost-based algorithm.
+    let hg_threshold = HISTOGRAM_THRESHOLD * (self.bit_depth as f64) / 8.0;
```
tdaede, line 289: We can do this separately, but we should probably rename the histogram method.

src/scenechange/mod.rs
```diff
@@ -354,67 +393,64 @@ impl<T: Pixel> SceneChangeDetector<T> {
   /// for coding this frame.
   #[hawktracer(cost_scenecut)]
   fn cost_scenecut(
-    &self, frame1: Arc<Frame<T>>, frame2: Arc<Frame<T>>, frameno: u64,
-    previous_keyframe: u64,
+    &self, frame1: Arc<Frame<T>>, frame2: Arc<Frame<T>>,
   ) -> ScenecutResult {
+    let frame1_ref2 = Arc::clone(&frame1);
     let frame2_ref2 = Arc::clone(&frame2);
+    let frame2_ref3 = Arc::clone(&frame2);
```
tdaede, line 400: Maybe instead name these after the functions they are created for?

src/scenechange/mod.rs
shssoichiro: Done

src/scenechange/mod.rs
tdaede: OK. Maybe we can revisit it later. It'll definitely create false positives in terms of "real" scenecuts, so the only way to really test it is with AWCY numbers rather than any sort of "keyframe ground truth".

tdaede (approved):

src/scenechange/mod.rs
```diff
@@ -194,51 +170,57 @@ impl<T: Pixel> SceneChangeDetector<T> {
...
-        previous_keyframe,
       );
     } else {
       self.deque_offset -= 1;
     }
 
     // Adaptive scenecut check
-    let scenecut = self.adaptive_scenecut();
+    let (mut scenecut, score) = self.adaptive_scenecut();
+    if let Some(result) = self.handle_min_max_intervals(distance) {
```
lu-zero, line 180: let scenecut = self.handle_min_max_intervals(distance).unwrap_or(scenecut);

maybe?

shssoichiro: That does seem a little cleaner, yeah.

lu-zero (approved):
The rest looks good :)

PR #2816, opened 2021-09-24
Explicitly check if NASM installed/reachable

master-of-zen:
Currently, it will just say that NASM version too old if NASM doesn't exist in the system.
This PR checks if NASM exists before checking its version.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -287,6 +287,20 @@ fn test_parse_nasm_version() {
 }
 
 #[cfg(feature = "asm")]
+fn is_nasm_installed_check() {
+  let mut cmd = std::process::Command::new("nasm");
+  cmd.arg("-h");
+  match cmd.status() {
+    Ok(_) => (),
+    Err(_) => {
+      panic!(
+        r#"Couldn't find NASM, make sure NASM installed and in PATH or disable "asm" feature"#
+      )
+    }
+  }
+}
+
+#[cfg(feature = "asm")]
 fn nasm_version_check() -> std::path::PathBuf {
   let nasm_path = std::path::PathBuf::from("nasm");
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -286,6 +286,18 @@ fn test_parse_nasm_version() {
...
 }
 
+#[cfg(feature = "asm")]
+fn is_nasm_installed_check() {
+  let mut cmd = std::process::Command::new("nasm");
+  cmd.arg("-h");
+  match cmd.output() {
+    Ok(_) => (),
+    Err(_) => {
+      panic!("Couldn't find NASM, make sure NASM installed and in PATH")
```
lu-zero, line 296: Might be good to mention that is possible to not set the `asm` feature to avoid installing it.

master-of-zen: Corrected

build.rs
```diff
@@ -286,6 +286,20 @@ fn test_parse_nasm_version() {
   assert_eq!((2, 14, 0), parse_nasm_version(ver_str));
 }
 
+#[cfg(feature = "asm")]
+fn is_nasm_installed_check() {
+  let mut cmd = std::process::Command::new("nasm");
+  cmd.arg("-h");
+  match cmd.output() {
```
shssoichiro, line 293: This is certainly minor in the overall scheme of things, but you can use `cmd.status()` instead which will be more performant as it avoids capturing stdout and stderr.

master-of-zen: Corrected

shssoichiro (approved):

tdaede (approved):
It might be nice to emphasize that installing nasm will yield a huge speedup over disabling asm in the message.

PR #2819, opened 2021-09-25
x86: Add SSE2 and AVX2 SAD intrinsics for `delta_in_planes`

redzic:
We can utilize the instruction psadbw on x86 to make `delta_in_planes`
faster for 8-bit input, which benefits fast scene detection. This change
makes `delta_in_planes` about 2x faster than the current implementation
for both SSE2 and AVX2.

Here are some measurements of the performance according to hawktracker, tested on a 240x240 resolution video on a Zen 1 Threadripper 1950x. 

Original Rust version (but compiled with `-C target-cpu=native` so it is autovectorized with AVX2):

image

New SSE2 version:

image

New AVX2 version:

image

### DIFF
--- a/src/asm/x86/mod.rs
+++ b/src/asm/x86/mod.rs
@@ -14,4 +14,5 @@ pub mod lrf;
 pub mod mc;
 pub mod predict;
 pub mod quantize;
+pub mod sad_row;
 pub mod transform;
--- a/src/asm/x86/sad_row.rs
+++ b/src/asm/x86/sad_row.rs
@@ -1,0 +1,182 @@
+// Copyright (c) 2021, The rav1e contributors. All rights reserved
+//
+// This source code is subject to the terms of the BSD 2 Clause License and
+// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+// was not distributed with this source code in the LICENSE file, you can
+// obtain it at www.aomedia.org/license/software. If the Alliance for Open
+// Media Patent License 1.0 was not distributed with this source code in the
+// PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+use crate::cpu_features::CpuFeatureLevel;
+use crate::sad_row::*;
+use crate::util::{Pixel, PixelType};
+
+#[cfg(target_arch = "x86")]
+use std::arch::x86::*;
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
+
+use std::hint::unreachable_unchecked;
+use std::mem;
+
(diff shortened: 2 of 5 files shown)


shssoichiro:
It looks like the avx2 and sse2 versions perform about identically, If that's the case, is there a reason to have the avx2 version? (We can have a comment on the sse2 saying it was tested that avx2 doesn't give any extra gains, but I'm asking for the reason that less code to maintain generally = better.)

redzic:
The CPU I tested on is a Zen 1 processor, where the AVX implementation is done via microcode and isn't as good as the implementation for Zen 2 and newer CPUs.

Quoted from [Agner's microarchitecture guide](https://www.agner.org/optimize/microarchitecture.pdf):

> 256-bit vector instructions are implemented differently on Zen 1 and Zen 2: 
 
> Zen 1: 256-bit AVX and AVX2 instructions are split into two μops that do 128 bits each. It is 
still more efficient to use 256-bit vector instructions than 128-bit instructions on Zen 1, due 
to efficient fetch and decoding.  
 
> There is no penalty for mixing AVX and non-AVX vector instructions on the Zen 1 because 
256-bit registers are split into two 128-bit registers. 
 
> Zen 2: The Zen 2 has full 256-bit execution units and data paths. This effectively doubles 
the throughput for 256-bit vector instructions on Zen 2 over Zen 1.

In theory the AVX2 version should be faster compared to the SSE2 version on newer CPUs. Maybe someone else with a newer x86 CPU could test this though.

src/sad.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2021, The rav1e contributors. All rights reserved
```
tdaede, line 1: We already have some sad functions in asm/dist/mod.rs. Some documentation on these functions about why this is separate (e.g. it's per line, doesn't operate on PlaneRegion) would be useful.

redzic: I renamed the module to `sad_row`, and I updated the doc comment for the `sad_row` function

src/asm/x86/sad.rs
```diff
@@ -0,0 +1,182 @@
...
+use std::arch::x86::*;
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
+
+use std::hint::unreachable_unchecked;
+use std::mem;
+
+/// SAFETY: src and dst must be the same length and less than 16 elements
+#[inline(always)]
+pub(crate) unsafe fn sad_below16_8bpc(src: &[u8], dst: &[u8]) -> i64 {
```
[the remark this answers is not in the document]
redzic: made all functions except `sad_row_internal` not pub

PR #2821, opened 2021-10-01
Fill in out-of-bounds area of the residual with edge extension.

### LINKED ISSUE #2817: Gray line at the bottom of the frame
**Describe the bug**
Encoding adds an artifact at the bottom edge of the image

Screenshot 2021-09-24 at 17 50 53

**To Reproduce**
Encode this image:
example

**Expected behavior**
Pixels on the left side of the last line should all be white.

**Required Information**
Rust 1.55. x86-64 macOS 11.6. NASM version 2.15.05 compiled on Aug 29 2020. rav1e 0.5.0-beta.2

tdaede:
This avoids a spatial discontinuity (and correspondingly large
DCT coefficients) when there is a large DC error in the prediction.

Fixes #2817

https://beta.arewecompressedyet.com/?job=master-2021-10-01_064210-48149f18&job=pad_extend-2021-10-01_064216-b67291a2


lu-zero (approved):
seems fine to me.

src/encoder.rs
```diff
@@ -1200,16 +1200,22 @@ pub fn encode_tx_block<T: Pixel, W: Writer>(
...
-  let initialized_area =
+  let initialized_area_idx =
     if visible_tx_w == 0 { 0 } else { tx_size.width() * visible_tx_h };
-  for a in residual[initialized_area..].iter_mut() {
-    *a = 0;
+  {
+    // do edge-extension padding of residual vertically
+    let (initialized_area, uninitialized_area) = residual.split_at_mut(initialized_area_idx);
+    for (i, a) in uninitialized_area.iter_mut().enumerate() {
+      *a = initialized_area[initialized_area_idx - tx_size.width() + (i % tx_size.width())];
```
lu-zero, line 1217: maybe move tx_size.width() outside to make the line shorter?

src/encoder.rs
```diff
@@ -1200,16 +1200,22 @@ pub fn encode_tx_block<T: Pixel, W: Writer>(
     );
     if visible_tx_w < tx_size.width() {
       for row in residual.chunks_mut(tx_size.width()).take(visible_tx_h) {
-        for a in &mut row[visible_tx_w..] {
-          *a = 0;
+        // do edge-extension padding of residual horizontally
+        let (initialized_area, uninitialized_area) = row.split_at_mut(visible_tx_w);
+        for a in &mut uninitialized_area.iter_mut() {
+          *a = initialized_area[visible_tx_w - 1];
```
lu-zero, line 1207: Since you wanted alternatives:
``` rust
unitialized_area.fill(initialized_area[visible_tx_w - 1])
```

tdaede:
I'm not entirely happy with this patch, as while it is a visual win everywhere I looked, it's a (small, 0.05%) coding gain loss. Not sure if that's inherent to edge extension as the padding mechanism or not.

barrbrain:
The code itself looks good to me. It doesn't look like metrics fully agree on any sequence, so I am not sure about that call.

tdaede:
My logic for this patch is that looking at the original picture, there are no prediction modes that can be a good DC predictor, as although the block is mostly white, the row of pixels above is solid blue. This means the residual will have a solid DC bias. However, zeroing the residual at the block edge creates a sharp discontinuity in the residual to be encoded, which once transformed and quantized, bleeds over into the visible area. It is a dark line because the prediction is too dark, but likewise it would be a light line if the prediction was too light. Edge extension is a simple way to not have this sharp discontinuity which gets spread out over many DCT coefficients - however it's far from ideal as it still adds energy far away from the edge that needs to be coded, which might be responsible for the 0.05% loss.

I apply this equally to residuals for inter and intra blocks because in theory there's nothing preventing the same problem happening for inter. For example, a full screen fade is a case where many predictors might have a DC offset, and the result could be transient lines around the edges of the frame. It might be interesting to contrive such an example.

PR #2825, opened 2021-10-06
Avoid bounds checks in copy_from_raw_u8 for HBD

shssoichiro:
The compiler cannot assume that `.chunks` will always produce
slices with 2 elements, because the final chunk may only
contain one element. However, we can assume that if the source
bytewidth is 2 bytes, that there must be an even number of bytes
and every chunk will have 2 elements.

Therefore, we can safely remove the bounds checks in the HBD path.
This improves performance of the HBD version of this function by 30%.
Approximately 1% overall encode performance at speed 10.

### DIFF
--- a/v_frame/src/plane.rs
+++ b/v_frame/src/plane.rs
@@ -422,11 +422,9 @@ impl<T: Pixel> Plane<T> {
             source_bytewidth
           );
           for (self_pixel, bytes) in
-            self_row.iter_mut().zip(source_row.chunks(2))
+            self_row.iter_mut().zip(source_row.chunks_exact(2))
           {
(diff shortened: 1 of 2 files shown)


v_frame/src/plane.rs
```diff
@@ -424,9 +424,12 @@ impl<T: Pixel> Plane<T> {
           for (self_pixel, bytes) in
             self_row.iter_mut().zip(source_row.chunks(2))
```
lu-zero, line 425: chunks_exact wouldn't help?

shssoichiro: Ah, I didn't know that method existed. I'll give that a try.

PR #2838, opened 2021-10-29
Demonstrate filling a frame from raw pixels in the simple_encoding example

### LINKED ISSUE #2799: example populating frame with actual (non zero) data
I cannot figure out how to populate a frame with data. The example explicitly calls out that it is sending 0'd data. I assume I'm supposed to use something in Plane to populate with yuv data, but it isn't clear to me how to do this, and the documentation there is thin. I also don't know if I'm supposed to continually assign to each plane in the frame after calling `new_frame()` or should i store the frame and copy data to it then do `send_frame`

An example encoding say, 1-10 frames of RGBA data (can just be random noise), using the Encoder api would be extremely helpful.

barrbrain:
Fixes #2799.

### DIFF
--- a/examples/simple_encoding.rs
+++ b/examples/simple_encoding.rs
@@ -23,7 +23,14 @@ fn main() {
 
   let mut ctx: Context<u16> = cfg.new_context().unwrap();
 
-  let f = ctx.new_frame();
+  let mut f = ctx.new_frame();
+
+  let pixels = vec![42; enc.width * enc.height];
+
+  for p in &mut f.planes {
+    let stride = (enc.width + p.cfg.xdec) >> p.cfg.xdec;
+    p.copy_from_raw_u8(&pixels, stride, 1);
+  }
 
   let limit = 30;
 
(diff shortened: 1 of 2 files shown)


examples/simple_encoding.rs
```diff
@@ -23,7 +23,13 @@ fn main() {
 
   let mut ctx: Context<u16> = cfg.new_context().unwrap();
 
-  let f = ctx.new_frame();
+  let mut f = ctx.new_frame();
+
+  let pixels = vec![42; enc.width * enc.height];
+
+  for p in &mut f.planes {
+    p.copy_from_raw_u8(&pixels, enc.width, 1);
```
lu-zero, line 31: would be better to show about luma vs chroma dimension maybe?

barrbrain: Something like this?
```suggestion
    let stride = (enc.width + p.cfg.xdec) >> p.cfg.xdec;
    p.copy_from_raw_u8(&pixels, stride, 1)
```

lu-zero: yes :)

lu-zero (approved):

PR #2840, opened 2021-10-31
CI: Run apt-get update before install

barrbrain:
If the apt index in the base image is outdated, fetching packages may fail with an HTTP 404 error.


lu-zero (approved):
I wonder if their new `apt` command is smarter

PR #2841, opened 2021-10-31
CI: Always use latest cargo-c and grcov versions

EwoutH:
Prevents having to update it manually all the time

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,7 +29,7 @@ jobs:
 
     - name: Install cargo-c
       run: |
-        $LINK = "https://github.com/lu-zero/cargo-c/releases/download/v0.8.0"
+        $LINK = "https://github.com/lu-zero/cargo-c/releases/latest/download"
         $CARGO_C_FILE = "cargo-c-windows-msvc"
         curl -LO "$LINK/$CARGO_C_FILE.zip"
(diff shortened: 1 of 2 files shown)


tmatth:
On one hand this would make CI builds arguably less reproducible, on the other hand I'm assuming it would log whatever version was used so maybe it's not such a bad tradeoff.

EwoutH:
That's indeed the trade-off. On the macOS builds we install cargo-c using homebrew by the way, which also picks the latest version.

@lu-zero Could you tell me a bit about the changed package names between cargo-c [v0.9.4](https://github.com/lu-zero/cargo-c/releases/tag/v0.9.4) and [v0.9.5](https://github.com/lu-zero/cargo-c/releases/tag/v0.9.5)? Noticed they are different, for example, the linux version is now a .zip instead a .tar.gz, and couldn't find anything in the release notes or commits.

lu-zero:
The deploy script failed also on cargo-c so it is a manual recover, please open an issue about it and I'll fix it tomorrow.

EwoutH:
Thanks for getting back so quickly! Opened an issue: https://github.com/lu-zero/cargo-c/issues/235

lu-zero:
Now it should be fixed, please try again.

EwoutH:
Thanks! I removed by last commit and rebased, and it looks like the cargo-c builds now pass.

lu-zero:
It needs a second rebase, I'll land it after.

lu-zero (approved):

PR #2843, opened 2021-11-02
CI: Move to Debian Unstable for aom and dav1d

barrbrain:
Now that they are available, pulling these dependencies from the Debian archive should result in less CI configuration churn thanks to a longer retention policy.


vibhoothi (approved):
LGTM, hopefully this can reduce some cycles of CI maintenance:)

PR #2846, opened 2021-11-05
Hide full search speed setting from public api

KyleSiefring:
Thoughts?

### DIFF
--- a/src/api/config/speedsettings.rs
+++ b/src/api/config/speedsettings.rs
@@ -84,15 +84,24 @@ pub struct SpeedSettings {
   /// Use fine directional intra prediction
   pub fine_directional_intra: bool,
 
-  /// Enable full search in some parts of motion estimation. Allowing full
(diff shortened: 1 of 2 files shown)


lu-zero (changes requested):
why do you want it hidden and why aren't you just making the field `pub(crate)`?

KyleSiefring:
Mostly want it hidden so that I can add and remove it without multiple api changes.

I felt that using pub(crate) might end up being too error prone. Example: accidentally exposing things by writing pub instead of pub(crate).

One alternative would be to save the speed level and seed internal settings inside the encoder, but not inside api exposed structures.

lu-zero:
Until I cut a release any API you put move or change is fine and expected to change. 

I'll make sure that the non-exhaustive and default() is in place tomorrow.

barrbrain:
It appears the proposed changes never happened and this setting became part of our speed ladder.

PR #2847, opened 2021-11-12
Better autovectorization of `spatiotemporal_scale`

redzic:
It turns out that avoiding a call to `Iterator::take` allows LLVM to autovectorize `spatiotemporal_scale`.

For a comparison of the code generated (simplified for example): https://godbolt.org/z/56Kec186G

### DIFF
--- a/src/rdo.rs
+++ b/src/rdo.rs
@@ -720,14 +720,33 @@ pub fn spatiotemporal_scale<T: Pixel>(
   let y1 = (y0 + bsize.height_imp_b()).min(fi.h_in_imp_b);
   let den = (((x1 - x0) * (y1 - y0)) as u64) << DistortionScale::SHIFT;
 
+  // calling this on each slice individually improves autovectorization
+  // compared to using `Iterator::take`
+  #[inline(always)]
+  fn take_slice<T>(slice: &[T], n: usize) -> &[T] {
+    if n >= slice.len() {
+      slice
+    } else {
+      &slice[..n]
+    }
+  }
+
   let mut sum = 0;
   for y in y0..y1 {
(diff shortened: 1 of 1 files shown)


barrbrain (commented):
I compared the generated code from this change against just removing the `.take(MAX_SB_IN_IMP_B)`. Both resulted in vectorised loops but this change was better because the output included such a loop for widths narrower than the maximum.

redzic:
For me it was never vectorized without this change, when compiling for znver1. Also the CI errors look unrelated to the change.

barrbrain:
The CI errors should be resolved by #2848. Rebasing this PR will apply that fix.

src/rdo.rs
```diff
@@ -720,14 +720,33 @@ pub fn spatiotemporal_scale<T: Pixel>(
...
+      &slice[..n]
+    }
+  }
+
   let mut sum = 0;
   for y in y0..y1 {
-    sum += fi.distortion_scales[y * fi.w_in_imp_b..][x0..x1]
-      .iter()
-      .zip(fi.activity_scales[y * fi.w_in_imp_b..][x0..x1].iter())
-      .take(MAX_SB_IN_IMP_B)
```
barrbrain, line 728: I checked again with `RUSTFLAGS="-C target-feature=+avx2,+fma" cargo asm -- rav1e::rdo::spatiotemporal_scale`.
Removing `.take(MAX_SB_IN_IMP_B)` here does result in auto-vectorisation but only for the maximum width.

barrbrain (approved):
This is better than the alternative of removing `.take()` as more widths receive auto-vectorised loops.

PR #2850, opened 2021-11-30
Use nasm-rs for nasm version check

kornelski:
nasm-rs already has a version check.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -126,7 +126,7 @@ fn build_nasm_files() {
     config_include_arg.push_str(&out_dir);
     config_include_arg.push('/');
     let mut nasm = nasm_rs::Build::new();
-    nasm.min_version(2,14,0);
+    nasm.min_version(2, 14, 0);
     for file in asm_files {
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -126,14 +126,18 @@ fn build_nasm_files() {
     config_include_arg.push_str(&out_dir);
     config_include_arg.push('/');
     let mut nasm = nasm_rs::Build::new();
+    nasm.min_version(2,14,0);
```
barrbrain, line 129: `cargo fmt` is failing here due to spacing.
```suggestion
    nasm.min_version(2, 14, 0);
```

barrbrain (approved):
Thank you, it appears fixes for all our workarounds have been merged upstream.

barrbrain:
Unfortunately, this did away with the less confusing error message in the case of missing nasm. We will follow up on that.

kornelski:
Fixed error message: https://github.com/medek/nasm-rs/pull/36

PR #2855, opened 2021-12-10
Refactor the SpeedSettings struct and impl

shssoichiro:
The impetus for this change is that the documentation for speed
settings, being a manually-maintained comment, would easily become out
of date and was difficult to maintain. The primary change this
introduces is that, rather than having a separate function for each
speed setting to determine when to enable it, the `from_preset` function
now sets all settings per speed level, with the `default()` being speed
0, and overriding as the speed increases. The idea is that this will
make it much easier to see which features are enabled at which speed
levels.

This also consolidates a couple of fields which were redundant, namely:
- `no_scene_detection` and `fast_scene_detection` are now one setting
  called `scene_detection_mode`
- `rdo_lookahead_frames` existed on EncoderConfig, but the function for
  setting it was in `SpeedSettings`. It has been moved to `SpeedSettings`,
  as even though it can be overridden by a CLI arg, it is primarily controlled
  by the speed preset.

### DIFF
--- a/src/api/channel/by_gop.rs
+++ b/src/api/channel/by_gop.rs
@@ -326,7 +326,7 @@ impl Config {
     self.validate()?;
 
     // TODO: make it user-settable
-    let input_len = self.enc.rdo_lookahead_frames as usize * 4;
+    let input_len = self.enc.speed_settings.rdo_lookahead_frames as usize * 4;
     let frame_limit = i32::MAX as u64;
 
     let (send_frame, receive_frame) = bounded(input_len);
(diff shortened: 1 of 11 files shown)


shssoichiro:
The build failures seem to be unrelated to this PR:

```
libaom0_1.0.0.errata1.ds-1_amd64.deb: FAILED
sha256sum: WARNING: 1 computed checksum did NOT match
```

lu-zero (changes requested):
Could you please split the moving of rdo_lookahead_frames from the rest and the same for removing no_scene_detection?

Beside that I'm fine with the changes, @barrbrain I'd like your opinion as well.

While at it, mark the struct non_exhaustive please.

shssoichiro:
I'm assuming separate commits within this PR is fine? If so, I'll get that pushed up here this morning.

lu-zero:
It is fine, we have to update the debs again meanwhile :)

Thank you !

shssoichiro:
All requested changes have been applied and rebased off of master @lu-zero @barrbrain

lu-zero (approved):
Thank you a lot!

PR #2862, opened 2021-12-13
Prepare 0.5.1

lu-zero:


tdaede:
Previously the Cargo.lock was on the Github release, not in the branch, I think?

lu-zero:
> Previously the Cargo.lock was on the Github release, not in the branch, I think?

It is in the release, historically was in the branch, but it is much simpler not having it around.

PR #2867, opened 2021-12-21
Improve autovectorization of `downscale_in_place`

redzic:
Removing the bounds check from each iteration of the loop allows LLVM to autovectorize `downscale_in_place`, resulting in about 15% better performance of `downscale_in_place` on yuv420p 1080p when compiling with AVX2 enabled.

Also adds an assertion that `box_pixels != 0` at the beginning of the function as it avoids later division-by-zero checks which happen per loop iteration.

Here you can see the difference in assembly generated with this change on godbolt: https://godbolt.org/z/4M4xnoTPn

Performance of `downscale_in_place` in current master:

image

Performance of `downscale_in_place` with this change:

image

This was tested on a zen 1 Threadripper 1950x.


shssoichiro (approved):
Thanks, I like it!

PR #2869, opened 2022-01-02
Update to clap3

lu-zero:


shssoichiro:
Looks like this requires bumping the min rustc to 1.52 (at least. the current error it's giving is a feature that was stabilized in 1.52).

shssoichiro (approved):

lu-zero:
1.54 ends up needed.

PR #2871, opened 2022-01-03
Display info for all y4m error types

shssoichiro:
A user on Discord reported that rav1e was reaching an unreachable
branch, which was in this section of the codebase. This commit
explicitly returns an error message for each potential type of y4m
error, instead of assuming that some are unreachable.


lu-zero (approved):
Sounds fine, we should consider using `anyhow` though.

tdaede (approved):

shssoichiro:
@lu-zero I agree, I can make a PR later to do that, but it's much larger than this since we aren't using it at all right now.

PR #2875, opened 2022-01-06
Add x86 assembly for cdef distortion

KyleSiefring:
Around ~7% faster on speed 2. It should be roughly the same for most
speed levels.

It would have been more, but some recent patches seem to have allowed
autovecterization to happen again.


shssoichiro (approved):
The Rust code looks good. I'm not sure if someone who knows ASM wants to comb through that portion of the PR.

PR #2877, opened 2022-01-08
Use const generics for hadamard transforms

redzic:
Using const generics allows the compiler to optimize the hadamard transforms significantly better: https://godbolt.org/z/YdWe7518h


shssoichiro (approved):
This is exactly what I hoped we could use const generics for once they were stabilized. I'd love to see more of this.

PR #2880, opened 2022-01-08
Bump minimum rust version to 1.54.0

lu-zero:


lu-zero:
Folding it back to the clap3 branch

PR #2883, opened 2022-01-11
Improve codegen of `Plane::downsampled`

redzic:
No `unsafe` code is actually needed to remove the bounds checks in this case, however the extra bounds check from PR #2824 helps ensure that the behavior is still exactly the same in all cases (since calling `zip` will basically truncate the entire iterator to the shortest one, rather than ensuring that there are always a certain number of elements beforehand).

Generated assembly difference: https://godbolt.org/z/eTG7Txh9P

### DIFF
--- a/v_frame/src/plane.rs
+++ b/v_frame/src/plane.rs
@@ -520,15 +520,17 @@ impl<T: Pixel> Plane<T> {
         &data_origin[(src.cfg.stride * row_idx * 2)..][..(2 * width)];
       let src_bottom_row =
         &data_origin[(src.cfg.stride * (row_idx * 2 + 1))..][..(2 * width)];
-      for (col, dst) in dst_row.iter_mut().take(width).enumerate() {
-        let mut sum = 0;
-        unsafe {
-          // SAFETY: We perform a bounds check above to ensure these are safe
-          sum += u32::cast_from(*src_top_row.get_unchecked(2 * col));
-          sum += u32::cast_from(*src_top_row.get_unchecked(2 * col + 1));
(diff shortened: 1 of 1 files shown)


v_frame/src/plane.rs
```diff
@@ -520,15 +520,25 @@ impl<T: Pixel> Plane<T> {
...
+      #[inline(always)]
+      fn take_slice_mut<T>(slice: &mut [T], n: usize) -> &mut [T] {
+        if n >= slice.len() {
+          slice
+        } else {
+          &mut slice[..n]
         }
+      }
+
+      for ((dst, a), b) in take_slice_mut(dst_row, width)
```
shssoichiro, line 533: We achieve similar codegen by using `dst_row[..width].iter_mut()` instead of creating a custom function. https://godbolt.org/z/co68hP8re

shssoichiro: Actually, the above only works if width is never greater than the size of the array. Do we have a guarantee of that?

I guess we'll find out if we make that change and it causes the tests to panic.

redzic: I didn't know either, so I just opted to go for the thing that ensures the exact same behavior as before. Even if the width is never greater than the size of the array, it would probably be better to use something like `take_slice_mut` so that the binary size is less.

redzic: I decided to opt for a `[..width.min(dst_row.len()]` (but with the length calculated separately because of unfortunate borrow checker limitations) since I think it's probably less obnoxious than an entire separate function.

lu-zero (approved):
It looks already much nicer, I'm pretty sure width cannot be larger than row, but we can doublecheck that later.

PR #2885, opened 2022-01-22
Simplify importance block sum in `estimate_importance_block_difference`

redzic:
As far as I could tell, 8x8 blocks are always used, so dynamically counting the number of pixels is not necessary.


redzic:
Tests seem to be failing for an unrelated reason, because of dav1d-sys I guess? Not sure what to do about that

shssoichiro (approved):
Nice. I'll look at the CI issue... sometimes things get updated in debian and then break our CI

shssoichiro:
@redzic Rebase and it should pass now

redzic:
@shssoichiro ping, I've rebased the PR

PR #2886, opened 2022-01-22
Update to Edition 2021

shssoichiro:
dav1d-sys has updated to edition 2021, forcing us to raise our minimum supported Rust version to 1.56. Since we are now raising the minimum version, it makes sense to update ourselves to edition 2021 at the same time.


lu-zero (approved):
I'm all for it

PR #2887, opened 2022-01-23
Add safety asserts for mc assembly callers

KyleSiefring:
put, prep, and avg now check bounds and width/height before calling
assembly functions.


shssoichiro:
Does this have any measurable impact on performance? I'm guessing it should be minimal, but these functions are called quite often so I'd like to confirm.

KyleSiefring:
Looking at perf, I don't see enough cpu cycles being used in the areas that this is being inlined into to be that concerned.

shssoichiro (approved):
Thanks, this looks good to me then

PR #2890, opened 2022-01-31
Speed up estimate_inter_costs again

redzic:
- Avoids downscaling and allocating hres and qres, since they are not used for inter cost estimation.
- Fixes optimization where allocation was supposed to be avoided, but was not completely eliminated because of padding.

Before this change (tested on yuv420p 1080p):

image

After this change:

image

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -418,14 +418,15 @@ impl<T: Pixel> FrameState<T> {
     )
   }
 
+  /// Does not create qres or hres versions of `frame`
   pub fn new_with_frame_and_me_stats_and_rec(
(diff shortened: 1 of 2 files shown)


src/encoder.rs
```diff
@@ -418,14 +418,15 @@ impl<T: Pixel> FrameState<T> {
     )
   }
 
+  /// Does not create qres or hres versions of `frame`
```
shssoichiro, line 421: Since this function behaves very specifically, I think it would be good to add a note of why this function exists.

shssoichiro (changes requested):

shssoichiro (approved):

PR #2891, opened 2022-02-01
Fix confusing scale factor

redzic:
It is not used for cost based scene detection, but it is still calculated for some reason.


shssoichiro (approved):
Probably from all the refactors a while back. Thanks!

PR #2897, opened 2022-02-25
Fix unnecessary auto-vectorization in residual loop of SAD intrinsics

redzic:
There is no way to force the compiler to not auto-vectorize a loop without changing the global compilation flags. However, it is possible with inline assembly which was stabilized in Rust 1.59.0.

Before, the compiler aggressively vectorized a loop that was supposed to be scalar since it was only a residual loop of an already vectorized function. With this PR, the correct assembly code is generated: https://godbolt.org/z/e6KTfqnxG

### DIFF
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -44,7 +44,7 @@ jobs:
       matrix:
         conf:
           - beta-build
-          - 1.56.0-tests
+          - 1.59.0-tests
           - aom-tests
           - dav1d-tests
           - no-asm-tests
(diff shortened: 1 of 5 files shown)


tdaede:
Interesting technique.

There is a lot of potential stuff to evaluate with the addition of inline assembly. One thing that comes to mind is the update_cdf function which has improved assembly in dav1d compared to our intrinsics (in theory).

shssoichiro:
Test failures are from the new clippy warnings in rustc stable. We'll have to do the typical clippy fixup PR.

shssoichiro:
you can rebase now to fix the CI

redzic:
@shssoichiro should be good to go, rebased the PR

shssoichiro:
Since this brings in inline asm, you'll also need to bump the minimum rustc version to 1.59. (We have a few places that do this e.g. build.rs, CI files, maybe a few others)

redzic:
@shssoichiro sorry about that, everything should be fixed now

shssoichiro (approved):
Thanks!

PR #2898, opened 2022-02-26
Fix clippy warnings

redzic:


redzic:
Oops, `cargo clippy --fix` changed all the usages of `black_box` to be incorrect.

PR #2900, opened 2022-03-01
Strongly type different variants of FrameInvariants

shssoichiro:
There are three distinct possible states for a FrameInvariants struct:
Normal frame, show existing frame, or invalid frame.
Currently, these are each represented by a separate boolean on the
struct. However, this results in some combinations which are not
valid, and wastes memory allocating fields that are not used for every
frame state.

This commit uses Rust's type system to define FrameInvariants as an enum
which is in one of the three valid states. In addition, this changes the
invalid state to an empty struct, and removes lookahead data from the
SEF struct, where it is not used. This results in a 3% reduction in
total encoder memory usage with default settings at speed 6, as well as
a 1-2% reduction in encoding time at speed 6 due to removal of
unnecessary allocations/clones.

Test against AWCY found output to be identical to previous.

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -204,23 +204,9 @@ impl InterConfig {
   }
 }
 
-// Thin wrapper for frame-related data
-// that gets cached and reused throughout the life of a frame.
-#[derive(Clone)]
-pub(crate) struct FrameData<T: Pixel> {
-  pub(crate) fi: FrameInvariants<T>,
-  pub(crate) fs: FrameState<T>,
-}
-
-impl<T: Pixel> FrameData<T> {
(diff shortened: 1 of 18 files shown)


shssoichiro:
Note that the changes to the tests are not due to a change in behavior: Actually, the tests were not correctly checking frame order before, as they would skip some reordered frames.

KyleSiefring:
How many places would placeholder occur? It seems like it would be better to define those places as optional FrameInvariants and have an optional value for lookahead data.

shssoichiro:
Hmm. True, that would avoid all the funkiness with calling `.base` and `.lookahead` due to how Rust handles enum destructuring. I'll try that out.

shssoichiro:
I'm going to close this PR then, since the one resulting from your suggestion would be considerably different from this one (and involve less code modification).

PR #2901, opened 2022-03-02
Fix pyramid and frame type tests

shssoichiro:
These tests were filtering out frames that may have been in between
other frames we wanted to check, leading to tests that don't fully check
what we expect them to check. To resolve this, the filtering has been
removed from the test verifications, and we now check all frames in the
set.

This also removes comments about the frame type for non-keyframes, as
these were not particularly useful and may have even been incorrect.

### DIFF
--- a/src/api/test.rs
+++ b/src/api/test.rs
@@ -318,28 +318,28 @@ fn output_frameno_low_latency_minus(missing: u64) {
       0 => {
         &[
           (0, true), // I-frame
-          (1, true), // P-frame
-          (2, true), // P-frame
-          (3, true), // P-frame
-          (4, true), // P-frame
+          (1, true),
+          (2, true),
+          (3, true),
+          (4, true),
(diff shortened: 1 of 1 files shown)


src/api/test.rs
```diff
@@ -976,84 +945,84 @@ fn output_frameno_incremental_reorder_minus(missing: u64) {
...
+          (0, true), // I-frame
+          (4, true),
+          (2, true),
+          (1, true),
+          (2, true),
+          (3, true),
+          (4, true),
+          (5, true), // I-frame
+          (5, false),
           (5, false), // 3rd last (missing)
```
tdaede, line 1010: did you maybe want to drop these comments as well?

shssoichiro: Good catch, thanks. I meant to drop all of them except the "I-frame" comments.

lu-zero (approved):

PR #2902, opened 2022-03-03
Improve storage of invalid and SEF frame invariants

shssoichiro:
Thie commit changes invalid frame invariants to be represented by a
`None` value, and for SEFs to no longer contain any additional data that
is only needed by coded frames (at the moment, this is only lookahead
data).

This results in a ~3% overall reduction in memory usage, and a ~1%
improvement in encode runtime at speed 6 (with low-latency=false).


shssoichiro:
Depends on #2901

lu-zero (approved):
It looks good

PR #2903, opened 2022-03-05
Avoid multiple inlining of residual loop in SAD intrinsics

redzic:
Before, `sad_scalar` was called in multiple different places for `sad_8bpc_avx2` and `sad_8bpc_sse2`, thus causing the residual loop to be inlined more than once. This PR improves the SAD intrinsics so that the residual loop is only called once for each of these functions, which ultimately avoids slightly bloated generated assembly.

Comparison of generated assembly: https://godbolt.org/z/xKzqv8Yss


shssoichiro (approved):
Looks like less branching as well. Thanks!

PR #2906, opened 2022-03-13
Move complex pred modes for keyframes to s2

shssoichiro:
Currently the speed gap for speeds 1-3 is very unbalanced.
Speed 2 is only 20% faster than speed 1, while speed 3 is 150% faster
than speed 2. After this change, speed 2 becomes 50% faster than speed
1, and speed 3 becomes 50% faster than speed 2. Impact on metrics at
speed 2 is very small: [AWCY](https://bit.ly/3q0sJrl)


lu-zero (approved):
Sounds good to me.

PR #2909, opened 2022-03-15
Fix memory leak when temporal RDO not in use

shssoichiro:
The cached intra costs ended up using a non-trivial amount of memory, so
we need to ensure they are not kept around when they are not needed.
In the case of temporal RDO, these are already removed from the cache
when we move them into the FrameInvariants.


shssoichiro:
@redzic this should also resolve the av1an scenechange memory leak mentioned in Discord (with an additional change to av-scenechange)

PR #2910, opened 2022-03-17
Remove a broken hawktracer annotation

shssoichiro:
Some change seems to have occurred which causes this specific hawktracer
annotation to fail to compile. For the time being, remove it so that
hawktracer support can compile again.

### DIFF
--- a/src/scenechange/standard.rs
+++ b/src/scenechange/standard.rs
@@ -8,7 +8,6 @@ use crate::{
   frame::Frame,
   me::FrameMEStats,
 };
-use rust_hawktracer::*;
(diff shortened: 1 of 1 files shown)


lu-zero:
What is the error?

lu-zero (approved):
Go for it.

shssoichiro:
I wish there was a fix for it instead, but I think the issue is related to the fact that we use `&mut self` inside of the different threads in that function. So I think the only way to keep this annotation would be to completely rewrite how we thread things in this function, which probably requires reducing the threading effectiveness.

PR #2911, opened 2022-03-17
rate: add assertion for q57 input range

tmatth:
Caveat: it may be preferable to wait for const function assertions to be available natively.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ simd_helpers = "0.1"
 wasm-bindgen = { version = "0.2.63", optional = true }
 rust_hawktracer = "0.7.0"
 arrayref = "0.3.6"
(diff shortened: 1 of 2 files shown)


shssoichiro:
Does this have any impact on performance? I'm not sure exactly how const fn assertions impact this, but a regular assertion here would break the ability to do autovectorization around this function. (although it looks like `cfn_debug_assert` is also an option.)

tmatth:
> Does this have any impact on performance? I'm not sure exactly how const fn assertions impact this, but a regular assertion here would break the ability to do autovectorization around this function. (although it looks like `cfn_debug_assert` is also an option.)

Yeah I think `cfn_debug_assert` would be preferable.

shssoichiro (approved):
Thanks!

PR #2913, opened 2022-03-24
Remove unnecessary clones from metrics calculation

shssoichiro:
Due to the interface of av-metrics, we were cloning two uncompressed
frames of data for each encoded frame, even when not doing any metrics
calculation. The interface to av-metrics has been changed to support
passing a `&Frame`, allowing us to avoid the clones entirely.

This reduces the total allocation count by about 6%, and peak
memory usage and encoding time by 1% at all speeds.


lu-zero (approved):
Looks good

PR #2914, opened 2022-03-24
Add a mock around rayon for disabling threading

shssoichiro:
Profiling becomes considerably more difficult to read when rayon is
involved, because rayon nests many layers of function calls which
obscure where CPU and memory usage is actually occurring. This uses the
maybe-rayon crate as a wrapper around rayon, which by default passes
through to rayon with zero overhead, but by disabling the "threading"
feature in rav1e, the maybe-rayon crate will become a mock which calls
functions serially.

Examples:

Memory flamegraph with rayon:
Screenshot_2022-03-24_10-49-12

Memory flamegraph without rayon:
Screenshot_2022-03-24_10-48-05

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,8 +37,9 @@ binaries = [
     "console",
     "av-metrics",
 ]
-default = ["binaries", "asm", "signal_support"]
+default = ["binaries", "asm", "threading", "signal_support"]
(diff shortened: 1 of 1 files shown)


tmatth (approved):
Yay determinism.

PR #2915, opened 2022-03-24
Avoid temporary allocation of block importance costs

shssoichiro:
When tested at speed 3, this `collect` call was responsible for 50% of
all allocations performed by rav1e. The `collect` was necessary for the
row multithreading here to satisfy the borrow checker. However, even
after the removal of row multithreading in the `update_block_importances`
function, this change produces the following improvements:

At speed 3:
- 1.5% wall time improvement
- 22% CPU time improvement
- 2% peak memory reduction

At speed 10:
- 5% wall time improvement
- 40% CPU time improvement

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -21,7 +21,6 @@ use crate::rate::{
   RCState, FRAME_NSUBTYPES, FRAME_SUBTYPE_I, FRAME_SUBTYPE_P,
   FRAME_SUBTYPE_SEF,
 };
-use crate::rayon::prelude::*;
 use crate::scenechange::SceneChangeDetector;
 use crate::stats::EncoderStats;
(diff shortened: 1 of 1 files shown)


rzumer (approved):

shssoichiro:
I've seen some confusion about the difference between wall time and CPU time, so I wanted to clarify here at the source:

Wall time is the time the rav1e binary spends encoding when it has access to all the CPU cores on the system.
CPU time is the sum of all the time spent by all of the cores on encoding. This will be equal to the time spent encoding if the encode is pinned to 1 thread, which is often the case with tools such as av1an. Therefore, you'll see the biggest speedup if you're using a chunked encoding tool like av1an.

Illustrated:

rav1e threading illustrated

PR #2920, opened 2022-04-04
Fix clippy const fn warnings in v_frame

shssoichiro:
Not really sure how these didn't get caught earlier, but this fixes them now.


shssoichiro:
Never mind, it's because I was on nightly. This will need to be shelved for a while.

PR #2921, opened 2022-04-08
Variance-based Spatial AQ

shssoichiro:
Co-authored by @cyanreg 

This changeset revives #2247 with a more minimal implementation, simply basing AQ off of the existing variance calculations per 8x8 block. This changeset does not have a noticeable impact on encoding speed.

As an immediate followup to this pull request, I plan to add luma-based AQ bias to provide more bits to dark areas and adapt AQ for HDR clips. Ideas for future improvements are also very welcome.

This changeset does harm metrics, as expected of a psy change. The AWCY links are posted below both for details and for aomanalyzer links to compare before/after images.

[AWCY](https://bit.ly/3joZ5Za)

### DIFF
--- a/src/activity.rs
+++ b/src/activity.rs
@@ -9,14 +9,18 @@
 
 use crate::frame::*;
 use crate::rdo::DistortionScale;
+use crate::segmentation::MAX_SEGMENTS;
 use crate::tiling::*;
 use crate::util::*;
 use itertools::izip;
 use rust_hawktracer::*;
+use statrs::statistics::Data;
+use statrs::statistics::OrderStatistics;
 
 #[derive(Debug, Default, Clone)]
 pub struct ActivityMask {
-  variances: Box<[u32]>,
+  pub variances: Box<[u32]>,
+  pub segments: Box<[u8]>,
 }
 
 impl ActivityMask {
@@ -52,7 +56,26 @@ impl ActivityMask {
         variances.push(variance);
       }
     }
-    ActivityMask { variances: variances.into_boxed_slice() }
+
+    // Compute initial segmentation choices for variance AQ
+    // We take the 95th percentile of variances here to get a less skewed sample.
+    // Otherwise segment 0 becomes heavily oversaturated.
+    let mut fdata = Data::new(
+      variances.iter().map(|&var| var as f64).collect::<Box<[f64]>>(),
+    );
+    let max_var = fdata.percentile(95);
+    let segments = variances
+      .iter()
+      .map(|var| {
+        clamp(
+          (*var as f64 / max_var * MAX_SEGMENTS as f64).floor(),
+          0.0,
+          (MAX_SEGMENTS - 1) as f64,
+        ) as u8
+      })
+      .collect();
+
+    ActivityMask { variances: variances.into_boxed_slice(), segments }
   }
 
   #[hawktracer(activity_mask_fill_scales)]
--- a/src/activity.rs
+++ b/src/activity.rs
@@ -21,7 +21,9 @@ pub struct ActivityMask {
 
 impl ActivityMask {
   #[hawktracer(activity_mask_from_plane)]
-  pub fn from_plane<T: Pixel>(luma_plane: &Plane<T>) -> ActivityMask {
+  pub fn from_plane<T: Pixel>(
+    luma_plane: &Plane<T>, bit_depth: usize,
+  ) -> ActivityMask {
     let PlaneConfig { width, height, .. } = luma_plane.cfg;
 
     // Width and height are padded to 8×8 block size.
@@ -48,7 +50,7 @@ impl ActivityMask {
         };
 
         let block = luma.subregion(block_rect);
-        let variance = variance_8x8(&block);
+        let variance = variance_8x8(&block, bit_depth);
         variances.push(variance);
       }
     }
@@ -68,9 +70,11 @@ impl ActivityMask {
 
 // Adapted from the source variance calculation in `cdef_dist_wxh_8x8`.
 #[inline(never)]
-fn variance_8x8<T: Pixel>(src: &PlaneRegion<'_, T>) -> u32 {
+fn variance_8x8<T: Pixel>(src: &PlaneRegion<'_, T>, bit_depth: usize) -> u32 {
   debug_assert!(src.plane_cfg.xdec == 0);
   debug_assert!(src.plane_cfg.ydec == 0);
+
+  let bd_shift = bit_depth - 8;
 
   // Sum into columns to improve auto-vectorization
   let mut sum_s_cols: [u16; 8] = [0; 8];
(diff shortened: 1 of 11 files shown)


shssoichiro:
Opened up hoping to get an initial review. It is fully functional, but I don't know if I'm satisfied with the results, so I wanted to get more eyes on it.

src/segmentation.rs
```diff
@@ -31,50 +35,202 @@ pub fn segmentation_optimize<T: Pixel>(
...
+  let coded_data = fi.coded_frame_data.as_ref().unwrap();
+  let avg_seg = coded_data.activity_mask.avg_seg;
+  let mut seg_bins = coded_data.activity_mask.seg_bins;
+  let threshold = coded_data.activity_mask.variances.len() / 6;
+
+  let mut num_neg = 0usize;
+  let mut num_pos = 0usize;
+
+  let mut tmp_delta = [0f64; 8];
+  for i in 0..8 {
```
lu-zero, line 80: Make the `8` a const with a name please

shssoichiro:
Now verified that this branch with `--aq-strength 0` results in the same output as master.

shssoichiro:
New AWCY up with the latest changes, Feedback welcome.

https://beta.arewecompressedyet.com/?job=master-s6%402022-04-06T22%3A13%3A18.476Z&job=aq-experiment1%402022-04-08T23%3A08%3A19.873Z

BlueSwordM:
So, I currently do think the spatial AQ helps with image integrity and fidelity overall, although there are a few outliers like the Foodmarket clip at higher qualities(Q80 and Q128) where the normal image seems to have the most consistent image in terms of quality, although that might come down to rav1e's segmentation and activity masking playing rather well with the high frequency detail and there been rather few low variance spots overall.

I'll test it on my side with some challenging clips to see what could be done, but I have the impression a change like doing variable AQ block size analysis(4x4-16x16) would benefit image quality consistency. That would make the underlying code more complex, but it could work that way.

shssoichiro:
It's late right now so I won't test this until morning, but it seems like `segmentation_optimize_aq` is also doing some suboptimal things with how it reduces the number of segments. It seems to always end up with about 90% of the blocks in segment 0, 9% in segment 6, and 1% in segment 7, even though the raw segment decisions are more like a downward curve.

```
[src/segmentation.rs:85] seg_bins = [
    27045,
    3749,
    1128,
    356,
    86,
    25,
    7,
    4,
]
[src/segmentation.rs:184] seg_bins = [
    31922,
    0,
    0,
    0,
    0,
    0,
    474,
    4,
]
```

So I'll look into how this can be done better. In the morning. :slightly_smiling_face:

shssoichiro:
Okay, I think I am now reasonably satisfied with the results. And as such I will be moving the PR out of draft.

Latest AWCY: https://bit.ly/3joZ5Za

src/segmentation.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2021, The rav1e contributors. All rights reserved
+// Copyright (c) AQ_BASE_SEGMENT_IDX018-2021, The rav1e contributors. All rights reserved
```
lu-zero, line 1: This is spurious ^^

shssoichiro:
I've gotten quite a few eyes on it, I think generally people are seeing a visual improvement. Although noted is a need to give more bits to low luma areas, which is a tweak I plan to make immediately following this PR.

That being said, I'd love to get a code review on this now. :)

src/bin/common.rs
```diff
@@ -735,6 +742,12 @@ fn parse_config(matches: &ArgMatches) -> Result<EncoderConfig, CliError> {
   if cfg.tune == Tune::Psychovisual {
     cfg.speed_settings.transform.tx_domain_distortion = false;
   }
+  if cfg.tune == Tune::Psnr {
+    cfg.aq_strength = 0.0;
+  }
+  if let Some(aq_str) = matches.value_of("AQ_STRENGTH") {
```
doctortheemh, line 748: Should this be possible in Tune::Psnr is requested?

shssoichiro: It would be weird. Typically in encoders, the finer grained options override broader options such as presets and tunes, so with that idea I decided that `--aq-strength` should take precedence. But maybe it should be disallowed? And print a warning to the user?

doctortheemh: Letting AQ_STRENGTH override the broader tune makes sense to me, especially if there's precedent. It could also be documented in the argument definition or program help?

src/activity.rs
```diff
@@ -52,7 +56,26 @@ impl ActivityMask {
         variances.push(variance);
       }
     }
-    ActivityMask { variances: variances.into_boxed_slice() }
+
+    // Compute initial segmentation choices for variance AQ
+    // We take the 95th percentile of variances here to get a less skewed sample.
+    // Otherwise segment 0 becomes heavily oversaturated.
```
tdaede, line 62: So one of the original problems with the initial patch is selecting segments this way isn't compatible with e.g. selecting segments for Temporal RDO.

I don't quite understand how this patch is trying to resolve that. I see it's creating a list of segments for the variance, and then trying to push around what segments of those are used for temporal RDO. This seems unlikely to produce comparable temporal RDO results when RDO is on. Is that correct?

shssoichiro: This portion of the code calculates the variance portion of the segmentation decision, but in `segmentation_optimize_aq` is where the final segments are calculated by weighting the variance and the temporal RDO decisions together.

tdaede: Basically, the thing I don't like here is the way we produce the segmentation binning. I would expect the initial segment splits (this is producing 8 of them split equally among the variance range, as I read) to be produced *after* the variances are combined with the temporal RDO decisions.

src/segmentation.rs
```diff
@@ -89,41 +279,111 @@ pub fn select_segment<T: Pixel>(
...
+    let frame_bo = PlaneBlockOffset(BlockOffset {
+      x: (i % w_in_imp_b) << 1,
+      y: (i / w_in_imp_b) << 1,
+    });
+
+    // Apply temporal RDO bias
+    let trdo_sidx =
+      get_temporal_rdo_sidx(fi, &ts, frame_bo, BlockSize::BLOCK_16X16, false);
+    // TODO: Use more gradual adjustments
+    match trdo_sidx {
```
tdaede, line 371: In particular this seems not equivalent to what we were doing before. Before for temporal RDO, we had 3 segments, +1, -1 and 0. Now, we are bumping segment numbers up and down amounts in the amount of aq_strength for temporal RDO, which seems wrong.

src/activity.rs
shssoichiro: I see what you're saying. I don't think the way I'm doing the binning is "wrong" per se, but your method does sound likely to be better. I'll try to put something together where we merge the variance and RDO results before binning.

tdaede: I guess another issue is us using the max variance of the frame. This is going to vary from frame to frame and end up with our segmentation map changing a lot (if I'm reading this right?). Using percentiles helps, but another option might just be to use a fixed max variance entirely (and therefore, effectively fixed segment map). Some frames with low variance might not use all the segments, but I think that's an ok tradeoff.

shssoichiro:
I'm going to close this PR for now, and soon I'll be opening a new one that takes a bit of a different approach. The activity level is already factored into the spatiotemporal_scale function, so I want to make use of that more directly like we are currently.

PR #2924, opened 2022-04-12
Implement photon noise based grain synthesis

shssoichiro:
This changeset adds two new command line options:

`--photon-noise` allows the user to specify a strength from 0-64 (0 = disabled) to enable grain synthesis in rav1e. This implementation of grain synthesis is based on the photon noise tables used by aomenc and SVT-AV1. rav1e internally generates a noise table based on the user's strength setting, the video resolution, and whether the video is HDR, and then applies it to the encode. This implementation does not do anything such as denoising, or matching grain to the amount of grain in the source. The amount of grain is based on the luma values of each block. Many users of the photon noise tables in aomenc have said that they prefer the way this looks over aomenc's traditional grain synth method.

`--photon-noise-table` is a more advanced version which allows the user to manually specify a path to a film grain table. These tables can be generated by aomenc's `photon_noise_table` tool, or by av1an. This allows the user to have more customization over the grain synthesis, enabling features such as chroma grain.

The recommended option for most users will be to use the simpler `--photon-noise`. It is disabled by default.

The patch has virtually no speed impact, and adds an overhead of about 40 bytes per frame with luma-only synth, or 90 bytes per frame with luma and chroma synth.

For now, this changeset will be behind the `unstable` feature flag, until denoising is implemented.


vibhoothi (commented):
Hi, thanks for this, 
Quick comment, license headers is missing for few files^^

BlueSwordM:
Very nice patch overall.

Certainly a great step forward, especially since photon noise grain synth has been working quite well for me and many people.

While a complete excellent grain synth implementation would be the best, having this is great honestly.

The grain synth implementations in aomenc and SVT-AV1 are currently sub-optimal. Tuned photon noise, while a bit less adaptive, has far more consistent and better performance in general. SVT-AV1 is certainly better than aomenc, but after a lot of experimentation, not that far ahead of aomenc honestly. 

There's also the advantage of having no runtime performance loss over even a very optimized implementation.

src/api/channel/by_gop.rs
```diff
@@ -52,7 +52,7 @@ impl<T: Pixel> SceneChange<T> {
     let seq = Arc::new(Sequence::new(enc));
 
     let detector = SceneChangeDetector::new(
-      *enc,
+      enc.clone(),
```
tdaede, line 55: I don't understand why this switched to being a clone().

tdaede (approved):
Looks fine.

FWIW I wouldn't even be opposed to exposing this more easily than unstable if that makes testing with av1an's film grain tools better, as long as we understand what those are.

src/api/channel/by_gop.rs
shssoichiro: Unfortunately there needed to be a field added for the film grain settings which is a Vec, which prevents it from being copy-able.

tdaede:
OK. (Could this be an ArrayVec like is used by other parts of the code? Maybe we don't want ArrayVec in the public api?)

shssoichiro:
A user could have a table with any number of different film grain segments, so unfortunately a Vec is needed. (Also, ArrayVec is unfortunately also `!Copy`)

Issue #2356, opened 2020-05-31
Support non-utf8 file systems


lu-zero:
`Path`s may not be utf8, currently we pass them around as `&str` in few places.

PR #2926, opened 2022-04-28
Treat IO paths as native os strings.

doctortheemh:
Treat input, output, first pass, second pass, and reconstruction IO paths as `OsString` to improve portability. This requires the corresponding arguments be allowed to be non-utf8. This is a first pass at the issue to support non-utf8 file systems (#2356).

### DIFF
--- a/src/bin/muxer/mod.rs
+++ b/src/bin/muxer/mod.rs
@@ -32,18 +32,21 @@ pub trait Muxer: Send {
   fn flush(&mut self) -> io::Result<()>;
 }
 
-pub fn create_muxer(
-  path: &str, overwrite: bool,
+pub fn create_muxer<P: AsRef<Path>>(
+  path: P, overwrite: bool,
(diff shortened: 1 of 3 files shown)


src/bin/muxer/mod.rs
```diff
@@ -32,18 +32,21 @@ pub trait Muxer: Send {
...
+    if path == "-" {
+      return IvfMuxer::open(path);
+    }
   }
 
-  let ext = Path::new(path)
+  let ext = path
+    .as_ref()
     .extension()
     .and_then(OsStr::to_str)
```
tdaede, line 51: You can probably drop this conversion to OsStr now.

doctortheemh: Do you mean the conversion to &str? We use it lower in match. Matching on OsStr is still in rfc.

tdaede: Ah I see.

tdaede (approved):

PR #2927, opened 2022-04-29
Use const generics for downscaling

redzic:
The main bottleneck of the fast scene detection mode is the internal downscaling which is performed. This PR changes the downscaling code to be specialized for each chosen scale factor (rather than one function which handles all scale factors), which in turn enables several optimizations in the compiler that are not possible for a general downscaling function. Overall, this PR's new downscaling is over 6x faster when compiling with AVX2 enabled, benchmarked on a 1280x720 yuv420p video.

Other changes:
- Remove rayon threading from `downscale_in_place`, as it seemed to prevent auto-vectorization by the compiler
- Allocate `frame_ref_buffer` and `downscaled_frame_buffer` on the stack instead of the heap
- Remove initialization checks with `unreachable_unchecked` (which originally existed because of borrow checker limitations)

Comparison of performance:

Before this PR:

image

After this PR:

image

With this PR, `fast_scenecut` in total (on the tested 720p clip) is about 5.9x faster (0.798 ms vs 0.135 ms average wall duration, which amounts to a difference of 1253 fps and 7407 fps throughput).

### DIFF
--- a/src/scenechange/fast.rs
+++ b/src/scenechange/fast.rs
@@ -5,11 +5,13 @@ use crate::{
   encoder::Sequence,
   frame::{Frame, Plane},
   sad_row,
+  scenechange::fast_idiv,
 };
+use debug_unreachable::debug_unreachable;
 use rust_hawktracer::*;
 use v_frame::pixel::Pixel;
 
-use super::{SceneChangeDetector, ScenecutResult};
+use super::{ScaleFunction, SceneChangeDetector, ScenecutResult};
 
 /// Experiments have determined this to be an optimal threshold
 pub(super) const FAST_THRESHOLD: f64 = 18.0;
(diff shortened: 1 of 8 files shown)


src/scenechange/fast.rs
```diff
@@ -86,7 +69,36 @@ impl<T: Pixel> SceneChangeDetector<T> {
           backward_adjusted_cost: delta as f64,
         }
       } else {
-        unreachable!()
+        // SAFETY: `frame_buffer` is unconditionally initialized at the start of this code block,
+        // and this branch is only reached when `frame_buffer` is `None`.
+        unsafe { std::hint::unreachable_unchecked() }
```
shssoichiro, line 74: I have mixed feelings about using this, although it does seem like we have it in a few places in ASM functions. It would actually be _really cool_ if there was a version of this that performed the check and panicked in debug mode, but ignored the check and allowed optimizations in release mode.

Edit: Actually it looks like such a crate exists: https://crates.io/crates/new_debug_unreachable

redzic: Should we replace all usages of `unreachable_unchecked` with that macro then?

redzic: I've replaced all usages of `unreachable_unchecked` with the macro from the crate you linked since there seems to be no downside. Thanks!

shssoichiro: Thanks! (I checked the macro code and I don't know any reason it shouldn't work :sweat_smile: )

shssoichiro (approved):

Issue #2928, opened 2022-05-01
Permission denied (os error 13)), exit_code: 101 }


jiapei100:
**Describe the bug**
Successfully build rav1e with `cargo build --release`, but failed to install with command `cargo cinstall --release`. And `sudo cargo cinstall --release` tells me 
```console
➜  rav1e git:(master) sudo cargo install --release
error: no override and no default toolchain set
```

**To Reproduce**
Just build and install

**Expected behavior**
Successfully built already, but how to install a package with `sudo cargo cinstall --release` successfully ?? 

**Required Information**
Toolchain (if is a build problem):
```
➜  ~ cargo --version
cargo 1.60.0 (d1fd9fe 2022-03-01)
➜  ~ rustc --version
rustc 1.60.0 (7737e0b5c 2022-04-04)
➜  ~ nasm --version
NASM version 2.15.05
```
Version:
```
➜  rav1e git:(master) ./target/release/rav1e --version
rav1e 0.5.0 (p20210601-215-gb5c76736) (release)
```
Operating system:
```
➜  ~ uname -r
5.15.0-27-generic
➜  ~ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04 LTS
Release:	22.04
Codename:	jammy
```

**Console Output**
Please paste the console output below if it is relevant,
<Details>
➜  rav1e git:(master) cargo cinstall --release
    Finished release [optimized + debuginfo] target(s) in 0.28s
  Installing pkg-config file
Error: CliError { error: Some(failed to copy `....../rav1e/target/x86_64-unknown-linux-gnu/release/rav1e.pc` to `/usr/local/lib/pkgconfig/rav1e.pc`

Caused by:
    Permission denied (os error 13)), exit_code: 101 }
</Details>

Can anybody give me a hand?

lu-zero:
Your rust setup seems incomplete and you seem to not have the right permissions, to install on `/usr/local`.
You may use something along the lines of
```
D=/tmp/rav1e-install
cargo cinstall --release --destdir=${D}
sudo cp -a ${D}/* /
```

FCLC:
Running into this currently

lu-zero:
install in a user-writable path and copy it over the system using sudo and either `cp -a` or `install` depending on your preferences.

PR #2929, opened 2022-05-01
Expand on how to install the library

lu-zero:
Suggest to use `--destdir` and then `sudo cp` the files.

See: #2928

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,11 @@ To build and install it you can use [cargo-c](https://crates.io/crates/cargo-c):
 
 ```sh
 cargo install cargo-c
-cargo cinstall --release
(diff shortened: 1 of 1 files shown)


tdaede:
Should probably use the `install` binary rather than cp to get proper permissions set.

There's also an incantation of cargo cinstall that I managed to install directly with, I can try to find it.

lu-zero:
cargo cinstall does not know about permission, I might add yet another feature to deal with that maybe.

lu-zero:
If you have ready the install command I can add it as well.

lu-zero:
It should be otherwise covered

PR #2930, opened 2022-05-01
Improve codegen of copy_from_raw_u8

redzic:
The optimizer cannot currently recognize that on little-endian architectures, the loop is equivalent to just a memcpy. However, when using `u16::from_le_bytes`, it recognizes the pattern correctly and emits a memcpy on little-endian architectures (and would generate efficient code to swap the bytes on big-endian architectures, at least that is what it does on x86 when using `u16::from_be_bytes` which requires an endianness swap since x86 itself is a little-endian architecture).

De-referencing a fixed-size slice of 2 elements seems to be crucial for the optimizer to emit the most efficient code, i.e. the code generated is not better than the current implementation when using `u16::from_le_bytes([bytes[0], bytes[1]])`. Passing fixed-size slices can be achieved safely by using `source_row.array_chunks::<2>()`, however since `array_chunks` is currently unstable, `slice::from_raw_parts` is used to achieve the same effect.

This PR also replaces chunks/chunks_exact with simple indexes for this function since chunks/chunks_exact uses slow integer division instructions when using a chunk size that is not known at compile time.

Comparison of generated assembly of inner loop: https://godbolt.org/z/PzaP71j5j

### DIFF
--- a/v_frame/src/plane.rs
+++ b/v_frame/src/plane.rs
@@ -428,10 +428,14 @@ impl<T: Pixel> Plane<T> {
     &mut self, source: &[u8], source_stride: usize, source_bytewidth: usize,
   ) {
     let stride = self.cfg.stride;
+
+    assert!(stride != 0);
+    assert!(source_stride != 0);
+
     for (self_row, source_row) in self
       .data_origin_mut()
-      .chunks_mut(stride)
-      .zip(source.chunks(source_stride))
+      .chunks_exact_mut(stride)
+      .zip(source.chunks_exact(source_stride))
     {
       match source_bytewidth {
         1 => {
@@ -443,14 +447,26 @@ impl<T: Pixel> Plane<T> {
         }
         2 => {
           assert!(
-            mem::size_of::<T>() >= 2,
+            mem::size_of::<T>() == 2,
             "source bytewidth ({}) cannot fit in Plane<u8>",
             source_bytewidth
           );
-          for (self_pixel, bytes) in
-            self_row.iter_mut().zip(source_row.chunks_exact(2))
-          {
-            *self_pixel = T::cast_from(bytes[1]) << 8 | T::cast_from(bytes[0]);
+
+          debug_assert!(T::type_enum() == PixelType::U16);
+
+          // SAFETY: because of the assert it is safe to assume that T == u16
+          let self_row: &mut [u16] = unsafe { std::mem::transmute(self_row) };
+          // SAFETY: we reinterpret the slice of bytes as a slice of elements of
+          // [u8; 2] to allow for more efficient codegen with from_le_bytes
+          let source_row: &[[u8; 2]] = unsafe {
+            std::slice::from_raw_parts(
+              source_row.as_ptr().cast(),
+              source_row.len() / 2,
+            )
+          };
+
+          for (self_pixel, bytes) in self_row.iter_mut().zip(source_row) {
+            *self_pixel = u16::from_le_bytes(*bytes);
           }
         }
 


v_frame/Cargo.toml
```diff
@@ -22,3 +22,4 @@ serde = { version = "1.0", features = ["derive"], optional = true }
 wasm-bindgen = { version = "0.2.63", optional = true }
 rust_hawktracer = "0.7.0"
 new_debug_unreachable = "1.0.4"
+arrayref = "0.3.6"
```
lu-zero, line 25: Maybe add a note that it will be unneeded from a future version of rust.

v_frame/src/plane.rs
```diff
@@ -428,11 +429,26 @@ impl<T: Pixel> Plane<T> {
     &mut self, source: &[u8], source_stride: usize, source_bytewidth: usize,
   ) {
     let stride = self.cfg.stride;
-    for (self_row, source_row) in self
-      .data_origin_mut()
-      .chunks_mut(stride)
```
lu-zero, line 433: how far is using `chunks_exact`  from this optimality?

tdaede:
Nice, glad that you caught the fact that it works with 2-byte slices.

I did also notice the ArrayChunks implementation in the stdlib is extremely simple, at its core it is just a recast(especially as we don't care about the remainder feature): https://doc.rust-lang.org/src/core/slice/mod.rs.html#1075

Possibly we could just put that in directly given we're already doing an unsafe transmute here. Seems slightly more straightforward than the macro, though otherwise I don't have a strong preference.

v_frame/src/plane.rs
```diff
@@ -443,14 +459,20 @@ impl<T: Pixel> Plane<T> {
...
+          debug_assert!(T::type_enum() == PixelType::U16);
+
+          // SAFETY: because of the assert it is safe to assume that T == u16
+          let self_row: &mut [u16] = unsafe { std::mem::transmute(self_row) };
+
           for (self_pixel, bytes) in
             self_row.iter_mut().zip(source_row.chunks_exact(2))
           {
-            *self_pixel = T::cast_from(bytes[1]) << 8 | T::cast_from(bytes[0]);
+            *self_pixel = u16::from_le_bytes(*array_ref!(bytes, 0, 2));
```
lu-zero, line 475: This looks fine, no matter the rest.

lu-zero (commented):
I'd land the use of `from_le_bytes` immediately, I'd like a second opinion (@shssoichiro ?) for the outer loop although I'm 65% fine with it.

Thank you a lot :)

shssoichiro:
I think the outer loop looks fine except that the variable names are potentially confusing. Given that `self_row_idx` and `source_row_idx` are the byte index of the first byte in the row relative to the entire plane. Maybe `self_start_idx` and `source_start_idx`, or `self_row_start` and `source_row_start` would be a clearer name?

redzic:
> I did also notice the ArrayChunks implementation in the stdlib is extremely simple, at its core it is just a recast(especially as we don't care about the remainder feature): https://doc.rust-lang.org/src/core/slice/mod.rs.html#1075

> Possibly we could just put that in directly given we're already doing an unsafe transmute here. Seems slightly more straightforward than the macro, though otherwise I don't have a strong preference.

You're right, I didn't even think of that. Doing a recast as `as_chunks` does seems simple, so I went for doing that instead as that also allows us to drop the arrayref dependency. 

@shssoichiro Thanks, I also renamed the variables to what you suggested since it does seem clearer than before.

v_frame/Cargo.toml
redzic: Following the advice of @tdaede I dropped arrayref altogether and we just use `slice::from_raw_parts` now

v_frame/src/plane.rs
redzic: `chunks_exact` uses a couple of integer division instructions at the start of the loop to get the amount of chunks, while the manual indexing does not require these operations. However, it seems like the approach with `step_by` has more instructions to carry out the inner loop, see here: https://godbolt.org/z/MTc8PG96W

I'll see if I can find something that avoids integer division while also having a small inner loop

redzic:
For now I just changed it to use `chunks_exact` since it seems to generate a much less complicated inner loop.

redzic:
I found a way to reduce the amount of integer divisions and get an even more efficient outer loop than `chunks_exact`, but the code is a bit complicated and uses even more `unsafe`

redzic:
Sorry for all the noise. I changed it back to `chunks_exact` again since on closer analysis I can't really tell if the weird pointer arithmetic version is actually better than `chunks_exact` version, and in any case I don't think it's worth micro-optimizing the outer loop anyways.

v_frame/src/plane.rs
redzic: I decided to just use chunks_exact

shssoichiro (approved):

PR #2931, opened 2022-05-02
Add pre-processing denoising

shssoichiro:
This changeset adds a `--denoise` CLI option which enables denoising prior to encoding. This takes a strength value from 0-50, where 0 disables denoising. The default is 0, or half of the `--photon-noise` setting if `--photon-noise` is enabled. `--denoise` can be set manually and will override the denoise strength chosen by `--photon-noise`.

The denoiser implemented is a FFT-based spatio-temporal denoiser based on the DFTTest plugin from Vapoursynth. This was chosen because it provides a reasonable balance of speed and quality.

This also moves the `--photon-noise` and `--photon-noise-table` args into the stable feature set, as was discussed in #2924.

This is currently marked as _draft_ due to performance concerns.


shssoichiro:
I think a large portion of the performance issue is due to the fact that each call to the FFT functions allocates a new scratch buffer. I'll need to make an addition to ndrustfft's API in order to support reusing an external scratch buffer, but I think that should bring performance of this patch considerably closer to the C version.

src/denoise.rs
```diff
@@ -0,0 +1,587 @@
+use crate::api::FrameQueue;
```
vibhoothi, line 1: Nit: License file is missing.

NSQY:
Have you seen this implementation? https://github.com/AmusementClub/vs-dfttest2

On my 3900X, with AVX2 enabled, it is more than twice as fast as https://github.com/AmusementClub/vs-dfttest2 in a synthetic VS script.

shssoichiro:
I have, come to think of it, it is an option... Since it seems to roll its own fft kernels instead of using fftw. The issue with this current PR, and something I want to keep in consideration for future PRs, is that currently rav1e has no dependencies on C libraries with the exception of nasm. To keep it simple to build, we'd prefer to keep it that way. However, the best available library for doing multi-dimensional FFTs in Rust at the moment is 10-20x slower than fftw, which by extension makes this implementation 10-20x slower than the Vapoursynth version of dfttest (which uses fftw).

src/denoise.rs
```diff
@@ -0,0 +1,607 @@
...
+      dftc[h].re += means[h].re;
+      dftc[h].im += means[h].im;
+    }
+  }
+
+  #[inline]
+  // Applies a generalized wiener filter
+  fn filter_coeffs(&self, dftc: &mut [Complex<f32>; COMPLEX_COUNT]) {
+    for h in 0..COMPLEX_COUNT {
+      let psd = dftc[h].re.mul_add(dftc[h].re, dftc[h].im.powi(2));
```
tdaede, line 424: Is this missing squaring the real part?

shssoichiro: No, it translates to `real * real + im^2`. The reason it's done like this is because `mul_add` is more efficient and accurate for floating point math.

PR #2933, opened 2022-05-07
Compute segmentation offsets dynamically

### LINKED ISSUE #752: Adaptive Quantization
Many encoders, including AOM, implement an adaptive quantization algorithm, which is a psychovisual algorithm designed to distribute bitrate among the macroblocks in a frame in a way that improves perceived quality.

AOM implements three different algorithms: 
- variance-based
- complexity-based
- cyclic refresh. 

There are also variations on these implemented by other encoders; e.g. x264 implements three variations of a variance-based AQ, one of which also uses a complexity mask. 

We should select which AQ algorithms we would like to implement, and select one to begin with as a "default".

shssoichiro:
This changeset, in lieu of reusing variance calculations to do spatial AQ, factors in the fact that the spatiotemporal scales we have _already consider_ both spatial and temporal importance. Therefore, this changeset uses the existing metric and uses an assessment of the distribution of that metric within a frame to determine how many segments to use and what the deltaq offset should be for each of them, rather than using a constant set of +21/-21/0.

The segment splitting code in this changeset is originally from @cyanreg in #2247.

As an immediate followup to this pull request, I still plan to add luma-based AQ bias to provide more bits to dark areas and adapt AQ for HDR clips. Ideas for future improvements are also very welcome.

[AWCY](https://beta.arewecompressedyet.com/?job=master-c23559d7bb0e2c97fac4603588d0f53477b2019a&job=spatial_aq_09%402022-05-07T01%3A42%3A16.301Z) shows promising results--neutral metrics and, in my opinion, good psychovisual improvements.

Closes #752

### DIFF
--- a/src/context/partition_unit.rs
+++ b/src/context/partition_unit.rs
@@ -249,7 +249,9 @@ impl<'a> ContextWriter<'a> {
     symbol_with_update!(self, w, skip as u32, cdf, 2);
   }
 
-  pub fn get_segment_pred(&self, bo: TileBlockOffset) -> (u8, u8) {
+  pub fn get_segment_pred(
+    &self, bo: TileBlockOffset, last_active_segid: u8,
+  ) -> (u8, u8) {
     let mut prev_ul = -1;
     let mut prev_u = -1;
     let mut prev_l = -1;
@@ -288,7 +290,8 @@ impl<'a> ContextWriter<'a> {
     } else {
       r = if prev_ul == prev_u { prev_u } else { prev_l };
     }
-    (r as u8, cdf_index)
+
+    ((r as u8).min(last_active_segid), cdf_index)
   }
 
   pub fn write_cfl_alphas<W: Writer>(&mut self, w: &mut W, cfl: CFLParams) {
@@ -434,7 +437,7 @@ impl<'a> ContextWriter<'a> {
     &mut self, w: &mut W, bo: TileBlockOffset, bsize: BlockSize, skip: bool,
     last_active_segid: u8,
   ) {
-    let (pred, cdf_index) = self.get_segment_pred(bo);
+    let (pred, cdf_index) = self.get_segment_pred(bo, last_active_segid);
     if skip {
       self.bc.blocks.set_segmentation_idx(bo, bsize, pred);
       return;
--- a/src/fuzzing.rs
+++ b/src/fuzzing.rs
@@ -232,6 +232,8 @@ impl Arbitrary for ArbitraryEncoder {
       tile_cols: u.int_in_range(0..=2)?,
       tile_rows: u.int_in_range(0..=2)?,
       tiles: u.int_in_range(0..=16)?,
+      aq_strength: *u
+        .choose(&[-1.0, -0.5, 0.0, 0.1, 0.5, 1.0, 1.5, 2.5, 69.0])?,
 
       chroma_sampling: *u.choose(&[
         ChromaSampling::Cs420,
--- a/src/segmentation.rs
+++ b/src/segmentation.rs
@@ -7,13 +7,22 @@
 // Media Patent License 1.0 was not distributed with this source code in the
 // PATENTS file, you can obtain it at www.aomedia.org/license/patent.
 
+use std::cmp::Ordering;
+
 use crate::context::*;
 use crate::header::PRIMARY_REF_NONE;
 use crate::partition::BlockSize;
+use crate::rdo::spatiotemporal_scale;
+use crate::rdo::DistortionScale;
 use crate::tiling::TileStateMut;
 use crate::util::Pixel;
 use crate::FrameInvariants;
 use crate::FrameState;
+
+pub const MAX_SEGMENTS: usize = 8;
+
+// This value was obtained through testing over multiple runs in AWCY.
+const BASE_AQ_MULT: f64 = -9.0;
 
 pub fn segmentation_optimize<T: Pixel>(
   fi: &FrameInvariants<T>, fs: &mut FrameState<T>,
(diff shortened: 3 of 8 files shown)


BlueSwordM:
Wow Soichiro.

This actually makes a positive difference in all clips, including Food Market, a difficult clip to encode and my reference for improving such performance.

BlueSwordM:
I do have some ideas regarding this: if possible, if you add the dark bias, also add a luma bias control and an AC quantizer bias control :P

BlueSwordM:
Oh I should have specified that yes, they should stay hidden as well.
The main reason to add these parameters would be to modify them in the future possibly based on quality ranges and different tuning for colorspaces as an example.

shssoichiro:
Bump

src/context/partition_unit.rs
```diff
@@ -288,7 +290,8 @@ impl<'a> ContextWriter<'a> {
     } else {
       r = if prev_ul == prev_u { prev_u } else { prev_l };
     }
-    (r as u8, cdf_index)
+
+    ((r as u8).min(last_active_segid), cdf_index)
```
tdaede, line 294: Was this a bug in our context code?

src/fuzzing.rs
```diff
@@ -232,6 +232,8 @@ impl Arbitrary for ArbitraryEncoder {
       tile_cols: u.int_in_range(0..=2)?,
       tile_rows: u.int_in_range(0..=2)?,
       tiles: u.int_in_range(0..=16)?,
+      aq_strength: *u
+        .choose(&[-1.0, -0.5, 0.0, 0.1, 0.5, 1.0, 1.5, 2.5, 69.0])?,
```
tdaede, line 236: nice

src/segmentation.rs
```diff
@@ -7,14 +7,23 @@
...
+use crate::rdo::DistortionScale;
 use crate::tiling::TileStateMut;
 use crate::util::Pixel;
 use crate::FrameInvariants;
 use crate::FrameState;
 
+pub const MAX_SEGMENTS: usize = 8;
+
+// This value was obtained through testing over multiple runs in AWCY.
+const BASE_AQ_MULT: f64 = -9.0;
```
tdaede, line 25: Is the idea that an AQ strength of 1 is optimal?

src/segmentation.rs
```diff
@@ -31,99 +40,226 @@ pub fn segmentation_optimize<T: Pixel>(
...
-    0
-  };
-
-  sidx..=sidx
+  // SAFETY: We know that every segment is between 0-7
+  unsafe {
+    let sidx = *ts
+      .segmentation
+      .segment_map
+      .get_unchecked(segment_idx_from_distortion(scale) as usize)
```
tdaede, line 248: will indexing with & 0x7 elide the bounds check here?

src/segmentation.rs
```diff
@@ -31,99 +40,226 @@ pub fn segmentation_optimize<T: Pixel>(
...
+) {
+  const AVG_SEG: f64 = 3.0;
+
+  let coded_data = fi.coded_frame_data.as_ref().unwrap();
+  let segments = &coded_data.imp_b_segments;
+  let mut seg_counts = [0usize; MAX_SEGMENTS];
+  segments.iter().for_each(|&seg| {
+    // SAFETY: When selecting the segments, we ensured they are all within the range of 0-7
+    unsafe {
+      *seg_counts.get_unchecked_mut(seg as usize) += 1;
```
tdaede, line 78: I think the Rusty way to do this would be to have seg be a custom type with a setter than guarantees this. Not sure if you're up for a Type System Quest or not.

src/segmentation.rs
```diff
@@ -31,99 +40,226 @@ pub fn segmentation_optimize<T: Pixel>(
...
+    let sidx = *ts
+      .segmentation
+      .segment_map
+      .get_unchecked(segment_idx_from_distortion(scale) as usize)
+      as u8;
+    sidx..=sidx
+  }
+}
+
+pub fn segment_idx_from_distortion(s: DistortionScale) -> u8 {
```
tdaede, line 254: Would be good to document where these numbers came from for someone who comes along and wants to improve it.

src/context/partition_unit.rs
shssoichiro: Kind of yes, kind of no. It previously never triggered any odd behavior due to the fixed number of segments/probably some amount of luck. With the dynamic segment selection, this was causing issues with skip blocks (see line 442).

src/segmentation.rs
shssoichiro: Yes, for the sake of user experience. A typical user would assume that the default strength should be "1.0", and then we try to make the default as optimal as possible.

src/segmentation.rs
shssoichiro: It seems that, yes, it does. That's a neat trick I'll have to keep in mind from now on.

I'm going to code it as a modulus because I think it makes the intention clearer, but according to Godbolt `%` and `&` generate the same asm: https://rust.godbolt.org/z/EnevWG6GG

src/segmentation.rs
shssoichiro: I'm wondering how useful it would be outside of this function. If it can be reusable, I think it's worthwhile, I'm just struggling to wrap my head around that at the moment.

However, the modulus trick can be used here to eliminate the `unsafe` block at least.

shssoichiro:
Bump 👀

src/activity.rs
```diff
@@ -52,6 +52,7 @@ impl ActivityMask {
         variances.push(variance);
       }
     }
+
```
lu-zero, line 55: this is stray?

lu-zero (approved):
The rest looks fine to me.

PR #2934, opened 2022-05-09
Remove uses of Arc::make_mut to avoid clones

shssoichiro:
Profiling revealed there to be a measurable number of clones coming out of `TileStateMut::new`, which was traceable to usage of `Arc::make_mut`. A characteristic of `Arc::make_mut` is that it will clone the inner data when a thread attempts to write to it. In theory, I have concerns that this could cause data inconsistencies, although that does not seem to have happened in practice in rav1e. However, the clones are avoidable, because each place where we use `Arc::make_mut` can be safely assumed to be independent--i.e. there is never more than one thread accessing the same portion of the data.

640x360 with 1 tile: 1.5% memory usage reduction
640x360 with 4 tiles: 4% memory usage reduction
1920x1080 with 4 tiles: 6% memory usage reduction

No measurable runtime improvement with 1 tile.
~1% runtime improvement with 4 tiles.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -47,6 +47,7 @@ use std::collections::VecDeque;
 use std::io::Write;
 use std::mem::MaybeUninit;
 use std::sync::Arc;
+use std::sync::Mutex;
 use std::{fmt, io, mem};
 
 use crate::rayon::iter::*;
@@ -438,6 +439,7 @@ pub struct FrameState<T: Pixel> {
   // these are stored per-tile for easier access.
   pub frame_me_stats: Arc<[FrameMEStats; REF_FRAMES as usize]>,
   pub enc_stats: EncoderStats,
+  pub active_tiles: Arc<Mutex<Vec<TileRect>>>,
 }
 
 impl<T: Pixel> FrameState<T> {
@@ -479,6 +481,7 @@ impl<T: Pixel> FrameState<T> {
       restoration: rs,
       frame_me_stats: me_stats,
       enc_stats: Default::default(),
+      active_tiles: Default::default(),
     }
   }
 
@@ -510,6 +513,7 @@ impl<T: Pixel> FrameState<T> {
       restoration: rs,
       frame_me_stats: FrameMEStats::new_arc_array(fi.w_in_b, fi.h_in_b),
       enc_stats: Default::default(),
+      active_tiles: Default::default(),
     }
   }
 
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -94,6 +94,10 @@ log = "0.4"
 console = { version = "0.14", optional = true }
 fern = { version = "0.6", optional = true }
 itertools = "0.10"
+# Used for creating global mutex-wrapped caches
+once_cell = "1.10"
+# Faster alternative to std::mutex
+parking_lot = "0.12"
```
lu-zero, line 100: the parking_lot mutex is [not always faster anymore](https://github.com/rust-lang/rust/pull/95035#issuecomment-1073966631) at least on linux since the futex-based mutex landed in std. it might worth further investigation for our specific workload.

shssoichiro: It looks like this is coming in 1.62. Maybe worth keeping parking_lot for the time being and then reevaluating? Or maybe we just rip it out, honestly tile creation is such a small portion of the time in rav1e that I doubt it matters.

lu-zero: I guess preparing a benchmark and see which between usync, parking_lot and std work best for us.

src/encoder.rs
```diff
@@ -992,13 +996,16 @@ impl<T: Pixel> FrameInvariants<T> {
...
-      if let Some(params) =
-        Arc::make_mut(&mut fi.config).get_film_grain_mut_at(cur_frame_time)
-      {
-        params.random_seed = params.random_seed.wrapping_add(3248);
-        if params.random_seed == 0 {
-          params.random_seed = DEFAULT_GS_SEED;
+      // SAFETY: We're not inside a tile, there's nothing else writing to the encoder config here.
+      unsafe {
+        // Increment the film grain seed for the next frame
+        if let Some(params) = arc_get_mut_unsafe(&mut fi.config)
```
lu-zero, line 1002: Wouldn't be better to move the film grain information to the FrameInvariants? Also who is accessing the `config` here?

src/encoder.rs
```diff
@@ -3242,12 +3256,16 @@ fn encode_tile_group<T: Pixel>(
...
       cdef_filter_tile(fi, &deblocked_frame, &blocks.as_tile_blocks(), rec);
     }
-    /* TODO: Don't apply if lossless */
-    fs.restoration.lrf_filter_frame(
-      Arc::get_mut(&mut fs.rec).unwrap(),
-      &deblocked_frame,
-      fi,
-    );
+    // SAFETY: We know no other threads are accessing this because we are done with tiles.
+    // The only references would be in ref frames, which are not being touched at the moment.
```
lu-zero, line 3260: What is the gain from doing this? Maybe it should be moved to a different patch.

lu-zero (changes requested):
`fs.frame_me_stats` hits `tile_iter_mut` with a count of 2, this is in itself fishy or possibly a bug that went unnoticed since a while.

Ideally get_mut() should always work.

lu-zero:
https://gist.github.com/lu-zero/ffc7152ab0aabc651f0fde3da42b4559 this fixes most of the problems
Still fail:
    test_encode_decode::bitrate_dav1d
    api::test::guess_frame_subtypes_assert
    api::test::log_q_exp_overflow
    api::test::max_quantizer_bounds_correctly
    api::test::min_quantizer_bounds_correctly

shssoichiro:
All the tests pass for me when I apply these changes against master. However, I still see the significant increase in memory usage, because this is basically the same as not having an `Arc` there--it's cloning the inner contents.

lu-zero:
You need 1 clone for the lookahead and 1 clone for the normal encoding in theory.

shssoichiro:
Closed for #2959

PR #2935, opened 2022-05-10
Enable source base code coverage on CI

Luni-4:
This PR uses the new [source-based code coverage](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#source-based-code-coverage) to compute code coverage

### DIFF
--- a/.github/actions-rs/grcov.yml
+++ b/.github/actions-rs/grcov.yml
@@ -1,4 +1,0 @@
-ignore-not-existing: true
-ignore:
-  - "/*"
-  - "../*"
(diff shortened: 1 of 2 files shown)


Luni-4:
Thank you @shssoichiro! 

I can see the last three attempts of today on coveralls: https://coveralls.io/repos/235642/builds
I guess the comment should appear on new PRs. Looking at the coveralls statistics we should ignore some directories in some way perhaps (like `target for example), but we can do that in a follow-up PR

shssoichiro (approved):

PR #2937, opened 2022-05-11
Do not consider some directories for coverage

Luni-4:
This PR shouldn't consider the `target` directory anymore for coverage


Luni-4:
It seems not to work...

shssoichiro:
Looks like it's working now?

I think we should also exclude `/tests` and `/src/test_encode_decode`

Luni-4:
Yep, seems ignored now, perhaps we just needed to wait for the update eh eh
Let me exclude the ones you have reported also

shssoichiro (approved):

PR #2938, opened 2022-05-12
Remove unused lifetime.

doctortheemh:
This lifetime is unused in this impl. This is a fix for an upcoming clippy lint.


tmatth (approved):
Nice catch.

PR #2939, opened 2022-05-14
Split `write_coeffs_lv_map` into smaller functions

shssoichiro:
This should make profiling and optimization easier, as well as improve
the code readability.

### DIFF
--- a/src/context/block_unit.rs
+++ b/src/context/block_unit.rs
@@ -1805,7 +1805,7 @@ impl<'a> ContextWriter<'a> {
     let coeffs = &mut coeffs_storage.data;
     coeffs.extend(scan.iter().map(|&scan_idx| coeffs_in[scan_idx as usize]));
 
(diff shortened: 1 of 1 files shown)


src/context/block_unit.rs
```diff
@@ -1850,7 +1851,19 @@ impl<'a> ContextWriter<'a> {
       );
     }
 
-    // Encode EOB
+    self.encode_eob(eob, tx_size, tx_class, txs_ctx, plane_type, w);
+    self.encode_coeffs(
+      coeffs, levels, scan, eob, tx_size, tx_class, txs_ctx, plane_type, w,
+    );
+    self.encode_coeff_signs(coeffs, w, plane_type, txb_ctx, &mut cul_level);
```
lu-zero, line 1858: You could return the `cul_level` and avoid the &mut.

tdaede (approved):
surprisingly hard to read diff

PR #2940, opened 2022-05-16
Rewrite x86 SAD row intrinsics in ASM

redzic:
This speeds up the sad_row functions mainly by avoiding a residual loop altogether, and avoiding mixing 128 bit and 256 bit vector instructions which causes slow downs on some CPUs. Instead of using a scalar residual loop to clean up the elements that don't fit into the SIMD register and unroll size, this code always uses vector instructions and just reads past the end of the array and masks out the irrelevant bytes (similar to glibc's AVX2 memchr implementation).

This also actually ends up de-duplicating some code between the SSE2 and AVX2 versions of the code, since a single macro is now used to generate both versions of the function.

Maybe in the future this function could just operate over 2 entire planes instead of just a row, this way there would be less looping overhead between rows (sad_row is only used on each row of an entire plane to get the SAD between 2 planes).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -108,6 +108,7 @@ fn build_nasm_files() {
     "src/x86/mc_sse.asm",
     "src/x86/mc16_sse.asm",
     "src/x86/me.asm",
+    "src/x86/sad_row.asm",
     "src/x86/sad_sse2.asm",
     "src/x86/sad_avx.asm",
     "src/x86/satd.asm",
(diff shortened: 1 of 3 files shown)


lu-zero (approved):
fine by me if the tests are happy.

shssoichiro:
For context: The reason sad_row (and a lot of other things) have to work at the row level instead of the plane level is because the visible area of a plane is not guaranteed to be a contiguous slice. There is often non-visible padding which we don't want included in the SAD calculation.

PR #2941, opened 2022-05-16
Remove explicit VEX coding from sad ASM

redzic:
I didn't initially see this problem in #2940. I believe executing VEX-encoded instructions on processors that don't support them (e.g., if it only supports sse2) will cause SIGILL.

### DIFF
--- a/src/x86/sad_row.asm
+++ b/src/x86/sad_row.asm
@@ -71,7 +71,7 @@ cglobal sad_row_8bpc, 3, 5, 8, p1, p2, len, \
   add     p2q, lenq
   sub     resid_simdq, lenq
   ; last register needed for unrolled loop
-  pxor   xm3, xm3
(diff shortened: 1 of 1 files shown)


shssoichiro (approved):
Basically, this is relying on nasm magic to choose the right one for the version we're compiling, right? As you described in your other issue.

redzic:
More specifically the macros defined in x86inc.asm, but yeah pretty much. `vpaddq`/`vpxor` is used on the avx2 version, and `paddq`/`pxor` is used on the sse2 version automatically.

redzic:
The coveralls bot seems broken by the way, I mean how does code coverage change based on adding a couple spaces?

shssoichiro:
I'm surprised coveralls is posting comments at all, given that their documentation said posting comments on PRs doesn't work with Github Actions. :shrug:

PR #2943, opened 2022-05-17
Reimplement sad_row as sad_plane

redzic:
This should remove any overhead of Rust having to loop through each row and call the ASM implementation of sad_row, and instead allow a new sad_plane function to iterate over the rows by itself.

### DIFF
--- a/src/x86/sad_plane.asm
+++ b/src/x86/sad_plane.asm
@@ -10,6 +10,8 @@
 %include "config.asm"
 %include "ext/x86/x86inc.asm"
 
+%if ARCH_X86_64
+
 SECTION_RODATA
 
 align 32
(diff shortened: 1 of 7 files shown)


redzic:
There seems to be an issue with strides that aren't powers of 2, looking into it

redzic:
Nevermind, that actually had to do with the allocations not being aligned sometimes in the test environment I wrote these functions in rather than there actually being a bug in the code. This should be fine for rav1e since Plane should already be sufficiently aligned.

shssoichiro:
This looks like it should be fine. I'm going to fire off an AWCY run just to make sure there is no change in output.

src/x86/sad_plane.asm
```diff
@@ -47,103 +49,132 @@ mask_lut: db \
...
+  lea     skip_ptrq, [jmp_table]
+  mov     skip_ptrq, [skip_ptrq + 8*resid_simdq]
+  ; shift back (for residual to load correct number of bytes)
+  shl     resid_simdq, ilog2(mmsize)
+  ; set pointer to point after end of width of first row
+  add     p1q, widthq
+  add     p2q, widthq
+  mov     width_unrllq, widthq
+  neg     widthq
+.loop_setup:
```
tdaede, line 107: maybe loop_row would make more sense as a label?

redzic: Seems reasonable, I changed the label to `loop_row`

PR #2944, opened 2022-05-19
Ignore some directories for code coverage analysis

Luni-4:
This PR ignores asm for code coverage analysis


shssoichiro:
Seems fine, since the Rust FFI should get covered for any assembly we use.

shssoichiro (approved):

PR #2953, opened 2022-05-28
Bump versions of aom and dav1d in CI

shssoichiro:
Builds were failing, so this became necessary.

Also deletes the .travis folder, which was unused.


.github/workflows/rav1e.yml
```diff
@@ -170,8 +170,11 @@ jobs:
           cd ~/.cache/debs
           xargs -a DEBS curl --remote-name-all
           sha256sum -c CHECKSUMS
+      # libdav1d6 depends on libc6 >= 2.32, which is not included in the docker image
       - name: Install packages
         run: |
+          apt update
+          apt --only-upgrade install libc6
```
barrbrain, line 177: We need `sudo` here:
```suggestion
          sudo apt update
          sudo apt --only-upgrade install libc6
```

barrbrain (approved):
Fetching packages succeeded, so just need to upgrade `libc6`.

barrbrain:
I think we need to move from `ubuntu-latest` to `ubuntu-22.04` to pick up a newer `libc6`. Here there be dragons.

shssoichiro:
Looks like it still didn't like that. I guess the required libc version just isn't in this ubuntu release.

This is why I use Arch.

shssoichiro:
In the interest of fixing CI, I'm going to revert to the previous dav1d version for now, since I _believe_ it was only libaom that was causing CI to fail.

barrbrain:
On the upside, all the current versions are in Ubuntu 22.04. So we could drop the current method and just:
```bash
sudo apt update
sudo apt install nasm libaom-dev libdav1d-dev
```

.github/workflows/rav1e.yml
```diff
@@ -114,54 +114,10 @@ jobs:
...
-        continue-on-error: true
-        id: debs
-        with:
-          path: ~/.cache/debs
-          key: debs-${{ hashFiles('CHECKSUMS') }}
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install nasm libaom-dev libdav1d-dev
       - name: Fetch packages
```
barrbrain, line 121: Need to also remove `Fetch packages` and `Install packages` steps.

PR #2954, opened 2022-05-28
x86: Fix relocations in `sad_plane`

redzic:
Should fix any linker failures that have been happening recently due to relocations in `sad_plane`.


redzic:
somehow using `mangle` was causing issues on macos, I think it's fixed now though

shssoichiro (approved):
Fingers crossed

PR #2955, opened 2022-05-31
Merge estimate_motion and motion_estimation into one function

shssoichiro:
They do very similar things but with slight tweaks, so it makes sense
for them to be one function instead of two to reduce code duplication
and remove naming confusion.

[AWCY](https://beta.arewecompressedyet.com/?job=master-295ad2e385366fe0f0a42957509275f43e941bf2&job=estimate-motion%402022-05-31T07%3A30%3A27.070Z) confirmed no change in output.


lu-zero (approved):
It looks fine to me, maybe wait for a second opinion from @barrbrain though.

PR #2956, opened 2022-06-02
Reuse me stats for lookahead

shssoichiro:
This partially addresses the comments in #2934 by removing the
unnecessary field `coded_data.lookahead_me_stats`. This data is never
modified between the time it is created in
`compute_lookahead_motion_vectors` and the time it is used in
`compute_block_importances`, so there is no need to keep a separate copy
of it. Instead, we can use the main copy of the data in
`fs.frame_me_stats`.

Results in ~1% peak memory usage reduction.


lu-zero:
Does it fix the broken invariant?

lu-zero (approved):

PR #2957, opened 2022-06-05
Reuse plane allocations in estimate_intra_costs

redzic:
As far as I understand, this frame buffer can be reused and is only read after written to, so the initial contents of the buffer should not matter.


shssoichiro:
Does this have an impact on performance? Given that it adds complexity to the code, I'd like to be sure it's actually reducing the number of large allocations and we're not just changing where the copies happen.

redzic:
According to hawktracer, there does some to be a measurable performance difference. `estimate_intra_costs` takes about 20% less time with this PR.

Current master:

image

With this PR:

image

redzic:
The `frame.planes[0].clone()` that you see in this PR only happens when the temporary frame buffer is not populated, and this should only happen once. After this field is populated, it is reused for the rest of the scene detection.

src/api/internal.rs
```diff
@@ -817,13 +817,27 @@ impl<T: Pixel> ContextInner<T> {
       .intra_costs
       .remove(&fi.input_frameno)
       .unwrap_or_else(|| {
+        let frame = self.frame_q[&fi.input_frameno].as_ref().unwrap();
+
+        let mut temp_plane =
+          if let Some(temp_plane) = self.keyframe_detector.temp_plane.take() {
```
shssoichiro, line 823: Is `take` necessary here to appease the borrow checker, or can this be done with a mutable ref via the `entry` API and avoid the insert at line 838? (Same with this code in `standard.rs`)

redzic: That's what it seems like to me, I think I tried some other way at first and the borrow checker was not happy about it. So the solution I came up with was to take ownership and later return it back. There might be a way to only swap pointers to the struct, so I'm definitely open to other solutions here. Although this shouldn't be *too* bad since it's moving around just the fields of the struct rather than the plane data itself, which obviously would be very expensive.

I searched the rust stdlib for `entry` but I only found stuff for hash maps and btrees, which I don't think applies here since we are just using an `Option` rather than any of those fancy data structures. I could've have missed it though, just let me know

shssoichiro: Oh, you're right, I got confused about the entry API. On an Option you'd want this: https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_with

And the compiler is _probably_ able to optimize this just fine. But it would be nice, if we can make it a little cleaner, to do so.

redzic: Ah, I see. Definitely an improvement over doing it the manual way, thanks!

redzic:
Oops, sorry I added some extra stuff to this PR that I didn't mean to. Working on fixing that now.

shssoichiro:
Thanks, this looks nicer! I'll approve after I've looked at Luca's PR, since that one seems like it may be harder to rebase, so I'd like to merge it first.

shssoichiro (approved):

PR #2958, opened 2022-06-05
Avoid taking ownership of Arc<Frame<T>> in scenechange API

redzic:
The current use of taking ownership of the passed in `Arc<Frame<T>>` in the scenechange module becomes problematic when trying to pass in a struct wrapped in `ManuallyDrop`, as it attempts to run the destructor and you can't coerce a `ManuallyDrop<Arc<Frame<T>>>` into an `Arc<Frame<T>>` that doesn't run the destructor.

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -853,7 +853,7 @@ impl<T: Pixel> ContextInner<T> {
     keyframe_detector: &mut SceneChangeDetector<T>,
     next_lookahead_frame: &mut u64, keyframes: &mut BTreeSet<u64>,
   ) {
-    if keyframes_forced.contains(&next_lookahead_frame)
+    if keyframes_forced.contains(next_lookahead_frame)
       || keyframe_detector.analyze_next_frame(
         lookahead_frames,
         *next_lookahead_frame,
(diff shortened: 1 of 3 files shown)


redzic:
There seems to be some issues with this PR when compiling with channel-api enabled. I'll try to look more into this tomorrow.

src/api/internal.rs
```diff
@@ -371,10 +371,22 @@ impl<T: Pixel> ContextInner<T> {
...
+          );
         }
       } else {
-        self.compute_keyframe_placement(&lookahead_frames);
+        Self::compute_keyframe_placement(
+          &lookahead_frames,
+          &self.keyframes_forced,
+          &mut self.keyframe_detector,
+          &mut self.next_lookahead_frame,
+          &mut self.keyframes,
```
shssoichiro, line 388: Did the borrow checker make you turn this into a list of parameters? 🙁

shssoichiro (approved):
A lot smaller changeset than expected. Looks good, thanks!

src/api/internal.rs
redzic: It's been a while so I can't remember the exact error, but yeah I think it was related to the borrow checker.

PR #2959, opened 2022-06-06
Tile rework

lu-zero:
This is some initial rework to remove the spurious Arc:make_mut()


shssoichiro (approved):
In the end this resulted in a 5% memory reduction across the board, as well as a slight speedup.

Issue #2960, opened 2022-06-23
msys2: cannot compile due to issue with crossbeam dependency


1480c1:
**Describe the bug**

Compile error due to crossbeam-epoch

```rust
error: linking with `i686-w64-mingw32-gcc` failed: exit code: 1
  |
  = note: "i686-w64-mingw32-gcc" ...
  = note: D:/media-autobuild_suite/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/12.1.0/../../../../i686-w64-mingw32/bin/ld.exe: D:\media-autobuild_suite\build\rav1e-git\target\i686-pc-windows-gnu\release\deps\rav1e-2bb7c42675a4c7b5.kb0hgx92w3aeqvs.rcgu.o: in function `ZN15crossbeam_epoch7default11with_handle17h689cff926aee7e0fE':
          D:/media-autobuild_suite/msys64/opt/cargo\registry\src\github.com-1285ae84e5963aae\crossbeam-epoch-0.9.9\src/internal.rs:(.text+0x30b): undefined reference to `_imp___ZN15crossbeam_epoch7default9COLLECTOR17hd32c8f6291a208e5E'
          collect2.exe: error: ld returned 1 exit status
          
  = help: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname)

warning: `rav1e` (bin "rav1e") generated 1 warning
error: could not compile `rav1e` due to previous error; 1 warning emitted
```

Possibly related to https://github.com/crossbeam-rs/crossbeam/commit/08e41e184832f9d0bd49d97e21be1d60c437a894, unsure how I can pin a dependency of a dependency since this seems to be pulled by stuff like rayon and such, so I do not know how I can test it. Reporting here just in case

**To Reproduce**
Steps to reproduce the behavior

get Windows with msys2, might happen with linux, haven't tested, git clone rav1e, and `CC=clang cargo build --release --target=i686-pc-windows-gnu `

**Expected behavior**

Successful compile

**Required Information**
Toolchain (if is a build problem):
```
cddeg@ccom-laptop /trunk
$ cargo --version
cargo 1.61.0 (a028ae42f 2022-04-29)
cddeg@ccom-laptop /trunk
[...]

PR #2963, opened 2022-07-14
Add a release-nolto profile

lu-zero:
Workaround for #2960, because of rust-lang/rust#98302

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -166,6 +166,11 @@ debug = true
 incremental = true
 lto = "thin"
 
+# windows-gnu should use it until rust-lang/rust#98302 is not fixed
+[profile.release-no-lto]
+inherits = "release"
+lto = "off"
+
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -166,6 +166,11 @@ debug = true
 incremental = true
 lto = "thin"
 
+# windows-gnu should use it until rust-lang/rust#98302 is not fixed
+[profile.release-nolto]
```
Luni-4: `[profile.release-no-lto]`

Luni-4 (changes requested):
Just a simple change, looks fine to me, thank you!

Luni-4 (approved):

Cargo.toml
lu-zero: Sure

shssoichiro (approved):

PR #2965, opened 2022-07-18
Optimize and add test for write_uleb128 function

shssoichiro:
Reduces the number of loops needed to encode the uleb128 bytes. Small global performance improvements (less than 1%).

I had to write this function for another crate anyway, so I figured it was worth porting here.


lu-zero (approved):
Looks fine

PR #2966, opened 2022-07-22
Use external crate for grain table generation and parsing

shssoichiro:
In order to facilitate sharing of the grain table parsing code between crates in the AV1-Rust community, the code has been moved to a new crate called `av1-grain`.


lu-zero (approved):
Sure :)

PR #2967, opened 2022-07-25
Optimize `BlockSize::from_width_and_height_opt`

redzic:
This should slightly reduce the overhead of calling ASM functions, since many of them are dispatched via block size and thus `BlockSize::from_width_and_height_opt` is called.

Comparison of generated assembly: https://godbolt.org/z/ja5c4zr6K

### DIFF
--- a/src/partition.rs
+++ b/src/partition.rs
@@ -10,6 +10,8 @@
 #![allow(non_camel_case_types)]
 #![allow(dead_code)]
 
+use std::num::NonZeroUsize;
+
 use crate::serialize::{Deserialize, Serialize};
 use thiserror::Error;
 
@@ -149,7 +151,7 @@ pub enum BlockSize {
   BLOCK_64X16,
 }
 
-#[derive(Debug, Error, Copy, Clone)]
+#[derive(Debug, Error, Copy, Clone, Eq, PartialEq)]
 pub struct InvalidBlockSize;
 
 impl Display for InvalidBlockSize {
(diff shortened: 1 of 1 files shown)


shssoichiro:
I love the direction of this, but I think it can be cleaned up to avoid writing our own `tzcnt` function. Rust has `NonZero` types in the stdlib which we can use to perform the check `tzcnt` is doing.

```rust
use std::num::NonZeroUsize;

pub fn from_wh(w: usize, h: usize) -> Option<BlockSize> {
    use BlockSize::*;

    static BLOCK_LUT: [[Option<BlockSize>; 6]; 6] = [/* sizes */];

    if (4..=128).contains(&w) && (4..=128).contains(&h) {
        // SAFETY: w/h are always nonzero due to previous range checks.
        unsafe {
            let w_zeroes = NonZeroUsize::new_unchecked(w).trailing_zeros();
            let h_zeroes = NonZeroUsize::new_unchecked(w).trailing_zeros();
            if 1 << w_zeroes == w && 1 << h_zeroes == h {
                let x = w_zeroes as usize - 2;
                let y = h_zeroes as usize - 2;

                return *BLOCK_LUT.get_unchecked(x).get_unchecked(y);
            }
        }
    }

    None
}
```

This actually results in even shorter assembly (bottom = NonZero, top = custom tzcnt): https://godbolt.org/z/GxWEdP3xv

redzic:
> Rust has NonZero types in the stdlib which we can use to perform the check tzcnt is doing.

Oh yeah, I didn't think of using that. That definitely results in much cleaner code than manually writing our own function, thanks!

```rust
            let w_zeroes = NonZeroUsize::new_unchecked(w).trailing_zeros();
            let h_zeroes = NonZeroUsize::new_unchecked(w).trailing_zeros();
```

However it looks like there's a small typo in the code you posted which caused the shorter assembly: `h_zeroes` is getting the `trailing_zeroes` of `w`, not `h`. This results in the compiler deleting the second tzcnt, since it was already computed, and thus you get the shorter assembly. When fixed, this is the assembly that is generated (which is actually 1 instruction longer than the original code): https://godbolt.org/z/7ndE45zxv 

After the fix, the one instruction difference seems to caused by computing both trailing zeroes, and then doing the shifts afterwards. Although in reality this is probably an extremely minor difference, I think doing it the original way (compute tzcnt one at a time) is probably ever so slightly better, since it exits earlier if `w` is a power of 2, and `h` isn't. `llvm-mca` also seems to agree that the early-exit version is better; it estimated 700 vs 1137 cycles in 100 iterations (you can click on "Add tool" on compiler explorer and select "llvm-mca" to view the analysis).

redzic:
Here's an updated link with the code in this PR and the code @shssoichiro posted: https://godbolt.org/z/Y99aY4zq8

shssoichiro:
Thanks, in the end the difference between one or two instructions is fine, I'm good with the way you've done it, and hugely appreciate you adding tests for it and catching my mistake. 🙂

shssoichiro:
P.S. This prompts me to wonder if there are other places in the code base we could take advantage of `NonZero*` for some free compiler optimizations 🤔

shssoichiro (approved):

PR #2968, opened 2022-07-26
Add T.35 metadata support.

tdaede:
This currently only adds a Rust API. Note that this is a breaking
change as it extends the publicly visible "params" struct.

The T.35 metadata support in AV1 is very flexible - this is
intended in particular for T.35 metadata attached to frames,
in the style of the AV1 HDR10+ metadata specification:

https://aomediacodec.github.io/av1-hdr10plus/

### DIFF
--- a/src/api/context.rs
+++ b/src/api/context.rs
@@ -88,6 +88,7 @@ impl<T: Pixel> Context<T> {
   /// let info = FrameParameters {
   ///   frame_type_override: FrameTypeOverride::Key,
   ///   opaque: None,
+  ///   ..Default::default()
   /// };
   ///
   /// // Send the plain frame data
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -10,7 +10,9 @@
 
 use crate::activity::ActivityMask;
 use crate::api::lookahead::*;
-use crate::api::{EncoderConfig, EncoderStatus, FrameType, Opaque, Packet};
+use crate::api::{
+  EncoderConfig, EncoderStatus, FrameType, Opaque, Packet, T35,
+};
 use crate::color::ChromaSampling::Cs400;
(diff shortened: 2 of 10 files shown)


shssoichiro:
Does this extend to DV support? Asking for a friend. This is amazing btw.

src/api/context.rs
```diff
@@ -88,6 +88,7 @@ impl<T: Pixel> Context<T> {
   /// let info = FrameParameters {
   ///   frame_type_override: FrameTypeOverride::Key,
   ///   opaque: None,
+  ///   t35_metadata: Box::new([]),
```
lu-zero, line 91: Better to add default() to FrameParameters.

lu-zero (approved):
I'd like to have `..Default::default()` instead of having to explicitly set `Box::new([])`

tdaede:
It probably works for DV, it is not something I have looked at though.

src/api/context.rs
tdaede: Updated.

I also considered making t35_metadata an option, though it's a bit redundant with an empty list. I think technically the Box::new([]) still needs a malloc, though.

redzic: Rust designed the containers in its standard library to not allocate memory until data is actually pushed into them. This mean that `Vec::new` and `Box::new([])` don't allocate and thus don't call malloc (or any other functions for that matter, with optimizations turned on).

See this for example: https://godbolt.org/z/Go5GbYvnW

tdaede: Cool. I remember some big debate about that a while back wrt custom allocators but I guess it's not relevant here.

tdaede:
If we're going to do a major release soon where we are ok with API breakage then let's land this.

lu-zero:
Let's land it :)

PR #2972, opened 2022-08-01
Add aarch64 NEON 8-bit SATD assembly for 4x4-transformed block sizes

redzic:
This PR only adds assembly for block sizes that use a 4x4 transform for now.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -198,6 +198,7 @@ fn build_asm_files() {
     "src/arm/64/ipred.S",
     "src/arm/64/ipred16.S",
     "src/arm/64/sad.S",
+    "src/arm/64/satd.S",
     "src/arm/tables.S",
   ];
 
--- a/src/arm/64/satd.S
+++ b/src/arm/64/satd.S
@@ -1,0 +1,579 @@
+; Copyright (c) 2022, The rav1e contributors. All rights reserved
+;
+; This source code is subject to the terms of the BSD 2 Clause License and
+; the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+; was not distributed with this source code in the LICENSE file, you can
+; obtain it at www.aomedia.org/license/software. If the Alliance for Open
+; Media Patent License 1.0 was not distributed with this source code in the
+; PATENTS file, you can obtain it at www.aomedia.org/license/patent.
+
+#include "src/arm/asm.S"
+#include "util.S"
+
+.macro butterfly r0, r1, r2, r3
+    add.8h  \r0, \r2, \r3
+    sub.8h  \r1, \r2, \r3
+.endm
+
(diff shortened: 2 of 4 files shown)


redzic:
I only have access to my M1 for the time being, so my other PR for adding avx2 hbd satd assembly will be a bit delayed

vibhoothi (commented):
Nice, would be nice to see any numbers showing the impact of the code^^

redzic:
I'll try to do actual benchmarking later, but some initial testing suggests around a ~2-3% overall reduction in encoding time (tested on speed 2, 4 tiles, 4:2:0 1080p on an M1 pro). This only affects speed <= 2 in practice, I guess because faster presets don't search small block sizes.

Hopefully once I add assembly for 8x8 transform-based blocks in another PR, the speedup will be higher.

lu-zero (approved):
I guess it could land as-is or you'd rather wait?

barrbrain (changes requested):
This doesn't compile with my aarch64 clang unless I change the comment format.
```
Alpine clang version 13.0.1
Target: aarch64-alpine-linux-musl
Thread model: posix
InstalledDir: /usr/bin
```
It doesn't compile at all with gcc even after fixing the comment format.
```
localhost:~/rav1e$ gcc --version                                                gcc (Alpine 11.2.1_git20220219) 11.2.1 20220219
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
The test fails on my device:
```
failures:

---- dist::test::get_satd_same_u8 stdout ----
thread 'dist::test::get_satd_same_u8' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
<1408
>1662

', src/dist.rs:507:7
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    dist::test::get_satd_same_u8

test result: FAILED. 219 passed; 1 failed; 24 ignored; 0 measured; 0 filtered out; finished in 8.33s
```

redzic:
The tests were failing because I used src_stride for dst in some places, and the tests actually tested when the 2 planes had different strides. Fixed now.

@barrbrain do you know what would be required to fix compilation with gcc? and is there a way to test gcc compilation on macos?

barrbrain:
> @barrbrain do you know what would be required to fix compilation with gcc? and is there a way to test gcc compilation on macos?

I found that I had to set `CFLAGS=-m64` with GCC as the default compiler in my environment.

barrbrain (approved):
This passes tests for me now with `CC=clang`. I will look into GCC issues in my environment.

PR #2975, opened 2022-08-02
Stabilize the --film-grain-table parameter

shssoichiro:
The grain table spec does not require denoising, and is consistent across all three major AV1 encoders. I believe it makes sense to stabilize this while the grain synth denoising is being worked out. This also keeps the old name `--photon-noise-table` for compatibility, but since grain tables are not specific to photon noise, the canonical name is changed to `--film-grain-table`.

Using a grain table already requires being a somewhat advanced user, given that it requires an external tool to generate one (or writing one by hand), so there is less risk of beginners accidentally stumbling upon it and having it not do what they expect.

### DIFF
--- a/src/api/config/encoder.rs
+++ b/src/api/config/encoder.rs
@@ -10,7 +10,6 @@
 use itertools::*;
 
 use crate::api::color::*;
-#[cfg(feature = "unstable")]
 use crate::api::config::GrainTableSegment;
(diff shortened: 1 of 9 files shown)


barrbrain (approved):
Looks good to me, particularly with the `GrainTableSegment` structure provided by the `av1_grain` crate.

PR #2979, opened 2022-08-04
Add enhanced version info

shssoichiro:
This readds the git commit hash to git builds, readds the profile name, and adds a "-MODIFIED" tag to builds made with a dirty git checkout.

This also introduces a long version text, accessible by typing out `--version`, which provides additional info on the compiled binary.

Example short info for release version:
```
rav1e 0.5.0 (Release)
```

Example short info for git version:
```
rav1e p20220802-4-ge9371bb (Release)
```

Example long info for git version:
```
rav1e p20220802-4-ge9371bb (Release)
rustc 1.64.0-nightly (62b272d25 2022-07-21) x86_64-unknown-linux-gnu
Optimizations: adx,aes,avx,avx2,bmi1,bmi2,cmpxchg16b,f16c,fma,fxsr,llvm14-builtins-abi,lzcnt,movbe,pclmulqdq,popcnt,rdrand,rdseed,sha,sse,sse2,sse3,sse4.1,sse4.2,sse4a,ssse3,xsave,xsavec,xsaveopt,xsaves
Assembly: Enabled
Threading: Enabled
Unstable Features: Disabled
Compiler Flags: -Ctarget-cpu=native
Built Thu, 04 Aug 2022 18:17:24 +0000
```


lu-zero (approved):
Looks good!

src/bin/common.rs
```diff
@@ -245,6 +257,43 @@ pub struct CliOptions {
...
+      if cfg!(debug_assertions) { "Debug" } else { "Release" }
+    )
+  });
+  &VERSION_STR
+}
+
+fn get_long_version() -> &'static str {
+  static LONG_VERSION_STR: Lazy<String> = Lazy::new(|| {
+    format!(
+      "{}\n{} {}\nOptimizations: {}\nAssembly: {}\nThreading: {}\nUnstable Features: {}\nCompiler Flags: {}\nBuilt {}",
```
redzic, line 282: I think it might be more accurate to say "CPU features" or something like that instead of "Optimizations"

redzic, line 282: Maybe it could also say "addition compiler flags"? If I just do `cargo build --release` it ends up with just an empty line after the "Compiler Flags:", which I think looks slightly strange. Maybe even something like `(none)` could be printed if there were no extra compiler flags, i.e. if the string was empty.

shssoichiro: Good call, I didn't even think to test that case 😅

PR #2981, opened 2022-08-09
Use frame-relative spatiotemporal scores

barrbrain:
This should enhance the effectiveness of dynamic segmentation by normalizing the scales.

AWCY results for [objective-fast-1 at default speed](https://beta.arewecompressedyet.com/?job=master-0bee40b875b694fe5cb85d65a309dbdd9edaa423&job=rate-control-frame-mean-rebased%402022-08-09T17%3A02%3A49.994Z)
| PSNR Y | PSNR Cb | PSNR Cr | CIEDE2000 |    SSIM | MS-SSIM | PSNR-HVS Y | PSNR-HVS Cb | PSNR-HVS Cr | PSNR-HVS |   VMAF | VMAF-NEG |
|   ---: |    ---: |    ---: |      ---: |    ---: |    ---: |       ---: |        ---: |        ---: |     ---: |   ---: |     ---: |
| 2.1192 |  0.0424 | -0.4626 |    0.2634 | -1.9722 | -1.9259 |     0.6551 |     -1.4601 |     -1.9457 |   0.6588 | 1.2593 |   1.3857 |

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -717,6 +717,34 @@ impl<T: Pixel> CodedFrameData<T> {
       imp_b_segments: Default::default(),
     }
   }
+
+  // Assumes that we have already computed activity scales and distortion scales
+  pub fn compute_spatiotemporal_scores(&mut self) -> u32 {
+    let mut scores = self
+      .distortion_scales
+      .iter()
+      .zip(self.activity_scales.iter())
+      .map(|(&d, &a)| d * a)
+      .collect::<Box<_>>();
+
+    let inv_mean = DistortionScale::inv_mean(&scores);
+
+    for score in scores.iter_mut() {
+      *score *= inv_mean;
+    }
+
+    let segments =
+      scores.iter().map(|&s| segment_idx_from_distortion(s)).collect();
+
+    for scale in self.distortion_scales.iter_mut() {
+      *scale *= inv_mean;
+    }
+
+    self.spatiotemporal_scores = scores;
+    self.imp_b_segments = segments;
+
+    inv_mean.0
+  }
 }
 
 pub(crate) const fn pos_to_lvl(pos: u64, pyramid_depth: u64) -> u64 {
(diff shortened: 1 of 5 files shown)


shssoichiro:
Overall I do think this improves perceptual quality. There's a bit more blur on things like grain or details in complex areas (food market, ducks) where you wouldn't notice without zooming in, but details on otherwise flat areas are considerably improved without zooming (e.g. aspen).

src/encoder.rs
```diff
@@ -717,6 +717,34 @@ impl<T: Pixel> CodedFrameData<T> {
...
+      .distortion_scales
+      .iter()
+      .zip(self.activity_scales.iter())
+      .map(|(&d, &a)| d * a)
+      .collect::<Box<_>>();
+
+    let inv_mean = DistortionScale::inv_mean(&scores);
+
+    for score in scores.iter_mut() {
+      *score = *score * inv_mean;
```
shssoichiro, line 733: Could use `score *= inv_mean` (and same for `scale` below), I actually thought there was a clippy lint for this.

src/rdo.rs
```diff
@@ -585,6 +585,15 @@ impl DistortionScale {
     )
   }
 
+  pub fn inv_mean(slice: &[Self]) -> Self {
+    let sum: u64 = slice.iter().map(|&s| s.0 as u64).sum();
+    assert_ne!(sum, 0);
```
shssoichiro, line 590: I wonder if there's any benefit from using `NonZeroU64` here. Maybe it makes the code look cleaner? I'm not sure if the compiler benefits from it or if it considers `assert_ne!(sum, 0)` equivalent for its optimization routine.

shssoichiro (changes requested):

src/encoder.rs
barrbrain: Right, I'll need to include another `impl` or possibly `Derive` it.

src/rdo.rs
barrbrain: I think there is an argument for changing the inner type of `DistortionScale` to `NonZeroU32`.

src/encoder.rs
shssoichiro: Ah, I see because it's adding DistortionScales. That makes sense. If you don't want to do that as part of this changeset then, that's fine.

src/rdo.rs
barrbrain: I opted to simply ensure that the outputs of the new operations are always in range.

shssoichiro (approved):

PR #2982, opened 2022-08-11
Fix aarch64 satd compilation on gcc

redzic:
gcc doesn't support arrangement specifiers after the instruction mnemonic like clang does, so they instead need to be specified after each register.


barrbrain (approved):
Confirmed with `gcc (Alpine 11.2.1_git20220219) 11.2.1 20220219` on aarch64 Linux.

PR #2984, opened 2022-08-12
Clean up encode_partition_topdown()

redzic:
Removes some unnecessary copies and usages of ArrayVec.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -2633,7 +2633,7 @@ fn encode_partition_bottomup<T: Pixel, W: Writer>(
   }
 
   let is_square = bsize.is_sqr();
-  let hbs = bsize.width_mi() >> 1;
+  let hbs = bsize.width_mi() / 2;
   let has_cols = tile_bo.0.x + hbs < ts.mi_width;
   let has_rows = tile_bo.0.y + hbs < ts.mi_height;
   let is_straddle_x = tile_bo.0.x + bsize.width_mi() > ts.mi_width;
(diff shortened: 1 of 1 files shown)


redzic:
looks like CI is failing due to unrelated network issues

src/encoder.rs
```diff
@@ -2933,22 +2933,22 @@ fn encode_partition_topdown<T: Pixel, W: Writer>(
       (bsize > fi.partition_range.min && is_square) || must_split
     };
 
-  let mut rdo_output =
-    block_output.clone().unwrap_or(PartitionGroupParameters {
+  let mut rdo_output = if let Some(block_output) = block_output {
+    block_output.clone()
+  } else {
+    PartitionGroupParameters {
```
barrbrain, line 2939: How does using `unwrap_or_else()` compare? It should be equivalent after optimisation but more concise.
```suggestion
  let mut rdo_output =
    block_output.clone().unwrap_or_else(|| PartitionGroupParameters {
```

src/encoder.rs
```diff
@@ -3156,7 +3156,7 @@ fn encode_partition_topdown<T: Pixel, W: Writer>(
             &Some(PartitionGroupParameters {
               rd_cost: mode.rd_cost,
               part_type: PartitionType::PARTITION_NONE,
-              part_modes: once(mode).collect(),
+              part_modes: ArrayVec::try_from(&[mode] as &[_]).unwrap(),
```
barrbrain, line 3159: Is the following equivalent?
```suggestion
              part_modes: [mode][..].try_into().unwrap(),
```

barrbrain (commented):
Thank you for cleaning out the cruft. :)

src/encoder.rs
redzic: It should be, since `try_into` calls `try_from` anyway.

src/encoder.rs
redzic: It does seem equivalent after optimization. For some reason I thought clone and then unwrap_or_else would be worse then cloning inside an if-let block, but I guess not.

src/encoder.rs
```diff
@@ -2934,21 +2934,19 @@ fn encode_partition_topdown<T: Pixel, W: Writer>(
...
 
   let mut rdo_output =
-    block_output.clone().unwrap_or(PartitionGroupParameters {
+    block_output.clone().unwrap_or_else(|| PartitionGroupParameters {
       part_type: PartitionType::PARTITION_INVALID,
       rd_cost: std::f64::MAX,
       part_modes: ArrayVec::new(),
-    });
+    }
+  };
```
barrbrain, line 2942: Oops, my suggestion didn't include reverting this hunk.
```suggestion
  });
```

barrbrain (approved):

PR #2989, opened 2022-08-13
Use fixed-sized array in luma/chroma pred mode counts

redzic:
According to #2236, this was converted to an ArrayVec since at that time Rust did not support deriving `Debug` for fixed-sized arrays above 32 elements, but this is no longer the case and thus ArrayVec is not needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,13 @@ check_asm = []
 capi = []
 tracing = ["rust_hawktracer/profiling_enabled"]
 scenechange = []
-serialize = ["serde", "toml", "v_frame/serialize", "arrayvec/serde"]
+serialize = [
+    "serde",
+    "toml",
+    "v_frame/serialize",
+    "arrayvec/serde",
+    "serde-big-array",
+]
 wasm = ["wasm-bindgen"]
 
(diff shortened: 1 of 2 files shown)


redzic:
Hmm, it builds locally for me, except when using `--features=serialize`. I'll look into it further

barrbrain:
We might need to supplement with https://github.com/est31/serde-big-array to cover `--features=serialize`.

Cargo.toml
```diff
@@ -50,7 +50,13 @@ check_asm = []
 capi = []
 tracing = ["rust_hawktracer/profiling_enabled"]
 scenechange = []
-serialize = ["serde", "toml", "v_frame/serialize", "arrayvec/serde"]
+serialize = [
+    "serde",
+    "toml",
+    "v_frame/serialize",
+    "arrayvec/serde",
```
barrbrain, line 57: Is this crate still in use?

barrbrain (approved):
Looks good but we might be able to drop `arrayvec/serde`.

Cargo.toml
redzic: Oh good catch, yeah it's not in use anymore.

PR #2990, opened 2022-08-16
Add T.35 metadata C API.

tdaede:
This is a copy-based API, assuming that the T.35 metadata will
be small. In fact, it does one extra copy internally that could
probably be eliminated with some work.

It allows for multiple T.35 metadata messages to be added to a
frame, but no method to inspect or remove them (for references
or reconstructed frames, for example).

### DIFF
--- a/src/capi.rs
+++ b/src/capi.rs
@@ -25,6 +25,7 @@ use std::sync::Arc;
 
 use std::ffi::CStr;
 use std::ffi::CString;
+use std::mem;
 use std::os::raw::c_char;
 use std::os::raw::c_int;
 use std::os::raw::c_void;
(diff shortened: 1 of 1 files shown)


src/capi.rs
```diff
@@ -1104,9 +1127,15 @@ pub unsafe extern fn rav1e_send_frame(
     (*frame).opaque.take().map(rav1e::Opaque::new)
   };
 
+  let t35_metadata = if frame.is_null() {
+    Box::new([])
+  } else {
+    (*frame).t35_metadata.clone().into_boxed_slice()
```
lu-zero, line 1133: alternatively [mem::take](https://doc.rust-lang.org/std/mem/fn.take.html) might be of use?

tdaede: seems OK as we hide the frame contents from the c api and there's no way to read back a frame after it's been sent, updated to use it

lu-zero (approved):

PR #2993, opened 2022-08-19
Compute segmentation map thresholds from the offsets

barrbrain:
AWCY results for [objective-1-fast at default speed](https://beta.arewecompressedyet.com/?job=master-3c05349095b4cdf1887c03f615c904799abe7ea3&job=segment-map%402022-08-19T06%3A01%3A31.082Z):

|  PSNR Y | PSNR Cb | PSNR Cr | CIEDE2000 |   SSIM | MS-SSIM | PSNR-HVS Y | PSNR-HVS Cb | PSNR-HVS Cr | PSNR-HVS |    VMAF | VMAF-NEG |
|    ---: |    ---: |    ---: |      ---: |   ---: |    ---: |       ---: |        ---: |        ---: |     ---: |    ---: |     ---: |
| -1.0439 | -1.2423 | -1.2902 |   -0.3670 | 0.9061 |  0.8534 |    -0.4362 |     -0.9269 |     -1.0861 |  -0.4454 | -0.6953 |  -0.7345 |

Includes reworking of `DistortionScale` construction so that constants may be defined from rational numbers.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -570,9 +570,26 @@ pub struct SegmentationState {
   pub last_active_segid: u8,
   pub features: [[bool; SegLvl::SEG_LVL_MAX as usize]; 8],
   pub data: [[i16; SegLvl::SEG_LVL_MAX as usize]; 8],
-  pub segment_map: [usize; 8],
+  pub threshold: [DistortionScale; 7],
   pub min_segment: u8,
   pub max_segment: u8,
+}
+
+impl SegmentationState {
+  pub fn update_threshold(&mut self, base_q_idx: u8, bd: usize) {
+    let base_ac_q = ac_q(base_q_idx, 0, bd) as u64;
+    let real_ac_q = ArrayVec::<_, MAX_SEGMENTS>::from_iter(
+      self.data[..=self.max_segment as usize].iter().map(|data| {
+        ac_q(base_q_idx, data[SegLvl::SEG_LVL_ALT_Q as usize] as i8, bd) as u64
+      }),
+    );
+    self.threshold.fill(DistortionScale(0));
+    for ((q1, q2), threshold) in
+      real_ac_q.iter().skip(1).zip(&real_ac_q).zip(&mut self.threshold)
+    {
+      *threshold = DistortionScale::new(base_ac_q.pow(2), q1 * q2);
+    }
+  }
 }
 
 // Frame Invariants are invariant inside a frame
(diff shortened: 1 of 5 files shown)


src/encoder.rs
```diff
@@ -570,11 +570,28 @@ pub struct SegmentationState {
...
   pub min_segment: u8,
   pub max_segment: u8,
 }
 
+impl SegmentationState {
+  pub fn update_threshold(&mut self, base_q_idx: u8, bit_depth: usize) {
+    use itertools::Itertools;
+    let base_ac_q = ac_q(base_q_idx, 0, bit_depth) as u64;
+    let threshold = (0..=self.max_segment as usize)
+      .map(|i| self.data[i][SegLvl::SEG_LVL_ALT_Q as usize] as i8)
```
lu-zero, line 583: maybe the two maps could be collapsed in one and you could iterate directly over a slice of data.

src/encoder.rs
```diff
@@ -570,11 +570,28 @@ pub struct SegmentationState {
...
+      .map(|delta_q| ac_q(base_q_idx, delta_q, bit_depth) as u64)
+      .collect::<ArrayVec<_, MAX_SEGMENTS>>()
```
barrbrain, line 585: Is this close to what you were thinking of?
```suggestion
    let real_ac_q = (0..=self.max_segment as usize)
      .map(|i| {
        let delta_q = self.data[i][SegLvl::SEG_LVL_ALT_Q as usize] as i8;
        ac_q(base_q_idx, delta_q, bit_depth) as u64
      })
      .collect::<ArrayVec<_, MAX_SEGMENTS>>();
    let threshold = real_ac_q
```

lu-zero: you can also feed the range directly to data, instead of using the range to iterate and then index.

src/rdo.rs
```diff
@@ -637,6 +637,14 @@ impl fmt::Debug for DistortionScale {
   }
 }
 
+impl From<f64> for DistortionScale {
+  #[inline]
+  fn from(scale: f64) -> Self {
+    let den = 1 << (Self::SHIFT + 1);
+    Self::new((scale * den as f64) as u64, den)
```
shssoichiro, line 644: Is this going to result in any rounding issues that we care about?

barrbrain: This is equivalent to the previous rounding behavior. However, there is an issue with overflow for extremely large values.

shssoichiro: What is the likelihood of any user encountering the overflow?

barrbrain: I'll fix it since u64::saturating_add is const

lu-zero (approved):
Looks fine to me.

Issue #886, opened 2019-01-26
Add rav1e to package managers


EwoutH:
Not directly related to the codebase, but it would be very useful to have rav1e available in the most used package managers. These for example:
 - [ ] [APT](https://wiki.debian.org/Apt) (Debian / Ubuntu)
 - [ ] [Homebrew](https://brew.sh/index_nl) (macOS)
 - [ ] [Chocolatey](https://chocolatey.org/) (Windows)

EwoutH:
@lu-zero Homebrew [requires](https://docs.brew.sh/Formula-Cookbook) at least one tagged version. Is there a release that can be tagged as 0.0.1 or something?

lu-zero:
I'm fine with the weekly tag approach, but I'd like to have everybody ack it.

lu-zero:
(once we have a tag we could also produce crav1e builds/packages)

lu-zero:
Here a [prerelease](https://github.com/xiph/rav1e/releases/tag/20190129) so we can see what is missing packaging-wise.

petterreinholdtsen:
I created https://bugs.debian.org/920842 asking for a Debian package.  Tagged releases will make it more likely. :)

jbeich:
Are you interested only in select few repositories or any? For example:
[![Packaging status](https://repology.org/badge/vertical-allrepos/rav1e.svg)](https://repology.org/metapackage/rav1e/versions)

EwoutH:
@jbeich The more the merrier :).

For CI the Ubuntu LTS versions (14.04, 16.04 and 18.04) and Debian Stable are the most important.

For general use Manjaro, Mint, MX Linux, elementary and Fedora would also be nice.

cuviper:
cc @ignatenkobrain for Fedora too.

vi:
Maybe at least publish pre-built executables for various linuxes to Github releases?

cromefire:
> Maybe at least publish pre-built executables for various linuxes to Github releases?

Having one for the latest Ubuntu LTS would be nice as it's pretty common

vi:
It it can be just a fully static musl build, like this:

https://vi-server.org/pub/rav1e_af9d4ab

    SHA256 (rav1e_af9d4ab) = 73c6f959ee38324608648a3e09b9cf6e04b02a2c359ddea84e20dc22f1e464fd

jbeich:
> https://vi-server.org/pub/rav1e_af9d4ab

[...]

PR #2995, opened 2022-08-20
vergen: Drop fork and upgrade to 6

barrbrain:
See https://github.com/xiph/rav1e/issues/886#issuecomment-1221317417


lu-zero:
We could just remove it and use the information from `built`, I think.

barrbrain:
I think this could be a minimal intermediate step toward replacement. It at least maintains the current behavior minus the local fork.

barrbrain:
Superseded by #2996.

PR #3000, opened 2022-08-25
Implement support for encoding HDR10+ from JSON metadata file

quietvoid:
The goal of this PR is to allow users to encode HDR video with HDR10+ metadata, in compliance with the [AV1 HDR10+ specification](https://github.com/AOMediaCodec/av1-hdr10plus).

In the current state, the changes make it possible to encode HDR10+ for both CLI and Rust API usage.

From CLI, the metadata is expected to be passed as a JSON metadata file, which follows the same format as encoders like `x265`.
The metadata parsing/encoding is done through the [hdr10plus](https://github.com/quietvoid/hdr10plus_tool/tree/main/hdr10plus) crate.
The library is probably less than ideal but it should be fine to parse things at init.
The CLI opt can be either `--hdr10plus-json` or `--dhdr10-info` (used by `x265`)

From Rust, the metadata must be encoded into the final payload bytes and provided for the frames that require it.

The code also avoids duplicating `ST2094-40` if some is already present in the `FrameParameters.t35_metadata` array.
It requires looking at the payloads to find out.

Output bitstream validated with the [AOM AV1 HDR10+ Validator](https://gpac.github.io/ComplianceWarden-wasm/av1hdr10plus.html).

Starting this as a draft, as some things need to be improved;
- Behaviour when some parsed metadata is invalid.
    - Either the encoder has to abort, or keeps going with an incomplete map of frame<->metadata.
- Need to make the `ST2094-40` prefix const.
- Improve comments.
- ... probably more, feedback welcome.


src/api/config/encoder.rs
```diff
@@ -15,6 +15,9 @@ use crate::api::{Rational, SpeedSettings};
 use crate::encoder::Tune;
 use crate::serialize::{Deserialize, Serialize};
 
+#[cfg(feature = "unstable")]
```
lu-zero, line 18: I think you can avoid adding the `unstable` while refining the patch.

quietvoid: Removed the `unstable` conditions.

quietvoid:
As is, this is probably ready for review.
I'm just not sure what how the CLI encoder should behave when the JSON metadata partially fails to parse.

Also, hopefully it also works in `rav1e-ch`, as long as the frame numbers are continuous across GOPs.

quietvoid:
It does appear that the current T.35 insertion code skips over some frames with `show_frame=0` and `show_existing_frame=1`.
At least compared to the reference stream in https://github.com/AOMediaCodec/av1-hdr10plus/wiki, which has the metadata OBUs for every frame.

Though the spec does say
> `HDR10+ Metadata OBUs` are not provided when `show_frame`=0.

So I'm not sure if the ref sample is correct.
cc  @tdaede

quietvoid:
As the specification seems to be  in a final state, I've reworked things to work properly for the OBU placement.
https://aomediacodec.github.io/av1-hdr10plus/#placement

The T.35 metadata is added at the end when present in the `EncoderConfig` `hdr10plus_payloads`.
And only if the existing frame T.35 metadata doesn't already contain one for HDR10+ metadata.

src/encoder.rs
```diff
@@ -1286,6 +1286,26 @@ impl<T: Pixel> FrameInvariants<T> {
...
+    self
+      .config
+      .hdr10plus_payloads
+      .as_ref()
+      .and_then(|payloads| payloads.get(&self.input_frameno))
+      .and_then(|payload| {
+        Some(T35 {
+          country_code: 0xB5,
+          country_code_extension_byte: 0x00,
+          data: payload.clone().into_boxed_slice(),
```
quietvoid, line 1305: This clone isn't ideal but I'm not sure if it would be better to store `T35` directly in the config, to avoid it.

quietvoid:
Doh. I was running clippy with `--all-features` and it was failing.

quietvoid:
It was pointed out on the IRC that the payloads should probably be handled on the CLI side only, as the API `EncoderConfig` already has T.35 metadata support.
However the current implementation has issues with missing metadata that I've yet to resolve, so I'll update this when the issues are worked out.

tricktron:
> However the current T.35 API implementation has issues with missing metadata that I've yet to resolve, so I'll update this when the issues are worked out.

@quietvoid  Could you resolve these issues?

quietvoid:
You can look at the implementation of #3250 to see how it should be done.

superyu1337:
Any news on this and #3250 ?

PR #3002, opened 2022-08-26
Eliminate floating-point paths from `segmentation_optimize()`

barrbrain:
Extract binary logarithm and exponentiation into `util::logexp` module:
* logexp: Add 64-bit binary exponentiation and logarithm in Q57.
* logexp: Add tests for `bexp64` and `blog64`
* logexp: Add conversions between 64-bit Q57 and 32-bit Q24.
* logexp: Add binary exponential in Q24 format, saturated to 47 bits.
* logexp: Add assertion for q57 input range
* rate: Pivot to `util::logexp` functions

Extend the `util::logexp` module and use it to improve `segmentation_optimize()`:
* logexp: Rewrite `blog64` so that it is `const`
* logexp: Add a fixed-point binary log in Q24 for u32 to i32
* segmentation: Avoid reallocation in `collect::<Box<[_]>>()`


src/util/logexp.rs
```diff
@@ -0,0 +1,300 @@
...
+  } else {
+    w = 1i64 << 62;
+  }
+  if ipart < 62 {
+    w = ((w >> (61 - ipart)) + 1) >> 1;
+  }
+  w
+}
+
+// Computes the binary log of w.
```
lu-zero, line 130: w?

lu-zero (approved):

PR #3005, opened 2022-08-30
Remove internal ivf crate and use av-ivf instead

shssoichiro:


shssoichiro:
This feels kind of... not really cleaner... Please provide feedback.

lu-zero:
It it paves the way to use the matroska crate (that we should publish) later I think it is a step in the right direction.

shssoichiro:
Based on feedback in IRC I think I'm going to close this and perhaps look at a different approach if we need it. I agree with the comments on IRC that there's too much overhead with the av-ivf crate.

PR #3010, opened 2022-09-03
Use sad_plane to avoid falling back to slow Rust SAD code

shssoichiro:
This gives approximately a 2% speed improvement at speed 2
on 1080p content. Currently only benefits 8-bit.

### DIFF
--- a/src/asm/x86/dist/mod.rs
+++ b/src/asm/x86/dist/mod.rs
@@ -13,6 +13,7 @@ pub use self::sse::*;
 use crate::cpu_features::CpuFeatureLevel;
 use crate::dist::*;
 use crate::partition::BlockSize;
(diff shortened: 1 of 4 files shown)


redzic:
Hmm, the tests pass for me locally. I wonder if the segfault is somehow due to the fact that sad_plane uses aligned loads for everything? But then again I wonder why it would only segfault on macos if that is the case.

lu-zero:
It needs a rebase.

lu-zero:
The macos failing test makes me wonder if there isn't a valgrind.

shssoichiro:
Well, the segfault does seem deterministic. It is indeed weird that it's only on Mac. I don't particularly have a Mac to run gdb on...

PR #3012, opened 2022-09-05
Bump several dependencies

shssoichiro:
Including `image`, `pretty_assertions`, and `av1-grain`. Not touching `arbitrary` and `libfuzzer-sys` in this patch because of breaking changes I don't feel like dealing with right now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,11 +103,11 @@ const_fn_assert = "0.1.2"
 nom = { version = "7.0.0", optional = true }
 new_debug_unreachable = "1.0.4"
 once_cell = "1.13.0"
-av1-grain = { version = "0.1.1", features = ["serialize"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -103,11 +103,11 @@ const_fn_assert = "0.1.2"
 nom = { version = "7.0.0", optional = true }
 new_debug_unreachable = "1.0.4"
 once_cell = "1.13.0"
-av1-grain = { version = "0.1.1", features = ["serialize"] }
+av1-grain = { version = "0.2.0", features = ["serialize"] }
```
barrbrain, line 106: It seems that additional changes are required to bump av1-grain.

shssoichiro: Should be fixed now. I always forget all the myriad of commands that the CI runs so I just let it run them all for me. :sweat_smile:

redzic (approved):

PR #3013, opened 2022-09-05
Calculate per-plane scales in Q14 and center `log(base_q)`

barrbrain:
The move to `DistortionScale` reduces `compute_distortion()` CPU time by about 10%.

AWCY results on [objective-1-fast at default speed](https://beta.arewecompressedyet.com/?job=master-8c260d2cd928fe3945ae41fa738b8e994fa5c68e&job=segment-opt%402022-09-06T05%3A03%3A46.038Z):
| PSNR Y | PSNR Cb | PSNR Cr | CIEDE2000 |   SSIM | MS-SSIM | PSNR-HVS Y | PSNR-HVS Cb | PSNR-HVS Cr | PSNR-HVS |   VMAF | VMAF-NEG |
|   ---: |    ---: |    ---: |      ---: |   ---: |    ---: |       ---: |        ---: |        ---: |     ---: |   ---: |     ---: |
| 0.0902 | -0.7627 | -0.4605 |   -0.0608 | 0.0927 |  0.0987 |     0.1329 |     -0.8583 |     -0.4014 |   0.1068 | 0.1648 |   0.1460 |

### DIFF
--- a/src/asm/x86/dist/sse.rs
+++ b/src/asm/x86/dist/sse.rs
@@ -11,6 +11,7 @@ use crate::cpu_features::CpuFeatureLevel;
 use crate::dist::*;
 use crate::encoder::IMPORTANCE_BLOCK_SIZE;
 use crate::partition::BlockSize;
+use crate::rdo::DistortionScale;
 use crate::tiling::PlaneRegion;
 use crate::util::*;
 
@@ -111,13 +112,15 @@ pub fn get_weighted_sse<T: Pixel>(
     std::mem::size_of::<T>()
   }
 
+  let den =
+    DistortionScale::new(1, 1 << rust::GET_WEIGHTED_SSE_SHIFT).0 as u64;
   let dist = match (bsize_opt, T::type_enum()) {
     (Err(_), _) => call_rust(),
     (Ok(bsize), PixelType::U8) => {
       match SSE_FNS[cpu.as_index()][to_index(bsize)] {
         // SAFETY: Calls Assembly code.
         Some(func) => unsafe {
-          (func)(
+          ((func)(
             src.data_ptr() as *const _,
             T::to_asm_stride(src.plane_cfg.stride),
             dst.data_ptr() as *const _,
@@ -125,6 +128,8 @@ pub fn get_weighted_sse<T: Pixel>(
             scale.as_ptr(),
             (scale_stride * size_of_element(scale)) as isize,
           ) as u64
+            + (den >> 1))
+            / den
         },
         None => call_rust(),
       }
@@ -133,14 +138,15 @@ pub fn get_weighted_sse<T: Pixel>(
       match SSE_HBD_FNS[cpu.as_index()][to_index(bsize)] {
         // SAFETY: Calls Assembly code.
         Some(func) => unsafe {
-          (func)(
+          ((func)(
             src.data_ptr() as *const _,
             T::to_asm_stride(src.plane_cfg.stride) as isize,
             dst.data_ptr() as *const _,
             T::to_asm_stride(dst.plane_cfg.stride) as isize,
             scale.as_ptr(),
             (scale_stride * size_of_element(scale)) as isize,
-          )
+          ) + (den >> 1))
+            / den
         },
         None => call_rust(),
       }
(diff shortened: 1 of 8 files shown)


src/rdo.rs
```diff
@@ -596,11 +596,18 @@ impl DistortionScale {
 
   /// Binary logarithm in Q11
   #[inline]
-  pub const fn blog32(self) -> i16 {
+  pub const fn blog16(self) -> i16 {
```
tdaede, line 599: was this named incorrectly before?

tdaede: ah nevermind I see how this works

barrbrain: Yes, I neglected to rename it when changing the output precision and word size.

src/x86/sse.asm
```diff
@@ -24,8 +24,8 @@ SECTION .text
     ; Multiply and shift using scalar code
     mov             scaled, [scaleq]
     imul               rax, scaleq
-    add                rax, 0x800
```
tdaede, line 27: Maybe I'm blind but I'm missing the corresponding rust change.

tdaede: Also I don't understand why we would want to lose precision at this stage if we're not currently overflowing anything.

barrbrain: The rust code always calculates the round from `DIstortionScale::SHIFT` as `1 << Self::SHIFT >> 1`.
https://github.com/barrbrain/rav1e/blob/20eacc69/src/rdo.rs#L573-L579
https://github.com/barrbrain/rav1e/blob/20eacc69/src/rdo.rs#L626

barrbrain: I don't think we are losing precision here, the change in shift matches the change in scale of the input. We have enough headroom for the additional precision.
However, I can see the argument for accumulating before shifting. Neither the rust nor assembly code do that yet.

barrbrain: Ah, there is a TODO specifically for deferring rounding in the assembly:
https://github.com/xiph/rav1e/blob/8c260d2c/src/x86/sse.asm#L42-L46

barrbrain: I drafted a compromise that avoided introducing overflow: https://github.com/barrbrain/rav1e/commit/0a50b08
AWCY results [on top of commits so far](https://beta.arewecompressedyet.com/?job=fixed-point%402022-09-05T05%3A19%3A26.619Z&job=fixed-point%402022-09-05T14%3A44%3A04.983Z):
|  PSNR Y | PSNR Cb | PSNR Cr | CIEDE2000 |   e  | MS-SSIM | PSNR-HVS Y | PSNR-HVS Cb | PSNR-HVS Cr | PSNR-HVS |    VMAF | VMAF-NEG |
|    ---: |    ---: |    ---: |      ---: |   ---: |    ---: |       ---: |        ---: |        ---: |     ---: |    ---: |     ---: |
| -0.1024 | -0.5120 | -0.4848 |   -0.4979 | 0.1456 |  0.1729 |    -0.0325 |     -0.6291 |     -0.5239 |  -0.1197 | -0.1653 |  -0.1696 |

src/x86/sse.asm
```diff
@@ -129,6 +129,7 @@ cglobal weighted_sse_4x4, 6, 7, 5, \
 
     pshuflw             m0, m1, q3232
     paddd               m0, m1
+    xor                rax, rax
```
redzic, line 132: movd clears the upper 32 bits of the output register, so I believe this xor shouldn't be necessary. Or perhaps it's to avoid a false dependency on the destination register? if so `xor eax, eax` should be less in code size but still also clear the upper 32 bits anyway, and is apparently the recommended zeroing method (https://stackoverflow.com/questions/33666617/what-is-the-best-way-to-set-a-register-to-zero-in-x86-assembly-xor-mov-or-and/33668295#33668295)

src/asm/x86/dist/sse.rs
```diff
@@ -111,20 +111,24 @@ pub fn get_weighted_sse<T: Pixel>(
     std::mem::size_of::<T>()
   }
 
+  use crate::rdo::DistortionScale;
+  let den = (DistortionScale::default().0 >> 8) as u64;
```
lu-zero, line 115: maybe this could be a self-describing function.

lu-zero (approved):
Thank you :)

Issue #2759, opened 2021-07-09
[Meta] Parity with aomenc


shssoichiro:
I wanted to create a meta issue to track features or changes we can implement to reach quality parity with aomenc. Right now, our speed 6 still trails aomenc's cpu-used 6 by about 30% BD-rate, while also being slower (assuming one tile and no parallel encoding) ([AWCY](https://beta.arewecompressedyet.com/?job=aom-cpu-6%402021-07-04T08%3A32%3A14.124Z&job=master-f553646d70fba8e265d436103a73520eb7adec8c)).

We come closer if we up rav1e to s0 ([AWCY](https://beta.arewecompressedyet.com/?job=aom-cpu-6%402021-07-04T08%3A32%3A14.124Z&job=master-c17ce82f8-s0)), where some clips even win over aomenc, but at the cost of rav1e being 2700% slower.

There's also the notable outlier of dark720, which is 200% worse MSSSIM BD Rate even at speed 0.

Here are the ideas so far:

- Adjust the luma to chroma balance to favor more luma bitrate. It looks like our chroma numbers are consistently better by a large amount, but luma usually has a larger impact on both objective and perceptual quality.
- aomenc uses non-square blocks even at speed 6 and 4x4 blocks for <1080p. They must have some way of doing this search without impacting speed too much. It looks like part of it involves limiting which sizes of blocks are allowed to be broken down into rectangles, which changes per speed setting.
- TODO: Try to fill out this list with more suggestions. I'm not extremely knowledgeable on the aomenc code base, and it's massive, so I'm preferring to welcome discussion from people who may be more knowledgeable.

tmatth:
Refs #845

BlueSwordM:
IIRC, the biggest reason that rav1e is slower than aomenc is because aomenc does a massive amount of search space pruning at the higher speed preset, particularly when it comes to motion estimation.

You can see it in low motion clips vs high motion clips: aomenc and rav1e at speed 6 are similar in high motion clips in terms of speed and visual quality, but once a low motion scenes comes in, aomenc speeds up a lot more than rav1e. rav1e has no search pruning in any manner.

That pruning also applies to block size selection and transform size partitioning, especially with rectangular partitions: at speed 6, they restriction partition selection from 8x8-32x32 transforms.

[...]

PR #3016, opened 2022-09-08
Preliminary work on complete next-gen CDEF implementation (early WIP)

BlueSwordM:
Hello again. It has been a while since I've made any relevant PRs to rav1e, so here goes.

Ever since a faithful meeting we had back in 2021, I knew that rav1e didn't have a higher quality full CDEF implementation, so I've always wanted rav1e to get it.

Following this issue made me think of it again today, which is why I started the work on it:
https://github.com/xiph/rav1e/issues/2759

My plan to implement the full CDEF implementation will follow multiple steps in its design:

### 1. Full CDEF strength selection implementation

Adding the full CDEF implementation based on distortion optimization with the full set of filter strengths being available first(0-15 for primary, 0-4 for secondary).

### 2. Implement curve based speed pruning: 
As quality increases(quantizer decrease), the CDEF strength filter search space is made smaller following a power curve function.
As speeds increase(s0 all the way to s10), the CDEF strength filter search space is made smaller quicker, and at much higher speeds, gets restricted entirely from the start.

This would mainly be for the psychovisual tune. The PSNR tune would only get static search spaces based on speed features.

### 3. Implement an entirely different superior CDEF_dist metric(long-term).

The current CDEF_dist metric, while still being considerably better than what aomenc and SVT-AV1 use(MSE as the dist metric), isn't the absolute best that can be currently used. My future plan is to add a subset of ssimulacra2 as a superior slower dist_metric for the psycho-visual tune, especially as it would penalize some classic encoder faults like excessive blurring and detail wiping. Repository can be found here:
https://github.com/cloudinary/ssimulacra2

Piping work for that:
1. Implementing a YUV > XYB and XYB > YUV crate for internal color conversion(could perhaps be used to replace YCbCr internally for rav1e, but that's on a completely different scope...)

2. Implementing everything ssimulacra2 related in Rust(biggest difficulty).

3. Adding some SIMD to make it as abominably fast as possible.

### 4. Making CDEF faster 

Simple as that: lower complexity, more SIMD for more architectures, etc.

That'll be all from me today.

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -174,6 +174,18 @@ pub struct CliOptions {
   /// Still picture mode
   #[clap(long, help_heading = "ENCODE SETTINGS")]
   pub still_picture: bool,
+  /// Quality tuning for the CDEF strength selection algorithm
+  /// ssimulacra2 for the highest quality(experimental, can be changed), default for the fastest mode
+  #[cfg(feature = "unstable")]
+  #[clap(
+    long,
+    value_parser,
+    hide = true,
+    default_value_t = default
+    help_heading = "ENCODE SETTINGS"
+  )]
+  pub cdef_tune,
+
   /// Uses grain synthesis to add photon noise to the resulting encode.
   /// Takes a strength value 0-64.
   #[cfg(feature = "unstable")]


tmatth:
Also resurfacing [this idea you brought up:](https://github.com/xiph/rav1e/issues/2759#issuecomment-1069766045)
> Furthermore, since CDEF can actually hurt fidelity when a lot of noise is present, a simple noise estimation algorithm could be 
used to disable CDEF filtering once enough noise reaches the threshold(also based on quantizer somewhat).

which definitely seems achievable if we can find a decent heuristic for "noise in source + quantizer will make CDEF harmful"

BlueSwordM:
Yes, and you just gave me an idea as well for further speedups and threading benefits.

If we could do it noise estimation at a transparent tile level(no actual tiles being made, just for the analysis), we could get great threading on that end since the noise estimation could easily be run in parallel, and this would also allow for greater CDEF control over the frame level method.

Now, that would require my implementation to go through tiles, so let's leave that for the end :)

PR #3017, opened 2022-09-13
bin: Bring `--photon-noise` out from under unstable feature

barrbrain:
Align the field name with the flag and the `av1-grain` API. This feature is external to `librav1e` and used only in binaries.

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -176,7 +176,6 @@ pub struct CliOptions {
   pub still_picture: bool,
   /// Uses grain synthesis to add photon noise to the resulting encode.
   /// Takes a strength value 0-64.
-  #[cfg(feature = "unstable")]
   #[clap(
     long,
     conflicts_with = "film-grain-table",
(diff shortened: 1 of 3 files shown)


shssoichiro:
I actually have a WIP PR here https://github.com/xiph/rav1e/pull/2931 that does this. The catch is that Lynne is dead set on not wanting this arg to go into stable unless there is denoising built in, and in the interest of being tired of arguing I attempted to implement denoising. The problem is it's currently too slow to be usable, something like 10-20x slower than the Vapoursynth implementation.

barrbrain:
I have no objection to splitting this into multiple steps. Film grain tables are already part of our stable API and this just adds a means to generate them to our binaries. Denoising is an open problem and orthogonal to grain synthesis which is outside the RDO loop. We most certainly should factor the grain parameters into our encoding decisions in a future enhancement.

barrbrain:
There's nothing stopping a user of the `rav1e` binary from using an external denoiser and setting film grain parameters via the CLI. When we make film grain parameter generation part of the `librav1e` API, it makes sense to couple it with denoising and RDO changes.

barrbrain:
Anyhow, as a compromise I suggest completing #2931 on top of this so that both a baseline denoiser and grain parameter generation are available through the CLI.

lu-zero (approved):
The patch is fine IMHO.

PR #3020, opened 2022-09-16
CI: Configure ClusterFuzzLite

barrbrain:
Override clusterfuzz's hard-coded maximum unit timeout of 25 seconds. Add a local `run_fuzzers` action that uses our Docker image. Follows the guide in `CONTRIBUTING.md` on fuzzing.


barrbrain:
This needs some additional testing for efficient configuration. We will also need to choose where to persist the corpus and coverage reports.

barrbrain:
Archiving this PR for now. The main takeaway is that we need to rework the fuzzing design to be more efficient.

Luni-4:
@barrbrain 

I propose [proptest](https://github.com/proptest-rs/proptest) as possible alternative

PR #2962, opened 2022-06-25
Rewrite argument parsing to use Clap's derives

shssoichiro:
This is something that has been on the radar for a while, but the recent
deprecation of everything from Clap 3.1 in Clap 3.2 has pushed our hand
a bit to get it done now with the new best practices from Clap 3.2.

PR #3022, opened 2022-09-19
Add back shorthand for `--verbose` and `--quiet`

redzic:
This shorthand flag appears to have been dropped in #2962, the removal of which broke some scripts.


barrbrain (approved):

barrbrain:
It appears that the short flag for verbose was also dropped. Perhaps check for others before merging this PR.

redzic:
> It appears that the short flag for verbose was also dropped. Perhaps check for others before merging this PR.

Nice catch, I looked at the code before #2962  and didn't find any other dropped short flags.

PR #3024, opened 2022-09-20
Keep the version line compatible with 0.5

lu-zero:
It would trip scripts parsing the version line otherwise.


shssoichiro (approved):

shssoichiro:
A script parsing the version line seems inherently fragile... however, I'm fine with landing this.

lu-zero:
there aren't alternatives sadly =/

PR #3033, opened 2022-10-04
Move v_frame to its own repository

shssoichiro:
Moved to https://github.com/rust-av/v_frame


barrbrain (approved):

vibhoothi (commented):
Nice, would be nice if we could preserve the history when moving to v_frame, not sure if it is too late now in the other end, I remember, we could separate a folder's history and files to new location(like we do for dav1d's assembly import).

shssoichiro:
We can update that separately from this PR, although I think it's a good idea. But any history revision we do in that repo won't affect any packages pulling it in, since this is pulling from crates.io.

PR #3034, opened 2022-10-05
Add low-luminance AQ bias

shssoichiro:
This adds a bias to give more bits to dark areas of the image. This idea originated in x264, and has generally been successful, so we wanted to add it into rav1e.

The approach we've taken here expands upon the basic premise by converting the YUV image to HSL, a colorspace which is optimal for measuring the lightness of an image. While the Y place in YUV has a loose correlation with lightness, a Y of 0 does not imply that an image is black (or even dark), and a Y of 255 does not imply that an image is white. The L channel in HSL directly corresponds to the human-perceptual lightness, and is linear, so it can be used directly for our purposes.

This conversion to HSL is also inherently HDR-aware, as long as the user has specified the matrix coefficients, etc. on the rav1e command line.

[AWCY results at speed 2](https://beta.arewecompressedyet.com/?job=default-s2%402022-10-25T21%3A25%3A57.683Z&job=s2-dark-aq%402022-10-27T03%3A14%3A52.677Z)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,7 +108,7 @@ new_debug_unreachable = "1.0.4"
 once_cell = "1.13.0"
 av1-grain = { version = "0.2.0", features = ["serialize"] }
 serde-big-array = { version = "0.4.1", optional = true }
-yuvxyb = "0.2.0"
+yuvxyb = "0.2.1"
 
 [dependencies.image]
 version = "0.24.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "rav1e"
 version = "0.6.0"
 authors = ["Thomas Daede <tdaede@xiph.org>"]
 edition = "2021"
-rust-version = "1.60.0"
+rust-version = "1.61.0"
 build = "build.rs"
 include = [
     "/Cargo.toml",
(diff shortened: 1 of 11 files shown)


shssoichiro:
Updated AWCY link with fixed commit. Minor differences from previous version on some clips, overall results about the same.

BlueSwordM:
The ways to make this better are 2-fold:

1. Integrate this with the activity masking framework(minor benefit).
2. Make the calculation work in linear light, as it will make it better psycho-visually speaking and be far stronger in areas like HDR.

shssoichiro:
I've converted this to draft state while I work on adding conversion to HSL to get pure linear, human-perceptual lightness values, which will be far more accurate than the luma plane, which only loosely correlates with human-perceptual lightness.

shssoichiro:
Updated using new approach, explained in both code comments and the initial PR description, with new AWCY. The improvement with the new methodology seems to be significantly more consistent across videos.

Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "rav1e"
 version = "0.6.0"
 authors = ["Thomas Daede <tdaede@xiph.org>"]
 edition = "2021"
-rust-version = "1.60.0"
+rust-version = "1.61.0"
```
shssoichiro, line 6: The `nalgebra` library which `yuvxyb` uses for matrix multiplication to convert to RGB requires rustc 1.61.

shssoichiro:
Closing this and will open a new PR once Blue and I have agreed on a better approach

PR #3038, opened 2022-10-15
Remove arrayref dependency

redzic:
Should slightly reduce compile times and dependency fetching.

meant to push this to my own fork... had a disk failure, git files got corrupted, recloned and pushed to 'origin' which used to be set to my own fork... sorry about that 😅 (for clarification, I mean, I still meant to open this PR, I just pushed my local changes to the wrong remote repository)


shssoichiro (approved):
Seems fine

PR #3039, opened 2022-10-16
Replace AVX2 HBD SATD intrinsics with ASM, for some block sizes

redzic:
I'm still working on writing ASM for 8x8-transformed block sizes, but I thought I might as well submit a PR for the 4x4-transformed block sizes that are currently used (16x4 and 4x16 seem unused for satd). This code should be significantly better than what the compiler generates, as the current intrinsics only explicitly vectorize the difference part and summing up the absolute value of the coefficients, while this ASM code also explicitly vectorizes the actual transform part in a much better way than the compiler.]

12-bit requires 32-bit intermediate precision for the coefficients while 10-bit only needs 16-bit. Not sure if satd for 10/12 bit should actually be separate functions, so I just followed what seemed to be the convention, which is to add a `bdmax` argument to the function to determine which code path to use.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -112,6 +112,7 @@ fn build_nasm_files() {
     "src/x86/sad_sse2.asm",
     "src/x86/sad_avx.asm",
     "src/x86/satd.asm",
+    "src/x86/satd16_avx2.asm",
     "src/x86/cdef_dist.asm",
     "src/x86/sse.asm",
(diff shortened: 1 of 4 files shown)


redzic:
From some preliminary tests the 10 bit version of the 8x4 satd was around 8x faster than the Rust code (which was already auto vectorized).

shssoichiro (approved):
Impressive

PR #3041, opened 2022-10-17
rate: Estimate target quantizer on first pass

barrbrain:


src/rate.rs
```diff
@@ -1274,13 +1274,12 @@ impl RCState {
...
-          .select_qi(ctx, output_frameno, FRAME_SUBTYPE_I, None, 0)
-          .log_base_q;
-      }
-    } else {
+    if self.twopass_state == PASS_SINGLE {
+      pass1_log_base_q = self
+        .select_qi(ctx, output_frameno, FRAME_SUBTYPE_I, None, 0)
+        .log_base_q;
+    }
+    if self.pass1_data_retrieved {
```
barrbrain, line 1282: > derf: the branch [...] is only meant to run once, during initialization, but that is no longer being structurally enforced. So you should just move it into a separate function that runs once, during initialization.

lu-zero (approved):
It looks fine.

PR #3044, opened 2022-10-23
Various minor cleanup

shssoichiro:


lu-zero (approved):

barrbrain:
For reference, the [current version of rustc in Ubuntu 22.04 LTS](https://packages.ubuntu.com/jammy/rustc) is `1.61.0`. So this shouldn't significantly impact back-porting.

PR #3047, opened 2022-10-25
Add developer CLI flags for tuning encoder internals

shssoichiro:
Currently adds the following:
- `--deblock-strength`
- `--deblock-sharpness`
- `--ip-qidx-ratio`
- `--temporal-rdo-strength`

These flags will be available only if the `devel` feature flag is active, as they are intended to be used to ease development and not to be tweaked by end users. rav1e continues to subscribe to the philosophy that end users should not need to be cargo culting their command lines to get optimal results, and we would prefer to tune the encoder internals so that users do not have to.


BlueSwordM:
Everything seems good on my end, as all flags seem to work(except --ip-qidx-ratio).
I'd wait until tomorrow to merge it since there might be some more stuff that I would want to add.

shssoichiro:
What is wrong with ip-qidx-ratio?

BlueSwordM:
I should have been clearer: I haven't tested it yet :P

shssoichiro:
Okay, well I will tell you from my testing in AWCY... It doesn't seem to work. :p at least it did not result in the output changing, which means it did not do what I wanted it to do.

src/encoder.rs
```diff
@@ -543,6 +543,19 @@ pub struct DeblockState {
   pub block_delta_multi: bool,
 }
 
+impl DeblockState {
+  pub fn new(config: &EncoderConfig) -> Self {
+    let mut state = DeblockState { ..Default::default() };
+    for level in &mut state.levels {
```
BlueSwordM, line 549: Since rav1e already codes keyframes very well for video and spends a lot of bits on it, it would be preferably to make the deblocking strength control only available to non intra refreshes. However, keeping it available for still picture mode is still advisable since lowering deblocking strength seems to provide higher fidelity results overall.

Here is how I would implement it:

if still_picture {

do deblocking_strength control }

else if !PRIMARY_REF_NONE(don't exactly know how to set this) {
do deblocking_strength control }

That way, we can actually get the best of both worlds.

BlueSwordM: That works for now. Thank you.

PR #3049, opened 2022-10-26
Optimize loading of data for satd kernels

shssoichiro:
Following the ASM update by redzic, this now only affects 4x16 and 16x4 transforms,
but since it is still in use, it is still worth pushing the more optimal assembly.

4x4 kernel Before: https://pastebin.com/GP4Vjb2z
4x4 kernel After: https://pastebin.com/Kr2nz7mZ

### DIFF
--- a/src/asm/x86/dist/hbd.rs
+++ b/src/asm/x86/dist/hbd.rs
@@ -61,9 +61,14 @@ macro_rules! satd_kernel_hbd_avx2 {
           let input1 = &mut Aligned::<[i32; 8 * 8]>::uninitialized().data[..size * size];
           let input2 = &mut Aligned::<[i32; 8 * 8]>::uninitialized().data[..size * size];
           for y in 0..(sizei) {
-            for x in 0..(sizei) {
-              input1.as_mut_ptr().offset(y * sizei + x).write(src.offset(y * src_stride + x).read() as i32);
-              input2.as_mut_ptr().offset(y * sizei + x).write(dst.offset(y * dst_stride + x).read() as i32);
+            if size == 8 {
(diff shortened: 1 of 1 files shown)


src/asm/x86/dist/hbd.rs
```diff
@@ -61,9 +61,14 @@ macro_rules! satd_kernel_hbd_avx2 {
           let input1 = &mut Aligned::<[i32; 8 * 8]>::uninitialized().data[..size * size];
           let input2 = &mut Aligned::<[i32; 8 * 8]>::uninitialized().data[..size * size];
           for y in 0..(sizei) {
-            for x in 0..(sizei) {
-              input1.as_mut_ptr().offset(y * sizei + x).write(src.offset(y * src_stride + x).read() as i32);
-              input2.as_mut_ptr().offset(y * sizei + x).write(dst.offset(y * dst_stride + x).read() as i32);
+            if size == 8 {
```
lu-zero, line 64: I hope this branch is elided by the compiler.

BlueSwordM:
I tested this patch, and for some reason, this periodically creates flat blocks within a frame, which is not normal behavior.

src/asm/x86/dist/hbd.rs
shssoichiro: It is, I verified that the compiler sees `size` as constant and elides out any branches. So the 8x8 version has the first branch only, and the 4x4 version has the second branch only.

shssoichiro:
Although apparently one of the tests is segfaulting... that's probably not good.

shssoichiro:
Made obsolete by #3051

PR #3050, opened 2022-10-26
Add SIMD helpers to speed up Rust get_sad

shssoichiro:
The Rust version of get_sad is still used during ME for regions that are not of an exact block size.
This was consuming about 2.3% of runtime at speed 2.

This change causes the function to iterate over chunks of 4 that can easily be vectorized,
and then only the remainder will go through the slower non-vectorized SAD process.

After the change, the time spend in Rust get_sad
appears to be reduced by about 2/3, and the overall runtime of a speed 2 is decreased by about 1% on an AVX2 CPU.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,7 +108,6 @@ new_debug_unreachable = "1.0.4"
 once_cell = "1.13.0"
 av1-grain = { version = "0.2.0", features = ["serialize"] }
 serde-big-array = { version = "0.4.1", optional = true }
-wide = "0.7.5"
 
 [dependencies.image]
 version = "0.24.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,6 +108,7 @@ new_debug_unreachable = "1.0.4"
 once_cell = "1.13.0"
 av1-grain = { version = "0.2.0", features = ["serialize"] }
 serde-big-array = { version = "0.4.1", optional = true }
+wide = "0.7.5"
 
 [dependencies.image]
 version = "0.24.3"
--- a/src/dist.rs
+++ b/src/dist.rs
@@ -46,22 +46,19 @@ pub(crate) mod rust {
       let mut iter_org = slice_org[..w].chunks_exact(4);
       let mut iter_ref = slice_ref[..w].chunks_exact(4);
       for (chunk_org, chunk_ref) in iter_org.by_ref().zip(iter_ref.by_ref()) {
-        let chunk_org = wide::i32x4::from([
-          i32::cast_from(chunk_org[0]),
-          i32::cast_from(chunk_org[1]),
-          i32::cast_from(chunk_org[2]),
-          i32::cast_from(chunk_org[3]),
-        ]);
-        let chunk_ref = wide::i32x4::from([
-          i32::cast_from(chunk_ref[0]),
-          i32::cast_from(chunk_ref[1]),
-          i32::cast_from(chunk_ref[2]),
-          i32::cast_from(chunk_ref[3]),
-        ]);
-        let sad = (chunk_org - chunk_ref).abs();
-        for i in sad.to_array() {
-          sum += i as u32;
-        }
+        let chunk_org = {
+          let mut i = chunk_org.iter();
+          [(); 4].map(|_| *i.next().unwrap()).map(i32::cast_from)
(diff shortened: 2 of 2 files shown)


src/dist.rs
```diff
@@ -43,10 +43,31 @@ pub(crate) mod rust {
     for (slice_org, slice_ref) in
       plane_org.rows_iter().take(h).zip(plane_ref.rows_iter())
     {
-      sum += slice_org
+      let mut iter_org = slice_org[..w].chunks_exact(4);
+      let mut iter_ref = slice_ref[..w].chunks_exact(4);
+      for chunk_org in iter_org.by_ref() {
+        let chunk_ref = iter_ref.next().expect("rows are same size");
```
lu-zero, line 49: why not using zip?

shssoichiro: The first time I tried it the compiler was yelling at me. I went back again and it works, I think I was missing the `by_ref`

barrbrain:
I don't think the `wide` crate is necessary here. If you use arrays and write the kernel idiomatically, LLVM will do the right thing.

src/dist.rs
```diff
@@ -43,10 +43,30 @@ pub(crate) mod rust {
...
+        let chunk_ref = wide::i32x4::from([
+          i32::cast_from(chunk_ref[0]),
+          i32::cast_from(chunk_ref[1]),
+          i32::cast_from(chunk_ref[2]),
+          i32::cast_from(chunk_ref[3]),
+        ]);
+        let sad = (chunk_org - chunk_ref).abs();
+        for i in sad.to_array() {
+          sum += i as u32;
+        }
```
barrbrain, line 64: Alternatively, without the `wide` crate:
```suggestion
        let chunk_org = [
          i32::cast_from(chunk_org[0]),
          i32::cast_from(chunk_org[1]),
          i32::cast_from(chunk_org[2]),
          i32::cast_from(chunk_org[3]),
        ];
        let chunk_ref = [
          i32::cast_from(chunk_ref[0]),
          i32::cast_from(chunk_ref[1]),
          i32::cast_from(chunk_ref[2]),
          i32::cast_from(chunk_ref[3]),
        ];
        sum += chunk_org
          .iter()
          .zip(chunk_ref)
          .map(|(a, b)| (a - b).abs() as u32)
          .sum::<u32>();
```

shssoichiro: Interestingly the compiler seems to be failing to accomplish this--at least given that I have `target-cpu=native` enabled, the `wide` crate is vectorizing with intrinsics and the generated asm is showing `vmovd` and `vpinsrd` instructions. The plain Rust code seems to be failing to autovectorize. (I think without `target-cpu=native` the compiler still defaults to sse2 which should still allow wide to use some intrinsics.)

barrbrain: For both 8-bit and 10-bit encodes, `get_sad` only accounts for 0.03% of CPU time after this change.
With the above suggestion, perf shows that the loop is auto-vectorized and hot:
```
  0.51 │285:   xor          %r14d,%r14d
  1.19 │       nop
  4.50 │290:   vmovd        (%rdi,%r14,1),%xmm0
 15.55 │       vmovd        (%r10,%r14,1),%xmm1
 14.23 │       vpsadbw      %xmm0,%xmm1,%xmm0
 15.73 │       vmovd        %xmm0,%ebx
 12.56 │       add          %ebx,%eax
  2.83 │       add          $0x4,%r14
  0.52 │       cmp          %r14,%rdx
  1.77 │     ↑ jne          290
```

shssoichiro: Interesting. I was using `cargo asm` to show the asm and it was not showing autovectorization.

Cargo.toml
```diff
@@ -108,6 +108,7 @@ new_debug_unreachable = "1.0.4"
 once_cell = "1.13.0"
 av1-grain = { version = "0.2.0", features = ["serialize"] }
 serde-big-array = { version = "0.4.1", optional = true }
+wide = "0.7.5"
```
barrbrain, line 111: We can drop this new dependency for now, as it is unused.

shssoichiro: Don't worry, I'll pull it back in shortly 🙂

barrbrain (approved):

PR #3051, opened 2022-10-27
Replace last 16x4 and 4x16 avx2 hbd satd intrinsics with asm

redzic:
16x4 for 10 bit is ~4x faster than the old intrinsics version.


barrbrain (approved):
Looks good, I'll let you fix any code formatting issues before merge.

shssoichiro (approved):
What an absolute boss

redzic:
looks like CI is failing due to network issues

PR #3052, opened 2022-10-27
Produced optimized binary versions from CI

shssoichiro:
There is up to a 10% performance improvement from compiling with `target-cpu=native`. We can provide the majority of this benefit to users of the pre-compiled binaries by providing avx2-optimized binaries which will be better autovectorized than the generic binaries.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -34,6 +34,7 @@ jobs:
     - name: Upload rav1e-ch msvc binary (unstable)
       uses: actions/upload-artifact@v3
       with:
+        name: rav1e-ch (unstable) (AVX2)
         path: target/release/rav1e-ch.exe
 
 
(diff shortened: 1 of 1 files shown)


barrbrain:
This would be a bit easier to review if the formatting were applied in a separate patch.

shssoichiro:
The easiest way was to split it into two commits so at least [the important stuff](https://github.com/xiph/rav1e/pull/3052/commits/394097d766b4b544bb5735dbb3e73b3f59d2d224) is viewable separately. Sometimes autoformatting is more of a pain than it is helpful...

lu-zero:
Can you please rebase it? I'd land it before tagging the alpha.

lu-zero (approved):

barrbrain:
Rebased in 3 passes:
* Regenerated the formatting patch with prettier; second patch applied cleanly
* Reverted indentation changes by prettier; resolved conflicts on second patch by applying the same indent
* Dropped formatting patches

The final patch is a little smaller than the original because some of the changes were already applied by `dependabot`.

PR #3054, opened 2022-11-01
tests: Add three-pass encode with target bitrate

barrbrain:
This was previously uncovered when this feature regressed during CLI rework for a library upgrade.


shssoichiro (approved):
Thanks!

PR #3057, opened 2022-11-09
Write level stored in sequence struct.

tdaede:


src/header.rs
```diff
@@ -287,12 +287,17 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
     self.write_bit(fi.sequence.still_picture)?; // still_picture
     self.write_bit(fi.sequence.reduced_still_picture_hdr)?; // reduced_still_picture_header
 
+    assert!(fi.sequence.level[0][0] <= 3);
+    assert!(fi.sequence.level[0][1] <= 7);
```
shssoichiro, line 291: Won't this fail because the default is [3, 9]?

src/header.rs
```diff
@@ -287,12 +287,17 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
...
     if fi.sequence.reduced_still_picture_hdr {
       assert!(!fi.sequence.timing_info_present);
       assert!(!fi.sequence.decoder_model_info_present_flag);
       assert_eq!(fi.sequence.operating_points_cnt_minus_1, 0);
       assert_eq!(fi.sequence.operating_point_idc[0], 0);
-      self.write(5, 31)?; // level
+      self.write(
+        5,
+        (fi.sequence.level[0][1] - 2) << 2 & fi.sequence.level[0][0],
```
shssoichiro, line 299: Might be a little pedantic, but since we use this in two places, it might be worthwhile to make this into a function.

shssoichiro (changes requested):

src/encoder.rs
```diff
@@ -227,15 +227,16 @@ impl Sequence {
...
     let mut tier: [usize; MAX_NUM_OPERATING_POINTS] =
       [0; MAX_NUM_OPERATING_POINTS];
 
     for i in 0..MAX_NUM_OPERATING_POINTS {
       operating_point_idc[i] = 0;
-      level[i][0] = 1; // minor
-      level[i][1] = 2; // major
+      // default to unconstrained
+      level[i][0] = 3; // minor
+      level[i][1] = 9; // major
```
rzumer, line 239: I understand that this isn't new in your PR, but since you are touching this, I find the array syntax a bit confusing. First because it's not clear without reading the code that the minor version comes before the major one. I'd expect the opposite. Second because special-purpose levels, like reserved values and unconstrained parameters, are not numbered in the spec, so having to "translate" unconstrained parameters to "9.3" is unintuitive. And indeed it has already caused an error in the assertion below.

src/header.rs
rzumer, line 299: If keeping the level as an array I would extract the index calculation to a variable for clarity.

src/encoder.rs
tdaede: I would actually prefer storing it as an index. I'll change it.

src/header.rs
tdaede: yeah my bad I added those asserts last minute

tdaede: removed

src/encoder.rs
tdaede: changed

src/header.rs
tdaede: changed to an index so this function is gone

tdaede: function gone entirely

src/header.rs
```diff
@@ -309,7 +314,10 @@ impl<W: io::Write> UncompressedHeader for BitWriter<W, BigEndian> {
       self.write_bit(false)?; // initial display delay present flag
       self.write(5, 0)?; // one operating point
       self.write(12, 0)?; // idc
-      self.write(5, 31)?; // level
+      self.write(
+        5,
+        (fi.sequence.level[0][1] - 2) << 2 & fi.sequence.level[0][0],
```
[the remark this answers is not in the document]
tdaede: gone

src/encoder.rs
```diff
@@ -197,8 +197,7 @@ pub struct Sequence {
   pub operating_point_idc: [u16; MAX_NUM_OPERATING_POINTS],
   pub display_model_info_present_flag: bool,
   pub decoder_model_info_present_flag: bool,
-  /// minor, major
-  pub level: [[usize; 2]; MAX_NUM_OPERATING_POINTS],
+  pub level_idx: [usize; MAX_NUM_OPERATING_POINTS],
```
rzumer, line 200: Wasn't this changed to u8 in the previous iteration? It would spare you from casting it when writing the value.

tdaede: last iteration didn't save level_idx here, u8 means I have to cast in other locations in the next patch but I'll go with u8 to save the ram

rzumer (approved):

shssoichiro (approved):

PR #3063, opened 2022-11-16
CI: Correct format of target_cpu for x86_64 in deploy workflow

barrbrain:
Confusingly, the LLVM target-cpu format is not the same as in the target triple.


shssoichiro (approved):
Huh.

PR #3064, opened 2022-11-22
Add C API for level.

tdaede:


src/capi.rs
```diff
@@ -1333,6 +1350,9 @@ mod test {
       rav1e_config_parse_int(rac, h.as_ptr(), 64);
       let s = CString::new("speed").unwrap();
       rav1e_config_parse_int(rac, s.as_ptr(), 10);
+      let l = CString::new("level").unwrap();
+      let lo = CString::new("unconstrained").unwrap();
+      rav1e_config_parse(rac, l.as_ptr(), lo.as_ptr());
```
lu-zero, line 1355: I'd test the with `6.2` and maybe add a fail test with `8.3`.

lu-zero (approved):
Looks fine.

src/capi.rs
tdaede: added

Issue #3066, opened 2022-11-24
Segfault in rav1e_idct_8x16_internal_16bpc_sse4


atourem3:
**Describe the bug**
Encoder crashes when trying to encode 10-bit video at --speed ≤ 1 on an Ivy Bridge Xeon E3-1245 v2.

**To Reproduce**
Run RAV1E_CPU_TARGET=SSE4.1 rav1e --speed 1 with 10-bit input.

**Expected behavior**
It should encode.

**Required Information**
Version:
```
$ rav1e --version
rav1e 0.6.0 (p20221123) (release)
rustc 1.65.0 (897e37553 2022-11-02) x86_64-unknown-linux-gnu
Compiled CPU Features: aes,avx,fxsr,pclmulqdq,popcnt,rdrand,sse,sse2,sse3,sse4.1,sse4.2,ssse3,xsave,xsaveopt
Assembly: Enabled
Threading: Enabled
Unstable Features: Disabled
Compiler Flags: -Ctarget-cpu=native
Built Thu, 24 Nov 2022 05:37:03 +0000
```
Operating system:
```
$ uname -a
Linux arch 6.0.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 10 Nov 2022 21:14:24 +0000 x86_64 GNU/Linux
```

**Console Output**
```
Thread 1 "rav1e" received signal SIGSEGV, Segmentation fault.
0x0000555555840432 in rav1e_idct_8x16_internal_16bpc_sse4 ()
```
<Details>
$ rust-gdb rav1e
GNU gdb (GDB) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from rav1e...
[...]

PR #3067, opened 2022-11-25
x86: Disable 8x16 16bpc inverse transforms for SSE4.1

barrbrain:
This avoids a segmentation fault specifically in `rav1e_idct_8x16_internal_16bpc_sse4`, a minimal fix for #3066.


lu-zero:
is this enough?

lu-zero (approved):

PR #3071, opened 2022-11-30
Fix the capi deps

lu-zero:


tmatth (approved):
Fixed the build here, thanks!

PR #3075, opened 2022-12-02
Specify default threads behavior

shssoichiro:


tdaede (approved):

tdaede:
Might also be nice to specify what can use threads. e.g. hint that tiles may be necessary to use a lot of threads.

tmatth (approved):

PR #3081, opened 2022-12-12
Drop BUILT_TIME_UTC

bmwiedemann:
Drop `BUILT_TIME_UTC`
to make builds reproducible.
See https://reproducible-builds.org/ for why this is good.

Alternatively someone could patch `built/src/lib.rs` to support [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/#rust)


lu-zero (approved):
It is fine by me, @shssoichiro are you ok with it?

PR #3086, opened 2023-01-14
Commit Rust 1.60 compatible Cargo.lock

shssoichiro:
As seen in #3085, not having the Cargo.lock file committed can lead to the inability to compile rav1e if some upstream crates do not respect semver. Many crates, unfortunately, do not consider a change in the minimum Rust version to be a breaking change, so this has been and will continue to be a recurring problem unless we commit Cargo.lock. This also generally is good for making builds reproducible.

The downside is we will have to manually update dependencies or use a tool like dependabot to update them for us (which will run through CI to verify the updates do not break our builds). I consider this preferable to having random build breakages that are outside of our control.


barrbrain (commented):
Can we keep Cargo.version for the cache key?

barrbrain (approved):

lu-zero:
You are trading hiding bugs for additional discomfort reviewing patches.

Please revert it and if you want to put workarounds add a patch section in Cargo.toml.

PR #3093, opened 2023-01-15
Enable dependabot to update Cargo dependencies

shssoichiro:
This will be particularly useful if we are committing the Cargo.lock file to aid in keeping us up-to-date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,9 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every week
       interval: "weekly"
+  # Maintain dependencies for Cargo
+  - package-ecosystem: "cargo"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,3 +6,9 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every week
       interval: "weekly"
+  # Maintain dependencies for Cargo
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
```
tmatth, line 13: nit: this comment is copy-pasted from the previous block and doesn't apply here, IMO it can just be dropped

Luni-4 (approved):

PR #3094, opened 2023-01-15
Hint that creating a region from an empty plane is unlikely

shssoichiro:
This is part of a series of commits authored by @maj160 to improve performance of rav1e.

This commit results in an overall performance improvement of <1% at speed 2 10-bit.

### DIFF
--- a/src/tiling/plane_region.rs
+++ b/src/tiling/plane_region.rs
@@ -144,21 +144,27 @@ macro_rules! plane_region_common {
       pub fn is_null(&self) -> bool {
         self.data.is_null()
       }
+
+      #[cold]
+      pub fn empty(plane_cfg : &'a PlaneConfig) -> Self {
+        return Self {
+          // SAFETY: This is actually pretty unsafe.
+          // This means we need to ensure that no other method on this struct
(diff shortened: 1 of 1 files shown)


tdaede (approved):

tdaede:
I guess another obvious thing to do is find out where empty planes are even used...

maj160:
> I guess another obvious thing to do is find out where empty planes are even used...

I think empty planes are only used in `api::lookahead::estimate_intra_costs` to create a cheap rec frame that gets thrown away.
Root cause seems to be that motion estimation takes a `TileStateMut`, which itself contains a non-optional `rec: TileMut`.
Solution is probably to decouple the read-only stuff that ME uses (mostly just tile bounds and hres/qres inputs?) from the mutable stuff (MEStats, Reconstructed Frame).

Cursory glance suggested it would be a lot of work to unpick and refactor, so I haven't really touched it.

tdaede:
OK. I think having empty planes is a useful thing to have around (also for 4:0:0 etc). It still might be nice to disallow PlaneRegions from them (panic paths will naturally be cold), but that still needs a bit of work too.

PR #3096, opened 2023-01-15
Minor optimization to take_slice

shssoichiro:


src/rdo.rs
```diff
@@ -478,11 +478,7 @@ pub fn spatiotemporal_scale<T: Pixel>(
   // compared to using `Iterator::take`
   #[inline(always)]
   fn take_slice<T>(slice: &[T], n: usize) -> &[T] {
-    if n >= slice.len() {
-      slice
-    } else {
-      &slice[..n]
-    }
+    slice.get(..n).unwrap_or(slice)
```
tdaede, line 481: Might be worth documenting the new possible panic (I think if frame_bo is too close to the edge)

maj160: I don't think this panics - if the get fails then the `unwrap_or(...)` (not `unwrap()`!) returns the original slice.

shssoichiro: The behavior should be identical. `slice.get` returns an `Option` which is `None` if the requested index is out of range.

tdaede: Indeed I misread that.

tdaede (approved):

PR #3097, opened 2023-01-16
Use region dimensions in SAD and ME

shssoichiro:
This avoids having to re-check bounds every time we perform SAD, as the region knows its own size. It also may save 2 usize's being passed around during ME. To enforce this, we also remove the w and h parameters from everywhere.

This is part of a series of commits authored by @maj160 to improve performance of rav1e.


shssoichiro:
This is going to be rather painful to debug given that it only seems to be failing on x64 Macos. Why that is, I have no clue... I'll fix it up when I have a moment.

PR #3103, opened 2023-01-18
Use a bit counter instead of a byte counter in EC

shssoichiro:
This allows us to avoid accounting for bits in WriterCounter, which is used heavily when estimating bit costs of various decisions.

### DIFF
--- a/src/ec.rs
+++ b/src/ec.rs
@@ -209,13 +209,14 @@ impl StorageBackend for WriterBase<WriterCounter> {
       stream_size: self.s.bits,
       backend_var: 0,
       rng: self.rng,
+      // We do not use `cnt` within Counter, but setting it here allows the compiler
+      // to do a 32-bit merged load/store.
       cnt: self.cnt,
     }
   }
   #[inline]
   fn rollback(&mut self, checkpoint: &WriterCheckpoint) {
     self.rng = checkpoint.rng;
-    self.cnt = checkpoint.cnt;
     self.s.bits = checkpoint.stream_size;
   }
 }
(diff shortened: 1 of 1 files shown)


src/ec.rs
```diff
@@ -193,22 +193,18 @@ impl StorageBackend for WriterBase<WriterCounter> {
...
-  fn stream_bytes(&mut self) -> usize {
-    self.s.bytes
+  fn stream_bits(&mut self) -> usize {
+    self.s.bits
   }
   #[inline]
   fn checkpoint(&mut self) -> WriterCheckpoint {
     WriterCheckpoint {
-      stream_bytes: self.s.bytes,
+      stream_bytes: self.s.bits,
```
tdaede, line 207: Can we rename stream_bytes, maybe to e.g. stream_counter? It is convenient that both Counter and Recorder use the same Checkpoint type I suppose, but at least a rename and comment would make it less confusing.

src/ec.rs
```diff
@@ -791,7 +787,7 @@ where
   fn tell(&mut self) -> u32 {
     // The 10 here counteracts the offset of -9 baked into cnt, and adds 1 extra
     // bit, which we reserve for terminating the stream.
-    (((self.stream_bytes() * 8) as i32) + (self.cnt as i32) + 10) as u32
+    (((self.stream_bits()) as i32) + (self.cnt as i32) + 10) as u32
```
tdaede, line 790: I think this technically should have an AWCY run, as we actually change the result of tell() for WriterCounter.

barrbrain: https://beta.arewecompressedyet.com/?job=master-90fdd3dec515600b7a9ab67ed071e5a88f804cee&job=maj-optim-a6%402023-01-18T12%3A50%3A49.710Z
I'll attach the results when it completes.

shssoichiro: I have been running each of these commits locally to ensure the output videos are identical. Looks like AWCY agrees, doesn't hurt to have it run for the extra thoroughness.

PR #3104, opened 2023-01-18
ec: Manually inline `msb()` and `ilog()` for clarity

barrbrain:
Also due to them not being forced inline in the `v_frame` crate.

### DIFF
--- a/src/ec.rs
+++ b/src/ec.rs
@@ -645,7 +645,7 @@ where
   fn count_quniform(&self, n: u32, v: u32) -> u32 {
     let mut bits = 0;
     if n > 1 {
-      let l = 32 - n.leading_zeros() as u32;
+      let l = 32 - n.leading_zeros();
       let m = (1 << l) - n;
       bits += (l - 1) << OD_BITRES;
       if v >= m {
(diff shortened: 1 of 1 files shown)


src/ec.rs
```diff
@@ -194,7 +193,7 @@ impl StorageBackend for WriterBase<WriterCounter> {
   #[inline]
   fn store(&mut self, fl: u16, fh: u16, nms: u16) {
     let (_l, r) = self.lr_compute(fl, fh, nms);
-    let d = 16 - ILog::ilog(r);
+    let d = r.leading_zeros() as usize;
```
tdaede, line 196: heh much more straightfoward

src/ec.rs
```diff
@@ -646,7 +645,7 @@ where
   fn count_quniform(&self, n: u32, v: u32) -> u32 {
     let mut bits = 0;
     if n > 1 {
-      let l = (msb(n as i32) + 1) as u32;
+      let l = 32 - n.leading_zeros() as u32;
```
barrbrain, line 648:
```suggestion
      let l = 32 - n.leading_zeros();
```

PR #3107, opened 2023-01-20
CI: Disable unstable sparse cargo registry

barrbrain:
Specifically, the fetch step before cargo-c build seems to be adversely impacted by this feature.

### DIFF
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -198,6 +198,8 @@ jobs:
           cargo check --features=unstable,channel-api
       - name: Run cargo-c
         if: matrix.conf == 'cargo-c'
+        env:
+          CARGO_UNSTABLE_SPARSE_REGISTRY: false
         run: |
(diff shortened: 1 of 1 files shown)


shssoichiro:
I couldn't seem to find info on this other than that it seems to be nightly-only and disabled by default as of June 2022. What issue is this attempting to resolve?

barrbrain:
It appears to be a recent change in `dtolnay/rust-toolchain@stable` 

> : enable Cargo sparse registry
> echo CARGO_UNSTABLE_SPARSE_REGISTRY=true >> $GITHUB_ENV

This caused the `cargo fetch` step of our `cargo-c` targets to silently fail.

shssoichiro:
Huh... interesting. I wonder why that was turned on globally for everyone using this Github action... seems like a bad plan, although I don't know enough about it to say that for certain.

shssoichiro (approved):

PR #3109, opened 2023-01-21
Fix rounding issue in HBD CDEF code

shssoichiro:
This resolves an issue where occasionally the ASM test for cdef_dist_hbd would fail due to the result being slightly off.


barrbrain (approved):
The reference implementation has grown in complexity to the point that it took multiple reads before I found the [matching lines in src/dist.rs](https://github.com/xiph/rav1e/blob/p20230117/src/dist.rs#L347-L354).

tmatth (approved):
Classic

PR #3110, opened 2023-01-21
ec: Simplify `lr_compute` function

barrbrain:
Not sure about performance impact, just trying to make the flow clear.


shssoichiro:
Looks better to me. I can't imagine the performance would be worse, it looks like less math overall.

shssoichiro (approved):

PR #3113, opened 2023-01-21
Move quant tables to separate file to improve organization

shssoichiro:


src/quantize/tables.rs
```diff
@@ -0,0 +1,154 @@
+pub const MINQ: usize = 0;
```
vibhoothi, line 1:
```suggestion
// Copyright (c) 2017-2023, The rav1e contributors. All rights reserved
//
// This source code is subject to the terms of the BSD 2 Clause License and
// the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
// was not distributed with this source code in the LICENSE file, you can
// obtain it at www.aomedia.org/license/software. If the Alliance for Open
// Media Patent License 1.0 was not distributed with this source code in the
// PATENTS file, you can obtain it at www.aomedia.org/license/patent.

pub const MINQ: usize = 0;
```

Nit: License headers seems to missing when making new file^^

shssoichiro:
@vibhoothi any chance you could take another look at this?

tmatth (approved):
Should probably be squashed (not worth having a commit without the license headers).

PR #3115, opened 2023-01-22
Optimize base quants using NonZero integers

shssoichiro:
This allows additional optimizations such as eliding division-by-zero checks.

Benchmarks with hyperfine seem to show a 3% overall improvement at speed 2.


shssoichiro:
Depends on #3113

lu-zero (approved):

vibhoothi (commented):
Would require to squash both commits before merging I believe.

PR #3116, opened 2023-01-27
Improvements to sgrproj HBD performance

shssoichiro:
sgrproj_sum_finish in particular was identified as a hotspot in HBD encoding. This changeset attempts to improve the performance of it and surrounding code by:

- Moving bounds checks out of a hot loop
- Use builtin saturating_sub method which is more efficient than performing multiple casts
- Use generics to allow compile-time computation of math dependent on bit-depth


lu-zero (approved):
We could do a pass and move const generics further up.

PR #3120, opened 2023-01-30
Minimize bounds checks in pred functions

shssoichiro:


src/predict.rs
```diff
@@ -821,7 +817,7 @@ pub(crate) mod rust {
     });
     let avg = T::cast_from((sum + (height >> 1) as u32) / height as u32);
     for line in output.rows_iter_mut().take(height) {
-      line[..width].iter_mut().for_each(|v| *v = avg);
+      line[..width].fill(avg);
```
shssoichiro, line 820: These currently generate equivalent ASM, but using `fill` gives us a guarantee that some compiler internals won't change the behavior of `for_each` in the future. (Also IMO it makes it clearer what we are doing--that this is equivalent to a memset)

barrbrain (approved):
The vintage of this code shows. Thank you for tidying up.

Issue #3162, opened 2023-03-19
Undefined behaviour in `CDFContextLogOps::rollback`


CodesInChaos:
`CDFContextLogOps::rollback` looks like this

``` rust
  #[inline(always)]
  fn rollback(
    log: &mut CDFContextLogBase, fc: &mut CDFContext, checkpoint: usize,
  ) {
    let base = fc as *mut _ as *mut u8;
    let mut len = log.data.len();
    // SAFETY: We use unchecked pointers here for performance.
    // Since we know the length, we can ensure not to go OOB.
    unsafe {
      let mut src = log.data.get_unchecked_mut(len) as *mut u16;
      while len > checkpoint {
        len -= Self::CDF_LEN_MAX + 1;
        src = src.sub(Self::CDF_LEN_MAX + 1);
        let offset = *src.add(Self::CDF_LEN_MAX) as usize;
        let dst = base.add(offset) as *mut u16;
        dst.copy_from_nonoverlapping(src, Self::CDF_LEN_MAX);
      }
      log.data.set_len(len);
    }
  }
```

The problematic part is:

``` rust
let mut src = log.data.get_unchecked_mut(len) as *mut u16;
```

This index is out-of-bounds, since it's equal to the length, not smaller than it. While the reference is never used, the documentation for [`slice::get_unchecked_mut`](https://doc.rust-lang.org/std/primitive.slice.html#method.get_unchecked_mut) states that this is still undefined behaviour:

> Calling this method with an out-of-bounds index is [undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) **even if the resulting reference is not used**.

I believe changing that line to:

``` rust
let mut src = log.data.as_mut_ptr().add(len);
```

would achieve the same effect while avoiding UB.

`push` has the same problem:

``` rust
let dst = log.data.get_unchecked_mut(len) as *mut u16;
```

lu-zero:
You are right, if you send a patch we'll land it.

PR #3165, opened 2023-03-19
Fix undefined behavior in CDFContextLogOps

CodesInChaos:
Fixes https://github.com/xiph/rav1e/issues/3162


barrbrain (approved):
This change better matches the ultimate intent of this code. `Vec` internals were replaced by carefully managed pointer operations. The use of `get_unchecked_mut()` is hanging over from the original implementation. I would like to explicitly name the undefined behaviour in the commit message, so I will adapt that from the Rust docs on merge.

PR #3167, opened 2023-03-19
Add debug-assertions and comments related to safety conditions

CodesInChaos:


src/context/cdf_context.rs
```diff
@@ -601,6 +601,10 @@ trait CDFContextLogOps: CDFContextLogSize {
       let capacity = log.data.capacity();
       debug_assert!(new_len <= capacity);
       let dst = log.data.as_mut_ptr().add(len);
+      // This reads beyond the end of `cdf`.
+      // Since it's part of a larger data structure, the out-of-bounds data is valid.
```
barrbrain, line 605: The larger data structure is `struct CDFContext` and we should have enough language features available now to rework this module.

src/context/cdf_context.rs
```diff
@@ -615,14 +619,20 @@ trait CDFContextLogOps: CDFContextLogSize {
   ) {
     let base = fc as *mut _ as *mut u8;
     let mut len = log.data.len();
+    debug_assert!(len % (Self::CDF_LEN_MAX + 1) == 0);
     // SAFETY: We use unchecked pointers here for performance.
     // Since we know the length, we can ensure not to go OOB.
     unsafe {
       let mut src = log.data.as_mut_ptr().add(len);
       while len > checkpoint {
+        debug_assert!(len >= Self::CDF_LEN_MAX + 1);
```
barrbrain, line 628: Clippy is being pedantic:
```
error: unnecessary `>= y + 1` or `x - 1 >=`
```
```suggestion
        debug_assert!(len > Self::CDF_LEN_MAX);
```

CodesInChaos: I think `>= Self::CDF_LEN_MAX + 1` represents the intent better, since we need to ensure that it's possible to remove `Self::CDF_LEN_MAX + 1` elements from the collection.

barrbrain: I might go ahead an rework the log structs with const generics. Then these assertions can be simplified or elided.

lu-zero: I prefer the way clippy suggests though.

src/context/cdf_context.rs
barrbrain: I have completed the rework in #3169 and #3172, please take a look and rebase.

src/context/cdf_context.rs
barrbrain: As above, please take a look at the latest code.

PR #3169, opened 2023-03-22
Use const generics for `CDFContextLog`

barrbrain:
The extra pointer arithmetic in `push()` is probably worth the cleaner generated code for `rollback()`.


lu-zero (approved):
Looks nice

PR #3174, opened 2023-03-23
Add clippy restrictions which we already conform to

shssoichiro:
These are allow-by-default lints which are useful for improving our code quality--we already conform to them, so this addition ensures they are checked in future changes.


barrbrain (approved):
It seems reasonable to commit to these lints given present conformance.
If we trip on them in future patches we can consider whether to allow them.

PR #3181, opened 2023-04-10
CI: Fix deploy workflow and address warnings

barrbrain:


.github/workflows/deploy.yml
```diff
@@ -95,11 +91,9 @@ jobs:
         7z e -y "$CARGO_C_FILE.zip" -o"${env:USERPROFILE}\.cargo\bin"
 
     - name: Install stable
-      uses: dtolnay/rust-toolchain@stable
+      uses: dtolnay/rust-toolchain@master
```
vibhoothi, line 94: Q, is it reliable to rely on the master branch?

barrbrain: See https://github.com/dtolnay/rust-toolchain#inputs

>  When passing an explicit toolchain as an input instead of `@rev`, you'll want to use "dtolnay/rust-toolchain@master" as the revision of the action.

vibhoothi: Great thank you

.github/workflows/rav1e.yml
```diff
@@ -368,12 +359,10 @@ jobs:
           curl -LO "$LINK/$SCCACHE_FILE.tar.gz"
           tar xzf "$SCCACHE_FILE.tar.gz"
           echo "$Env:GITHUB_WORKSPACE/$SCCACHE_FILE" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
-      - name: Install ${{ matrix.toolchain }}
+      - name: Install ${{ matrix.target }} target
         uses: dtolnay/rust-toolchain@stable
         with:
-          profile: minimal
-          toolchain: stable-${{ matrix.target }}
```
barrbrain, line 375: It turns out that this is the pattern for setting a default target that behaves as expected.

vibhoothi (approved):
Lets try to deploy this and see how it unfolds:), 

Thanks for the PR

PR #3185, opened 2023-04-17
CI: deploy library artifacts, add i686 linux and aarch64 mac targets

fdintino:
This adds artifacts to the deploy workflow for the cargo-c library and headers. I have created [a release on my fork](https://github.com/fdintino/rav1e/releases/tag/p20230414) that includes the new release artifacts.

I have held off on opening this pull request in the hope that cross would add support for custom subcommands (cross-rs/cross#716) but that doesn't seem to be happening anytime soon. I'm using a pretty ugly hack to get around this limitation: I run `cross` with a custom dockerfile whose entrypoint rewrites `cargo build` to `cargo cbuild` in the `CMD` arguments.

I have also added targets for i686 linux and aarch64 mac.

I was motivated to do this work so that I would not need to maintain cross-platform builds of rav1e when packing pillow-avif-plugin. The librav1e artifacts on my fork are now being used [here](https://github.com/fdintino/pillow-avif-plugin/blob/34c3c4b9a79dacdfba458037a1041e076a84f309/wheelbuild/config.sh#L321-L342).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -73,7 +73,7 @@ jobs:
          - conf: gnu
            name: gnu-avx2
            toolchain: stable
-           profile: release
+           profile: release-strip
            target_cpu: x86-64-v3
 
     if: github.repository_owner == 'xiph'
@@ -99,10 +99,6 @@ jobs:
       env:
         RUSTFLAGS: "-C target-cpu=${{ matrix.target_cpu }}"
       run: cargo build --profile ${{ matrix.profile }}
-
-    - name: Strip rav1e gnu-binary
-      if: matrix.conf == 'gnu'
-      run: strip target/${{ matrix.profile }}/rav1e.exe
 
     - name: Run cargo-c
       env:
@@ -161,8 +157,10 @@ jobs:
 
   linux-binaries:
     strategy:
+      fail-fast: false
       matrix:
         target:
+         - i686-unknown-linux-musl
          - x86_64-unknown-linux-musl
          - aarch64-unknown-linux-musl
         target_cpu:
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -200,96 +198,155 @@ jobs:
...
+           startsWith(github.ref, 'refs/tags/v') ||
+           github.event_name == 'schedule')
       run: |
-        TAR_FILE=rav1e-${{ steps.tagName.outputs.version }}-${{ matrix.name }}
-        cd target/${{ matrix.target }}/release
-        tar -czvf $GITHUB_WORKSPACE/$TAR_FILE.tar.gz ${{ matrix.binaries }}
-
-    - name: Upload pre-release binaries
+        mkdir -p dist/{include/rav1e,lib/pkgconfig}
+        cp target/${{ matrix.target }}/release-strip/rav1e.h dist/include/rav1e/
```
lu-zero, line 309: `cargo cinstall` wouldn't work better?

.github/workflows/deploy.yml
```diff
@@ -298,48 +355,82 @@ jobs:
...
+           startsWith(github.ref, 'refs/tags/v') ||
+           github.event_name == 'schedule')
       run: |
-        ZIP_FILE=rav1e-${{ steps.tagName.outputs.version }}-macos.zip
-        cd target/release
-        zip -9 $GITHUB_WORKSPACE/$ZIP_FILE rav1e
-
-    - name: Upload pre-release binaries
+        mkdir -p dist/{include/rav1e,lib/pkgconfig}
+        cp target/${{ matrix.target }}/release-strip/rav1e.h dist/include/rav1e/
```
lu-zero, line 419: use `cargo cinstall` instead as well

cross/entrypoint-build-libs.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+for arg in "$@"; do
+  arg=$(echo $arg | perl -pne \
+    's/(?:(?<=\s)|^)build(?=\s)/cbuild --library-type staticlib --library-type cdylib/')
```
lu-zero, line 5: this should be `cargo cinstall` I think.

lu-zero (commented):
It looks very good, if you could try to use direct `cargo cinstall` would be great.

lu-zero (approved):
If it works for you, it looks fine to me.

fdintino:
Thanks. I've just reconfirmed that these artifacts still work with the `cinstall` change.

lu-zero:
Please rebase and I'll land it immediately :)

PR #3206, opened 2023-05-20
Remove obsolete Wasm-specific code

kleisauke:
This is now part of the maybe-rayon crate introduced in PR #2914.


lu-zero (approved):
Sounds right. Thank you!

PR #3211, opened 2023-05-26
Separate and tune temporal models

barrbrain:
| PSNR Y | PSNR Cb | PSNR Cr | CIEDE2000 |   SSIM | MS-SSIM | PSNR-HVS Y | PSNR-HVS Cb | PSNR-HVS Cr | PSNR-HVS |    VMAF | VMAF-NEG |
|   ---: |    ---: |    ---: |      ---: |   ---: |    ---: |       ---: |        ---: |        ---: |     ---: |    ---: |     ---: |
| 1.1495 |  1.4326 |  0.1374 |    0.6160 | 0.0110 | -0.1630 |     0.0230 |      1.0351 |      0.0387 |  -0.0508 | -0.1188 |  -0.2709 |

### DIFF
--- a/src/api/internal.rs
+++ b/src/api/internal.rs
@@ -20,7 +20,7 @@ use crate::encoder::*;
 use crate::frame::*;
 use crate::partition::*;
 use crate::rate::{
-  RCState, FRAME_NSUBTYPES, FRAME_SUBTYPE_I, FRAME_SUBTYPE_P,
+  DynRelQ, RCState, FRAME_NSUBTYPES, FRAME_SUBTYPE_I, FRAME_SUBTYPE_P,
   FRAME_SUBTYPE_SEF,
 };
 use crate::scenechange::SceneChangeDetector;
@@ -670,7 +670,7 @@ impl<T: Pixel> ContextInner<T> {
         output_frameno,
         fti,
         self.maybe_prev_log_base_q,
-        0,
+        DynRelQ::Static,
       )
     };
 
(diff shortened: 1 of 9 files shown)


src/api/config/speedsettings.rs
```diff
@@ -92,7 +92,7 @@ impl Default for SpeedSettings {
         reduced_tx_set: false,
         // TX domain distortion is always faster, with no significant quality change,
         // although it will be ignored when Tune == Psychovisual.
-        tx_domain_distortion: true,
+        tx_domain_distortion: false,
```
shssoichiro, line 95: Is this disabled for testing, or has it become worse as a default? (If the latter, we should update or remove the above comment.)

src/api/internal.rs
```diff
@@ -1240,6 +1233,47 @@ impl<T: Pixel> ContextInner<T> {
...
+      let mut ref_log_isqrt_mean_scale = fi.ref_log_isqrt_mean_scale;
+      if output_framenos.len() > 1 {
+        let fi = &mut self
+          .frame_data
+          .get_mut(&output_framenos[1])
+          .unwrap()
+          .as_mut()
+          .unwrap()
+          .fi;
+        if keyframe && fi.frame_type != FrameType::KEY {
```
shssoichiro, line 1246: Would this ever be true?

src/rate.rs
```diff
@@ -66,19 +66,19 @@ const INTER_DELAY_TARGET_MIN: i32 = 10;
...
+  (0.9931640625 * (1 << 12) as f64) as i32,
   (1.0 * (1 << 12) as f64) as i32,
-  (1.0 * (1 << 12) as f64) as i32,
-  (1.0 * (1 << 12) as f64) as i32,
-  (1.0 * (1 << 12) as f64) as i32,
+  (1.000732421875 * (1 << 12) as f64) as i32,
+  (1.0009765625 * (1 << 12) as f64) as i32,
 ];
 
 // The ratio 33_810_170.0 / 86_043_287.0 was derived by approximating the median
```
shssoichiro, line 75: This comment will need updated

src/api/internal.rs
barrbrain: Scoping makes this condition confusing, so I should make it clearer. The second clause refers to the next output frame. It is true unless there are consecutive keyframes.

src/api/config/speedsettings.rs
barrbrain: This is primarily for testing. I think both are true at the moment because temporal RDO is disabled when transform-domain distortion is enabled. The quality change will determined by the difference between static and dynamic temporal models. Note that the static model is frame-level.

barrbrain:
This needs to be decomposed into incremental improvements. The current compound model provides some insurance for poor content-dependent behavior. We may need to improve the temporal modelling itself first.

PR #3250, opened 2023-08-07
Implement support for encoding Dolby Vision from RPU file

quietvoid:
The changes allow to encode AV1 files with Dolby Vision metadata for playback on supported devices.

As there is no official specification, the minimum I've tested is that the metadata is recognized and used correctly on playback.
The metadata OBU placement is based on #3000 (the AV1 HDR10+ specification), and I've tested it to be working fine with 2 GOPs.

From CLI, the metadata is expected to be passed as a RPU binary file, which follows the same format as encoders like `x265`.
The metadata parsing/encoding is done through the [dolby_vision](https://crates.io/crates/dolby_vision) crate.
The CLI opt can be either `--dovi-rpu` or `--dolby-vision-rpu` (used by `x265`)

From Rust, the metadata must be encoded into the final T.35 and provided for the frames that require it.

For muxing, it's possible to use the GPAC utilities but it currently requires patching for AV1: https://github.com/gpac/gpac/issues/2549
`mkvmerge` will also have support for raw OBU. IVF has to be added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,9 +83,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.71"
+version = "1.0.72"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9c7d0618f0e0b7e8ff11427422b64564d5fb0be1940354bfe2e0529b18a9d9b8"
+checksum = "3b13c32d80ecc7ab747b80c3784bce54ee8a7a0cc4fbda9bf4cda2cf6fe90854"
 
 [[package]]
 name = "aom-sys"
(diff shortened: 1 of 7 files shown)


quietvoid:
There are some issues with the current T35 code..
- With `--keyint 1`, some frames are missing metadata because the input frame number is incorrect when fetching the metadata from `t35_q`.
- With inter frames, the lookahead causes some frame invariants to be created twice for the same frame number, with the second time having `t35_metadata` overwritten since the metadata has been removed from `t35_q` on the first time.

I don't know if the changes I made are any correct but they seem to fix the issues for me.

quietvoid:
There doesn't seem to be an easy way of fixing the T35 metadata without cloning from the queue, and removing later.
The alternative would be breaking the API and fixing the `FrameInvariants` by setting `t35_metadata` prior to actually encoding the packet instead of in `build_frame_properties`.

PR #3253, opened 2023-09-11
ci: Set MACOSX_DEPLOYMENT_TARGET to 10.10 for x86_64, 11.0 for arm64

fdintino:
I noticed that rav1e is compiling against the OS X 12 SDK, whereas the binaries I distribute for pillow-avif-plugin target 10.10 and 11 for x86_64 and arm64, respectively—which is pretty common practice for python wheels. I've confirmed that cargo-c respects the `MACOSX_DEPLOYMENT_TARGET` environment variable and that the artifacts are in fact built against the desired SDKs.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -335,14 +335,19 @@ jobs:
         include:
          - target: x86_64-apple-darwin
            name: macos
+           macosx_deployment_target: "10.10"
          - target: aarch64-apple-darwin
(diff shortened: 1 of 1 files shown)


fdintino:
I'll rebase off of master to re-run the actions.

lu-zero:
Let see if the auto-rebase works as intended. I eventually fixed the CI, hopefully.

PR #3255, opened 2023-09-27
ci: Bundle LICENSE file in librav1e deploy artifacts

fdintino:
Adding the LICENSE to the library archives makes it easier for dependent projects to include it in their build artifacts.


lu-zero:
Thank you

PR #3261, opened 2023-10-03
Reduce uses of uninitialized arrays

kornelski:
I'm trying to remove UB of `Aligned::uninitialized()` — #1572

### DIFF
--- a/src/asm/shared/predict.rs
+++ b/src/asm/shared/predict.rs
@@ -39,18 +39,10 @@ mod test {
 
   fn pred_matches_inner<T: Pixel>(cpu: CpuFeatureLevel, bit_depth: usize) {
     let tx_size = TxSize::TX_4X4;
-    // SAFETY: We write to the array below before reading from it.
-    let mut ac: Aligned<[i16; 32 * 32]> = unsafe { Aligned::uninitialized() };
-    for i in 0..ac.data.len() {
-      ac.data[i] = i as i16 - 16 * 32;
-    }
-    // SAFETY: We write to the array below before reading from it.
-    let mut edge_buf: Aligned<[T; 4 * MAX_TX_SIZE + 1]> =
-      unsafe { Aligned::uninitialized() };
-    for i in 0..edge_buf.data.len() {
-      edge_buf.data[i] =
-        T::cast_from(((i ^ 1) + 32).saturating_sub(2 * MAX_TX_SIZE));
-    }
+    let ac: Aligned<[i16; 32 * 32]> = Aligned::from_fn(|i| i as i16 - 16 * 32);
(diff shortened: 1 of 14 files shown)


kornelski:
CI fails due to https://github.com/xiph/rav1e/issues/3263

kornelski:
Any comments for this PR? Is the direction good? Can it be merged despite the CI issue?

lu-zero:
Lots of fixes should land since they are good. I hope we'll fix the dav1d situation so we can have back the tests :/

src/asm/shared/transform/inverse.rs
barrbrain, line 1: If all the changes in this PR were like this file, I would approve it in a heartbeat.
This is the pattern we should be following to avoid undefined behavior and double initialization.

kornelski: I don't understand the comment. Does that mean other files than inverse.rs are not acceptable? What do I need to change?

kornelski: BTW, I've wanted to use a pattern of:

```rust
let initialized = function(&mut maybe_uninit);
```

but that required reliably tracking the length inside the function, which would regress performance. This lead me to try some loop optimizations like #3262 but I found it wasn't always possible to have guaranteed length for free. Plus there's A LOT of uninitialized to deal with.

There are more functions using which I haven't converted yet, because I'm not actually sure they fully initialize their data!

src/asm/shared/predict.rs
```diff
@@ -39,18 +39,10 @@ mod test {
...
-    let mut edge_buf: Aligned<[T; 4 * MAX_TX_SIZE + 1]> =
-      unsafe { Aligned::uninitialized() };
-    for i in 0..edge_buf.data.len() {
-      edge_buf.data[i] =
-        T::cast_from(((i ^ 1) + 32).saturating_sub(2 * MAX_TX_SIZE));
-    }
+    let ac: Aligned<[i16; 32 * 32]> = Aligned::from_fn(|i| i as i16 - 16 * 32);
+    let edge_buf: Aligned<[T; 4 * MAX_TX_SIZE + 1]> = Aligned::from_fn(|i| {
+      T::cast_from(((i ^ 1) + 32).saturating_sub(2 * MAX_TX_SIZE))
+    });
```
barrbrain, line 45: This is really nice and reason enough to add `Aligned::from_fn()`.

src/asm/shared/transform/inverse.rs
kornelski: The only cases of zero-fill (I assume you mean `from_fn(|| 0)`) are in benchmarks, outside of the benchmarked section. I assume they're not important, since they're neither in the library, nor performance-critical. So I've just changed them for simplicity and determinism.

barrbrain (approved):
Looks good to me. Sorry that I missed the zero-fill changes were limited to tests and benchmarks.

PR #3262, opened 2023-10-04
Optimized diff iterator

kornelski:
Hoisted checks that before the loop


kornelski:
It looks like the `libdav1d` hack with an upgrade-in-place to Ubuntu Lunar doesn't work any more, and it won't upgrade `libc6`.

lu-zero:
the set doesn't seem bad though.

kornelski:
What do you mean?

kornelski:
Included in #3261

PR #3269, opened 2023-10-18
Integrate lossless inverse transform assembly functions

barrbrain:


lu-zero (approved):
It seems fine, I now wonder if the various dispatching functions could be marked inline.

PR #3270, opened 2023-10-19
Integrate lossless forward transform intrinsics

barrbrain:
* Introduce `TxType::WHT_WHT` and so unify the `forward_transform` and `inverse_transform_add` functions.
* Plumb `fwht4` through the existing forward transform code and gain an `AVX2` implementation.
* Improvements to round trip and  transform tests
* Found and fixed a bug with aarch64 assembly integration

### DIFF
--- a/src/transform/mod.rs
+++ b/src/transform/mod.rs
@@ -485,9 +485,8 @@ mod test {
     let coeff_area: usize = av1_get_coded_tx_size(tx_size).area();
     let mut src_storage = [T::cast_from(0); 64 * 64];
     let src = &mut src_storage[..tx_size.area()];
-    // dynamic allocation: test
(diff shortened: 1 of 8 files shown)


lu-zero:
Why did we lose a test? that codepath isn't really exercised by our encode-decode tests?

src/transform/mod.rs
```diff
@@ -557,6 +557,7 @@ mod test {
 
   fn roundtrips<T: Pixel>() {
     let combinations = [
+      (TX_4X4, WHT_WHT, 0),
```
barrbrain, line 560: This replaces `test_lossless_roundtrip()` as it can now be expressed by `test_roundtrip()`.

lu-zero (approved):

PR #3271, opened 2023-10-21
Reduce risk of uninit ac in luma_ac and encode_coeffs

kornelski:
## luma_ac

In various `pred_cfl_ac` I've added checks that `ac[..].len()` must be exactly as long as needed, i.e. `plane_bsize.width() * plane_bsize.height()`. The multiplication got repetitive, so I've added `plane_bsize.area()`.

Those prediction functions are in assembly. I haven't checked whether they initialize exactly `plane_bsize.area()` and nothing else, but I don't see a reason why they wouldn't, it would be a bug otherwise. Anyway, it's not a safety issue in pratice, since the allocated memory behind `ac` is the same size as before.

## encode_coeffs

`get_nz_map_contexts` initializes context up to `eob` which is `u16`. Other code used `usize` for `eob`. To be type-safe sure it can't be longer than `u16` anywhere (therefore reading past what `get_nz_map_contexts` could init), I've changed `eob` to be consistently `u16` everywhere. It's never more than 1024 in reality.

`get_nz_map_contexts` used to write `coeffs[scan[i]]`, but that had two issues:

1. It could not guarantee that all coeffs are initialized *and contiguous* to make an initialized slice.
2. It was unnecessary performance cost. Writing at `scan`'s positions was a needless step, because they were also always read as `coeffs[scan[i]]`. Since all uses of the `coeffs` wanted a coefficient for `scan[i]`, I've made them get it from `coeffs[i]` rather than `coeffs[scan[i]]`. `scan` acted only as an unnecessary shuffle. This is faster, guarantees a contiguous slice, and does not change the behavior.

### DIFF
--- a/src/asm/aarch64/predict.rs
+++ b/src/asm/aarch64/predict.rs
@@ -30,7 +30,7 @@ macro_rules! decl_cfl_ac_fn {
     extern {
       $(
         fn $f(
-          ac: *mut i16, src: *const u8, stride: libc::ptrdiff_t,
+          ac: *mut MaybeUninit<i16>, src: *const u8, stride: libc::ptrdiff_t,
           w_pad: libc::c_int, h_pad: libc::c_int,
           width: libc::c_int, height: libc::c_int,
         );
@@ -50,7 +50,7 @@ macro_rules! decl_cfl_ac_hbd_fn {
     extern {
       $(
         fn $f(
-          ac: *mut i16, src: *const u16, stride: libc::ptrdiff_t,
+          ac: *mut MaybeUninit<i16>, src: *const u16, stride: libc::ptrdiff_t,
           w_pad: libc::c_int, h_pad: libc::c_int,
           width: libc::c_int, height: libc::c_int,
         );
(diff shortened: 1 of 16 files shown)


lu-zero:
It looks very nice, but I'd like to have @barrbrain opinion :)

kornelski:
Thanks for x86 fixes. As you can see, I'm non ARM, and conditionally-compiled x86 is a blind spot for me.

vibhoothi (commented):
maybe doing a quick AWCY run is a good idea too:)

kornelski:
I don't expect this to change outputs at all. Hopefully tests already cover this, but you can verify by running compression and check file hashes before/after.

barrbrain:
> maybe doing a quick AWCY run is a good idea too:)

Unfortunately, we have build issues on AWCY presently.
I have confirmed unchanged output on  test sequence at default settings:
```bash
ffmpeg -loglevel error -f lavfi -i testsrc=duration=10:size=320x240:rate=30 \
  -pix_fmt yuv420p -f yuv4mpegpipe - | rav1e -o >(sha256sum) -y -
```

barrbrain (approved):
The changes relating to scan order required close review. Everything else looked good to me at a cursory glance. Thank you for this effort.

PR #3273, opened 2023-10-23
Consistent setup for round trip and inverse transform tests

barrbrain:
This inconsistency was identified while trying to integrate and test WHT functions. We have an unnecessary allocation in one case and a mismatch in stride in the other case. This change will give terse error messages when refactoring inverse transform code.


barrbrain:
Included in #3270.

PR #3274, opened 2023-10-23
Move WHT inverse transforms to common lookup table for `x86_64` and `aarch64`

barrbrain:
* Explicitly elide bounds checks for `TxSize` and `TxType` as array indices.
* Expand transform function tables to make room for `WHT_WHT` implementations.
* Replace branch for `WHT_WHT` with table lookup for `aarch64` and `x86_64`.

Note: The `x86_64` table generation code is getting unwieldy.


lu-zero (approved):
Seems nice, I'm unsure if implementing IndexMut would make the code even simpler, but since it is used only in macros, I guess won't change much the end result.

PR #3276, opened 2023-10-25
Initialize edge buf

kornelski:
Continuation of #1572.

https://github.com/xiph/rav1e/blob/master/src/partition.rs#L624

Right-aligned pixels suggests that not all of the `left` pixels are always initialized, so `get_intra_edges` was returning `edge_buf` without initializing all of its elements. That's technically UB, and an example of [a safety comment being misleading about safety](https://github.com/xiph/rav1e/blob/3d43ecac0145688ac9d17712e685e80ee4eefbe2/src/partition.rs#L617).

I don't think I could simply change `edge_buf` to be a contiguous slice of pixels (of variable width) or something like a `PlaneRegion`, because in `dispatch_predict_intra` there's assembly that seems to rely on `edge_buf` having a particular layout. It's a bit scary to change `edge_buf` layout, since code elsewhere has assumptions like `let edge_ptr = edge_buf.data.as_ptr().offset(2 * MAX_TX_SIZE as isize) as *const _;`.

So if `edge_buf` can't be contiguous, it can't have uninitialized elements left, so I've just initialized it. AFAIK that's 256 zeroed bytes. I took it out of `get_intra_edges` to avoid initializing it every time inside the function, so the initialization can be amortized in the outer scope.

I think longer term it could be worth trying to make `edge_buf` a newtype or a contiguous array, and change assembly to match, but that's a bit more than I can do now.

### DIFF
--- a/src/api/lookahead.rs
+++ b/src/api/lookahead.rs
@@ -1,6 +1,6 @@
 use crate::api::internal::InterConfig;
 use crate::config::EncoderConfig;
-use crate::context::{BlockOffset, FrameBlocks, TileBlockOffset};
+use crate::context::{BlockOffset, FrameBlocks, TileBlockOffset, MAX_TX_SIZE};
 use crate::cpu_features::CpuFeatureLevel;
 use crate::dist::get_satd;
 use crate::encoder::{
@@ -12,6 +12,7 @@ use crate::partition::{get_intra_edges, BlockSize};
 use crate::predict::{IntraParam, PredictionMode};
 use crate::tiling::{Area, PlaneRegion, TileRect};
 use crate::transform::TxSize;
+use crate::util::Aligned;
 use crate::Pixel;
 use rayon::iter::*;
 use rust_hawktracer::*;
@@ -44,6 +45,8 @@ pub(crate) fn estimate_intra_costs<T: Pixel>(
   let w_in_imp_b = plane.cfg.width / IMPORTANCE_BLOCK_SIZE;
   let mut intra_costs = Vec::with_capacity(h_in_imp_b * w_in_imp_b);
 
+  let mut edge_buf = Aligned::new([T::zero(); 4 * MAX_TX_SIZE + 1]);
+
   for y in 0..h_in_imp_b {
     for x in 0..w_in_imp_b {
       let plane_org = plane.region(Area::Rect {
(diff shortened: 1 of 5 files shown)


src/api/lookahead.rs
```diff
@@ -44,6 +45,9 @@ pub(crate) fn estimate_intra_costs<T: Pixel>(
   let w_in_imp_b = plane.cfg.width / IMPORTANCE_BLOCK_SIZE;
   let mut intra_costs = Vec::with_capacity(h_in_imp_b * w_in_imp_b);
 
+  let mut edge_buf: Aligned<[T; 4 * MAX_TX_SIZE + 1]> =
+    Aligned::from_fn(|_| T::zero());
```
barrbrain, line 49: There is a terse pattern for fully initialized `Aligned<[T; N]>`:
```suggestion
  let mut edge_buf = Aligned::new([T::zero(); 4 * MAX_TX_SIZE + 1]);
```

barrbrain:
With the backing buffer passed in, it could be uninitialized and we may return a new struct of `{ left: &[T], topleft: &T, top: &[T] }` for consumption by `predict_intra()`.

barrbrain:
After poking around, my major insight is that if we modify `get_intra_edges` to always initialize `top_left`, then the initialized region is always contiguous and contains the anchor reference for the assembly.

src/partition.rs
```diff
@@ -833,7 +829,6 @@ pub fn get_intra_edges<T: Pixel>(
       }
     }
   }
```
barrbrain, line 831:
```suggestion
    }
    (&left[left_uninit..], top_left, &above[..above_init])
  }
```

barrbrain:
I made an attempt in #3277 to identify the plumbing required for an `IntraEdge` type.

barrbrain:
> I made an attempt in #3277 to identify the plumbing required for an `IntraEdge` type.

I made a few wrong turns but eventually found an ergonomic type for `edge_buf`.

> I think longer term it could be worth trying to make `edge_buf` a newtype or a contiguous array, and change assembly to match, but that's a bit more than I can do now.

The good news is that we can have tuple struct of slice references backed by a contiguous array region and maintain the layout requirements of the assembly. :tada:

I will try to prepare #3277 for review, in preference to this intermediate change. Thank you for highlighting the issues to be resolved.

barrbrain:
Superseded by #3277, which has merged.

PR #3277, opened 2023-10-26
Well-defined lazy initialization for `get_intra_edges`

barrbrain:
Inspired by #3276. Introduce new types to capture the requirements of intra edge buffers and avoid undefined behavior.
Analysis from @kornelski quoted below:

```rust
// left pixels are ordered from bottom to top and right-aligned
let (left, not_left) = edge_buf.data.split_at_mut(2 * MAX_TX_SIZE);
let (top_left, above) = not_left.split_at_mut(1);
```

> Right-aligned pixels suggests that not all of the `left` pixels are always initialized, so `get_intra_edges` was returning `edge_buf` without initializing all of its elements. That's technically UB, and an example of [a safety comment being misleading about safety](https://github.com/xiph/rav1e/blob/3d43ecac0145688ac9d17712e685e80ee4eefbe2/src/partition.rs#L617).
> 
> I don't think I could simply change `edge_buf` to be a contiguous slice of pixels (of variable width) or something like a `PlaneRegion`, because in `dispatch_predict_intra` there's assembly that seems to rely on `edge_buf` having a particular layout. It's a bit scary to change `edge_buf` layout, since code elsewhere has assumptions like `let edge_ptr = edge_buf.data.as_ptr().offset(2 * MAX_TX_SIZE as isize) as *const _;`.
> 
> So if `edge_buf` can't be contiguous, it can't have uninitialized elements left, so I've just initialized it. AFAIK that's 256 zeroed bytes. I took it out of `get_intra_edges` to avoid initializing it every time inside the function, so the initialization can be amortized in the outer scope.
> 
> I think longer term it could be worth trying to make `edge_buf` a newtype or a contiguous array, and change assembly to match, but that's a bit more than I can do now.

### DIFF
--- a/benches/predict.rs
+++ b/benches/predict.rs
@@ -12,25 +12,19 @@ use rand::{Rng, SeedableRng};
 use rand_chacha::ChaChaRng;
 use rav1e::bench::cpu_features::CpuFeatureLevel;
 use rav1e::bench::frame::*;
-use rav1e::bench::partition::BlockSize;
+use rav1e::bench::partition::{BlockSize, IntraEdge};
 use rav1e::bench::predict::*;
 use rav1e::bench::transform::TxSize;
 use rav1e::bench::util::*;
 
 pub const BLOCK_SIZE: BlockSize = BlockSize::BLOCK_32X32;
 
-pub fn generate_block<T: Pixel>(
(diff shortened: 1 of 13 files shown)


lu-zero (approved):
Probably fine, I'm debating if it is better to hide the buffer allocation in the function and return (buffer, ref).

PR #3311, opened 2023-12-18
CI: Fix artifact names for manual pre-release tag

barrbrain:
Follow-up to #3309 and #3308, which necessitated renaming some artifacts.


barrbrain:
See https://github.com/barrbrain/rav1e/actions/runs/7250793422/job/19752191241 for a test run on a fork.

lu-zero (approved):

PR #3316, opened 2023-12-28
Prepare for release

lu-zero:


barrbrain (approved):

MarijnS95:
Thanks! Can we also get a release of `ivf` to get rid of an older `bitstream-io 0.1` dependency?

It looks like the crate was already [bumped to `0.1.2`](https://github.com/xiph/rav1e/commit/0a4eb807b327cdf7589cd54d85ef608eb959b2d0) before [that `bitstream-io 0.2` upgrade happened](https://github.com/xiph/rav1e/commit/1bbe17850fd154ba17e72b36e25a79079d6c2cf6), but was never pushed to crates.io either :)

PR #3320, opened 2023-12-29
Simplify fast scenechange code a bit

FreezyLemon:
`downscaled_frame_buffer` is an `Option<([Plane<T>; 2], bool)>`. The bool is called `is_initialized` and is supposed to show whether the memory in the `[Plane<T>; 2]` is initialized or not.

As far as I can tell, there is no code path that results in the Planes being uninitialized: `downscaled_frame_buffer` starts out as a `None`, the first buffer access allocates & initializes the memory, and the memory should always be initialized after that. The bool never seems to be `false` from what I can tell, which would support my assumption, I think.

There's also a field called `frame_ref_buffer` which caches the `Arc<Frame<T>>` when no downscaling is being done. Unless I am missing something, there's no real benefit to caching these pointers (because refs to both frames will always be passed into the function anyways), so i removed the buffer.

I tested my changes to the best of my ability, and didn't notice any change in behaviour or performance. LMK what you think

### DIFF
--- a/src/scenechange/fast.rs
+++ b/src/scenechange/fast.rs
@@ -24,40 +24,23 @@ impl<T: Pixel> SceneChangeDetector<T> {
   ) -> ScenecutResult {
     if let Some(scale_func) = &self.scale_func {
       // downscale both frames for faster comparison
-      if let Some((frame_buffer, is_initialized)) =
-        &mut self.downscaled_frame_buffer
+      if let Some(frame_buffer) = &mut self.downscaled_frame_buffer
       {
-        let frame_buffer = &mut *frame_buffer;
-        if *is_initialized {
-          frame_buffer.swap(0, 1);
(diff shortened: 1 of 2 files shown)


barrbrain:
@shssoichiro What are the implications for https://github.com/rust-av/av-scenechange/ ?

shssoichiro:
It looks like this shouldn't affect the scene detection code in either av-scenechange or av1an, as far as I can tell.

shssoichiro:
As for why the code is like this, at one point in time I believed there to be cases where we may be re-running a comparison on the same pair of frames twice. However, it's possible that is no longer true (or never was true) if there is no measurable difference in performance. This shouldn't affect behavior either way, as it was intended to be a performance optimization, but I'm okay with removing it.

shssoichiro (approved):

PR #3322, opened 2023-12-30
Remove duplicate assume_slice_init_mut

FreezyLemon:
The two functions are equivalent except for an additional `T: Copy` bound on one function. That bound is always fulfilled at the moment, so I left it as-is.

I preferred the name `slice_assume_init_mut` over the other one because [std uses the same name](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.slice_assume_init_mut).


shssoichiro:
Sad that the std version is nightly only.

shssoichiro (approved):

PR #3323, opened 2023-12-31
Fix test-specific clippy lints

shssoichiro:
These aren't checked for in CI, but it's not a bad idea to fix them anyway.


barrbrain (approved):
Looks fine to me. Full `clippy` checks could be set up but I think that is non-trivial.

PR #3324, opened 2023-12-31
Ignore IDE config folders

shssoichiro:
These include developer-specific IDE customizations and generally aren't intended to be committed.


barrbrain (approved):
These are the most popular IDEs for Rust. Neovim with rust-analyzer doesn't leave cruft by default. 😃

PR #3327, opened 2024-01-01
Remove rustc_version_check from build.rs

FreezyLemon:
This was added in #1552 but is no longer necessary. The cfg value `nightly_rustc` does not seem to be used anywhere, and [rust-version](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) in Cargo.toml allows a built-in way of checking the MSRV.

`rust-version` is only supported in toolchains >=1.56, but all older toolchains will fail compilation for other reasons (mostly edition = "2021").

Examples:
```
$ cargo +1.55 c
error: failed to parse manifest at `/.../Development/rav1e/Cargo.toml`

Caused by:
  feature `edition2021` is required
```

```
$ cargo +1.60 c
error: package `clap_complete v4.4.5` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.60.0
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1311,7 +1311,6 @@ dependencies = [
  "quickcheck_macros",
  "rand",
  "rand_chacha",
- "rustc_version",
  "scan_fmt",
  "semver",
  "serde",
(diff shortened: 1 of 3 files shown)


lu-zero (approved):

lu-zero:
Thank you!

PR #3328, opened 2024-01-01
Remove crate `quickcheck_macros`

FreezyLemon:
This is the last dependency that requires version 1 of syn (see #3254):

```
$ cargo tree | grep -c "syn v1"
0
```

`quickcheck_macros` isn't actually required, and `quickcheck` is already on version 2 of syn. The code change that is needed to remove `quickcheck_macros` is very small, so I don't see a reason not to do it.


shssoichiro (approved):
Makes sense

PR #3332, opened 2024-01-05
Remove `allow(clippy::undocumented_unsafe_blocks)` where not necessary

FreezyLemon:
There were two bugs ([1](https://github.com/rust-lang/rust-clippy/issues/9142), [2](https://github.com/rust-lang/rust-clippy/issues/8264)) that made these necessary. Both bugs are fixed in 1.70, so the `allow`s can be removed now.


shssoichiro (approved):
Thanks!

PR #3333, opened 2024-01-05
Don't use host's strip command when cross-compiling

kornelski:
On macOS, `cargo build --target=x86_64-unknown-linux-musl` makes rustc panic in compiler/rustc_codegen_llvm/src/back/archive.rs:397:67, because the `librav1easm.a` file is invalid and unusable. This is because running aarch64-darwin's `strip` command run on an x86-64-musl archive ruins it.

I've made rav1e not run `strip` during cross-compilation, unless `CARGO_TARGET_{target}_STRIP` or `STRIP` env var is set.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -51,9 +51,9 @@ fn hash_changed(
     }
   }
 
-  let strip = env::var("STRIP").unwrap_or_else(|_| "strip".to_string());
-
-  hasher.write(strip.as_bytes());
+  if let Some(cmd) = strip_command() {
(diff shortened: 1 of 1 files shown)


shssoichiro:
Looks good. Could you please run `cargo fmt` and push to make the CI happy?

lu-zero:
I wonder if would be possible to tell nasm to not emit those symbols to begin with now. The patch seems fine.

lu-zero (approved):

PR #3334, opened 2024-01-05
Replace cdf! macro with const fns

FreezyLemon:
Notable changes:
- `cdf!` always required the parameter count to be `== CDF_LEN - 1`, the new `cdf` function accepts any parameter count `< CDF_LEN`. Any values after the last parameter will be zero. (e.g. if the output has a length of 8 and the parameter count is 5, then the output values 6 to 8 will be zero). This means:
  ```rust
  let test: [u16; 4] = cdf!(123); // Error: Number of parameters must be exactly (4 - 1)
  let test: [u16; 4] = cdf([123]); // Works: test == [some_value, 0, 0, 0]
  ```
- Added helper functions for multi-dimensional arrays (up to 5D) to avoid spamming the `cdf` macro/fn
  ```rust
  [
    cdf!(123, 456, 789),
    cdf!(987, 654, 321),
    // ...
  ]
  ```

  will become

  ```rust
  cdf_2d([
    [123, 456, 789],
    [987, 654, 321],
    // ...
  ])
  ```

  and so on.
- Check all input parameters to be in the range `[0, 32768]` to avoid underflows in `32768 - val`
- Fully replaced macro with generic code. Type safety is ensured through the function signatures and compile-time `assert!` statements

The code for `cdf_2d`, `cdf_3d` etc. is a little messy. But it's the best way I found to keep these `const` without some sort of macroized implementation (which seems a bit overkill for just 5 similar functions).


lu-zero (approved):
Looks very good! Thank you

PR #3346, opened 2024-01-27
Add whitespace to compiler flags in long version string

FreezyLemon:
From the [reference regarding environment variables](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) set by Cargo:

> **CARGO_ENCODED_RUSTFLAGS** — extra flags that Cargo invokes rustc with, separated by a 0x1f character (ASCII Unit Separator).

Before:
```
Compiler Flags: -Ctarget-cpu=native-Clink-arg=-fuse-ld=mold
```

After:

```
Compiler Flags: -C target-cpu=native -C link-arg=-fuse-ld=mold
```

### DIFF
--- a/src/bin/common.rs
+++ b/src/bin/common.rs
@@ -273,10 +273,14 @@ fn get_version() -> &'static str {
 
 fn get_long_version() -> &'static str {
   static LONG_VERSION_STR: Lazy<String> = Lazy::new(|| {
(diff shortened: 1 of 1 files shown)


shssoichiro:
Interesting. tbh I didn't realize the previous version didn't even have a space between the different flags.

shssoichiro (approved):

PR #3352, opened 2024-02-20
Make apply_ssim_boost a const fn

FreezyLemon:
The `ILog` trait was introduced in v_frame before `{integer}::ilog` became stable. `const fn` doesn't do much here, but it's simple to do, so one might as well.

Note that the functions are similar but not equal in behavior:
`ILog::ilog` is effectively

```rs
if x == 0 {
  0
} else {
  x.ilog2() + 1
}
```

`x.ilog2()` will panic if `x == 0`, but that is already assumed to not be the case here (and it looks like the check for zero is optimized away in all callsites).

### DIFF
--- a/src/activity.rs
+++ b/src/activity.rs
@@ -106,11 +106,11 @@ struct RsqrtOutput {
 }
 
 /// Fixed point `rsqrt` for `ssim_boost`
-fn ssim_boost_rsqrt(x: u64) -> RsqrtOutput {
+const fn ssim_boost_rsqrt(x: u64) -> RsqrtOutput {
   const INSHIFT: u8 = 16;
   const OUTSHIFT: u8 = 14;
 
-  let k = ((ILog::ilog(x) - 1) >> 1) as i16;
+  let k = (x.ilog2() >> 1) as i16;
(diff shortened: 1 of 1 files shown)


shssoichiro (approved):

src/activity.rs
```diff
@@ -106,11 +106,11 @@ struct RsqrtOutput {
 }
 
 /// Fixed point `rsqrt` for `ssim_boost`
-fn ssim_boost_rsqrt(x: u64) -> RsqrtOutput {
+const fn ssim_boost_rsqrt(x: u64) -> RsqrtOutput {
   const INSHIFT: u8 = 16;
   const OUTSHIFT: u8 = 14;
 
-  let k = ((ILog::ilog(x) - 1) >> 1) as i16;
```
shssoichiro, line 113: Just making sure, is the removal of the `- 1` intentional here? My guess is yes given the docs in v_frame: `floor(log2(self)) + 1, or 0 if self == 0`, but want to be certain this is bit-identical to the prior output.

FreezyLemon: Yes, it's intentional and I did make sure they're equivalent for x != 0: https://rust.godbolt.org/z/Gh5TxbEYK

shssoichiro (approved):

PR #3354, opened 2024-02-25
Remove redundant imports

FreezyLemon:
These result in warnings on [1.78](https://github.com/rust-lang/rust/pull/117772) (current nightly). For example:

```rs
warning: the item `Pixel` is imported redundantly
  --> src/bin/stats.rs:14:21
   |
10 | use av_metrics::video::*;
   |     -------------------- the item `Pixel` is already imported here
...
14 | use rav1e::{Packet, Pixel};
   |                     ^^^^^
```


barrbrain (approved):
Most of these are clearly redundant from the local context.

PR #3359, opened 2024-03-10
Update dependencies which do not bump minimum rustc

shssoichiro:


Cargo.lock
```diff
@@ -251,15 +260,15 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.14.0"
+version = "3.15.4"
```
FreezyLemon, line 263: This breaks the WASM build. Bumpalo 3.15.0 [changed the MSRV to 1.73.](https://github.com/fitzgen/bumpalo/blob/main/CHANGELOG.md#3150)

FreezyLemon (changes requested):

Cargo.lock
shssoichiro: I would be really happy if more Rust maintainers would consider a compiler version change as breaking... anyway I guess I'll add a CI step for wasm, looks like we're missing them.

FreezyLemon:
BTW I recommend using [cargo-hack](https://github.com/taiki-e/cargo-hack) to validate MSRV (and other things):
`cargo +1.70 hack --each-feature check`

Might be worth adding something to CI, although a job like that would take very long to complete...

shssoichiro:
We do have CI jobs for validating minimum rust version. Just not one for wasm currently. hack might be useful for local testing before pushing up to CI.

Cargo.lock
FreezyLemon: MSRV bump being considered semver breaking or not seems to be a very contentious issue..
It would be a lot better if cargo update honored the MSRV, I think a lot of issues would be inherently solved by that

shssoichiro:
It looks like adding wasm tests in CI here may be a bit more complicated than in v_frame. But I'm also not an expert at Github actions. Anyway, for now I'll create an issue to follow up on adding wasm tests to CI, so we don't forget it.

FreezyLemon:
On second thought, maybe add a comment in the Cargo.toml that clap is held back for MSRV reasons?

lu-zero (approved):

lu-zero:
Meh, let's bump to 1.74 and be done with that :)

PR #3361, opened 2024-03-10
Add new NASM pragma regarding macro parameter expansion

### LINKED ISSUE #3330: NASM warnings about empty macro parameters
**Describe the bug**
There are multiple NASM warnings when compiling rav1e with a recent version of NASM (>=2.16, maybe 2.15):

```
cdef_rav1e.asm:262: warning: dropping trailing empty parameter in call to multi-line macro `ACCUMULATE_TAP' [-w+pp-macro-params-legacy]
```

The [documentation](https://www.nasm.us/doc/nasmdoca.html) has some more information on this warning:

> pp-macro-params-legacy warns about multi-line macros being invoked with the wrong number of parameters, but for bug-compatibility with NASM versions older than 2.15, NASM tried to fix up the parameters to match the legacy behavior and call the macro anyway. This can happen in certain cases where there are empty arguments without braces, sometimes as a result of macro expansion.
> 
> The legacy behavior is quite strange and highly context-dependent, and can be disabled with:
> 
>      %pragma preproc sane_empty_expansion true
> 
> It is highly recommended to use this option in new code.

Some of these warnings do exist because of other macros expanding, so it would be hard to fix these warnings by changing the macro. I think adding this pragma would make sense, but it is only available on NASM >=2.15, so it would require a bump in the minimum version.

**Required Information**
NASM version: 2.16.01

FreezyLemon:
Fixes #3330.

This raises the minimum NASM version to 2.15.0 which was released in [June 2020](https://repo.or.cz/nasm.git/tag/refs/tags/nasm-2.15). The minimum Rust version 1.70 was released in [June 2023](https://releases.rs/docs/1.70.0/), so I don't think this is a problem.

I also spent a bit of time refactoring the build script slightly.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -135,20 +135,17 @@ fn build_nasm_files() {
   if let Some((hash, hash_path)) =
     hash_changed(asm_files, &out_dir, &dest_path)
   {
-    let mut config_include_arg = String::from("-I");
-    config_include_arg.push_str(&out_dir);
-    config_include_arg.push('/');
-    let mut nasm = nasm_rs::Build::new();
-    nasm.min_version(2, 15, 0);
-    for file in asm_files {
-      nasm.file(file);
-    }
-    nasm.flag(&config_include_arg);
-    nasm.flag("-Isrc/");
-    let obj = nasm.compile_objects().unwrap_or_else(|e| {
-      println!("cargo:warning={e}");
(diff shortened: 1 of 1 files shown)


shssoichiro (approved):
Makes sense to me

PR #3367, opened 2024-03-13
Fix build on wasm32-unknown-unknown

FreezyLemon:
- `crate-type = "cdylib"` is needed for `wasm-pack` to work
- There is no need for a `wasm` feature, everything can be auto-detected (`-F wasm` is now a no-op)
- `wasm-bindgen` should only be included for `wasm32-unknown-unknown`. `wasm32-wasi` does not need it. I'm not sure about emscripten

`wasm-pack build` and other commands should now work. `--no-default-features` is necessary because the `binaries` feature does not compile on wasm.

Note that testing on wasm32-unknown-unknown can't just be done with `cargo test` and requires more work. I'll do that in a separate PR (incl. CI).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,10 @@ tracing = [
 ]
 scenechange = []
 serialize = ["serde", "toml", "v_frame/serialize", "serde-big-array", "av1-grain/serialize"]
-wasm = ["wasm-bindgen"]
+
+# Left here for compatibility reasons.
+# Does not do anything, the wasm target is auto-detected.
+wasm = []
 
 # Enables debug dumping of lookahead computation results, specifically:
 # - i-qres.png: quarter-resolution luma planes,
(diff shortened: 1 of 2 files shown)


FreezyLemon:
This seems to break fuzzing somehow. I'll take a look and try to fix it

Cargo.toml
```diff
@@ -167,6 +173,8 @@ bench = false
 
 [lib]
 bench = false
+# cdylib is required for wasm32-unknown-unknown
```
lu-zero, line 176: This is potentially annoying, https://github.com/rustwasm/wasm-pack/issues/1297 should fix it.

lu-zero (commented):
I'd wait for https://github.com/rustwasm/wasm-pack/issues/1297 to get addressed and/or check if the unreleased wasm-pack doesn't address it otherwise.

I do not consider acceptable to add cdylib unconditionally.

FreezyLemon:
I see. wasm-pack's master branch doesn't seem to include any changes that would help work around this problem. Not much to do except wait.

I'll close this. When this is fixed in wasm-pack, it should be easy to reimplement.

lu-zero:
https://crates.io/crates/trunk might be an alternative.

PR #3368, opened 2024-03-14
Add CI jobs for `wasm32-wasi`

FreezyLemon:


lu-zero (approved):

FreezyLemon:
It looks like `wasm32-wasi` is going to be renamed to `wasm32-wasip1` ("WASI Preview 1") at some point. It's possible that targetting `wasm32-wasi` will fail at some point in the future.

PR #3371, opened 2024-03-29
Replace AlignedBoxedSlice with aligned_vec::ABox

### LINKED ISSUE #3370: AlignedBoxedSlice is unsound
`AlignedBoxedSlice<T>` is unsound in multiple places:

1. The safety invariants of `Layout::from_size_align_unchecked` are not always fulfilled: The `size` parameter can overflow `isize` for some lengths
2. `NonNull::new_unchecked(alloc(...))`:  `alloc` can return a null pointer on allocation failure
3. The initialization code in `AlignedBoxedSlice::new` creates a mutable slice from uninitialized memory (instead of ptr::write or `MaybeUninit`)

Pretty much the same problems [also existed in v_frame](https://github.com/rust-av/v_frame/issues/14). Those were fixed, more issues were found and the implementation was ultimately replaced and `aligned-vec` was used instead. I think it could be used in rav1e too. Maybe `Aligned<T>` and the entirety of `align.rs` could be removed then because it also has an [unsound function](https://github.com/xiph/rav1e/issues/1572#issuecomment-575329601).

FreezyLemon:
Closes #3370.

`ABox<[T]>` generally works the same way as `AlignedBoxedSlice<T>`, but uses a different alignment by default:

AlignedBoxedSlice: 8 on wasm, 64 everywhere else
ABox: Generally two cachelines, 64 bytes for unknown archs ([exact rules](https://github.com/sarah-ek/aligned-vec/blob/b52b2e78d940cda6622f99ffb8deb4c505e2635b/src/lib.rs#L30))

I am pretty sure the wasm allocation bug mentioned in the code comment has been [fixed](https://github.com/rust-lang/rust/pull/70585), so using a higher alignment should be fine on the wasm32 arch.

It's possible to configure aligned_vec to reproduce the old behaviour exactly, but I think it's probably better to stick to the crate defaults.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "aligned-vec"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4aa90d7ce82d4be67b64039a3d588d38dbcc6736577de4a847025ce5b0c468d1"
+
+[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1256,6 +1262,7 @@ dependencies = [
 name = "rav1e"
 version = "0.7.0"
 dependencies = [
+ "aligned-vec",
  "aom-sys",
  "arbitrary",
  "arg_enum_proc_macro",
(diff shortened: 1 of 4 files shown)


barrbrain (approved):
The `aligned_vec` crate doesn't appear to be popular enough to have made its way into distributions yet. As it is a transitive dependency via `v_frame`, this is no worse than the current state.

PR #3383, opened 2024-05-25
Update nasm-rs to 0.3.0

FreezyLemon:
nasm-rs versions before 0.3 ignored the number of jobs set by cargo's job server. This means that even with `cargo b -j1` or `NUM_JOBS=1 cargo b`, the rav1e build script uses 100% CPU.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -795,9 +795,9 @@ dependencies = [
 
 [[package]]
 name = "jobserver"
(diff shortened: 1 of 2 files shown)


FreezyLemon:
Also bumped `jobserver` while I'm at it. It has some small safety improvements.

shssoichiro (approved):

PR #3384, opened 2024-05-31
Disable compiler warnings for known custom `cfg` values

FreezyLemon:
[Context](https://blog.rust-lang.org/2024/05/06/check-cfg.html).

TL;DR: Custom `cfg`s are considered unexpected by `rustc`, so they will produce a warning (on 1.80+) unless they're added to a list of known custom cfgs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -162,6 +162,11 @@ rand_chacha = "0.3"
 [target.'cfg(any(decode_test, decode_test_dav1d))'.dependencies]
 system-deps = "6"
 
+[lints.rust.unexpected_cfgs]
+level = "warn"
+# These custom cfgs are expected, so tell rustc not to output warnings for them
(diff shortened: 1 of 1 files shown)


shssoichiro:
I think this looks fine, but I'm curious, is this not solvable in some way other than ignoring? For example, can we fix this in our [build.rs](https://blog.rust-lang.org/2024/05/06/check-cfg.html#custom-cfgs-in-build-scripts)?

FreezyLemon:
Yes that's possible too, but it looks like the [lints] entry is meant for cfgs that are statically known while the build.rs approach is more meant for dynamic addition of cfgs. At least it reads that way to me, but I really don't have a preference one way or the other

lu-zero (approved):
Overall I like the Cargo.toml better since it is easier to discover

PR #3387, opened 2024-06-20
Cleanup custom lints

shssoichiro:
- Move custom lints from lib.rs/main.rs into Cargo.toml
- Remove most custom lints, as this is overkill and difficult to maintain
- Fix new lints from most recent Rust version


lu-zero (approved):

lu-zero:
Looks like few more need to be addressed

PR #3398, opened 2024-09-14
Drop `binary` integration test from coverage

### LINKED ISSUE #3394: Fix or remove grcov-coverage CI job
The CI job for code coverage has been broken for quite some time. It is allowed to fail, but still annoying to show as a red X on every PR. We should either fix the job or remove it.

Example: https://github.com/xiph/rav1e/actions/runs/10392532069/job/28780308261

FreezyLemon:
Should fix #3394.


FreezyLemon:
Had to include a clippy fix so the job actually runs.

lu-zero (approved):

lu-zero:
Thank you!

PR #3407, opened 2024-12-23
Bump msrv-compatible dependencies

lu-zero:


barrbrain (approved):
Thank you for this. I will find a less brittle solution for arm64 tests.

lu-zero:
We can test on `macos-latest` iirc...

PR #3424, opened 2025-04-20
Preparation for rav1e 0.8.0

lu-zero:


shssoichiro (approved):
Nice. LGTM!

lu-zero:
We can consider this the branch point or fix more optional code that might not work now.

PR #3430, opened 2025-05-21
Migrate to pastey

shssoichiro:
There has been a security warning that the `paste` crate is officially marked as unmaintained, and users should migrate away from it. `pastey` is a fork of the `paste` crate with an identical interface, so it is the easiest migration path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1279,12 +1279,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b3a8cb46bdc156b1c90460339ae6bfd45ba0394e5effbaa640badb4987fdc261"
 
 [[package]]
-name = "pastey"
-version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


lu-zero:
I'd wait for a security issue to crop up or have comfirmation pastey is actually getting contributions from enough people that we wouldn't have to switch to another. Ideally in a couple of releases we should be able to paste idents as we need just using [this](https://github.com/rust-lang/rust/issues/124225)

lu-zero:
You release av-scenechange with pastey, the next rav1e release shall sync with it?

lu-zero (approved):

Timmmm:
Hey I'm just passing through but I had a thought - surely it's more risky to switch to a "maintained" fork than to stick with the unmaintained one? I'm sure it's fine in this case but I think a pretty easy supply chain attack in Rust would be to watch all popular crates to see which ones get marked as unmaintained, then as soon as they do create a maintained fork and encourage people to switch to it - all of the security tools will encourage people too!

You easily get into the supply chain without the actual work of having to create a useful library yourself (which I think is the main thing stopping supply chain attacks at the moment; not many people want to spend a ton of time creating a useful library and then just burn that on a probably-easily-discovered attack).

Anyway carry on; just a thought.

Issue #3415, opened 2025-03-09
The paste crate is unmaintained


theoparis:
Output from `cargo deny`:
```

error[unmaintained]: paste - no longer maintained
    ┌─ /Users/theo/Documents/wgpu-mc/Cargo.lock:281:1
    │
281 │ paste 1.0.15 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
    │
    ├ ID: RUSTSEC-2024-0436
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0436
    ├ The creator of the crate `paste` has stated in the [`README.md`](https://github.com/dtolnay/paste/blob/master/README.md)
      that this project is not longer maintained as well as archived the repository
    ├ Announcement: https://github.com/dtolnay/paste
    ├ Solution: No safe upgrade is available!
    ├ paste v1.0.15
      └── rav1e v0.7.1
          └── ravif v0.11.11
              └── image v0.25.5
                  └── wgpu-mc v0.0.1
                      ├── wgpu-mc-demo v0.1.0
                      └── wgpu-mc-jni v0.1.0
```

Issue #3418, opened 2025-04-02
Migrate away from paste


lu-zero:
Once https://github.com/rust-lang/rust/issues/124225 hits stable

onkoe:
`paste` is also unmaintained, which causes `cargo-deny` to fail in CI:

```console
$ cargo deny
# (...snip)

error[unmaintained]: paste - no longer maintained
    ┌─ /Users/barrett/Documents/projects/raves-project/backdrop/Cargo.lock:219:1
    │
219 │ paste 1.0.15 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
    │
    ├ ID: RUSTSEC-2024-0436
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0436
    ├ The creator of the crate `paste` has stated in the [`README.md`](https://github.com/dtolnay/paste/blob/master/README.md)
      that this project is not longer maintained as well as archived the repository
    ├ Announcement: https://github.com/dtolnay/paste
    ├ Solution: No safe upgrade is available!
    ├ paste v1.0.15
      └── rav1e v0.7.1
          └── ravif v0.11.12
              └── image v0.25.6
                  └── backdrop v0.1.0

advisories FAILED, bans ok, licenses FAILED, sources ok
```

lu-zero:
You can fix the cargo-deny configuration to simply warn about unmaintained crates.

PR #3442, opened 2025-06-21
Migrate from paste to pastey

pepa65:
Removed paste and added pastey from dependencies, `cargo update` and changed `paste` to `pastey` to all the files in `src`.


pepa65:
Fixes #3415 and #3418, but I now see there is another Pull request already for the same issue: #3430

PR #3453, opened 2025-09-21
fix: re-enable asm for scenechange detection

shssoichiro:
This was accidentally disabled when migrating the scenechange code to a
separate crate, because the feature for assembly was disabled.

This also fixes clippy lints which are required for CI to pass.


shssoichiro:
Clearly updating av-scenechange is going to be a pain, so I'll back that out for now.

lu-zero (approved):

PR #3461, opened 2025-12-23
Allow providing hermetic nasm binary

dzbarsky:
This lets this crate build with Bazel more easily without relying on system dependencies.


lu-zero (approved):

lu-zero:
I wonder why the crate doesn't provide this by itself...

dzbarsky:
> I wonder why the crate doesn't provide this by itself...

haha https://github.com/medek/nasm-rs/issues/43
I can do a followup to fix it there as well (https://github.com/medek/nasm-rs/pull/45)

PR #3477, opened 2026-08-28
Update to `rand` v0.10

Expyron:
- Update `rand` to v0.10
- Replace `rand_chacha` with `chacha20`
- Various mechanical fixes due to a few small breaking changes (rename Rng to RngExt, ChaChaRng to ChaCha20Rng)
- Increase MSRV to 1.85 to be compatible

### DIFF
--- a/.github/workflows/rav1e.yml
+++ b/.github/workflows/rav1e.yml
@@ -46,8 +46,8 @@ jobs:
           version: 2.15
 
       # Minimum version in Cargo.toml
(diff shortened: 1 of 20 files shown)


barrbrain:
`README.md` has a stale reference to the MSRV and we should update that too. Bumping the MSRV is worth a mention in the commit message or as a separate commit.

Expyron:
I split the change into two commits for clarity.

barrbrain (approved):

